PR #2, opened 2015-12-17
Don't do DNS lookups each time we want to send metrics

spladug:
socket.sendto() was resolving the name of the statsd server each time we
needed to send a batch of metrics. This was obviously quite slow. By
doing an upfront socket.connect() call, the DNS lookup only happens
once.

Here's what it looked like when I put the IP address instead of hostname
in the configs:

render 3

:eyeglasses: @kjoconnor @bsimpson63


kjoconnor:
:fish: programming!!!

bsimpson63:
:fish:

Issue #4, opened 2015-12-18
Thrift pool exceptions don't have good error messages


spladug:
Because of the `TException` constructor, `message` is `None`. This makes for rather annoying log messages.

spladug:
Improved in 2895dc2a8f9b5e692a05b3d3d78757c5a8a2feb7.

PR #9, opened 2015-12-21
Various small improvements

### LINKED ISSUE #3: Remove rounding from metrics
Though [the spec](https://github.com/b/statsd_spec) says "For compatibility all values should be integers in the range (-2^53^, 2^53^)." rounding the metrics to an integer is making for really really ugly graphs when things are quick.  It looks like everything we're using is float-safe, so let's just let the raw numbers through.

screenshot from 2015-12-18 09 38 01

spladug:
These fix #3 and #4. I'll do retries next.

:eyeglasses: @bsimpson63 @kjoconnor

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -73,7 +73,7 @@ class RawTransport(object):
         self.socket.connect(endpoint.address)
 
     def send(self, serialized_metric):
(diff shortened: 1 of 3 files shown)


bsimpson63:
:fish:

kjoconnor:
:fish:

PR #12, opened 2016-02-08
Additions for robin and clicktracker

spladug:
These are a few additions required for the click tracker and robin projects. I'll be working on event sending from baseplate next as needed for click tracker. Each one is pretty independent, so it might be best to look at individual commits.

:eyeglasses: @florenceyeun @MelissaCole @dellis23


dellis23:
:fish:

MelissaCole:
:fish: Looks good to me. Thanks for doing this!

PR #13, opened 2016-02-16
Message queues

spladug:
This will be helpful for communicating with sidekick processes on the same machine, e.g. events. I'm adding this here first and then moving the event collectors over to this from their SysV IPC queues. Once that's proven out, then the goal is to have r2 and baseplate app servers send messages to the collector via local-system queues and a small daemon, rather than going through Rabbit which is currently getting close to max throughput.

(this also includes a fix for a bug i introduced by trying to be too clever in constant_time_compare. i've made sure the code is identical to r2's copy again. i should've known better, sorry)

:eyeglasses: @bsimpson63 @ketralnis @MelissaCole

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ virtualenv:
 before_install:
     - sudo add-apt-repository -y ppa:reddit/ppa
     - sudo apt-get update -q
-    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster
+    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc
 
 install:
     - pip install -e ".[gevent,pyramid]"
--- a/baseplate/crypto.py
+++ b/baseplate/crypto.py
@@ -19,7 +19,7 @@ if hasattr(hmac, "compare_digest"):
     # This was added in Python 2.7.7 and 3.3
     constant_time_compare = hmac.compare_digest
 else:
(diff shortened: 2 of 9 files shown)


bsimpson63:
:nail_care: just the one question

spladug:
:haircut: question hopefully answered. please let me know if that's confusing or unhelpful or if a comment/code clarification is called for.

ketralnis:
This looks like it causes the sender to block when the message queue is full. Do you think it makes sense to make this configurable? I think we'd rather drop events than block r2

spladug:
@ketralnis the `put` and `get` methods take a `timeout` parameter which if set to 0 will immediately raise without blocking if the queue is full. You can see this in action in the event-collector error publishing https://github.com/reddit/event-collector/pull/10/files#diff-23b50895818eee443d54c2fc562a765dR149

(I didn't do it for non-errors, but I could. For things like r2, they'd retry if a 500 were returned. Unfortunately browsers and mobile devices sending screenview events would just drop the message forever so maybe it's better to hold the request open for a little while? That'd only happen if kafka's been down for quite a while though)

ketralnis:
Gotcha, that makes sense. :fish:

MelissaCole:
:fish: Looks good to me too. 

Was trying to think about if you should check for `ExistentialError` in the get/put in case something closed the MessageQueue from somewhere else but I don't think that would be an issue

spladug:
> Was trying to think about if you should check for ExistentialError in the get/put in case something closed the MessageQueue from somewhere else but I don't think that would be an issue

Ah neat, hadn't thought of that. We should actually be safe from that as the kernel won't delete a queue from existence while any processes have an open queue descriptor.

bsimpson63:
:fish:

PR #14, opened 2016-02-26
Add a system for sending events to the event collector

spladug:
This is split into two parts: a system for building and queing events,
and a daemon which reads from a queue, batches, and publishes events to
the collector. The goal is to make it so we can send a _lot_ of events
without overwhelming a central server like Rabbit.

It is immediately useful for @MelissaCole's clicktracker service, but I'm also
planning on getting r2 cut over to this ASAP so we can continue adding more
instrumentation and A/B testing without worrying about overwhelming the Rabbit
box.

:eyeglasses: @MelissaCole @umbrae @KeyserSosa 

cc: @Kaitaan @jophuds

### DIFF
--- a/baseplate/events/publisher.py
+++ b/baseplate/events/publisher.py
@@ -1,0 +1,203 @@
+import argparse
+import email.utils
+import gzip
+import hashlib
+import hmac
+import logging
+import time
+
+import requests
+
+from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
+from .. import config, make_metrics_client
+from .. _compat import configparser, BytesIO
+from .. message_queue import MessageQueue, TimedOutError
+
+
+logger = logging.getLogger(__name__)
+
+
+# seconds to wait for the event collector
+POST_TIMEOUT = 3
+# maximum time (seconds) an event can sit around while we wait for more
+# messages to batch
+MAX_BATCH_AGE = 1
+# seconds to wait between retries when the collector fails
+RETRY_DELAY = 2
+
+
+class Batcher(object):
+    """Time-aware batching producer.
+
+    A batch-consumer gives necessary context to this object. It must provide:
+        * batch_size_limit - the maximum size of an individual batch
+        * batch_size_overhead - base size cost of a batch
+        * get_item_size() - given an item, return its batch-relevant size
+        * consume_batch() - consume a fully formed batch
+
+    A flush may occur if the batch reaches the maximum specified size during
+    add() or if explicitly flush()ed.
+
+    """
+    def __init__(self, consumer):
+        self.consumer = consumer
+        self.batch = []
+        self.batch_size = self.consumer.batch_size_overhead
+        self.batch_start = None
+
+    @property
+    def batch_age(self):
+        """Return the age in seconds of the oldest item in the batch.
+
+        If there are no items in the batch, 0 is returned.
+
+        """
+        if not self.batch_start:
+            return 0
+        return time.time() - self.batch_start
+
+    def add(self, item):
+        """Add an item to the batch, potentially flushing."""
+        item_size = self.consumer.get_item_size(item)
+        if self.batch_size + item_size > self.consumer.batch_size_limit:
+            self.flush()
+        self.batch.append(item)
+        self.batch_size += item_size
+        if not self.batch_start:
+            self.batch_start = time.time()
+
+    def flush(self):
+        """Explicitly flush the batch if any items are enqueued."""
+        if self.batch:
+            self.consumer.consume_batch(self.batch)
+            self.batch = []
+            self.batch_size = self.consumer.batch_size_overhead
+            self.batch_start = None
+
+
+def gzip_compress(content):
+    buf = BytesIO()
+    with gzip.GzipFile(fileobj=buf, mode="wb", compresslevel=9) as f:
+        f.write(content)
+    return buf.getvalue()
+
+
+class BatchConsumer(object):
+    batch_size_overhead = 2  # the [] that wrap the json list
+    batch_size_limit = 500 * 1024
+
+    def __init__(self, metrics_client, cfg):
+        self.metrics = metrics_client
+        self.url = "https://%s/v1" % cfg.collector.hostname
+        self.key_name = cfg.key.name
+        self.key_secret = cfg.key.secret
+        self.session = requests.Session()
+
+    @staticmethod
+    def get_item_size(item):
+        # the item is already serialized, so we'll just add one byte for comma
+        return len(item) + 1
+
+    def _sign_payload(self, payload):
+        digest = hmac.new(self.key_secret, payload, hashlib.sha256).hexdigest()
+        return "key={key}, mac={mac}".format(key=self.key_name, mac=digest)
+
+    def consume_batch(self, items):
+        logger.info("sending batch of %d items", len(items))
+        payload = b"[" + b",".join(items) + b"]"
+        compressed_payload = gzip_compress(payload)
+        headers = {
+            "Date": email.utils.formatdate(usegmt=True),
+            "User-Agent": "baseplate-event-publisher/1.0",
+            "Content-Type": "application/json",
+            "X-Signature": self._sign_payload(payload),
+            "Content-Encoding": "gzip",
+        }
+
+        while True:
+            try:
+                with self.metrics.timer("post"):
+                    response = self.session.post(
+                        self.url,
+                        headers=headers,
+                        data=compressed_payload,
+                        timeout=POST_TIMEOUT,
+
+                        # http://docs.python-requests.org/en/latest/user/advanced/#keep-alive
+                        stream=False,
+                    )
+                response.raise_for_status()
+            except requests.HTTPError as exc:
+                self.metrics.counter("error.http").increment()
+                logger.exception("HTTP Request failed")
+
+                # we should crash if it's our fault
+                if getattr(exc, "response", None) and exc.response.status_code < 500:
+                    raise
+            except IOError:
+                self.metrics.counter("error.io").increment()
+                logger.exception("HTTP Request failed")
+            else:
+                self.metrics.counter("sent").increment(len(items))
+                return
+
+            time.sleep(RETRY_DELAY)
+
+
+def publish_events():
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument("config_file", type=argparse.FileType("r"),
+        help="path to a configuration file")
+    arg_parser.add_argument("--queue-name", default="main",
+        help="name of event queue / publisher config (default: main)")
+    arg_parser.add_argument("--debug", default=False, action="store_true",
+        help="enable debug logging")
+    args = arg_parser.parse_args()
+
+    if args.debug:
+        level = logging.DEBUG
+    else:
+        level = logging.WARNING
+    logging.basicConfig(level=level)
+
+    config_parser = configparser.ConfigParser()
+    config_parser.readfp(args.config_file)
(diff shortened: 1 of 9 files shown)


baseplate/events/queue.py
```diff
@@ -0,0 +1,139 @@
...
+        :param str key: The name of the field.
+        :param value: The value to set the field to. Should be JSON
+            serializable.
+        :param bool obfuscate: Whether or not to put the field in the obfuscated
+            section. This is used for sensitive info like IP addresses that must
+            be treated with care.
+
+        """
+        if not obfuscate:
+            self.payload[key] = value
```
MelissaCole: Currently the [eventcollector](https://github.com/reddit/reddit-public/blob/master/r2/r2/lib/eventcollector.py#L1146) checks to see if the `value` is `None` and won't add it to the payload if it is. Changing this would send some empty fields that I don't think are meant to be in there (and lead to a lot more lines in the code to fix this).

spladug: Good point. Part of me has been thinking there's some forwards compatibility issue with omitting nulls, but the more I think about it the less I think it's real. I'll bring that back.

spladug: d713618002f610a1db4d220261e335afbe4f9792

tests/unit/events/publisher_tests.py
```diff
@@ -0,0 +1,112 @@
...
+        self.batcher.flush()
+        self.assertEqual(self.batcher.batch_start, None)
+
+    def test_manual_flush(self):
+        self.batcher.add("a")
+        self.batcher.add("b")
+        self.batcher.flush()
+        self.consumer.consume_batch.assert_called_once_with(["a", "b"])
+
+    def test_flush_when_full(self):
```
Deimos: :no_bell: This is probably just a general testing thing more than anything, but should this use `self.batch_size_limit` instead of having hardcoded numbers so that it wouldn't need updating if that limit is changed? For example:

```
def test_flush_when_full(self):
    for i in range(self.batch_size_limit + 2):
        self.batcher.add(str(i))
    self.consumer.consume_batch.assert_called_once_with(
        [str(i) for i in range(self.batch_size_limit)])
```

tests/unit/events/queue_tests.py
```diff
@@ -0,0 +1,101 @@
...
+        self.assertEqual(event.topic, "topic")
+        self.assertEqual(event.event_type, "type")
+        self.assertEqual(event.timestamp, 333000)
+        self.assertEqual(event.id, "1-2-3-4")
+
+    def test_timestamp_specified(self):
+        timestamp = datetime.datetime(2016, 2, 23, 0, 3, 17)
+        event = Event("topic", "type", timestamp=timestamp)
+        self.assertEqual(event.timestamp, 1456185797000)
+
```
Deimos: :no_bell: Would a test for a non-UTC timestamp be a good idea?

Deimos:
Just a couple tiny testing questions, otherwise looks great to me. :fish:

tests/unit/events/queue_tests.py
spladug: Yeah, it would. I just don't know a good way to do that without either a) requiring `pytz` to run the test suite, b) making my own timezone class, c) only testing that on py3.2+ which has a timezone in the stdlib. Thoughts?

tests/unit/events/publisher_tests.py
spladug: I'm never quite clear on where to draw the line on being too tied to internal implementation vs. being easier to update. :/

tests/unit/events/queue_tests.py
Deimos: Ah, hmm. I guess the way to do it without adding `pytz` or another timezone library would be your own class that inherits from `datetime.tzinfo`. It looks like it would probably just take something like this:

```
class FakeTZ(datetime.tzinfo):
    def utcoffset(self, dt):
        return datetime.timedelta(hours=2)
```

Not too bad, might be worth adding since an error that starts allowing non-UTC timestamps could cause a pretty big mess in the data.

KeyserSosa:
this is awesome :fish:

tests/unit/events/queue_tests.py
spladug: alrighty, done!

Deimos:
Excellent, thanks. :fish:

MelissaCole:
:fish: Really awesome job. Thanks!

baseplate/events/publisher.py
```diff
@@ -0,0 +1,203 @@
...
+        return len(item) + 1
+
+    def _sign_payload(self, payload):
+        digest = hmac.new(self.key_secret, payload, hashlib.sha256).hexdigest()
+        return "key={key}, mac={mac}".format(key=self.key_name, mac=digest)
+
+    def consume_batch(self, items):
+        logger.info("sending batch of %d items", len(items))
+        payload = b"[" + b",".join(items) + b"]"
+        compressed_payload = gzip_compress(payload)
```
umbrae: I'm sure you already thought about this, but is there a chance that by always gzipping we're actually losing out and that we should consider not gzipping for items below a certain threshold? (I think we just did something with this somewhere else recently.) This talks about it a bit: http://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-gzip-performance-benefits

Maybe we feel confident that even a single event would be big enough to always gzip?

baseplate/events/publisher.py
```diff
@@ -0,0 +1,203 @@
...
+                if getattr(exc, "response", None) and exc.response.status_code < 500:
+                    raise
+            except IOError:
+                self.metrics.counter("error.io").increment()
+                logger.exception("HTTP Request failed")
+            else:
+                self.metrics.counter("sent").increment(len(items))
+                return
+
+            time.sleep(RETRY_DELAY)
```
umbrae: :no_bell: OK I'm just gonna think through this out loud, I think I've got it but I want to make sure there are no queue sharing issues and maybe it'll be helpful to someone else later:

Each consumer daemon has a single batcher. Each batcher has a maximum size and age.

The daemon will fill the batcher with messages it consumes from the queue.

If the batcher reaches the maximum size or age, it will flush synchronously. That flush is an http POST. If that flush returns a 500, it will sleep, halting the consumer, and retry in 2 seconds. As this happens all other consumers are free to batch as they like.

They may consume and send just fine, which means that events may be sent out of order. (Which is fine and we should expect).

If the event collector starts returning 500s for everyone, every consumer daemon will reach this state, retrying every 2 seconds. This means that nothing will be consuming from the top level queues, but they still may be added to. The event collector is likely receiving POSTs on success from every consumer every second or two anyhow, so this load will be expected. A fixed retry of 2seconds is probably fine.

If we've been in a 500 state for a minute or two and have a whole bunch of full queues, we'll start consuming them and batching them right away as we hit max size. This may overwhelm the event collector server, but if it does it will start to return 500 level responses (or timeouts) and we'll retry in two seconds, effectively slowing us down.

OK, phew. I think I now understand this better and this feels great. :+1:

spladug: :+1: fwiw, given the current rate of events from r2 servers and the size of the queues we should be good for at least half an hour of consecutive collector downtime before the r2 queues start filling up

baseplate/events/publisher.py
spladug: Yeah, that's definitely possible. I think in practice the server-side events are always gonna be big enough. For example, a login event on my local install is 696 bytes uncompressed and ends up 451 compressed. Then you get to batching (which will likely happen every batch as on average we're doing about 4 events per second per server right now) and you're really getting benefit from compression.

baseplate/events/publisher.py
spladug: also: in current puppet work, there's only one daemon per queue so the out-of-order stuff won't really happen. the message queues do leave that possibility open (and the kafka producers on the collectors themselves run multiple) so it's good to think about.

baseplate/events/queue.py
```diff
@@ -0,0 +1,146 @@
+"""Client library for sending events to the data processing system.
+
+This is for use with the event collector system. Events generally track
+something that happens in production that we want to instrument for planning
+and analytical purposes.
+
+Events are serialized and put onto a message queue on the same server. These
+serialized events are then consumed and published to the remote event collector
+by a separate daemon.
```
umbrae: Do you think we need to worry about losing events on app kill from autoscaling?

baseplate/events/queue.py
```diff
@@ -0,0 +1,146 @@
...
+import json
+import time
+import uuid
+
+from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
+from baseplate.message_queue import MessageQueue, TimedOutError
+
+
+class Event(object):
+    """An event."""
```
umbrae: :no_bell: for some reason this made me lol :)

baseplate/events/queue.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, topic, event_type, timestamp=None, id=None):
+        self.topic = topic
+        self.event_type = event_type
+        if timestamp:
+            if timestamp.tzinfo and timestamp.utcoffset().total_seconds() != 0:
+                raise ValueError("Timestamps must be in UTC")
```
umbrae: yaaaaayyyy to being strict about this

tests/unit/events/publisher_tests.py
umbrae: I think what Chad means here is that you specifically specify `7`, even though if someone changed `batch_size_limit` in your test on line 24 to `10`, this would fail. He's wondering if you should avoid that by specifically setting this range to be 1 (or 2) larger than the batch size limit you defined on line 24.

tests/unit/events/publisher_tests.py
```diff
@@ -0,0 +1,112 @@
...
+        for i in range(7):
+            self.batcher.add(str(i))
+        self.consumer.consume_batch.assert_called_once_with(
+            ["0", "1", "2", "3", "4"])
+
+
+class CompressTests(unittest.TestCase):
+    def test_compress(self):
+        input = b"test"
```
umbrae: tiniest of nits: `input` is already a keyword :3

umbrae:
:fish: yaaayyy, this looks super fantastic. Top work spladug!

baseplate/events/queue.py
spladug: In regular operation? no more so than we already do (`amqp.add` adds an item to a `Queue.Queue` in-process which gets drained by the worker thread that talks to amqp synchronously).

When things are going poorly? yes, that's definitely a concern. We can pause the scaler during an extended downtime of the collector.

baseplate/events/queue.py
spladug: A comment.

:P

tests/unit/events/publisher_tests.py
spladug: ah, gotcha. that sounds good. done.

tests/unit/events/publisher_tests.py
spladug: how's this look?

umbrae: much better, thanks :)

spladug:
:haircut: just the changes you can see in individual commits. i had some inspiration last night on how to clean up the os side of things and it needed the `--create` code.

Deimos:
:fish:

umbrae:
:fish:

spladug:
a5a20363a31788e1ef42f3deca4cf58fcf9b22e8...6b1bc5b984f5d549591188fda57e671b4b3ad28b

PR #15, opened 2016-02-29
Allow pyramid apps to generate their own trace information

spladug:
This is important for external-facing pyramid applications so that trace
headers don't get polluted with client-sent garbage.

:eyeglasses: @kjoconnor @xilvar

### DIFF
--- a/baseplate/integration/pyramid.py
+++ b/baseplate/integration/pyramid.py
@@ -9,7 +9,10 @@ An abbreviated example of it in use::
         configurator = Configurator()
 
         baseplate = Baseplate()
-        baseplate_config = BaseplateConfigurator(baseplate)
+        baseplate_config = BaseplateConfigurator(
+            baseplate,
+            trust_trace_headers=True,
+        )
         configurator.include(baseplate_config.includeme)
 
         return configurator.make_wsgi_app()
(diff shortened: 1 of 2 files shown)


baseplate/integration/pyramid.py
```diff
@@ -43,9 +58,14 @@ def _on_new_request(self, event):
             # TODO: some metric for 404s would be good
             return
 
-        trace_id = request.headers.get("X-Trace", "no-trace")
-        parent_id = request.headers.get("X-Parent", "no-parent")
-        span_id = request.headers.get("X-Span", "no-span")
+        if self.trust_trace_headers:
+            trace_id = request.headers.get("X-Trace", "no-trace")
```
kjoconnor: Not really sure, but would we want to make our headers for this slightly more unique? i.e. X-Reddit-Trace or something? Just in case there's something else out there that does tracing and would stomp on others. This might not be a real concern, so who knows!

kjoconnor:
:fish: , just a question of naming

baseplate/integration/pyramid.py
spladug: Nothing's really using this yet, so we could definitely change them if it's an issue. I don't really feel strongly.

xilvar:
:fish: cool, looks good... I want to see those trace visualizations :)

spladug:
a5a20363a31788e1ef42f3deca4cf58fcf9b22e8

PR #17, opened 2016-03-08
Adding prose doc that walks through using MessageQueue.

gtaylor:
This could probably still use some polishing, but I figured I'd toss this up here as a start. This seemed like a neat little convenience class that I could see myself using in the future, and I love having examples with supporting prose for when I inevitably forget how something works.

Also: First fork! :cake: :pizza: :tada:


spladug:
This is super cool! Thank you for writing these docs up. I've been thinking of moving the docs more towards how the Python stdlib docs work: prose mixed with full API documentation. What do you think of merging these examples/notes into the lead-in section of `api/baseplate/message_queue.rst` and still linking it from index? Also, it'd be nice to have the code samples tested with doctest so they can't silently bitrot.

gtaylor:
Hello @spladug!

I can probably find the time to make these changes sometime in the next two nights. If you get ambitious and just want to get it out of the way, let me know if you start so I don't duplicate your effort.

Thanks for the quick review!

gtaylor:
I moved these to the API docs as suggested. Initially, I replaced the link to the old prose-only docs with the API docs, but the title of the article is "baseplate.message_queue", which looked pretty weird on the index. We can rename the article, but then it looks out of place in the API reference index.

Unless there's a way to override the name within a toctree that I don't know about. That would definitely solve this.

doctests are in as well.

Once this is in good shape, I can rebase/squash it before merge so we don't junk up your pretty git history.

spladug:
This looks great, thanks! I'll merge tomorrow and get cracking soon on narrativizing the rest of the docs.

spladug:
that^ is a :fish: btw (making it explicit to clean up my [review dashboard](https://github.com/reddit/reddit/wiki/The-Code-Salon))

gtaylor:
:dolphin: looking forward to it!

spladug:
Merged as 0083efbeed2d5c5fb720602c72d4922941b71a72. Thanks!

PR #20, opened 2016-03-27
Documentation restructuring and cleanup

### LINKED ISSUE #16: Document or include Sphinx requirements
I needed to install the following to get the docs to build locally:

* sphinx
* sphinxcontrib-spelling
* pyenchant

I have traditionally followed the `docs-requirements.txt`/`dev-requirements.txt` pattern for stuff like this, but didn't want to presume this is where you were heading with a premature PR.

I can toss a PR over the fence if you let me know how you'd like to handle it (or not handle it).

Looking good so far!

spladug:
This is a bunch of restructuring and (hopefully) improvements to the docs. I've also got a shiny little favicon now!

For easier reading, I've uploaded the docs in their new form to S3: http://spladug.s3.amazonaws.com/baseplate/index.html

cc: @gtaylor (fixes #16 and does what we talked about in #17)


gtaylor:
I'll have play with some of these other modules enough to comment intelligently on them, but this is starting to look very nice.

docs/baseplate/core.rst
```diff
@@ -1,19 +1,62 @@
...
+
+       ... snip ...
 
-The heart of Baseplate is its diagnostics system.
+When a request is made which routes to the ``do_something`` handler, a
+:py:class:`~baseplate.core.RootSpan` is automatically created. If the incoming
+request has trace headers, the root span will be identical to the child span in
+the upstream service. When we call ``request.some_redis_client.ping()`` in the
+handler, Baseplate will create a child :py:class:`~baseplate.core.Span` object
+to represent the time taken talking to redis.
```
dellis23: Nice!  Definitely getting a better grasp of it now.

While I think I get the gist of most of this, what does it mean for a Span object to be created?  Sure, it represents some time taken, but what does it do?  Is that Span object then processed in some way later?  Does it automatically perform some action when it's created?  Or does it just keep the information there to be accessed later?

I see below you say the span nonsense means baseplate automatically sends metrics based on them.  Is this through the observers?

It's getting much clearer! :-)  All of this might be explained below, but since you're adding an explanation here, it feels *almost* complete for a high-level overview.

dellis23:
:nail_care: Cool!  Thanks for doing this.

docs/baseplate/core.rst
spladug: thanks for the feedback. i've updated the following paragraph, lemme know what you think.

spladug:
:haircut:

dellis23:
:fish: Awesome, thanks!

spladug:
Merged as f5dd8522c609d8628ca90fedf5cd8b2ce67fe9be...ac3e4b19893d918fb23e6fff715cb5dc6529e532

docs will always be improving, so please feel free to comment if anything continue to be confusing.

PR #21, opened 2016-03-27
MessageSigner: Return some info about valid signatures

spladug:
This is useful if we want to collect metrics on how old signatures are
when they are used, even if valid.

:eyeglasses: @MelissaCole @umbrae 

cc: WEB-1622

### DIFF
--- a/tests/unit/crypto_tests.py
+++ b/tests/unit/crypto_tests.py
@@ -34,6 +34,17 @@ class SignatureTests(unittest.TestCase):
         self.signer.validate_signature(self.message, signature)
 
     @mock.patch("time.time")
+    def test_signature_info(self, time):
+        time.return_value = 0
+        signature = self.signer.make_signature(self.message, max_age=datetime.timedelta(seconds=30))
+
+        time.return_value = 10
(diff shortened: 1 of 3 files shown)


tests/unit/crypto_tests.py
```diff
@@ -34,6 +34,17 @@ def test_roundtrip(self):
         self.signer.validate_signature(self.message, signature)
 
     @mock.patch("time.time")
+    def test_signature_info(self, time):
+        time.return_value = 0
+        signature = self.signer.make_signature(self.message, max_age=datetime.timedelta(seconds=30))
+
+        time.return_value = 10
```
MelissaCole: Why is this line necessary? The expiration won't change as the time changes, so I'm just curious

MelissaCole:
:fish: lgtm

tests/unit/crypto_tests.py
spladug: iiit's not. good point. removed!

spladug:
f5dd8522c609d8628ca90fedf5cd8b2ce67fe9be

Issue #19, opened 2016-03-24
Record success/failure counts in child spans


spladug:
r2 activity service stuff is doing a lot of `g.stats.simple_event("activity_service.write.fail")` in a try/except. This should be automatic as part of the child span stuff.

spladug:
This is now possible as of 7c263cfa79768a54437edb9653f7a865055e5d81 and I've taken care of the activity stuff mentioned in the issue over in r2.

PR #23, opened 2016-03-28
Pass exception information to span observers when spans end

spladug:
This allows us to take different actions depending on if the span exited
cleanly or exceptionally, such as automatically sending a "this request
failed" metric or the like.

Note: this is an API-breaking change. r2 is currently the only
subclasser of Span outside of baseplate, so that should be safe to fix
up before rolling this out.

This is in aid of #19 


dellis23:
:fish:

spladug:
7c263cfa79768a54437edb9653f7a865055e5d81

PR #26, opened 2016-04-06
Unix sockets and script cleanup

spladug:
This does a restructuring of the scripts and then updates the server and healthchecker for UNIX domain socket support. 

Einhorn support for UNIX domain sockets comes in reddit/einhorn#1

:eyeglasses: @dellis23 @ckwang8128 @bsimpson63

### DIFF
--- a/baseplate/requests.py
+++ b/baseplate/requests.py
@@ -15,6 +15,9 @@ import urllib
 import requests.adapters
 import urllib3.connectionpool
 
+# the adapter code below is inspired by similar code in
+# https://github.com/msabramo/requests-unixsocket/blob/master/requests_unixsocket/adapters.py
+# https://github.com/docker/docker-py/blob/master/docker/unixconn/unixconn.py
 
 class _UNIXConnection(urllib3.connectionpool.HTTPConnection):
     def __init__(self, url):
@@ -45,4 +48,5 @@ class _UNIXAdapter(requests.adapters.HTTPAdapter):
 
 
 def add_unix_socket_support(session):
+    """Add support to a Requests session for HTTP over UNIX domain sockets."""
     session.mount("http+unix://", _UNIXAdapter())
--- a/baseplate/requests.py
+++ b/baseplate/requests.py
@@ -1,0 +1,48 @@
+"""Internal helpers for the requests HTTP client library.
+
+This stuff is not stable yet, so it's only for baseplate-internal use.
+
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import socket
+import urlparse
+import urllib
+
+import requests.adapters
+import urllib3.connectionpool
+
+
+class _UNIXConnection(urllib3.connectionpool.HTTPConnection):
+    def __init__(self, url):
+        urllib3.connectionpool.HTTPConnection.__init__(self, "localhost")
+        self.url = urlparse.urlparse(url)
+
+    def connect(self):
+        socket_path = urllib.unquote(self.url.netloc)
+
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.sock.settimeout(1)
+        self.sock.connect(socket_path)
+
+
+class _UNIXConnectionPool(urllib3.connectionpool.HTTPConnectionPool):
+    def __init__(self, url):
+        super(_UNIXConnectionPool, self).__init__(host="localhost")
+        self.url = url
+
+    def _new_conn(self):
+        return _UNIXConnection(self.url)
+
+
+class _UNIXAdapter(requests.adapters.HTTPAdapter):
+    def get_connection(self, url, proxies=None):
+        assert not proxies, "proxies are not supported"
+        return _UNIXConnectionPool(url)
+
+
+def add_unix_socket_support(session):
+    session.mount("http+unix://", _UNIXAdapter())
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -74,6 +74,23 @@ def configure_logging(debug):
 
 
 def make_listener(endpoint):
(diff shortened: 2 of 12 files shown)


baseplate/server/healthcheck.py
```diff
@@ -0,0 +1,74 @@
+"""Check health of a baseplate service on localhost."""
```
dellis23: I know you're just moving this script, so maybe it's out of scope for this PR, but it'd be nice to have an explanation of what the health checks do, how they should be run, what the output is when they fail, and how they are intended to be used.

baseplate/requests.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class _UNIXAdapter(requests.adapters.HTTPAdapter):
+    def get_connection(self, url, proxies=None):
+        assert not proxies, "proxies are not supported"
+        return _UNIXConnectionPool(url)
+
+
+def add_unix_socket_support(session):
+    session.mount("http+unix://", _UNIXAdapter())
```
bsimpson63: What's the deal with the "http+unix" protocol, I'm not having any luck googling. Is that a custom convention you're using?

baseplate/server/__init__.py
```diff
@@ -75,7 +75,7 @@ def configure_logging(debug):
 
 def make_listener(endpoint):
     try:
-        fileno = int(os.environ["EINHORN_FDS"])
+        fd_count = int(os.environ["EINHORN_FD_COUNT"])
     except KeyError:
```
dellis23: What's the catch on the `KeyError` for here?  To detect whether or not we're running under einhorn?

baseplate/requests.py
ckwang8128: I also had a bit of trouble with this. AFAICT, it's custom so the custom requests adapter can be mounted and used for it to differentiate from http over INET. 

Adding a comment about this would be good for readability.

baseplate/server/__init__.py
```diff
@@ -83,7 +83,10 @@ def make_listener(endpoint):
         sock.listen(128)
         return sock
     else:
-        return socket.fromfd(fileno, socket.AF_INET, socket.SOCK_STREAM)
+        assert fd_count > 0, "Running under Einhorn  but no sockets were bound."
+        fileno = int(os.environ["EINHORN_FD_0"])
```
dellis23: Why do we use the first fd from einhorn over the others?  Or: what's going on with the fd count and specific fd logic?

baseplate/server/healthcheck.py
```diff
@@ -0,0 +1,74 @@
...
+        url = "http+unix://{path}/health".format(path=quoted_path)
+    else:
+        raise ValueError("unrecognized socket family %r" % endpoint.family)
+
+    session = requests.Session()
+    add_unix_socket_support(session)
+    response = session.get(url, timeout=TIMEOUT)
+    response.raise_for_status()
+    response.json()
+
```
ckwang8128: is there any reason why the thrift and http service checks don't return the same things for healthy and unhealthy? if not, I think consistency would be good (might be out of scope for this). this could also be handled at run_healthchecks.

ckwang8128:
💅

dellis23:
:nail_care: Just some potential comments / documentation things.  Up to you as to which are worthwhile.

baseplate/server/healthcheck.py
spladug: neither returns anything. the scripts either succeed (in which case they return a successful exit code and print out "OK!") or they fail and print out an error (and a fail exit code).

baseplate/server/__init__.py
spladug: I've added a docstring to this function and a few others in this file to hopefully explain the Einhorn logic. Do these answer your questions?

baseplate/server/__init__.py
spladug: (same note as below re: docstrings)

baseplate/requests.py
spladug: I've added some comments/docstrings in this file. Do these answer your questions?

(I'm being deliberately obtuse in comments because I want the code to speak for itself to posterity who might not have this pull request to read)

baseplate/server/healthcheck.py
spladug: totally valid. don't ever hold back on asking for docs/explanation! :)

i've added another file to the docs explaining this tool, does that help?

spladug:
i've added a bunch of docs and done some tweaks. thanks for keeping me on my toes, i definitely got lazy on the docs and clarity in this pull request. :haircut:

baseplate/server/healthcheck.py
ckwang8128: ugh, I forgot python doesn't implicitly return (better explicit than implicit after all...). That makes sense. healthcheck.rst clarifies this as well, thanks!

baseplate/requests.py
ckwang8128: 💯 / 💯

ckwang8128:
🐟

bsimpson63:
:fish:

dellis23:
:fish: Thanks!

docs/cli/healthcheck.rst
```diff
@@ -0,0 +1,34 @@
...
+
+   baseplate-healthcheck2 wsgi /run/myservice.sock
+
+Results
+-------
+
+If the service is healthy, the tool will exit with a status code indicating
+success (0) and print "OK!". If the service is unhealthy, the tool will exit
+with a status code indicating failure (1) and print an error message explaing
+what went wrong.
```
dellis23: Awesome!  Very helpful.

spladug:
0734b29d0c2a3cc72f531cf54f0ab77b270793a2...e563495062e9212a347966006603aca5881ac90c

PR #27, opened 2016-04-10
Pull Einhorn client code into separate module

spladug:
This makes a more clearly defined interface for interacting with Einhorn
and is a little more robust with respect to detecting if we're a child
of Einhorn.


dellis23:
:fish: Cool, even cleaner!

ckwang8128:
:fish: I love good testing

PR #29, opened 2016-04-11
Add unset_field and unset_fields to Event

lan17:
Wrote unit tests.

They pass.

### DIFF
--- a/baseplate/events/queue.py
+++ b/baseplate/events/queue.py
@@ -80,24 +80,15 @@ class Event(object):
             obfuscated_payload = self.payload.setdefault("obfuscated_data", {})
             obfuscated_payload[key] = value
 
-    def unset_field(self, key):
+    def unset_field(self, key, obfuscated=False):
         """Remove the key from payload if it exists.
-        :param key:
+        :param str key: The name of the field
+        :param bool obfuscated: Whether to look for the field in the obfuscated
+            payload
         :return: value of key that was removed.  None otherwise
         """
-        return self.payload.pop(key, None)
-
-    def unset_fields(self, keys):
-        """Iterate over keys and remove each key from payload
-        :param keys:  iterable containing keys to remove
-        :return: number of items that were removed
-        """
-        counter = 0
-        for key in keys:
-            val = self.unset_field(key)
-            if val is not None:
-                counter += 1
-        return counter
+        payload = self.payload if not obfuscated else self.payload.get("obfuscated_data")
+        return payload.pop(key, None)
 
     def serialize(self):
         return json.dumps({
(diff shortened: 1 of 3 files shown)


lan17:
👓  @bsimpson63 @spladug

spladug:
The tests are failing on the spellchecker:

```
baseplate/events (line   1) param 
baseplate/events (line   1) param 
baseplate/events (line   1) iterable 
```

(you can ignore the spam about various packages not being installed, that's just the documentation generator complaining).

baseplate/events/queue.py
```diff
@@ -80,6 +80,25 @@ def set_field(self, key, value, obfuscate=False):
...
             obfuscated_payload[key] = value
 
+    def unset_field(self, key):
+        """Remove the key from payload if it exists.
+        :param key:
+        :return: value of key that was removed.  None otherwise
+        """
+        return self.payload.pop(key, None)
+
+    def unset_fields(self, keys):
```
spladug: I'm not really sure we need this function, if a caller wants to do this they can loop themselves.

baseplate/events/queue.py
```diff
@@ -80,6 +80,25 @@ def set_field(self, key, value, obfuscate=False):
             obfuscated_payload = self.payload.setdefault("obfuscated_data", {})
             obfuscated_payload[key] = value
 
+    def unset_field(self, key):
+        """Remove the key from payload if it exists.
+        :param key:
```
spladug: I think what's going on with the spellchecker is that rST is pretty picky about line spacing (and even if it weren't it'd be good to be consistent with other docstrings in this package). The param should also actually be documented. How about the following for docs for consistency sake:

```python
"""Remove a field from the event if it exists.

:param str key: The name of the field.
:return: The value of the removed field or :py:data:`None`.

"""
```

this function should also probably support the `obfuscated` flag like `get_field` does.

tests/unit/events/queue_tests.py
```diff
@@ -77,6 +77,38 @@ def test_serialize(self, time, uuid):
             },
         })
 
+    def test_unset_field(self):
+        event = Event("topic", "type")
+        event.set_field("foo", "bar")
+
+        # foo should point to bar
```
spladug: :+1: :)

spladug:
:nail_care: thanks!

baseplate/events/queue.py
lan17: I was going to use it in my other PR for convenience.

baseplate/events/queue.py
lan17: fixed

tests/unit/events/queue_tests.py
lan17: added more tests for obfuscated.

baseplate/events/queue.py
lan17: removed from this PR.

lan17:
💇 
fixed the typos!

spladug:
going to close this for now as discussed in slack.

PR #31, opened 2016-04-18
Retries!

spladug:
There are two big things going on here: the `baseplate.retry` module and then the sugar in `baseplate.context.thrift`. 

The `baseplate.retry` module is a low-level library with "guts exposed". It provides common retry policies in a composable manner. That commit also refactors several places inside baseplate which did their own retry logic to use this module. 

The second part, in `baseplate.context.thrift`, adds some syntactic sugar to the thrift client. By default, the client will continue to do no retries (a policy of `attempts=1`) on remote operations because it cannot know if the operation being done is idempotent or not. The new sugar adds a context manager that allows a retry policy to be applied to service calls made within a `with` statement.

```python
with context.activity_service.retrying(attempts=3) as activity_service:
    return activity_service.count_activity(sr._fullname)
```

I looked at a bunch of prior art and tried various options before settling on this structure. I've written up [some documentation on my thoughts behind the API design](https://gist.github.com/spladug/dc587162fb718d8a9aca2b4eeeb6df69) in case it's helpful.

I also restructured the exception hierarchy from the `thrift_pool` module. I'm taking the tack now that it shouldn't make its own exceptions as it's essentially acting as a "missing piece" from the thrift library itself and so it should integrate with that exception hierarchy properly instead. r2's currently the only user of this module and its use will not be broken by this change, though it could be scoped down further to prevent masking bugs down the line.

There are a couple of other commits here that do minor lint-related stuff that came up while working on this.

:eyeglasses: @bsimpson63 @KeyserSosa @ckwang8128

### DIFF
--- a/baseplate/context/thrift.py
+++ b/baseplate/context/thrift.py
@@ -3,10 +3,14 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import contextlib
 import functools
 import inspect
 
+from thrift.transport.TTransport import TTransportException
+
 from . import ContextFactory
+from ..retry import RetryPolicy
 
 
 class ThriftContextFactory(ContextFactory):
@@ -21,6 +25,15 @@ class ThriftContextFactory(ContextFactory):
     :param baseplate.thrift_pool.ConnectionPool pool: The connection pool.
     :param client_cls: The class object of a Thrift-generated client class,
         e.g. ``YourService.Client``.
+
+    The proxy object has a ``retrying`` method which takes the same parameters
+    as :py:meth:`RetryPolicy.new <baseplate.retry.RetryPolicy.new>` and acts as
+    a context manager. The context manager returns another proxy object where
+    Thrift service method calls will be automatically retried with the
+    specified retry policy when transient errors occur::
+
+        with context.my_service.retrying(attempts=3) as svc:
+            svc.some_method()
 
     """
     def __init__(self, pool, client_cls):
@@ -52,25 +65,48 @@ def _enumerate_service_methods(client):
 class PooledClientProxy(object):
     """A proxy which acts like a thrift client but uses a connection pool."""
 
-    def __init__(self, client_cls, pool, root_span, namespace):
+    def __init__(self, client_cls, pool, root_span, namespace, retry_policy=None):
         self.client_cls = client_cls
         self.pool = pool
         self.root_span = root_span
         self.namespace = namespace
+        self.retry_policy = retry_policy or RetryPolicy.new(attempts=1)
 
         for name in _enumerate_service_methods(client_cls):
             setattr(self, name, functools.partial(
                 self._call_thrift_method, name))
 
+    @contextlib.contextmanager
+    def retrying(self, **policy):
+        yield PooledClientProxy(
+            self.client_cls,
+            self.pool,
+            self.root_span,
+            self.namespace,
+            retry_policy=RetryPolicy.new(**policy),
+        )
+
     def _call_thrift_method(self, name, *args, **kwargs):
         trace_name = "{}.{}".format(self.namespace, name)
+        last_error = None
 
-        with self.root_span.make_child(trace_name) as span:
-            with self.pool.connection() as prot:
-                prot.trans.set_header("Trace", str(span.trace_id))
-                prot.trans.set_header("Parent", str(span.parent_id))
-                prot.trans.set_header("Span", str(span.id))
+        for _ in self.retry_policy:
+            try:
+                with self.root_span.make_child(trace_name) as span:
+                    with self.pool.connection() as prot:
+                        prot.trans.set_header("Trace", str(span.trace_id))
+                        prot.trans.set_header("Parent", str(span.parent_id))
+                        prot.trans.set_header("Span", str(span.id))
 
-                client = self.client_cls(prot)
-                method = getattr(client, name)
-                return method(*args, **kwargs)
+                        client = self.client_cls(prot)
+                        method = getattr(client, name)
+                        return method(*args, **kwargs)
+            except TTransportException as exc:
+                last_error = exc
+                continue
+        else:
+            raise TTransportException(
+                type=TTransportException.TIMED_OUT,
+                message="retry policy exhausted while attempting operation, "
+                        "last error was: {}".format(last_error),
+            )
--- a/baseplate/message_queue.py
+++ b/baseplate/message_queue.py
@@ -5,9 +5,10 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import select
-import time
 
 import posix_ipc
+
+from .retry import RetryPolicy
 
 
 class MessageQueueError(Exception):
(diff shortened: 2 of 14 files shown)


baseplate/context/thrift.py
```diff
@@ -52,25 +65,48 @@ def _enumerate_service_methods(client):
 class PooledClientProxy(object):
     """A proxy which acts like a thrift client but uses a connection pool."""
 
-    def __init__(self, client_cls, pool, root_span, namespace):
+    def __init__(self, client_cls, pool, root_span, namespace, retry_policy=None):
```
KeyserSosa: `retry_policy=None` leading to `self.retry_policy=RetryPolicy.new(attempts=1)` seems potentially surprising. Maybe beef up the class docstring?

baseplate/context/thrift.py
```diff
@@ -52,25 +65,48 @@ def _enumerate_service_methods(client):
...
+            try:
+                with self.root_span.make_child(trace_name) as span:
+                    with self.pool.connection() as prot:
+                        prot.trans.set_header("Trace", str(span.trace_id))
+                        prot.trans.set_header("Parent", str(span.parent_id))
+                        prot.trans.set_header("Span", str(span.id))
+
+                        client = self.client_cls(prot)
+                        method = getattr(client, name)
+                        return method(*args, **kwargs)
```
KeyserSosa: 🔕  This is mostly for my edification: which parts of this block can actually raise the `TTransportException`?  I'm guessing only the connection context manager and the method call? 

Related questions:
- Replaying this block has no bad side effects if it fails at other points?
- Do you really need to re-allocate a new span on every retry or could you recycle the previously created on?

baseplate/context/thrift.py
spladug: `RetryPolicy.new(attempts=1)` is equivalent to "don't retry, just do it once". So I can understand what to document, what part is confusing there?

baseplate/context/thrift.py
spladug: The `with self.pool.connection() as prot` can raise `TTransportException` [here](https://github.com/reddit/baseplate/blob/retries/baseplate/thrift_pool.py#L72) if other requests are using all the pooled connections, or [here](https://github.com/reddit/baseplate/blob/retries/baseplate/thrift_pool.py#L100) if the pool exhausts its retry budget attempting to connect. The `method` can then raise `TTransportException` if the connection times out or otherwise breaks during its service call.

The only unsafe thing in here is `method` which might be a non-idempotent service method on the remote service. The goal of the `with` statement syntax for overriding the policy is that it forces the user to think about whether or not it's safe to retry the operation. We'd need to extend thrift somehow to indicate idempotency if we wanted this to be checked automatically.

I dithered on whether the span should live outside or inside the retry logic, but ended on it being within. The reason I chose this is that each `make_child` call generates a new span ID for the child span and this means that instead of sending the same span id to the remote service N times, we send a different one for each retry. This is all rather theoretical right now as we still don't have Zipkin running, but it seems to me that'd be a more useful representation to the remote service since it has no reason to care that the N calls it just got were part of the same retry.

baseplate/message_queue.py
```diff
@@ -76,16 +54,16 @@ def get(self, timeout=None):
             duration of the call.
 
         """
-        selector = _CumulativeTimeoutSelector(timeout)
-
-        while True:
+        for time_remaining in RetryPolicy.new(budget=timeout):
```
bsimpson63: this is pretty nice

baseplate/retry.py
```diff
@@ -0,0 +1,126 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import time
+
+
+class RetryPolicy(object):
+    """A policy for retrying operations.
+
+    Policies are meant to be used as an iterable::
```
bsimpson63: is the :: intentional?

spladug: Yeah, it's a reStructuredText shorthand that starts a block after the current paragraph.

tests/unit/retry_tests.py
```diff
@@ -0,0 +1,72 @@
...
+    TimeBudgetRetryPolicy,
+)
+
+from .. import mock
+
+
+class RetryPolicyTests(unittest.TestCase):
+    def test_maximum_attempts(self):
+        base_policy = mock.MagicMock()
+        base_policy.__iter__.return_value = itertools.repeat(1.2)
```
ckwang8128: Any particular reason for 1.2 here?

spladug: Nope. Just wanted to see a value make it through.

bsimpson63:
🐟

baseplate/retry.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.yield_attempts()
+
+    @staticmethod
+    def new(attempts=None, budget=None, backoff=None):
+        """Create a new retry policy with the given constraints.
+
+        :param int attempts: The maximum number of times the operation can be
+            attempted.
+        :param float budget: The maximum amount of time, in seconds, that the
+            local service will wait for the operation to succeed.
```
ckwang8128: It might be worth making this "The maximum amount of time, in seconds, that the local service will wait for the operation to succeed, **regardless of the number of attempts**". This might be interpreted as budget per attempt, which is common API behavior (although in this case `budget` is used instead of the more common `timeout` parameter I've seen to indicate that). However, if I'm reading the code correctly, `budget` is a hard cap.

In the event that only `attempts` is specified, the timeout per attempt is managed at in the connection pool definition?

spladug: > It might be worth making this "The maximum amount of time, in seconds, that the local service will wait for the operation to succeed, regardless of the number of attempts". This might be interpreted as budget per attempt, which is common API behavior (although in this case budget is used instead of the more common timeout parameter I've seen to indicate that). However, if I'm reading the code correctly, budget is a hard cap.

You're correct in your interpretation. Would `timeout` be a better name than `budget`?

> In the event that only attempts is specified, the timeout per attempt is managed at in the connection pool definition?

Yes. That's correct.

tests/unit/retry_tests.py
```diff
@@ -0,0 +1,72 @@
...
+
+        next(retries)
+        sleep.assert_called_with(.4)
+
+        next(retries)
+        sleep.assert_called_with(.8)
+
+        # the base policy's lower time remaining takes over here
+        next(retries)
+        sleep.assert_called_with(.9)
```
ckwang8128: As a followup to my other comment about RetryPolicy's attempts + budget interaction, having a test here to show the expected behavior would be good.

baseplate/context/thrift.py
ckwang8128: " it seems to me that'd be a more useful representation to the remote service since it has no reason to care that the N calls it just got were part of the same retry."

Agreed. This way explicitly highlights retries from the local service and the full end-to-end trace should still be maintained with `trace_id`.

ckwang8128:
:nail_care:

baseplate/context/thrift.py
KeyserSosa: Oh.  I read that as "1 retry".  Well that explains it then.  PEBKAC.  I'm content with the current behavior.

KeyserSosa:
:fish: Thanks for the answers!

tests/unit/retry_tests.py
spladug: added a test and found a bug in the process. thanks!

spladug:
:haircut:

baseplate/retry.py
ckwang8128: > Would `timeout` be a better name than `budget`?

discussed in person, `budget` is actually better I think.

tests/unit/retry_tests.py
ckwang8128: nice!

ckwang8128:
:fish: thanks, this is awesome!

KeyserSosa:
Free re-:fish: with added unit test.

spladug:
merged as d7cca431d7acc4e72c1351465434e696c8057191...29f44eeb71c5a2e4f1e14281a58ab65c7246b0f1

PR #32, opened 2016-04-19
Make time budgeted retries always execute at least once

spladug:
If a time budget of zero is given, this would previously not run the
retried code at all. This is awkward if we want to say "don't block" but
do want to try at least once.

This was breaking EventQueue which uses MessageQueue.put(..., timeout=0)
for non-blocking sends.

Do you think this messes up the concept of a total time budget too much? or is it
more surprising that the code wouldn't run at all?

:eyeglasses: @bsimpson63 @ckwang8128

### DIFF
--- a/baseplate/retry.py
+++ b/baseplate/retry.py
@@ -107,7 +107,7 @@ class TimeBudgetRetryPolicy(RetryPolicy):
             time_remaining = self.budget - elapsed
             if i > 0 and time_remaining <= 0:
                 break
-            yield time_remaining
+            yield max(0, time_remaining)
 
 
 class ExponentialBackoffRetryPolicy(RetryPolicy):
--- a/baseplate/retry.py
+++ b/baseplate/retry.py
@@ -102,10 +102,10 @@ class TimeBudgetRetryPolicy(RetryPolicy):
     def yield_attempts(self):
         start_time = time.time()
 
-        for _ in self.subpolicy:
+        for i, _ in enumerate(self.subpolicy):
             elapsed = time.time() - start_time
             time_remaining = self.budget - elapsed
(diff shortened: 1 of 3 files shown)


baseplate/retry.py
```diff
@@ -102,10 +102,10 @@ def __init__(self, policy, budget):
...
         start_time = time.time()
 
-        for _ in self.subpolicy:
+        for i, _ in enumerate(self.subpolicy):
             elapsed = time.time() - start_time
             time_remaining = self.budget - elapsed
-            if time_remaining <= 0:
+            if i > 0 and time_remaining <= 0:
                 break
             yield time_remaining
```
bsimpson63: Will anything weird happen as a result of this being a negative number?

bsimpson63:
I think it's more surprising that the code wouldn't run at all.

I don't think it's too weird for 0 to be a special case. I think it might make sense for it to be treated as the same as the no budget case (infinite retries).

baseplate/retry.py
spladug: Yup. Good catch. I'll have to rethink this. Closing. Thanks.

spladug:
:haircut: i don't think negative is actually gonna be an issue since the first execution will be usually almost instant between the two calls to `time.time` but either way i've added a `max(0, time_remaining)` to guard against the case. i tried a few alternatives and still feel this is probably right.

i agree that not executing at all is more surprising. i'm not sure that infinite retries would be a good idea with a time budget of zero though (and that'd specifically break the message_queue stuff)

bsimpson63:
🐟

spladug:
67bb8dc5d2f65170e3551fcfed14d3dff695ea96

PR #33, opened 2016-04-26
Redis pipelines + bonus

spladug:
:eyeglasses: @kjoconnor @dellis23

### DIFF
--- a/baseplate/context/redis.py
+++ b/baseplate/context/redis.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import redis
+import redis.client
 
 from . import ContextFactory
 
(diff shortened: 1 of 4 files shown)


baseplate/context/redis.py
```diff
@@ -4,55 +4,83 @@
...
-        raise NotImplementedError
+    def pipeline(self, name, transaction=True, shard_hint=None):
+        """Create a pipeline.
+
+        This returns an object on which you can call the standard redis
+        commands. Execution will be deferred until ``execute`` is called. This
+        is useful for saving round trips.
+
+        :param str name: The name to attach to diagnostics for this pipeline.
+        :param bool transaction: Whether or not the commands in the pipeline
```
kjoconnor: So will this work currently? I see `NotImplementedError`, but is that only if you're using `pipeline` directly?

kjoconnor:
EXCELLENT

💅

baseplate/context/redis.py
spladug: That other `transaction` method is separate. This will work.

spladug:
:haircut:

dellis23:
🐟  More diagnostics!  Yay!

baseplate/context/redis.py
```diff
@@ -4,55 +4,83 @@
...
+    """Redis connection that collects diagnostic information.
 
+    This connection acts like :py:class:`redis.StrictRedis` except that all
+    operations are automatically wrapped with diagnostic collection.
+
+    The interface is the same as that class except for the
+    :py:meth:`~baseplate.context.redis.MonitoredRedisConnection.pipeline`
+    method.
+
+    .. note:: Locks and pubsub are currently unsupported.
```
kjoconnor: According to your other note transactions are still unsupported right? Only in pipelines?

kjoconnor:
I am excite 🐟

Issue #7, opened 2015-12-18
Add dev-mode auto-restarter for servers


spladug:
This should watch for changes in any related source or config files and restart itself. Useful for quick development.

spladug:
Added in a1a53a516c70b0cd7c1b438bc957ae00f1587840.

PR #34, opened 2016-05-09
Add automatic reloader for development

spladug:
This is just a nice quality of life thing when developing.

Implements #7.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals
 
 import argparse
 import collections
+import fcntl
 import importlib
 import logging
 import signal
@@ -17,7 +18,7 @@ import socket
 import sys
 import traceback
 
-from . import einhorn
+from . import einhorn, reloader
 from .._compat import configparser
 from ..config import Endpoint
 
@@ -33,6 +34,8 @@ def parse_args(args):
 
     parser.add_argument("--debug", action="store_true", default=False,
         help="enable extra-verbose debug logging")
+    parser.add_argument("--reload", action="store_true", default=False,
+        help="restart the server when code changes (development only)")
     parser.add_argument("--app-name", default="main", metavar="NAME",
         help="name of app to load from config_file (default: main)")
     parser.add_argument("--server-name", default="main", metavar="NAME",
@@ -79,6 +82,10 @@ def make_listener(endpoint):
         return einhorn.get_socket()
     else:
         sock = socket.socket(endpoint.family, socket.SOCK_STREAM)
+
+        flags = fcntl.fcntl(sock.fileno(), fcntl.F_GETFD)
+        fcntl.fcntl(sock.fileno(), fcntl.F_SETFD, flags | fcntl.FD_CLOEXEC)
+
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         sock.bind(endpoint.address)
         sock.listen(128)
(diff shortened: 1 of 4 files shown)


baseplate/server/__init__.py
```diff
@@ -79,6 +82,10 @@ def make_listener(endpoint):
         return einhorn.get_socket()
     else:
         sock = socket.socket(endpoint.family, socket.SOCK_STREAM)
+
+        flags = fcntl.fcntl(sock.fileno(), fcntl.F_GETFD)
+        fcntl.fcntl(sock.fileno(), fcntl.F_SETFD, flags | fcntl.FD_CLOEXEC)
```
bsimpson63: what's going on here?

spladug: added comment. how's that?

baseplate/server/reloader.py
```diff
@@ -0,0 +1,65 @@
...
+    initial_mtimes = {}
+    while True:
+        for filename in _get_watched_files(extra_files):
+            try:
+                current_mtime = os.path.getmtime(filename)
+            except os.error:
+                continue
+
+            initial_mtimes.setdefault(filename, current_mtime)
+            if initial_mtimes[filename] < current_mtime:
```
bsimpson63: I think this is probably the right behavior but just checking--when a new file is created it won't trigger a reload because `current_mtime` will be set into `initial_mtimes`?

baseplate/server/__init__.py
bsimpson63: great, thanks

baseplate/server/reloader.py
```diff
@@ -0,0 +1,65 @@
...
+                logger.debug("Reloading, %s changed", filename)
+                os.execl(sys.executable, sys.executable, *sys.argv)
+
+        time.sleep(.25)
```
bsimpson63: no leading 0? how bold

spladug: fresh new taste

baseplate/server/reloader.py
spladug: Right, because "a new file is created" is actually more like "a new module was imported" which could happen normally from just hitting a new codepath that imports something new. Not necessarily indicative of anything changing.

bsimpson63:
🐟

baseplate/server/reloader.py
dellis23: Any reason not to use `inotify` here?  I've never worked with it before, but I'd assumed that was the defacto way of doing it rather than looping and syscalls.

spladug: It pulls in a new dependency that is not otherwise present (and one that isn't terribly gevent friendly at that). This is actually very similar code to that in Paste and Gunicorn, so it's not too crazy to do it with polling.

baseplate/server/reloader.py
```diff
@@ -0,0 +1,65 @@
...
+        for filename in _get_watched_files(extra_files):
+            try:
+                current_mtime = os.path.getmtime(filename)
+            except os.error:
+                continue
+
+            initial_mtimes.setdefault(filename, current_mtime)
+            if initial_mtimes[filename] < current_mtime:
+                logger.debug("Reloading, %s changed", filename)
+                os.execl(sys.executable, sys.executable, *sys.argv)
```
dellis23: How does this actually reload the code?  I've never heard of `execl`, but the docs make it looks like it's just re-launching the interpreter.

dellis23:
🐟  This'll make development a lot nicer!

baseplate/server/reloader.py
spladug: Exactly. `exec` and friends replace the current process by running another executable (e.g. your shell runs a program by forking then execing it). In this case, we replace ourselves with the same command line that started ourselves which is like dying and being restarted but in-place.

spladug:
Merged as a1a53a516c70b0cd7c1b438bc957ae00f1587840.

PR #36, opened 2016-05-22
Diagnostics support for SQLAlchemy and some refactoring

spladug:
This does a bit of API cleanup and then adds support for SQLAlchemy in the diagnostics framework. This automatically generates trace spans for each query executed against the database (meaning we get graphite metrics and eventually zipkin tracing). Additionally, we automatically add the trace and span ids to the query statement as a comment, e.g. `-- trace:13145633593015559345,span:12869606375677030388`. This is intended for use in slow query logs and the like.

This should help with the new services we talked about last week, since we'll be wanting to talk to Postgres in them.


dellis23:
🐟 Thanks for this!

spladug:
Merged as 8a0471b08b50c635239c3ad7e2616e9c3dd55f58...fdefda9ac9be41dc705e3212bf21b03c1ac7e178

PR #38, opened 2016-06-09
Fix negative time remaining in first run of TimeBudgetRetryPolicy

spladug:
The first execution of a zero budget retry will have a negative time
remaining because time passes between the start_time being captured and
the elapsed time being calculated, even if it seems like it's immediate.

This makes it so we do the "always execute at least once" aspect of the
policy up front rather than relying on the math working out for us.

This also adds a test that previously failed and now passes. :tada: 

This is prong two of the three pronged attack on the event publishing errors (IO-245).

:eyeglasses: @bsimpson63 @dellis23


dellis23:
🐟

bsimpson63:
🐟

spladug:
15330b435bb611fac1a66ae6a3eaa237ef52fac0

PR #42, opened 2016-06-16
Add Vagrantfile for running tests in development

dwick:
👓 @ckwang8128 

cc @spladug for visibility

Guess, check, google, repeat on puppet so feel free to be hard on me!

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,0 +1,64 @@
+# Installation of initial dependencies and baseplate project
+# in a development environment.
+# Please add additional dependencies and env setup code as needed.
+class baseplate {
+  $project_path = '/home/vagrant/baseplate'
+
+  Exec { path => [ '/usr/bin', '/usr/sbin', '/bin', '/usr/local/bin' ] }
+
+  # Add reddit package apt-repo
+  exec { 'reddit-repo-add':
+    command => 'sudo add-apt-repository ppa:reddit/ppa -y',
+    unless  => 'apt-cache policy | grep reddit',
+    notify  => Exec['update-apt'],
+  }
+
+  # Update apt
+  exec { 'update-apt':
+    command     => 'sudo apt-get update -q',
+    refreshonly => true,
+  }
+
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -0,0 +1,72 @@
...
+     'python3-posix-ipc',
+    ]:
+      ensure  => installed,
+      require => [
+        Exec['reddit-repo-add'], # Add the reddit ppa
+        Exec['update-apt'],      # The system update needs to run
+      ],
+  } ->
+
+  package {
```
spladug: Why have these guys come from pip?

manifests/init.pp
```diff
@@ -0,0 +1,72 @@
+# Installation of initial dependencies and baseplate project
+# in a development environment.
+# Please add additional dependencies and env setup code as needed.
+class baseplate {
+  $project_path = '/home/vagrant/baseplate'
+
+  group { 'puppet':
```
spladug: What's this for?

spladug:
:nail_care: just a couple tiny questions. looks great. would you mind running `puppet-lint` on the puppetfile for style please?

manifests/init.pp
ckwang8128: this was copied from the manifest I linked as an example in the cookiecutter template https://github.com/reddit/reddit-cookiecutter-pyramid/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/manifests/init.pp

I think that was erroneously copied from an example I used as I was learning how to write puppet manifests properly. it shouldn't be needed here (or there).

manifests/init.pp
ckwang8128: just checked, there are deb packages available for all of these for python2 and python3 so it should be safe to add them to the above stanza!

dwick: i was looking at the travis config for deps, wasn't sure why these were pip there either :)

dwick:
💇

spladug:
:fish:

PR #43, opened 2016-06-16
Interana

dwick:
👓 @spladug 

I also tackled a couple of other issues that became more apparent with the addition of a second payload section and brought test coverage up to 100% for this module. 🎉 

The actual payload still uses `interana_excluded` since that's what works right now, but if we want to change that in the future we can without causing any breaking changes to the API.  I think that's inline with what you wanted. //cc @jophuds


tests/unit/events/queue_tests.py
```diff
@@ -76,6 +76,18 @@ def test_serialize(self, time, uuid):
...
 
+    @mock.patch("uuid.uuid4")
+    @mock.patch("time.time")
+    def test_get_field(self, time, uuid):
+        time.return_value = 333
+        uuid.return_value = "1-2-3-4"
+
+        event = Event("topic", "type")
+        event.set_field("normal", "value1")
+        event.set_field("obfuscated", "value2")
```
spladug: should `obfuscate=True` be here?

baseplate/events/queue.py
```diff
@@ -74,11 +87,31 @@ def set_field(self, key, value, obfuscate=False):
         if value is None or value == "":
             return
 
-        if not obfuscate:
-            self.payload[key] = value
+        if obfuscate and high_cardinality:
```
spladug: Another API option that might be less confusing/error prone is to have a single parameter which takes an enum of "field type".  `def set_field(self, key, value, kind=NORMAL)`. Then we're explicit that the types are all mutually exclusive (instead of having some combinations of flags that can yell at the user) and we're more open to future extension if necessary.

We'd want to support `obfuscate` still as a deprecated param though, like https://github.com/reddit/baseplate/commit/6f0981bccebdf847b39211fd17016fd7a63edc58#diff-e23a2f0193db987b0509d66964675524R180

baseplate/events/queue.py
```diff
@@ -65,6 +75,9 @@ def set_field(self, key, value, obfuscate=False):
         :param bool obfuscate: Whether or not to put the field in the obfuscated
             section. This is used for sensitive info like IP addresses that must
             be treated with care.
+        :param bool high_cardinality: Whether or not to put the field in the
+            interana_excluded section.  This excludes the field from being
```
spladug: I'm kinda torn here. I really do prefer the high cardinality (or similar) naming rather than explicitly calling out interana, since that's very leaky of current tooling. But I also understand that we want to be obvious to users that this is how they avoid putting stuff into interana since that's what they're using. Maybe the high cardinality option should be described as "for fields that should not be indexed due to high cardinality (e.g. not used in Interana)"?

baseplate/events/queue.py
```diff
@@ -39,24 +39,34 @@ def __init__(self, topic, event_type, timestamp=None, id=None):
         self.id = id or uuid.uuid4()
         self.payload = {}
 
-    def get_field(self, key, obfuscated=False):
+    def get_field(self, key):
```
spladug: :+1: good call on this change.

baseplate/events/queue.py
```diff
@@ -39,24 +39,34 @@ def __init__(self, topic, event_type, timestamp=None, id=None):
...
 
         """
 
-        if not obfuscated:
-            payload = self.payload
-        else:
-            payload = self.payload.get("obfuscated_data", {})
-        return payload.get(key, None)
+        payload = self.payload
+        obfuscated_data = self.payload.get("obfuscated_data", {})
```
spladug: I'm wondering if I made a bad choice in mimicking the serialized form of the event while building it and keeping the obfuscated fields in the regular payload dict. As a crazy alternative, and assuming we have an enum of field types like mentioned down in `set`, what if payload became a dict from type -> fields, so `set` would just do `self.payloads_by_type[field_type][key] = value` (with sanity checking) and get would just loop through the payload types looking for the key in each dict. `serialize` would need a bit more work, but overall I think it might be a bit cleaner?

baseplate/events/queue.py
```diff
@@ -39,24 +39,34 @@ def __init__(self, topic, event_type, timestamp=None, id=None):
...
+        interana_excluded = self.payload.get("interana_excluded", {})
+
+        # Check main payload section, the obfuscated section, then
+        # the interana excluded section of the payload before
+        # defaulting to `None`.
+        return payload.get(
```
spladug: nit: I find this kinda hard to read, especially since `get` is never gonna be able to have more than 2 args, so diff-friendliness isn't super important. The above thing might help a bit, but yeah.

spladug:
nit: it looks like CI's failing because the spellchecker doesn't like "interana". If it's gonna stay in the docstring, can you add the word to `docs/words.txt` to appease it please?

Might want to capitalize too.

spladug:
:nail_care: thanks a bunch for working on this. very nice improvements.

.travis.yml
```diff
@@ -16,7 +16,7 @@ virtualenv:
 before_install:
     - sudo add-apt-repository -y ppa:reddit/ppa
     - sudo apt-get update -q
-    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc
+    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 python3-enum34
```
dwick: not sure if this is needed here.

dwick: unsure if this is needed

debian/control
```diff
@@ -9,6 +9,7 @@ Build-Depends: debhelper (>= 8),
                python3-all,
                python3-setuptools,
                python-nose, python3-nose, python-mock,
+               python-enum34 python3-enum34,
```
dwick: added this for support of python 3.4 enums in earlier versions: https://pypi.python.org/pypi/enum34

.travis.yml
dwick: not sure if this is needed

debian/control
dwick: added this for support of python 3.4 enums in earlier versions: https://pypi.python.org/pypi/enum34

baseplate/events/queue.py
```diff
@@ -21,8 +21,25 @@
...
+from enum import Enum
+
 from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from baseplate.message_queue import MessageQueue, TimedOutError
+from baseplate._utils import warn_deprecated
+
+
+class FieldType(Enum):
+    NORMAL = None
+    """
```
dwick: unsure if this is the way to do this

baseplate/events/queue.py
```diff
@@ -21,8 +21,25 @@
...
+    For fields containing sensitive information like IP addresses that must
```
dwick: is this the proper way to document an enum? the only example i could find was on [stackoverflow](http://stackoverflow.com/questions/20227051/how-to-document-a-module-constant-in-python)

dwick:
💇 put some questions in the fixups commit for things i was unsure about.

baseplate/events/queue.py
spladug: good question! what's the output look like when running `make docs`?

debian/control
spladug: :+1: cool. didn't know this was already in trusty too!

do we need a python3 version since it's in the stdlib?

baseplate/events/queue.py
```diff
@@ -21,8 +21,25 @@
...
 import uuid
 
+from enum import Enum
+
 from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from baseplate.message_queue import MessageQueue, TimedOutError
+from baseplate._utils import warn_deprecated
+
+
+class FieldType(Enum):
```
spladug: since this goes into a param called `kind`, maybe `FieldKind` would be more consistent?

baseplate/events/queue.py
spladug: oh, i think you'll also need to add an `autoclass` thing to docs/baseplate/events.rst to get this class into the docs

.travis.yml
spladug: can't hurt

debian/control
dwick: for versions less than 3.4 you do. does that apply?

baseplate/events/queue.py
```diff
@@ -65,6 +77,9 @@ def set_field(self, key, value, obfuscate=False):
         :param bool obfuscate: Whether or not to put the field in the obfuscated
```
spladug: let's just drop this field from the docs since it's deprecated.

debian/control
spladug: i think there are other places in baseplate assuming that py3 will be 3.4+, so i think it's ok to continue that here.

baseplate/events/queue.py
```diff
@@ -74,19 +89,34 @@ def set_field(self, key, value, obfuscate=False):
...
+        if obfuscate:
+            kind = FieldKind.OBFUSCATED
+            warn_deprecated("Passing obfuscate to set_field is deprecated in"
+                            " favor of passing a FieldKind value as kind.")
+
+        self.payload[key] = value
+        self.payload_types[key] = kind
 
     def serialize(self):
+        payload = {
```
spladug: would it be possible to only add these keys if there's a field that needs to be in them? i think data generally prefers minimalism here.

baseplate/events/queue.py
```diff
@@ -21,8 +21,28 @@
...
+from enum import Enum
+
 from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from baseplate.message_queue import MessageQueue, TimedOutError
+from baseplate._utils import warn_deprecated
+
+
+class FieldKind(Enum):
+    """Enum of field kinds."""
+    NORMAL = None
```
spladug: can you docstring this guy too please?

baseplate/events/queue.py
```diff
@@ -38,33 +58,28 @@ def __init__(self, topic, event_type, timestamp=None, id=None):
...
+    def set_field(self, key, value, obfuscate=False, kind=FieldKind.NORMAL):
         """Set the value for a field in the event.
 
         :param str key: The name of the field.
         :param value: The value to set the field to. Should be JSON
             serializable.
-        :param bool obfuscate: Whether or not to put the field in the obfuscated
-            section. This is used for sensitive info like IP addresses that must
-            be treated with care.
+        :param :py:class:`FieldKind` kind: The kind the field is. Used
```
spladug: the :py:class: is unnecessary here: `:param baseplate.events.queue.FieldKind kind: blah blah`

right now the docs render as

screenshot from 2016-06-17 15 21 09

baseplate/events/queue.py
```diff
@@ -74,19 +89,32 @@ def set_field(self, key, value, obfuscate=False):
...
+            warn_deprecated("Passing obfuscate to set_field is deprecated in"
+                            " favor of passing a FieldKind value as kind.")
+
+        self.payload[key] = value
+        self.payload_types[key] = kind
 
     def serialize(self):
+        payload = {}
+
+        for key, value in self.payload.iteritems():
```
spladug: `iteritems` doesn't exist in py3. `make tests` shows a failure here on the py3 version.

baseplate/events/queue.py
spladug: also something's going funky when the docs generate: it says `None` for each of them

spladug:
:nail_care: just for the last little remaining things. once you've got them fixed up, can you make sure the build passes please?

thanks again. looks great!

spladug:
Oh, also, can you make sure `enum34` gets added to `install_requires` in setup.py when in py2 please?

baseplate/events/queue.py
dwick: looks ok to me, am i looking at the wrong thing?

screen shot 2016-06-20 at 2 39 51 pm

dwick:
💇 build passes 🎉

spladug:
:fish: !!!

this looks fantastic. thanks for putting up with all the review! :)

dwick:
no worries, code is better for it.

PR #47, opened 2016-08-17
A few minor tweaks that have come out of snoo's day projects

spladug:
:eyeglasses: @dellis23 @pacejackson


setup.py
```diff
@@ -42,7 +42,14 @@
 
 setup(
     name="baseplate",
+    description="A library to build services on",
+    long_description=open("README.rst").read(),
+    author="Neil Williams",
```
dellis23: can we put my name here instead

pacejackson:
Neat, lgtm 🐟

setup.py
spladug: i actually don't really know what to put here. should it be `reddit inc`?

spladug: otherwise, yeah, we could put your name here, or maybe putting `Albert Einstein` would lend us some cred

enmatt:
Looks good.

siemiatj:
Good to merge

spladug:
Thanks for all the support, Dan's friends!

PR #48, opened 2016-10-03
OpenTracing standardization and exception catching fixes

spladug:
The core of this pull request is to rework the tracing / diagnostics APIs in baseplate to match the terminology and functionality of the [OpenTracing](http://opentracing.io) standard. This takes the form of:

* renaming `RootSpan` to `ServerSpan` since there's only one root span in the whole trace
* renaming the `annotate` method on spans `set_tag` to match the standard
* adding a `log` method to spans
* renaming `stop` to `finish` to match the standard
* adding some standard tags to spans in the thrift and pyramid integrations

This is mostly about standardizing and getting ready for actually capturing trace data (in Zipkin) as well as error info in Sentry.

Speaking of Sentry, the other part of this pull request is fixing the way we capture exceptions in spans in two ways: passing the full traceback information to the span observers to allow more introspection, and fixing the pyramid integration's capture of these exceptions.

Note: the Thrift integration doesn't properly capture exceptions at this point. I'm going to pass on fixing that in this pull request because I've got a pull request to switch thrift implementations to thriftpy coming soon (but I didn't want to make this pull request any more complicated).

### DIFF
--- a/baseplate/diagnostics/metrics.py
+++ b/baseplate/diagnostics/metrics.py
@@ -42,7 +42,7 @@ class MetricsSpanObserver(SpanObserver):
     def on_annotate(self, key, value):  # pragma: nocover
         pass
 
-    def on_stop(self, error):
+    def on_stop(self, exc_info):
         self.timer.stop()
 
 
@@ -51,6 +51,6 @@ class MetricsRootSpanObserver(MetricsSpanObserver):
         observer = MetricsSpanObserver(self.batch, "clients." + span.name)
         span.register(observer)
 
-    def on_stop(self, error):
-        super(MetricsRootSpanObserver, self).on_stop(error)
+    def on_stop(self, exc_info):
+        super(MetricsRootSpanObserver, self).on_stop(exc_info)
         self.batch.flush()
--- a/baseplate/diagnostics/metrics.py
+++ b/baseplate/diagnostics/metrics.py
@@ -25,10 +25,10 @@ class MetricsBaseplateObserver(BaseplateObserver):
     def __init__(self, client):
         self.client = client
 
-    def on_root_span_created(self, context, root_span):
+    def on_server_span_created(self, context, server_span):
         context.metrics = self.client.batch()
-        observer = MetricsRootSpanObserver(context.metrics, "server." + root_span.name)
-        root_span.register(observer)
+        observer = MetricsServerSpanObserver(context.metrics, "server." + server_span.name)
+        server_span.register(observer)
 
 
 class MetricsSpanObserver(SpanObserver):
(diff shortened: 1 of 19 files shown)


spladug:
:eyeglasses: @bsimpson63 @pacejackson @wting

spladug:
(Let me know if you want an in-person review for this; it looks pretty dense but it shouldn't be too bad)

pacejackson:
An in-person review would be nice.  It seems to me that 90 - 95% of this PR is _just_ changing naming conventions so it would be nice to see if there are any highlights or things to be careful of.

spladug:
Sure, I can set one up if the others want it too. In the meantime,

> see if there are any highlights or things to be careful of.

Should all be covered in the PR body. Nothing too surprising here, I hope.

baseplate/diagnostics/metrics.py
```diff
@@ -39,18 +39,15 @@ def __init__(self, batch, name):
     def on_start(self):
         self.timer.start()
 
-    def on_annotate(self, key, value):  # pragma: nocover
```
bsimpson63: what this originally unneeded, since it was duplicating `SpanObserver.on_annotate`?

spladug: yeah, i think i was just shutting a lint up.

baseplate/integration/pyramid.py
```diff
@@ -83,10 +86,12 @@ def _start_server_span(self, request, name, trace_info=None):
 
     # pylint: disable=no-self-use
     def _on_new_response(self, event):
-        if not event.request.matched_route:
+        request = event.request
+        if not request.matched_route:
             return
 
-        event.request.trace.finish()
+        request.trace.set_tag("http.status_code", request.response.status_code)
```
bsimpson63: cool stuff

bsimpson63:
🐟 looks good

baseplate/integration/pyramid.py
```diff
@@ -71,26 +91,23 @@ def _on_new_request(self, event):
...
-            return
-
-        event.request.trace.stop()
-
     def includeme(self, config):
-        config.add_subscriber(self._on_new_request, ContextFound)
-        config.add_subscriber(self._on_new_response, NewResponse)
+        config.add_subscriber(self._on_new_request, pyramid.events.ContextFound)
+        config.add_tween("baseplate.integration.pyramid._make_baseplate_tween",
+                         under=pyramid.tweens.EXCVIEW)
```
pacejackson: 🔕 
Purely educational question. I'm curious about when you use these different but similar parts of pyramid.  It seems like you want to use a tween here because you want to respond to something but not change it.  As you said in the commit message, a NewResponse subscription was too late in the flow to do what you wanted.  It also seems like you _could_ use add_view with context=Exception but you don't because I'm guessing it would mess up if a baseplate service (like the gateway) added an Exception view as well and adding a view is for when you actually want to change the response and in this case, you just want this to be a side effect.

pacejackson:
🐟  I think I'll need to make one small change in the gateway code to go along with this, lgtm though.

baseplate/integration/pyramid.py
spladug: Great question. Honestly, I wasn't aware of `context=Exception` on `add_view`. Let me dig into that and get back to you.

pacejackson: We use add_view for exceptions in the gateway, you can see it in reddit_service_gateway.__init__.py if you're curious.

spladug: Ah ok, exception views. I see. Yeah, we just want to silently observe the exception not claim any ability to render a useful response etc. So this lets us inject ourselves between the app and the pyramid layer that handles exceptions and then just watch them fly by without anyone knowing better.

pacejackson: Cool, thanks!

spladug:
19f2b6f052d7641521fc4b1e842a7df5f59167de...ac83b3b2c34695d21bcd415318f1a645671c9313

PR #50, opened 2016-10-04
thrift: Fix binary protocol responses from header protocol server

spladug:
THeaderTransport detects the protocol of the client and can accept raw
binary protocol (without headers). It's supposed to then use the same
protocol when responding. Because we were using a different instance of
THeaderTransport for the input and output streams, the detected format
was not being used when responding. This lead to binary protocol clients
hanging as they got invalid data in response. This resolves the issue by
using the same transport and protocol instances in both directions.

:eyeglasses: @dellis23 @krishnan-chandra

### DIFF
--- a/baseplate/server/thrift.py
+++ b/baseplate/server/thrift.py
@@ -40,22 +40,18 @@ class GeventServer(StreamServer):
         client = TSocket()
         client.setHandle(client_socket)
 
-        itrans = self.transport_factory.getTransport(client)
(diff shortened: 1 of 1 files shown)


dellis23:
🎺 💀

krishnan-chandra:
:fish: 

Once you've merged/built a deb I can test this out by switching our stuff over to the latest version of baseplate instead of our custom one. Thanks for the fix!

PR #53, opened 2016-10-13
Fix missed annotate() calls in OpenTracing rename

spladug:
Sorry about that. I should really get some tests on this module :(

:eyeglasses: @pacejackson


pacejackson:
🐟

PR #55, opened 2016-10-20
debian: Don't install test suite

spladug:
This fixes the find_packages() call so we don't install the test suite
as a top-level "tests" package which is pretty messy.


ckwang8128:
:fish:

dellis23:
🐟

Issue #54, opened 2016-10-13
Improve/get test coverage in context helpers


spladug:

spladug:
Fixed in https://github.com/reddit/baseplate/compare/970b287a61eb2b6d7824b67789a03073c61dac6d...f48c8cf070d881942f9ce30e560d99b9dd9e1dce

PR #57, opened 2016-10-28
Use kwargs in call to execute_async

pacejackson:
Looks like we are setting trace=True in all requests because the third argument in Session.execute_async is `trace` rather than `timeout`:

https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L2000

### DIFF
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -89,7 +89,11 @@ class CassandraSessionAdapter(object):
         span.start()
         # TODO: include custom payload
         span.set_tag("statement", query)
-        future = self.session.execute_async(query, parameters, timeout)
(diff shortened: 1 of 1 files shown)


pacejackson:
👓 @kjoconnor @spladug Neil, you better not look at this until you get back from vacation!

kjoconnor:
oops 🐟  good find!

ckwang8128:
:fish: any chance we're doing this elsewhere in the cassandra driver wrapper?

pacejackson:
No, there's only 2 places we use CassandraSessionAdapter.session, in that fix and we call `session.prepare` in CassandraSessionAdapter.prepare

spladug:
:fish: sorry :( really need to do #54

PR #60, opened 2016-11-18
Add basic Zipkin-compatible tracing observer

ckwang8128:
The diagnostics.tracing module adds span recording to Baseplate
via a set of observers for both a request's server span and any
child spans. The observer records spans with the basic annotations
required for simple Zipkin querying and visualization, namely the
client_send, client_receive, server_send, and server_receive
annotations.

TraceSpanObservers set `cs` and `cr` annotations for outgoing request spans, and TraceServerSpanObservers set `sr` and `ss` annotations for inbound requests the service handles.

This first pass does *not* include functionality for adding arbitrary span tags ('binary annotations' in Zipkin language) and is missing some optional endpoint information in the base annotations but both should be pretty straightforward with `on_set_tag` and are coming in followup work. This is mainly for feedback on the core implementation details, e.g. the use and structure of the observers.

:eyeglasses: @spladug  @pacejackson

### DIFF
--- a/baseplate/diagnostics/tracing.py
+++ b/baseplate/diagnostics/tracing.py
@@ -1,0 +1,276 @@
+"""Components for processing Baseplate spans for service request tracing."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import calendar
+
+import json
+import logging
+import socket
+import threading
+import time
+from datetime import datetime
+
+import requests
+
+from .._compat import queue
+from ..core import BaseplateObserver, SpanObserver, TraceInfo
+
+
+logger = logging.getLogger(__name__)
+
+
+# Span annotation types
+ANNOTATIONS = {
+    'CLIENT_SEND': 'cs',
+    'CLIENT_RECEIVE': 'cr',
+    'SERVER_SEND': 'ss',
+    'SERVER_RECEIVE': 'sr',
+}
+
+
+def current_epoch_microseconds():
+    """Return current UTC time since epoch in microseconds."""
+    epoch_ts = datetime.utcfromtimestamp(0)
+    return int((datetime.utcnow() - epoch_ts).
+               total_seconds() * 1000 * 1000)
+
+
+class TraceBaseplateObserver(BaseplateObserver):
+    """Distributed tracing observer.
+
+    This observer handles Zipkin-compatible distributed tracing
+    instrumentation for both inbound and outbound requests.
+    Baseplate span-specific tracing observers (TraceSpanObserver
+    and TraceServerSpanObserver) are registered for tracking,
+    serializing, and recording span data.
+
+    :param str service_name: The name for the service this observer
+        is registered to.
+    :param str tracing_endpoint: A 'hostname:port' destination to record
+        span data.
+    :param int max_conns: pool size for remote recorder connections.
+    """
+    def __init__(self, service_name, tracing_endpoint=None,
+                 max_conns=100, max_span_queue=10000):
+        self.service_name = service_name
+
+        if tracing_endpoint:
+            self.recorder = RemoteRecorder(tracing_endpoint,
+                                           max_conns=max_conns,
+                                           max_span_queue=max_span_queue)
+        else:
+            self.recorder = NullRecorder()
+
+    def on_server_span_created(self, context, server_span):
+        observer = TraceServerSpanObserver(self.service_name,
+                                           server_span,
+                                           self.recorder)
+        server_span.register(observer)
+
+
+class TraceSpanObserver(SpanObserver):
+    """Span recording observer for outgoing request child spans.
+
+    This observer implements the client-side span portion of a
+    Zipkin request trace.
+    """
+    def __init__(self, service_name, span, recorder):
+        self.service_name = service_name
+        self.recorder = recorder
+        self.span = span
+        self.start = None
+        self.end = None
+        self.elapsed = None
+        self.binary_annotations = []
+        super(TraceSpanObserver, self).__init__()
+
+    def on_start(self):
+        self.start = current_epoch_microseconds()
+        self.client_send = self.start
+
+    def on_finish(self, exc_info):
+        self.end = current_epoch_microseconds()
+        self.elapsed = self.end - self.start
+        self.record()
+
+    def _create_time_annotation(self, annotation_type, timestamp):
+        """Create Zipkin-compatible Annotation for a span.
+
+        This should be used for generating span annotations with a time component,
+        e.g. the core "cs", "sr", "ss", and "sr" Zipkin Annotations
+        """
+        endpoint_info = {
+            'serviceName': self.service_name,
+            'ipv4': socket.gethostbyname(socket.gethostname()),
+        }
+        return {
+            'endpoint': endpoint_info,
+            'timestamp': timestamp,
+            'value': annotation_type,
+        }
+
+    def _create_binary_annotation(self, ):
+        """Create Zipkin-compatible BinaryAnnotation for a span.
+
+        This should be used for generating span annotations that
+        do not have a time component, e.g. URI, arbitrary request tags
+        """
+        # TODO: Implement.
+        raise NotImplementedError
+
+    def _to_span_obj(self, annotations, binary_annotations):
+        span = {
+            "traceId": self.span.trace_id,
+            "name": self.span.name,
+            "id": self.span.id,
+            "timestamp": self.start,
+            "duration": self.elapsed,
+            "annotations": annotations,
+            "binary_annotations": binary_annotations,
+        }
+
+        span['parentId'] = self.span.parent_id or 0
+        return span
+
+    def _serialize(self):
+        """Serialize span information into Zipkin-accepted format."""
+        annotations = []
+
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['CLIENT_RECEIVE'],
+                self.start,
+            )
+        )
+
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['CLIENT_SEND'],
+                self.end,
+            )
+        )
+
+        return self._to_span_obj(annotations, self.binary_annotations)
+
+    def record(self):
+        """Record serialized span."""
+        serialized = self._serialize()
+        self.recorder.send(serialized)
+
+
+class TraceServerSpanObserver(TraceSpanObserver):
+    """Span recording observer for incoming request spans.
+
+    This observer implements the server-side span portion of a
+    Zipkin request trace
+    """
+
+    def __init__(self, service_name, span, recorder):
+        self.service_name = service_name
+        self.span = span
+        self.recorder = recorder
+        super(TraceServerSpanObserver, self).__init__(service_name,
+                                                      span,
+                                                      recorder)
+
+    def on_start(self):
+        self.start = current_epoch_microseconds()
+
+    def on_child_span_created(self, child_span):
+        child_span_observer = TraceSpanObserver(self.recorder)
+        child_span.register(TraceSpanObserver(self.recorder))
+
+    def on_finish(self, exc_info):
+        self.end = current_epoch_microseconds()
+        self.elapsed = self.end - self.start
+        self.record()
+
+    def _serialize(self):
+        """Serialize span information into Zipkin-accepted format."""
+        annotations = []
+
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['SERVER_RECEIVE'],
+                self.start,
+            )
+        )
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['SERVER_SEND'],
+                self.end,
+            )
+        )
+
+        return self._to_span_obj(annotations, [])
+
+
+class NullRecorder(object):
+    def __init__(self):
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+    def send(self, span):
+        """Write a set of spans to debug log."""
+        self.logger.debug("Span recording for trace_id=%s: %s",
+                          span['traceId'],
+                          span)
+
+
+class RemoteRecorder(object):
+    """Interface for recording spans to a remote collector.
+
+    The RemoteRecorder adds spans to an in-memory Queue for a background
+    thread worker to process. It currently does not shut down gracefully -
+    in the event of parent process exit, any remaining spans will be discarded.
+    """
+    def __init__(self, endpoint, max_conns, max_queue_size=10000):
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+        adapter = requests.adapters.HTTPAdapter(pool_connections=max_conns,
+                                                pool_maxsize=max_conns)
+        self.session = requests.Session()
+        self.session.mount('http://', adapter)
+        self.endpoint = "http://%s/api/v1/spans" % endpoint
+
+        self.span_queue = queue.Queue(maxsize=max_queue_size)
+        self.flush_worker = None
+        self.flush_worker = threading.Thread(target=self._flush_spans)
+        self.flush_worker.name = "span remote recorder"
+        self.flush_worker.daemon = True
+        self.flush_worker.start()
+
+    def _flush_spans(self):
+        """Send a set of spans to remote collector.
+
+        This reads a batch of at most 10 spans off the recorder queue
+        and sends them to a remote recording endpoint. If the queue
+        empties while being processed before reaching 10 spans, we flush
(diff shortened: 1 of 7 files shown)


baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
...
+ANNOTATIONS = {
+    'CLIENT_SEND': 'cs',
+    'CLIENT_RECEIVE': 'cr',
+    'SERVER_SEND': 'ss',
+    'SERVER_RECEIVE': 'sr',
+}
+
+
+def current_epoch_microseconds():
+    """Returns current UTC time since epoch in microseconds."""
```
spladug: nit: pep-257, should be "Return"

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
...
+    epoch_ts = datetime.utcfromtimestamp(0).replace(tzinfo=pytz.utc)
+    return int((datetime.utcnow().replace(tzinfo=pytz.utc) - epoch_ts).
+               total_seconds() * 1000 * 1000)
+
+
+class TraceBaseplateObserver(BaseplateObserver):
```
spladug: it'd be good to get a docstring on here and register the observer in the docs: https://raw.githubusercontent.com/reddit/baseplate/master/docs/baseplate/diagnostics/index.rst

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
+from __future__ import absolute_import
```
spladug: module docstring please!

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def on_finish(self, exc_info):
+        self.end = current_epoch_microseconds()
+        self.elapsed = self.end - self.start
+        self.record()
+
+    def _create_annotation(self, annotation_type, timestamp):
+        """Create Zipkin-compatible annotation for a span.
+
+        This is limited to creating protocol-specific annotations.
```
spladug: what's a protocol-specific annotation?

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        logger.debug(span)
+
+
+class RemoteRecorder(object):
+    def __init__(self, endpoint):
+        self.session = requests.Session()
+        self.endpoint = "http://%s/api/v1/spans" % endpoint
+
+    def send(self, span):
```
spladug: am i understanding right that this will be called directly in the server loop by the server span being stopped? would it be possible to pull this out of the request loop and into its own greenlet/daemon fed by queue so zipkin never affects production?

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class NullRecorder(object):
+    def __init__(self):
+        pass
+
+    def send(self, span):
+        """Write a set of spans to debug log.
+        """
+        logger.debug(span)
```
spladug: might be good to prefix the log message somehow

spladug:
:nail_care: looks awesome! so excited to see this go out

baseplate/diagnostics/tracing.py
ckwang8128: fixed!
...also do you have pep memorized?

baseplate/diagnostics/tracing.py
ckwang8128: I was referencing the core Zipkin annotations here: http://zipkin.io/pages/instrumenting.html

Upon closer reading, the format I'm actually targeting is "any annotations with a timestamp component", I wanted to make sure developers didn't look to this for creating binary annotations and only used this for core supported annotations. Making this a bit more clear.

baseplate/diagnostics/tracing.py
ckwang8128: >am i understanding right that this will be called directly in the server loop by the server span being stopped? 
it'll be called as the last step of `on_finish`; `on_finish` in both observers makes a call to the associated recorder's `record`, which would make a remote request in the case of the `RemoteRecorder` 

I looked to how the metrics collecting system worked for this implementation, I can definitely make a more explicit pool for this; I think I erroneously assumed observer processing happened out of band of the spans they register to

baseplate/diagnostics/tracing.py
ckwang8128: Added some identifying information

baseplate/diagnostics/tracing.py
ckwang8128: done!

baseplate/diagnostics/tracing.py
ckwang8128: done

baseplate/diagnostics/tracing.py
spladug: haha, no, just a stickler for the consistency-enforcing conventions :)

(there's a pep257 tool you can run to check docstrings)

baseplate/diagnostics/tracing.py
spladug: ah, yeah, because the statsd thing happens inband. it might be worthwhile to change that, but it's less of a problem because it's udp and so there won't be much wait; mostly fire-and-forget except when dns goes wrong.

ckwang8128:
:haircut: 

CI is currently failing because we don't have a gevent package for python3 which is causing import errors. Any objection to upgrading to gevent 1.1.2 which includes a python3 built? We could also wrap the async execution call and choose between gevent & python3's asyncio, or just note that this is only currently compatible with python2 baseplate services at the moment.

pacejackson:
@ckwang8128 I have a backport of gevent 1.1.0 here https://launchpad.net/~andrew.boyle.reddit/+archive/ubuntu/testppa

We can copy that into the reddit PPA if we want.

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    'SERVER_SEND': 'ss',
+    'SERVER_RECEIVE': 'sr',
+}
+
+
+def current_epoch_microseconds():
+    """Return current UTC time since epoch in microseconds."""
+    epoch_ts = datetime.utcfromtimestamp(0)
+    return int((datetime.utcnow() - epoch_ts).
+               total_seconds() * 1000 * 1000)
```
pacejackson: I have a collection of functions in the gateway that do things like this, I wonder if it would be better to collect them into a util in baseplate?  They look like this:

```python
EPOCH = datetime(1970, 1, 1, tzinfo=pytz.UTC)


def utc_now():
    return datetime.now(pytz.UTC)


def epoch_timestamp(dt):
    """Returns the number of seconds from the epoch to date."""
    return (dt - EPOCH).total_seconds()


def to_epoch_milliseconds(dt):
    """Returns the number of milliseconds from the epoch to date."""
    return int(math.floor(epoch_timestamp(dt) * 1000.))


def from_epoch_milliseconds(ms):
    """Convert milliseconds from the epoch to UTC datetime.

    :param int ms: milliseconds since the epoch
    :rtype: :py:class:`datetime.datetime`
    """
    seconds = int(ms / 1000.)
    microseconds = (ms - 1000 * seconds) * 1000.
    return EPOCH + timedelta(seconds=seconds, microseconds=microseconds)
```

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+        e.g. the core "cs", "sr", "ss", and "sr" Zipkin Annotations
+        """
+        endpoint_info = {
+            'serviceName': self.service_name,
+            'ipv4': socket.gethostbyname(socket.gethostname()),
+        }
+        return {
+            'endpoint': endpoint_info,
+            'timestamp': timestamp,
+            'value': annotation_type
```
pacejackson: nit: trailing comma

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+        span['parentId'] = self.span.parent_id or 0
+        return span
+
+    def _serialize(self):
+        """Serialize span information into Zipkin-accepted format."""
+        annotations = []
+
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['CLIENT_RECEIVE'],
```
pacejackson: 🔕 I've found that it is nice to use something like the Storage class in r2 for this so you can write these sort of constanst like `ANNOTATIONS.CLIENT_RECEIVE`.

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    def on_finish(self, exc_info):
+        self.end = current_epoch_microseconds()
+        self.elapsed = self.end - self.start
+        self.record()
+
+    def _serialize(self):
+        """Serialize span information into Zipkin-accepted format."""
+        annotations = []
+
+        annotations.append(self._create_time_annotation(ANNOTATIONS['SERVER_RECEIVE'],
```
pacejackson: I have a feeling this might be a pep8 violation.

spladug:
re: gevent, if you use the stdlib `threading` module it'll be monkeypatched when running in gevent and work as a normal thread when not. that should resolve the issues with running tests.

i'd like to get gevent 1.1 going so we can get py3 services out, but i'm not sure how safe/mature it is at this point.

pacejackson:
fwiw, Wily and above use gevent 1.1 as the default apt install.

```
>>> import gevent
>>> gevent.__version__
'1.1b1'
```

pacejackson:
💅 Seems good overall, I noticed several formatting issues, trailing comma's mostly.  Could you give it a pass over for those sort of things?

baseplate/diagnostics/tracing.py
ckwang8128: thanks!

baseplate/diagnostics/tracing.py
ckwang8128: This is a good idea, thanks for the additional functions! I'll add a followup PR for a time-utility library so we can provide consistency across baseplate services.

ckwang8128:
thanks for the `threading` tip, I forgot about the implicit monkeypatching. also fixed commas + formatting! 

:haircut:

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,232 @@
...
+class TraceBaseplateObserver(BaseplateObserver):
+    """Distributed tracing observer.
+
+    This observer handles Zipkin-compatible distributed tracing
+    instrumentation for both inbound and outbound requests.
+    Baseplate span-specific tracing observers (TraceSpanObserver
+    and TraceServerSpanObserver) are registered for tracking,
+    serializing, and recording span data.
+
+    :param string service_name: The name for the service this observer
```
spladug: nit: `str` is the type in python, not `string`

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,232 @@
...
+                self.end
+            )
+        )
+
+        return self._to_span_obj(annotations, self.binary_annotations)
+
+    def record(self):
+        """Record serialized span."""
+        serialized = self._serialize()
+        t = threading.Thread(target=self.recorder.send,
```
spladug: Cool, OK, so this'll fire off a thread/greenlet for each request to zipkin, right? I was originally thinking more along the lines of a single worker thread that reads from a `queue.Queue` and sends to zipking and then the observers would just append to the queue. What are the pros/cons of these two approaches?

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,232 @@
...
+            'value': annotation_type,
+        }
+
+    def _create_binary_annotation(self, ):
+        """Create Zipkin-compatible BinaryAnnotation for a span.
+
+        This should be used for generating span annotations that
+        do not have a time component, e.g. URI, arbitrary request tags
+        """
+        # TODO: Implement.
```
spladug: could just `raise NotImplementedError`

spladug:
:nail_care: just to discuss the threading approach.

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,232 @@
...
+            self._create_time_annotation(
+                ANNOTATIONS['CLIENT_RECEIVE'],
+                self.start
+            )
+        )
+
+        annotations.append(
+            self._create_time_annotation(
+                ANNOTATIONS['CLIENT_SEND'],
+                self.end
```
pacejackson: nit: still some missing commas

baseplate/diagnostics/tracing.py
ckwang8128: Having an intermediary queue + worker adds an extra level of complexity to manage. I'm working under the assumption that greenlets will be the default here so that this will remain relatively inexpensive under high request load, and that this is a simple enough operation that greenlets will not be long-lived. 

That being said, it might be safer from a resource perspective to use a queue + worker as system resources will remain constant under higher load, aside from the queue memory impact due to back up. 

There are a couple of zipkin implementations on both sides of this but after thinking a bit more I'm going to err on the side of safety for this first pass.

baseplate/diagnostics/tracing.py
ckwang8128: should I be adding trailing commas to arg list in a method call?

pacejackson: Yea, if you break it into multiple lines.  Once you do that you want to make it so if you add an argument, the diff for what was previously the last argument is not affected.  The case where this does not apply is if you are actually passing `**kwarg` in which case you can't add a trailing comma (and it has to be the last parameter anyways).

ckwang8128: gotcha, okay!

ckwang8128:
:haircut: 

I added a worker thread and queue for remote recording of spans. the worker thread operates as a daemon thread, so it won't gracefully shut down. There doesn't appear to be a model yet for triggering Observer behaviors in the process of handling signals. I was thinking about a solution where Baseplate observers could have signal handlers and propagate to any resources as part of shutdown. In this case, we could set a flag in the remote recorder which could terminate the queue process worker gracefully. Thoughts?

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,277 @@
...
+import calendar
+
+import json
+import logging
+import socket
+import threading
+import time
+try:
+    import queue
+except ImportError:
```
spladug: you can import this as `from .._compat import queue`

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,277 @@
...
+        adapter = requests.adapters.HTTPAdapter(pool_connections=max_conns,
+                                                pool_maxsize=max_conns)
+        self.session = requests.Session()
+        self.session.mount('http://', adapter)
+        self.endpoint = "http://%s/api/v1/spans" % endpoint
+
+        self.span_queue = queue.Queue(maxsize=max_queue_size)
+        self.flush_worker = None
+        self.flush_worker = threading.Thread(target=self._flush_spans)
+        self.flush_worker.name = "span remote recorder"
```
spladug: nice!

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,277 @@
...
+            try:
+                while len(spans) < 10:
+                    spans.append(self.span_queue.get(False, timeout=.1))
+            except queue.Empty:
+                pass
+            finally:
+                if spans:
+                    recording_req = requests.Request(
+                        'POST', self.endpoint,
+                        data=json.dumps(spans),
```
spladug: :no_bell: there's also a `json` kwarg that does encoding for you. http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests

baseplate/diagnostics/tracing.py
```diff
@@ -0,0 +1,277 @@
...
+                        headers={
+                            'Content-Type': 'application/json',
+                        }).prepare()
+                    self.session.send(recording_req, timeout=2)
+
+    def send(self, span):
+        try:
+            self.span_queue.put_nowait(span)
+        except Exception as e:
+            self.logger.error("Failed adding span to recording queue: %s" % e)
```
spladug: nit: the thing with the logging parameters again :)

spladug:
:fish: modulo queue import and logging parameter. looks great! super excited.

baseplate/diagnostics/tracing.py
ckwang8128: sweet, changed

pacejackson:
🐟 
Awesome!  Although the CR was less fun without random people popping in...

PR #63, opened 2016-12-15
Add binary annotations and span sampling + flags

ckwang8128:
This adds two features to Baseplate span handling, split across two commits:

1. Add binary annotation generation for on_set_tag events - This is relatively straightforward.

2. Add passing of sampling information and flags in trace information across spans - This adds handling of two more headers for trace information propagation: 
   * "Sampled": If this value is present, it communicates whether or not a request has already been deemed to be sampled or not upstream. If not present, we calculate it at the handling service based on a defined sampling rate. For requests that are not sampled, either by our choice or the upstream-defined option, we do not remotely record the span in baseplate via the observer. 
   * "Flags": This is a 64-bit integer used as a set of bitflags for feature flagging across a distributed tracing system. Currently the only  defined use is for the setting of a "Debug" flag, which allows a request to bypass sampling and "force" tracing. 

Because these pieces of information need to be passed downstream to child spans and processed at request-handling/span creation by way of header parsing, it made the most sense to add this functionality in the TraceInfo and Span objects themselves (rather than try to isolate in the tracing observer). 

More tests are incoming, but wanted to get immediate feedback on implementation. 

:eyeglasses: @spladug

### DIFF
--- a/baseplate/context/thrift.py
+++ b/baseplate/context/thrift.py
@@ -98,7 +98,11 @@ class PooledClientProxy(object):
                         prot.trans.set_header("Trace", str(span.trace_id))
                         prot.trans.set_header("Parent", str(span.parent_id))
                         prot.trans.set_header("Span", str(span.id))
-
+                        if span.sampled is not None:
+                            sampled = "1" if span.sampled else "0"
+                            prot.trans.set_header("Sampled", sampled)
+                        if span.flags:
+                            prot.trans.set_header("Flags", str(span.flags))
                         client = self.client_cls(prot)
                         method = getattr(client, name)
                         return method(*args, **kwargs)
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -63,7 +63,8 @@ class ServerSpanObserver(SpanObserver):
         pass
 
 
-_TraceInfo = collections.namedtuple("_TraceInfo", "trace_id parent_id span_id")
+_TraceInfo = collections.namedtuple("_TraceInfo",
+                                    "trace_id parent_id span_id sampled flags")
 
 
 class TraceInfo(_TraceInfo):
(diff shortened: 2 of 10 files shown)


baseplate/context/thrift.py
```diff
@@ -98,7 +98,10 @@ def _call_thrift_method(self, name, *args, **kwargs):
                         prot.trans.set_header("Trace", str(span.trace_id))
                         prot.trans.set_header("Parent", str(span.parent_id))
                         prot.trans.set_header("Span", str(span.id))
-
+                        if span.sampled:
+                            prot.trans.set_header("Sampled")
```
spladug: no value for these?

baseplate/core.py
```diff
@@ -83,15 +84,18 @@ def new(cls):
...
 
     @classmethod
-    def from_upstream(cls, trace_id, parent_id, span_id):
+    def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
         """Build a TraceInfo from individual headers.
 
         :param int trace_id: The ID of the trace.
         :param int parent_id: The ID of the parent span.
         :param int span_id: The ID of this span within the tree.
+        :param str sampled: Boolean flag of value "0" or "1" to determine request sampling.
```
spladug: can this be an actual true/false boolean in python rather than being stringly typed?

baseplate/core.py
```diff
@@ -105,7 +109,15 @@ def from_upstream(cls, trace_id, parent_id, span_id):
         if parent_id is None or not 0 <= parent_id < 2**64:
             raise ValueError("invalid parent_id")
 
-        return cls(trace_id, parent_id, span_id)
+        if sampled is not None and (sampled not in ["0", "1"]):
+            raise ValueError("invalid sampled header")
+
+        if flags is not None:
+            flags = int(flags)
```
spladug: also good to require the caller to make this an int. worth doing an enum for values and checking against that? or shall we let things fly by that we don't understand for better compatibility?

baseplate/diagnostics/tracing.py
```diff
@@ -81,12 +91,26 @@ def __init__(self, service_name,
                 batch_wait_interval=span_batch_interval,
             )
 
+    def force_sampling(self, span):
+        return span.flags and (span.flags & FLAGS['DEBUG'])
+
+    def should_sample(self, span):
+        if span.sampled is None:
+            return (random.uniform(0, 1) < self.sample_rate or
```
spladug: just curious: how does `random.uniform(0, 1)` differ from `random.random()`?

baseplate/context/thrift.py
ckwang8128: see this is why I need to write tests

baseplate/diagnostics/tracing.py
ckwang8128: `uniform(0,1)` gives you a value in `[0,1]` whereas `random.random()` gives you a value in `[0, 1)` ..... which is what I actually want here since I do a `<` check. Fixing!

baseplate/core.py
ckwang8128: done. re: values, I think letting things fly by is better for flexibility + compatibility

ckwang8128:
thanks for the quick feedback! updated using standard types while working with trace info in baseplate, and added tests 

:haircut:

docs/words.txt
```diff
@@ -8,12 +8,16 @@ configurables
 Cookiecutter
 crypto
 Einhorn
+Enum
```
spladug: thanks for fixing the sorting ;)

ckwang8128:
woo the system worked! bit again by byte conversion. :haircut:

PR #65, opened 2016-12-30
Synapse file_output integration, config.File, and some cleanup

spladug:
(definitely go commit-by-commit on this one. It's not so bad, but the overall diff is unnecessarily daunting)

The core of this pull request is a module that can read and understand synapse-generated JSON service inventories. This is needed for the upcoming websocket auto-discovery stuff (IO-1056) and potentially down the line for things like Cassandra clients.

I intend to replace `r2.lib.utils.weighted_lottery` with the implementation here to further standardize.

The `config.File` parser came up in both authn and thing services, so it felt like a good time to get it in here.

While doing this work, I ran into some issues with tests and some cleanliness issues that pyflakes in my editor caught but `make lint` (and therefore CI) didn't; so I've cleaned those up and switched that lint step over to `flake8` for more thoroughness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ virtualenv:
 before_install:
     - sudo add-apt-repository -y ppa:reddit/ppa
     - sudo apt-get update -q
-    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 pep8 pylint python-redis python-gevent
+    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 python-flake8 pylint python-redis python-gevent
 
 install:
     - pip install -e ".[pyramid]"
--- a/baseplate/config.py
+++ b/baseplate/config.py
@@ -21,8 +21,14 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
     from baseplate import config
     from baseplate._compat import configparser
+    from tempfile import NamedTemporaryFile
     config_parser = configparser.ConfigParser()
     config_parser.readfp(open("docs/config_example.ini"))
+
+    tempfile = NamedTemporaryFile()
+    tempfile.write("cool")
+    tempfile.flush()
+    config_parser.set("app:main", "some_file", tempfile.name)
 
 .. doctest::
 
@@ -39,6 +45,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     ...             "deep": config.Timespan,
     ...         },
     ...     },
+    ...     "some_file": config.File(mode="r"),
     ...     "optional": config.Optional(config.Integer, default=9001),
     ... })
 
@@ -50,6 +57,14 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
     >>> print(cfg.nested.really.deep)
     0:00:03
+
+    >>> cfg.some_file.read()
+    'cool'
+    >>> cfg.some_file.close()
+
+.. testcleanup::
+
+    tempfile.close()
 
 """
 from __future__ import absolute_import
@@ -165,6 +180,26 @@ def Base64(text):
         return base64.b64decode(text)
     except TypeError as exc:
         raise ValueError(*exc.args)
+
+
+def File(mode="r"):
+    """A path to a file.
+
+    This takes a path to a file and returns an open file object, like
+    returned by :py:func:`open`.
+
+    :param str mode: an optional string that specifies the mode in
+        which the file is opened.
+
+
+
+    """
+    def open_file(text):
+        try:
+            return open(text, mode=mode)
+        except IOError:
+            raise ValueError("could not open file: %s" % text)
+    return open_file
 
 
 def Timespan(text):
--- a/baseplate/diagnostics/tracing.py
+++ b/baseplate/diagnostics/tracing.py
@@ -4,8 +4,6 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import calendar
-
 import json
 import logging
 import random
@@ -17,7 +15,7 @@ from datetime import datetime
 import requests
 
 from .._compat import queue
-from ..core import BaseplateObserver, SpanObserver, TraceInfo
+from ..core import BaseplateObserver, SpanObserver
 
 
 logger = logging.getLogger(__name__)
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -1,0 +1,64 @@
+"""Extensions to the standard library `random` module."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import random
+
+
+class WeightedLottery(object):
+    """A lottery where items can have different chances of selection.
+
+    Items will be picked with chance proportional to their weight relative to
+    the sum of all weights, so the higher the weight, the higher the chance of
+    being picked.
+
+    :param items: A sequence of items to choose from.
+    :param weight_key: A function that takes an
+        item in ``items`` and returns a non-negative integer
+        weight for that item.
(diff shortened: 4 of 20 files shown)


spladug:
:eyeglasses: @ckwang8128 @dellis23

baseplate/service_discovery.py
```diff
@@ -0,0 +1,135 @@
...
+
+        :rtype: :py:class:`Backend`
+        :raises: :py:exc:`NoBackendsAvailableError` if the inventory
+            has no available endpoints.
+
+        """
+        backends = self.get_backends()
+        if not backends:
+            raise NoBackendsAvailableError
+        return weighted_choice(backends, lambda b: b.weight)
```
ckwang8128: This is going to be an O(backends) operation every time its called; In a world where there are few backends and another service doing actual routing (Synapse + haproxy) this isn't an issue, but I'm concerned about this for a use-case like websocket URL generation. I think the case where `get_backend` is called often on a large set of backends might call for pre-calcing CDF + binary search but maybe that's premature optimization (no crazy large clusters yet). Perhaps the answer is to just be prudent with this usage and maybe add a small note on efficiency of this operation.

tests/unit/config_tests.py
```diff
@@ -94,6 +95,31 @@ def test_valid(self):
...
+    def test_read_file(self):
+        file_opener = config.File()
+        the_file = file_opener(self.tempfile.name)
+        self.assertEqual(the_file.read(), "test")
+
+    def test_write_file(self):
+        file_opener = config.File(mode="w")
+        the_file = file_opener(self.tempfile.name)
+        the_file.write("cool")
+
```
ckwang8128: A followup read would also verify the file was in fact written to

baseplate/diagnostics/tracing.py
```diff
@@ -17,7 +15,7 @@
 import requests
 
 from .._compat import queue
-from ..core import BaseplateObserver, SpanObserver, TraceInfo
+from ..core import BaseplateObserver, SpanObserver
```
ckwang8128: Thanks for fixing ^_^;

ckwang8128:
:nail_care: looks great! would like thoughts on `get_backend` usage and efficiency and a small suggestion for the file writing test.

tests/unit/config_tests.py
spladug: added

spladug:
:haircut_man: how's that new interface look? should also be a little faster since we cache the initial setup, but most importantly it leaves us open interface-wise to optimizing down the line.

ckwang8128:
great! good call on the interface change, I like it a lot. :fish:

.travis.yml
```diff
@@ -16,7 +16,7 @@ virtualenv:
 before_install:
     - sudo add-apt-repository -y ppa:reddit/ppa
     - sudo apt-get update -q
-    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 pep8 pylint python-redis python-gevent
+    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 python-flake8 pylint python-redis python-gevent
```
dellis23: Nice, I love all the stuff flake8 gives.

setup.cfg
```diff
@@ -4,3 +4,7 @@ with-coverage=1
 cover-package=baseplate
 cover-html=1
 cover-html-dir=build/coverage
+
+[flake8]
+ignore = E501,E128,E226
```
dellis23: 🔕  :( but I _like_ `E501`

dellis23: 🔕  Also not an E226 fan?

spladug: E226 actually doesn't jive with the real PEP-8 doc. The actual rule is 

> Always surround these binary operators with a single space on either side: assignment ( = ), augmented assignment ( += , -= etc.), comparisons ( == , < , > , != , <> , <= , >= , in , not in , is , is not ), Booleans ( and , or , not ).

while E226 enforces for all operators. This shows up in baseplate for some math like https://github.com/reddit/baseplate/blob/master/baseplate/core.py#L103

spladug: (note these ignores aren't new)

i'd like to get 501 going, that's a good one. the three outstanding long-lines when setting max length to 100 are all URLs in comments though :(

baseplate/random.py
```diff
@@ -0,0 +1,57 @@
...
+    An example of usage:
+
+    .. doctest::
+
+        >>> words = ["apple", "banana", "cantelope"]
+        >>> weighted_choice(words, weight_key=lambda word: len(word))
+        'banana'
+
+    """
+    tickets = 0
```
dellis23: I've gone over this implementation thoroughly and can't find any off by one errors, inclusivity / exclusivity issues, or potentially misweighted results.  Thanks for nothing.

baseplate/random.py
```diff
@@ -0,0 +1,57 @@
...
+    :param weight_key: A function that takes an
+        item in ``items`` and returns a non-negative integer
+        weight for that item.
+
+    :return: The chosen item.
+
+    :raises: :py:exc:`ValueError` if any weights are negative or there are no
+        items.
+
+    .. testsetup::
```
dellis23: Neat.  Didn't know doctests had this functionality available.

dellis23: Also, very great doc in general.  Super easy to understand.

tests/unit/random_tests.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_empty_list(self):
+        with self.assertRaises(ValueError):
+            random.weighted_choice([], lambda i: i)
+
+    def test_negative_weight(self):
+        with self.assertRaises(ValueError):
+            random.weighted_choice(["a"], lambda i: -1)
+
+    @mock.patch("random.random")
+    def test_iterator(self, random_random):
```
dellis23: Maybe use `patched_random` here?  It's what I tend to see in the `mock` docs, iirc.

baseplate/service_discovery.py
dellis23: Unless I'm missing something, even if the cluster size were 100k, we're just adding numbers together to find the weight.  That happens in a couple of milliseconds on my machine.

spladug: Yeah, let me do some timing of the function and class versions of the lottery. I think there's probably still some benefit to doing it the new way since it skips allocating a new list on every `choose` call, but it might not be as useful beyond that.

tests/unit/random_tests.py
spladug: looks like they go with `mock_random`. i'll do that.

baseplate/random.py
spladug: thanks :)

baseplate/service_discovery.py
ckwang8128: Right, it depends on how much those couple of milliseconds matters in the grand scheme of things for you/another system using this service discovery mechanism. 3 milliseconds -> 11 microseconds might be a huge jump in performance in certain scenarios.

baseplate/config.py
```diff
@@ -167,6 +167,24 @@ def Base64(text):
         raise ValueError(*exc.args)
 
 
+def File(mode="r"):
+    """A path to a file.
+
+    This takes a path to a file and returns an open file object, like
```
dellis23: Hmm, so this doesn't actually return an open file object, it returns a function that returns an open file object, right?  So they'd still need to do something like `with config['foo']() as f:`?  (This is more a note on the docstring and maybe making the usage a bit clearer).

I guess I'm wondering if this really provides enough functionality to justify the abstraction.  In other words: why not just let the author of the application take the path in the config and `open` it and avoid the potential confusion of whether or not the file needs to be closed?

dellis23:
🐟  up to you on the file thing

baseplate/config.py
spladug: I've added some example usage for the file thingy. The general goal is to fail fast if any config is bad.

dellis23: Cool, thanks!

setup.cfg
spladug: https://github.com/reddit/baseplate/issues/68

PR #66, opened 2017-01-04
Allow Thrift connection pool to talk to framed/buffered servers

krishnan-chandra:
Explanation for this PR: When using Thrift with Java we don't have the benefits of gevent, so to achieve concurrency in requests we use nonblocking server types which have a thread pool for receiving requests and a separate thread pool for processing them. Unfortunately, these servers require the use of a framed/buffered transport since they are processing asynchronously. This poses a problem when using Baseplate libraries/services, since the thread pool assumes the use of a plain TSocket.

I'm open to suggestions on this change - if it would be better to use an alternative method instead of making the change here, that's fine too.

👓  @spladug

### DIFF
--- a/baseplate/thrift_pool.py
+++ b/baseplate/thrift_pool.py
@@ -37,14 +37,15 @@ from .retry import RetryPolicy
 logger = logging.getLogger(__name__)
 
 
-def _make_protocol(endpoint):
+def _make_transport(endpoint):
     if endpoint.family == socket.AF_INET:
         trans = TSocket.TSocket(*endpoint.address)
     elif endpoint.family == socket.AF_UNIX:
         trans = TSocket.TSocket(unix_socket=endpoint.address)
     else:
         raise Exception("unsupported endpoint family %r" % endpoint.family)
-    return THeaderProtocol.THeaderProtocol(trans)
+
+    return trans
 
 
 class ThriftConnectionPool(object):
@@ -60,17 +61,20 @@ class ThriftConnectionPool(object):
         RPC call can take before a TimeoutError is raised.
     :param int max_retries: The maximum number of times the pool will attempt
         to open a connection.
+    :param thrift.transport.TTransport.TTransportFactoryBase transport_wrapper:
+        Factory class that wraps the thrift connection in a framed/buffered transport if necessary.
 
     All exceptions raised by this class derive from
     :py:exc:`~thrift.transport.TTransport.TTransportException`.
 
     """
     # pylint: disable=too-many-arguments
-    def __init__(self, endpoint, size=10, max_age=120, timeout=1, max_retries=3):
+    def __init__(self, endpoint, size=10, max_age=120, timeout=1, max_retries=3, protocol_factory=THeaderProtocol.THeaderProtocol):
         self.endpoint = endpoint
         self.max_age = max_age
         self.retry_policy = RetryPolicy.new(attempts=max_retries)
         self.timeout = timeout
+        self.protocol_factory = protocol_factory
 
         self.pool = queue.LifoQueue()
         for _ in range(size):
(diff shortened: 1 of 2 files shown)


spladug:
I'm wondering if it'd be best to just move everything to framed transport right now before more services are using thrift. It seems there are definitely advantages to it and the downside is only a tiny bit of overhead on each message.

krishnan-chandra:
Happy to make that change if you'd prefer to change things over wholesale - I can take a crack at it.

spladug:
Doing some digging into THeaderProtocol right now, it's a bit more complicated than just swapping things to TFramedTransport.

baseplate/thrift_pool.py
```diff
@@ -37,13 +37,22 @@
...
+    BufferedTransport or FramedTransport if necessary.
+    """
     if endpoint.family == socket.AF_INET:
         trans = TSocket.TSocket(*endpoint.address)
     elif endpoint.family == socket.AF_UNIX:
         trans = TSocket.TSocket(unix_socket=endpoint.address)
     else:
         raise Exception("unsupported endpoint family %r" % endpoint.family)
+
+    if transport_wrapper is not None:
```
spladug: I'm not sure it's actually safe to put TFramedTransport beneath THeaderProtocol in terms of actually getting bytes on the wire that make sense to any server. Instead, I think we should move the instantiation of the `THeaderProtocol` into the default `transport_wrapper` (which would be better called `protocol_factory`) and let your application override that to just `TFramedTransport` being wrapped by `TBinaryProtocol`. 


```python
def __init__(self, endpoint, size..., transport_factory=THeaderProtocol.THeaderProtocol):
    ...
    self.protocol_factory = protocol_factory

def _acquire(self):
    ...
    trans = _make_transport(self.endpoint)
    prot = self.protocol_factory(transport)
    ...

def _make_transport(endpoint):
    if endpoint.family ...
    ...
    return trans
```

```python
def framed_protocol_factory(trans):
    trans = TFramedTransport(trans)
    return TBinaryProtocol(trans)

my_framed_pool = ThriftConnectionPool(endpoint, protocol_factory=framed_protocol_factory)
```

krishnan-chandra: So FWIW, this change did work with a Java server requiring framed transport - that being said, happy to make this fix so we don't break other things.

spladug: i must have missed something getting lost in THeaderProtocol. cool! :)

i think this would probably be a cleaner API though if that's cool

krishnan-chandra:
:haircut_man:, I fixed up the API according to your suggestions and fixed up the tests as well (since all the tests referenced the old API)

baseplate/server/healthcheck.py
```diff
@@ -46,6 +60,7 @@ def check_http_service(endpoint):
 
 CHECKERS = {
     "thrift": check_thrift_service,
+    "thrift-framed": check_framed_thrift_service,
```
spladug: I'm kinda torn here. I see the benefit in this being here for your server, but a baseplate server is never going to use a different protocol like this and so it feels a little confusing/cluttery. Maybe just having the ability to use the other factory (the rest of this patch) is sufficient and we can use that code in a custom healthcheck on the puppet side for your stuff?

krishnan-chandra: That's fair, I can take this out. I mainly just need the ability to run that custom healthcheck, it's immaterial to me where it actually lives.

krishnan-chandra:
:haircut_man: :crossed_fingers:

baseplate/server/healthcheck.py
```diff
@@ -46,7 +45,7 @@ def check_http_service(endpoint):
 
 CHECKERS = {
     "thrift": check_thrift_service,
-    "wsgi": check_http_service,
+    "wsgi": check_http_service
```
spladug: my trailing commas!!! :'(

baseplate/server/healthcheck.py
```diff
@@ -16,7 +16,6 @@
 from baseplate.thrift import BaseplateService
 from baseplate.thrift_pool import ThriftConnectionPool
 
-
```
spladug: ?

krishnan-chandra: I don't get this one either tbh, must have been my text editor

krishnan-chandra:
💇  and I hope you like the commit message

PR #67, opened 2017-01-04
Use bisect for weighted choice implementation.

dellis23:
👓 @spladug @ckwang8128 

You nerd sniped me.  Couldn't stop thinking about it on the way home.

### DIFF
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -4,6 +4,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import bisect
 import random
 
 
@@ -39,26 +40,28 @@ class WeightedLottery(object):
     """
 
     def __init__(self, items, weight_key):
-        self.tickets = 0
-        self.items_with_weights = []
+        self.weights = []
+        self.items = list(items)
+        if len(self.items) == 0:
+            raise ValueError("items must not be empty")
 
-        for item in items:
+        # Build a running list of current total weight.  This allows us to do a
+        # binary search for a random weight to get a choice quickly.
+        for i, item in enumerate(self.items):
+            try:
+                prev_weight = self.weights[i - 1]
+            except IndexError:
+                prev_weight = 0
             weight = weight_key(item)
             if weight < 0:
                 raise ValueError("weight for %r must be non-negative" % item)
-            self.tickets += weight
-            self.items_with_weights.append((item, weight))
+            self.weights.append(prev_weight + weight)
 
-        if self.tickets == 0:
+        if self.weights[-1] < 0:
             raise ValueError("at least one item must have weight")
 
     def pick(self):
         """Pick a random element from the lottery."""
-        winning_ticket = random.random() * self.tickets
-        current_ticket = 0
-        for item, weight in self.items_with_weights:
-            current_ticket += weight
-            if current_ticket > winning_ticket:
-                return item
-        else:  # pragma: nocover
-            raise RuntimeError("weighted_choice failed unexpectedly")
+        winning_ticket = random.random() * self.weights[-1]
+        winning_idx = bisect.bisect(self.weights, winning_ticket)
+        return self.items[winning_idx]
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -47,17 +47,15 @@ class WeightedLottery(object):
 
         # Build a running list of current total weight.  This allows us to do a
         # binary search for a random weight to get a choice quickly.
+        accumulated_weight = 0
         for i, item in enumerate(self.items):
-            try:
-                prev_weight = self.weights[i - 1]
-            except IndexError:
-                prev_weight = 0
             weight = weight_key(item)
             if weight < 0:
                 raise ValueError("weight for %r must be non-negative" % item)
-            self.weights.append(prev_weight + weight)
+            accumulated_weight += weight
+            self.weights.append(accumulated_weight)
 
-        if self.weights[-1] < 0:
+        if accumulated_weight < 0:
             raise ValueError("at least one item must have weight")
 
     def pick(self):
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -42,7 +42,7 @@ class WeightedLottery(object):
     def __init__(self, items, weight_key):
         self.weights = []
         self.items = list(items)
-        if len(self.items) == 0:
+        if not self.items:
             raise ValueError("items must not be empty")
 
         # Build a running list of current total weight.  This allows us to do a
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -48,7 +48,7 @@ class WeightedLottery(object):
         # Build a running list of current total weight.  This allows us to do a
         # binary search for a random weight to get a choice quickly.
         accumulated_weight = 0
-        for i, item in enumerate(self.items):
+        for item in self.items:
             weight = weight_key(item)
             if weight < 0:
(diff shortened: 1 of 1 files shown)


dellis23:
If you guys like this I can update the docs to make the underlying search time more clear, as well as perhaps the generator exhaustion.

baseplate/random.py
```diff
@@ -39,26 +40,28 @@ class WeightedLottery(object):
     """
 
     def __init__(self, items, weight_key):
-        self.tickets = 0
-        self.items_with_weights = []
+        self.weights = []
+        self.items = list(items)
+        if len(self.items) == 0:
```
spladug: can this just be `if not self.items` ?

baseplate/random.py
```diff
@@ -39,26 +40,28 @@ class WeightedLottery(object):
...
+            raise ValueError("items must not be empty")
 
-        for item in items:
+        # Build a running list of current total weight.  This allows us to do a
+        # binary search for a random weight to get a choice quickly.
+        for i, item in enumerate(self.items):
+            try:
+                prev_weight = self.weights[i - 1]
```
spladug: could you do this like:

```python
accumulated_weight = 0
for item in self.items:
    weight = weight_key(item)
    if weight < 0:
        raise ...
    accumulated_weight += weight
    self.weights.append(accumulated_weight)

if accumulated_weight < 0:
    raise ...
```

baseplate/random.py
```diff
@@ -39,26 +40,28 @@ class WeightedLottery(object):
...
-        winning_ticket = random.random() * self.tickets
-        current_ticket = 0
-        for item, weight in self.items_with_weights:
-            current_ticket += weight
-            if current_ticket > winning_ticket:
-                return item
-        else:  # pragma: nocover
-            raise RuntimeError("weighted_choice failed unexpectedly")
+        winning_ticket = random.random() * self.weights[-1]
+        winning_idx = bisect.bisect(self.weights, winning_ticket)
```
spladug: beautiful :tada:

baseplate/random.py
dellis23: Yes!

baseplate/random.py
dellis23: Nice, I like that much better.

dellis23:
:haircut_man:

baseplate/random.py
```diff
@@ -39,26 +40,26 @@ class WeightedLottery(object):
...
+        self.weights = []
+        self.items = list(items)
+        if not self.items:
+            raise ValueError("items must not be empty")
 
-        for item in items:
+        # Build a running list of current total weight.  This allows us to do a
+        # binary search for a random weight to get a choice quickly.
+        accumulated_weight = 0
+        for i, item in enumerate(self.items):
```
spladug: `i` isn't used anymore, is the `enumerate` still necessary?

baseplate/random.py
```diff
@@ -39,26 +40,26 @@ class WeightedLottery(object):
     """
 
     def __init__(self, items, weight_key):
-        self.tickets = 0
-        self.items_with_weights = []
+        self.weights = []
+        self.items = list(items)
```
spladug: :no_bell: the 3.6 stdlib does `tuple` here. i wonder if that's significantly better in some way?

baseplate/random.py
dellis23: Doh, thanks!

dellis23:
:haircut_man:

ckwang8128:
:fish: great implementation!

baseplate/random.py
ckwang8128: More nerd sniping happening: I did some brief reading on this, and the possibilities that come to mind are:

1. `tuple` construction is generally faster than `list` construction and accessing tuples is sometimes faster as well (http://stackoverflow.com/questions/3340539/why-is-tuple-faster-than-list) 
2. may be slightly more space efficient as lists over-allocate memory IIRC
3. since we're only iterating through the data (no sorting, mutation) there's no functionality loss.

pacejackson:
:no_bell: If we're worried about the performance of this, would it be worthwhile to use numpy for it?

https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.choice.html

ckwang8128:
I think this updated implementation is pretty solid and is basically what numpy does - generate CDF and then binary search (https://github.com/numpy/numpy/blob/b94c2b01ff7ef5b8dc44726512cfa232e9054882/numpy/random/mtrand/mtrand.pyx#L1017) - without a lot of overhead

spladug:
I did some timing.

## Current master

(my naive linear implementation)

```
.....1000 iterations of picks from even weight lottery with 10 choices, 0.62μs per iteration
.....1000 iterations of picks from even weight lottery with 100 choices, 2.95μs per iteration
.....1000 iterations of picks from even weight lottery with 1000 choices, 28.01μs per iteration
.....1000 iterations of picks from even weight lottery with 100000 choices, 2777.58μs per iteration
```

## This branch

(dan's logarithmic implementation, massive improvement)

```
.....1000 iterations of picks from even weight lottery with 10 choices, 0.48μs per iteration
.....1000 iterations of picks from even weight lottery with 100 choices, 0.56μs per iteration
.....1000 iterations of picks from even weight lottery with 1000 choices, 0.62μs per iteration
.....1000 iterations of picks from even weight lottery with 100000 choices, 1.43μs per iteration
```

Doing `tuple` instead of `list` has no significant effect.

## Numpy

Oddly enough `numpy` does pretty poorly, even worse than the original linear one (please check my implementation here)

```
.....1000 iterations of picks from even weight lottery with 10 choices, 48.06μs per iteration
.....1000 iterations of picks from even weight lottery with 100 choices, 61.58μs per iteration
.....1000 iterations of picks from even weight lottery with 1000 choices, 167.05μs per iteration
.....1000 iterations of picks from even weight lottery with 100000 choices, 12475.12μs per iteration
```

I used this re-implementation of the `WeightedLottery` interface to test numpy: https://gist.github.com/spladug/80d0872b40a46bfd6d1ec1cc60214645 . It passes the distribution test, so it should provide correct results (but it fails some of the other tests because the `mock.patch` on `random.random` doesn't work on numpy)

## Methodology:

This test script was used unmodified for each test: https://gist.github.com/spladug/79ad34db3481a1373b87f017553aab5e

spladug:
(note that linear was pretty quick in the ranges we'll be using it immediately, but it's good to see the logarithmic solution doing better at high volume)

pacejackson:
Cool!

pacejackson:
If I had to guess, numpy probably rebuilds the underlying data structure each time you call it, so that would slow things down a lot as the list got bigger. Also, that build time probably isn't included in the timings for this branch. 

_Regardless_ though, this seems way better especially if your use case is same thing like "make this once when you start the app and use it some indeterminate number of times".

spladug:
Yeah, that's definitely a possibility. If I move the `lottery = WeightedLottery(...)` construction line into the statement and out of setup, this branch gets way slower as expected:

```
.....1000 iterations of picks from even weight lottery with 10 choices, 3.89μs per iteration
.....1000 iterations of picks from even weight lottery with 100 choices, 20.72μs per iteration
.....1000 iterations of picks from even weight lottery with 1000 choices, 185.43μs per iteration
.....1000 iterations of picks from even weight lottery with 100000 choices, 17917.62μs per iteration
```

dellis23:
Awesome discussion!  Thanks for doing that timing stuff @spladug :)

kjoconnor:
nerds

PR #72, opened 2017-01-18
Fix arg passing from configure_tracing to TraceBaseplateObserver

ckwang8128:
:eyeglasses: @spladug


spladug:
3ad5787f842fab1549306aef943c5e8b995f802d

PR #74, opened 2017-01-19
Fix binary annotations field name in tracing request

ckwang8128:
binary annotations aren't being properly recorded because the field name in the tracing API request is wrong. See http://zipkin.io/pages/data_model.html for the binary annotations key `binaryAnnotations`

:eyeglasses: @bsimpson63


bsimpson63:
🐟

PR #75, opened 2017-01-27
Add support for local spans in tracing

ckwang8128:
This adds baseplate support for tracing local in-request spans. It adheres to the "spec" outlined in the following:

https://github.com/openzipkin/zipkin/issues/808
https://github.com/openzipkin/zipkin/pull/821

This interface is intended to be able to generate local spans in various contexts from the main server span. The differences in serialization, use case, and explicit annotations (or lack thereof) were enough for me to lean towards making a new observer and API to specifically handle local spans. This allows us the freedom to customize Baseplate local spans as we see fit moving forward, e.g. force all local spans to carry certain tags, etc.

### DIFF
--- a/baseplate/context/__init__.py
+++ b/baseplate/context/__init__.py
@@ -27,7 +27,10 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from ..core import BaseplateObserver
+from ..core import (
+    BaseplateObserver,
+    LocalSpan,
+)
 
 
 class ContextFactory(object):
@@ -53,3 +56,8 @@ class ContextObserver(BaseplateObserver):
     def on_server_span_created(self, context, server_span):
         context_attr = self.context_factory.make_object_for_context(self.name, server_span)
         setattr(context, self.name, context_attr)
+
+    def on_child_span_created(self, child_span):
+        if isinstance(child_span, LocalSpan):
+            context_attr = self.context_factory.make_object_for_context(self.name, child_span)
+            setattr(child_span.context, self.name, context_attr)
--- a/baseplate/context/sqlalchemy.py
+++ b/baseplate/context/sqlalchemy.py
@@ -9,7 +9,10 @@ from sqlalchemy import event
 from sqlalchemy.orm import Session
 
 from ..context import ContextFactory
-from ..core import ServerSpanObserver
+from ..core import (
+    ServerSpan,
+    SpanObserver,
+)
 
 
 class SQLAlchemyEngineContextFactory(ContextFactory):
@@ -95,18 +98,29 @@ class SQLAlchemySessionContextFactory(SQLAlchemyEngineContextFactory):
     :param sqlalchemy.engine.Engine engine: A configured SQLAlchemy engine.
 
     """
-    def make_object_for_context(self, name, server_span):
-        engine = super(SQLAlchemySessionContextFactory,
-            self).make_object_for_context(name, server_span)
-        session = Session(bind=engine)
-        server_span.register(SQLAlchemySessionServerSpanObserver(session))
+    def make_object_for_context(self, name, span):
+        if isinstance(span, ServerSpan):
+            engine = super(SQLAlchemySessionContextFactory,
+                           self).make_object_for_context(name, span)
+            session = Session(bind=engine)
+        else:
+            # Reuse session in the existing context
+            #  There should always be one passed down from the
+            #  root ServerSpan
+            session = getattr(span.context, name)
+        span.register(SQLAlchemySessionSpanObserver(session, span))
         return session
 
 
-class SQLAlchemySessionServerSpanObserver(ServerSpanObserver):
+class SQLAlchemySessionSpanObserver(SpanObserver):
     """Automatically close the session at the end of each request."""
-    def __init__(self, session):
+    def __init__(self, session, span):
         self.session = session
+        self.span = span
 
     def on_finish(self, exc_info):
-        self.session.close()
+        # A session is passed down to child local spans
+        #   in a request pipeline so only close the session
+        #  if the parent ServerSpan is closing.
+        if isinstance(self.span, ServerSpan):
+            self.session.close()
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -6,6 +6,8 @@ from __future__ import unicode_literals
 import collections
 import random
 
+from .integration.wrapped_context import WrappedRequestContext
+
 
 class BaseplateObserver(object):
     """Interface for an observer that watches Baseplate."""
@@ -225,9 +227,10 @@ class Baseplate(object):
         if trace_info is None:
             trace_info = TraceInfo.new()
 
+        context = WrappedRequestContext(context)
         server_span = ServerSpan(trace_info.trace_id, trace_info.parent_id,
                                  trace_info.span_id, trace_info.sampled,
-                                 trace_info.flags, name)
+                                 trace_info.flags, name, context)
         for observer in self.observers:
             observer.on_server_span_created(context, server_span)
         return server_span
@@ -236,8 +239,7 @@ class Baseplate(object):
 class Span(object):
     """A span represents a single RPC within a system."""
 
-    # pylint: disable=invalid-name
-    def __init__(self, trace_id, parent_id, span_id, sampled, flags, name):
+    def __init__(self, trace_id, parent_id, span_id, sampled, flags, name, context):
         self.trace_id = trace_id
         self.parent_id = parent_id
         self.id = span_id
@@ -317,19 +319,65 @@ class Span(object):
         else:
             self.finish()
 
+    def make_child(self, name, local=False, component_name=None):
+        """Return a child Span whose parent is this Span."""
+        raise NotImplementedError
 
-class ServerSpan(Span):
+
+class LocalSpan(Span):
+    def __init__(self,
+                 trace_id,
+                 parent_id,
+                 span_id,
+                 sampled,
+                 flags,
+                 name,
+                 context):
+        super(LocalSpan, self).__init__(trace_id, parent_id, span_id, sampled,
+                                        flags, name, context)
+        self.context = context
+
+    def make_child(self, name, local=False, component_name=None):
+        """Return a child Span whose parent is this Span.
+
+        The child span can either be a local span representing an in-request
+        operation or a span representing an outbound service call.
+
+        In a server, a local span represents the time spent within a
+        local component performing an operation or set of operations.
+        The local component is some grouping of business logic,
+        which is then split up into operations which could each be wrapped
+        in local spans.
+
+        :param str name: Name to identify the operation this span
+            is recording.
+        :param bool local: Make this span a LocalSpan if True, otherwise
+            make this span a base Span.
+        :param str component_name: Name to identify local component
+            this span is recording in if it is a local span.
+        """
+        span_id = random.getrandbits(64)
+
+        if local:
+            context_copy = self.context.clone()
+            span = LocalSpan(self.trace_id, self.id, span_id, self.sampled,
+                             self.flags, name, context_copy)
+            if component_name is None:
+                raise ValueError("Cannot create local span without component name.")
+            span.component_name = component_name
+            context_copy.trace = span
+        else:
+            span = Span(self.trace_id, self.id, span_id, self.sampled, self.flags, name, self.context)
+        for observer in self.observers:
+            observer.on_child_span_created(span)
+        return span
+
+
+class ServerSpan(LocalSpan):
     """A server span represents a request this server is handling.
 
     The server span is available on the :term:`context object` during requests
     as the ``trace`` attribute.
 
     """
-
-    def make_child(self, name):
-        """Return a child span representing an outbound service call."""
-        span_id = random.getrandbits(64)
-        span = Span(self.trace_id, self.id, span_id, self.sampled, self.flags, name)
-        for observer in self.observers:
-            observer.on_child_span_created(span)
-        return span
+    pass
--- a/baseplate/diagnostics/metrics.py
+++ b/baseplate/diagnostics/metrics.py
@@ -3,7 +3,11 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from ..core import BaseplateObserver, SpanObserver
+from ..core import (
+    BaseplateObserver,
+    LocalSpan,
+    SpanObserver,
+)
 
 
 class MetricsBaseplateObserver(BaseplateObserver):
@@ -45,7 +49,10 @@ class MetricsSpanObserver(SpanObserver):
 
 class MetricsServerSpanObserver(MetricsSpanObserver):
     def on_child_span_created(self, span):  # pragma: nocover
-        observer = MetricsSpanObserver(self.batch, "clients." + span.name)
+        if isinstance(span, LocalSpan):
+            observer = MetricsSpanObserver(self.batch, "clients." + span.name)
+        else:
+            observer = MetricsSpanObserver(self.batch, span.component_name + "." + span.name)
         span.register(observer)
 
     def on_finish(self, exc_info):
--- a/baseplate/diagnostics/tracing.py
+++ b/baseplate/diagnostics/tracing.py
@@ -15,7 +15,11 @@ from datetime import datetime
 import requests
 
 from .._compat import queue
-from ..core import BaseplateObserver, SpanObserver
+from ..core import (
+    BaseplateObserver,
+    LocalSpan,
+    SpanObserver,
+)
 
 
 logger = logging.getLogger(__name__)
@@ -27,6 +31,7 @@ ANNOTATIONS = {
     'CLIENT_RECEIVE': 'cr',
     'SERVER_SEND': 'ss',
     'SERVER_RECEIVE': 'sr',
+    'LOCAL_COMPONENT': 'lc',
 }
 
 # Feature flags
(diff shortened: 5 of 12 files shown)


ckwang8128:
:eyeglasses: @spladug @bsimpson63

tests/unit/diagnostics/tracing_tests.py
```diff
@@ -276,6 +278,66 @@ def test_on_child_span_created(self):
...
 
+class TraceLocalSpanObserverTests(unittest.TestCase):
+    def setUp(self):
+        self.recorder = NullRecorder()
+        self.span = ServerSpan('test-id',
+                               'test-parent-id',
+                               'test-span-id',
+                               None,
+                               0,
+                               'test')
```
bsimpson63: missing a newline

tests/unit/diagnostics/tracing_tests.py
```diff
@@ -276,6 +278,66 @@ def test_on_child_span_created(self):
...
+            'test-host',
+            self.span,
+            self.recorder,
+        )
+        self.assertEqual(
+            local_trace_observer.local_component,
+            'test-component',
+        )
+
+    def test_init_local_component(self):
```
bsimpson63: these tests have the same name, is that ok?

spladug:
General question: why treat local spans as a completely different thing altogether? the difference is in annotations only, right? should client spans made inside a local span be children of the local span? Should graphite metrics etc. apply to the spans too (via the regular span infrastructure)?

:nail_care: for discussion

tests/unit/diagnostics/tracing_tests.py
ckwang8128: fixed

tests/unit/diagnostics/tracing_tests.py
ckwang8128: nope, copy/pasta error. thanks for catching!

ckwang8128:
> why treat local spans as a completely different thing altogether? the difference is in annotations only, right?

Along with the lack of server/client-specific annotations (cs, cr, ss, sr), the local span binary annotation with component and operation names need to be added directly. There's also a slight metadata difference with the `duration` being calculated and passed (Zipkin would use the timestamps of the CS/CR/SS/SR annotations to do that server-side for the other span types). There were enough serialization and use-case differences that I felt making a separate observer would be more simple and extensible than pushing that differentiation into the Span object itself. Having all of the local span logic isolated in the custom observer means that if there's future differentiation and customization of local spans vs. other span types it'll be more straightforward to add in. 

> should client spans made inside a local span be children of the local span? 
Yes they should for purposes of modeling and visualization. I can't say for sure, but https://github.com/openzipkin/brave/pull/114 is leading me to believe weird things will happen if we have client spans in local ones without explicitly setting the local span as the parent of the client span. I had only considered local spans as being used for non-client-call blocks of code, but this case should be supported. This means we'll need to do a bit more work to support this entirely (local spans without client requests inside them will be fine as things are right now) since server spans are used directly to make child spans in Baseplate contexts. Initial thoughts are that we'd need to allow all Span objects, not just ServerSpans, to make child spans, and then pass along the notion of being in a local span as opposed to just the server span to contexts without breaking the API too badly. 

> Should  graphite metrics etc. apply to the spans too (via the regular span infrastructure)?

Not sure what you mean by this, could you provide an example?

ckwang8128:
:haircut:

bsimpson63:
🐟 lgtm

spladug:
> Having all of the local span logic isolated in the custom observer means that if there's future differentiation and customization of local spans vs. other span types it'll be more straightforward to add in.

Yeah, that makes sense here. It ends up making sense in the metrics case (below) too since we'd likely want those in a separate timer namespace (not `clients`). :+1: 


> Initial thoughts are that we'd need to allow all Span objects, not just ServerSpans, to make child spans, and then pass along the notion of being in a local span as opposed to just the server span to contexts without breaking the API too badly.

Yup. Agreed. Before having local spans, there was no reason to have multi-level span ancestry in the same service since all spans were considered remote. But if we have local spans, it seems like we should allow for nesting; if you don't want to do this right now we can punt it and make it happen during the near-future structural improvements push. 

>> Should graphite metrics etc. apply to the spans too (via the regular span infrastructure)?

> Not sure what you mean by this, could you provide an example?

Right now, anything wrapped in a span gets a timer metric sent to graphite automatically via the MetricsObserver stuff. It'd be nice to have the metrics observer support this new span type so we get automatic timers for local spans as well.

:nail_care: to continue discussion. feel free to tell me you want to move ahead like this since I think we'd only be making additive changes to the public API.

ckwang8128:
Thanks for the feedback and explanations - I think it's straightforward to get everything in, and I'd rather the next Baseplate release have all of that functionality included. To be continued...

ckwang8128:
"I think it's straightforward to get everything in" one month later....

Major changes for local spans:

* The base Span class can now create local and child spans. I didn't explicitly subclass anything but might add some attributes to signal if a span is a local or child span so that it's explicitly not possible to create a local span from a child span. Open to thoughts here. 

* Spans now require as an argument and track the context in which they were created. This is mainly to simplify the `make_local_span` call and not require the caller to pass in the context at creation. 

* Request contexts are wrapped in the Pyramid + Thrift integration level OR at the server span creation level into a `WrappedRequestContext` object (in cases where full integrations aren't used, like r2) which proxies to the underlying wrapped request context and has the ability to copy itself. The copying is necessary for when we want to attach new local-span-aware clients to a context for use within the span, but want to preserve the outside-of-the-span context elsewhere. Open q: Is it worth doing the request context wrapping in the framework integration code or just have it in the server span creation part and make it known that the context returned from a make_local_span request is a wrapped version of the "native" context object?

* The  ContextFactory class handles local span creation events by attaching a new span-aware client to the provided context, which should be a copy of the outer WrappedRequestContext object.

* Local-span-aware context objects are returned as part of the `make_local_span` call; code within the local span will need to enforce using that context.  

This is a lot of surgery, more testing is needed. Also I might have missed a simpler way to handle all this, open to any and all suggestions!

ckwang8128:
:haircut: opening this up for initial feedback as I continue testing.

baseplate/context/__init__.py
```diff
@@ -53,3 +53,7 @@ def __init__(self, name, context_factory):
     def on_server_span_created(self, context, server_span):
         context_attr = self.context_factory.make_object_for_context(self.name, server_span)
         setattr(context, self.name, context_attr)
+
+    def on_local_span_created(self, context, span, component_name):
+        context_attr = self.context_factory.make_object_for_context(self.name, span)
```
spladug: It looks like this would probably confuse the SQLAlchemySessionContextFactory: https://github.com/reddit/baseplate/blob/master/baseplate/context/sqlalchemy.py#L102 as it'd close the session at the end of a local span before the server span had finished.

We probably need to look at the type of the span (ew?) and only do the session closing on server spans or something else smarter there.

baseplate/integration/wrapped_context.py
```diff
@@ -0,0 +1,21 @@
...
+    """Given two dicts, merge them into a new dict as a shallow copy."""
+    z = x.copy()
+    z.update(y)
+    return z
+
+
+class WrappedRequestContext(object):
+    def __init__(self, context, trace=None):
+        self._context = context
+        self.__dict__.update(self._context.__dict__)
```
spladug: Why copy over the dict? Doesn't this make the `__getattr__` do nothing essentially? I think doing just the getattr would be cleanest since it keeps the wrapper free of any stuff in upstream stuff (e.g. does pyramid's request have a `clone` that would shadow `WrappedRequestContext`'s native clone?)

baseplate/core.py
```diff
@@ -317,19 +333,44 @@ def __exit__(self, exc_type, value, traceback):
         else:
             self.finish()
 
+    def make_local(self, component_name, op_name):
```
spladug: i keep dithering on if this should be its own method or just a boolean flag on `make_child`. that'd mean we'd need to figure out a new signature for `make_child` which just returns the new context or something and you have to get at the span from `context.trace`? whatcha think?

baseplate/diagnostics/metrics.py
```diff
@@ -48,6 +48,10 @@ def on_child_span_created(self, span):  # pragma: nocover
         observer = MetricsSpanObserver(self.batch, "clients." + span.name)
         span.register(observer)
 
+    def on_local_span_created(self, context, span, component_name):  # pragma: nocover
+        observer = MetricsSpanObserver(self.batch, component_name + "." + span.name)
```
spladug: Given the issue with the sqlalchemy guy, maybe we can just rely on checking types of spans for generating the namespace? then we probably don't need a separate observer method for just local spans which frees up the class hierarchy a bit.

baseplate/integration/wrapped_context.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __getattr__(self, attr):
+        return getattr(self._context, attr)
+
+    def clone(self):
+        new_context = copy.copy(self._context)
```
spladug: do we actually want to copy the underlying object ever? or just keep referencing it since we're just a neat wrapper.

i'm concerned that we'd do something like

```python
with request.trace.make_local("...") as (local_request, local):
    local_request.response_code = 302
```

and that would get lost because we were copying the underlying request object instead of just referencing.

(it'd probably never happen so dumbly in real code, but i could see injecting a local span around a block of code that passes a request object around various function calls getting messed up)

baseplate/integration/wrapped_context.py
ckwang8128: This was mainly to support passing in the wrapped context to existing framework integrations (which make some direct calls that were troublesome). I think I understand enough of how all of this is working now to no longer need this + move to only having wrapped contexts in baseplate components.

baseplate/integration/wrapped_context.py
ckwang8128: Yeah, I think this was just me running circles around myself. I was worried about the flip side -> innocuous changing of underlying context attributes on accident. I think it's better to keep the reference to existing context object and do some logging every time we access the underlying context instead of changing our wrapped attrs or adding new ones.

baseplate/core.py
ckwang8128: I think there are enough differences between the two types that its own method is warranted. Child spans and local spans are more sibling types than subtypes, and I think using the boolean flag implies local spans are a specific type of child span.

baseplate/diagnostics/metrics.py
ckwang8128: Agreed, updating types now to see how it all feels.

baseplate/core.py
ckwang8128: I changed my mind. More explanation to follow in commit

baseplate/context/__init__.py
ckwang8128: Since we'd  be creating a new session when a local span is created ( https://github.com/reddit/baseplate/blob/master/baseplate/context/sqlalchemy.py#L101 ) and a new observer to go with it to close it, this shouldn't inadvertently close the server span's session I think?

ckwang8128: I think with my changes all of the references to `server_span` and `ServerSpan` in the context class definitions should be `Span` since objects for contexts won't just be created for server spans anymore.

ckwang8128:
Updates for this revision pass:

* Spans are more hierarchically organized, Base Span -> Local Span -> Server Span:
   -> Base Spans represent outgoing request wrapping; the `Span` class cannot create either child base spans or local spans. 
   -> LocalSpans are for local sets of operations within the server request handling pipeline and can create both child base spans (for outgoing requests) or local spans (for sub-division of operations). 
   -> ServerSpans wrap an entire request handled by the server and can create both child base spans and local spans

Having `ServerSpan` subclass `LocalSpan` seemed a bit weird, but I didn't want to overcomplicate the hierarchy and logically it technically works -> the server span can be seen as a top level local span for the server. In the future this might need to be split out a bit more but I'm okay with this. 

* `make_local` has been consolidated with `make_child` with the addition of some flags to specify whether or not a local span is needed. I think this is a better abstraction moving forward as it communicates that the span being created is always a child of the span `make_child` is invoked on. 

* Context wrapping has been cleaned up -> contexts are wrapped initially in the server span and passed on from there to children. Attribute setting of the core context object is also propagated. 

Context object factories still need some variable renaming; `server_span` doesn't apply to most of the `make_object_for_context` calls since local spans might also be passed in.

ckwang8128:
:haircut: 

tests/comments are needed, would appreciate feedback on changes as that's happening!

baseplate/context/sqlalchemy.py
```diff
@@ -95,15 +95,15 @@ class SQLAlchemySessionContextFactory(SQLAlchemyEngineContextFactory):
...
 
     """
-    def make_object_for_context(self, name, server_span):
+    def make_object_for_context(self, name, span):
         engine = super(SQLAlchemySessionContextFactory,
-            self).make_object_for_context(name, server_span)
+            self).make_object_for_context(name, span)
         session = Session(bind=engine)
-        server_span.register(SQLAlchemySessionServerSpanObserver(session))
+        span.register(SQLAlchemySessionSpanObserver(session))
```
spladug: I think making a new session for local spans would be pretty confusing, particularly if they rollback or whatever at the end of the local span which might be pretty unexpected behaviour. It also makes local spans much more heavyweight of an operation if you have sqlalchemy set up. What do you think of this happening only for server spans?

baseplate/core.py
```diff
@@ -317,19 +324,59 @@ def __exit__(self, exc_type, value, traceback):
...
 
-    """
+        """
+        raise NotImplementedError
 
-    def make_child(self, name):
-        """Return a child span representing an outbound service call."""
+
+class LocalSpan(Span):
+    # pylint: disable=invalid-name
```
spladug: huh, what's the invalid name?

baseplate/core.py
```diff
@@ -317,19 +324,59 @@ def __exit__(self, exc_type, value, traceback):
...
+    def make_child(self, name, local=False, component_name=None):
         span_id = random.getrandbits(64)
-        span = Span(self.trace_id, self.id, span_id, self.sampled, self.flags, name)
+
+        if local:
+            context_copy = self.context.clone()
+            span = LocalSpan(self.trace_id, self.id, span_id, self.sampled,
+                             self.flags, name, context_copy)
+            if component_name is None:
+                raise SpanError("Cannot create local span without component name.")
```
spladug: `ValueError` should be fine here

> Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

what _is_ `component_name` anyway? (it'd be good to have param docs here)

baseplate/integration/wrapped_context.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, context, trace=None):
+        self.__dict__['_context'] = context
+        self.__dict__['logger'] = logging.getLogger(self.__class__.__name__)
+
+    def __getattr__(self, attr):
+        return getattr(self._context, attr)
+
+    def __setattr__(self, attr, value):
+        if attr in self._context.__dict__:
+            self.logger.debug("Assigning attr=%s of wrapped request context.")
```
spladug: This might be a bit spammy for running apps in debug mode. Do you think it's generally useful?

baseplate/core.py
```diff
@@ -317,19 +324,59 @@ def __exit__(self, exc_type, value, traceback):
...
+            span.component_name = component_name
+            context_copy.trace = span
```
spladug: it looks like `context_copy` isn't actually going anywhere?

baseplate/core.py
ckwang8128: Hmm, I don't think there's one. This was copypasta'd from the definition in `Span`, but I just removed and ran linting and there wasn't complaining. Removing.

baseplate/core.py
ckwang8128: Good call on `ValueError`. Those docs got accidentally removed while shuffling things around, adding back in. To answer here, the `component_name` is a secondary namespace of the local span. The span name itself should be the specific operation, but you might want to query similar operations under different namespaces ("components"), e.g. if you had "query" named spans for both "cassandra" and "mysql" components. http://javadox.com/io.zipkin.java/zipkin/1.1.4/zipkin/Constants.html#LOCAL_COMPONENT

baseplate/core.py
ckwang8128: It's getting passed into the span (https://github.com/reddit/baseplate/pull/75/files/b92742b01897b2921a5ae78f065b5ee0dab6a8b9#diff-e23a2f0193db987b0509d66964675524R362) which is how the copied context will be accessed within the span block.

baseplate/integration/wrapped_context.py
ckwang8128: I think the case where we're logging for changing the wrapped request context will definitely be spammy (since we assign a ton of context objects when a new wrapped context gets created for local spans), removing that one. The reason I added the one for setting attributes of the underlying context is to provide explicit messaging in case developers are inadvertently messing with the context itself when they want to manipulate WrappedContext attrs and trying to debug whether or not that's the case.

baseplate/context/__init__.py
```diff
@@ -53,3 +56,8 @@ def __init__(self, name, context_factory):
     def on_server_span_created(self, context, server_span):
         context_attr = self.context_factory.make_object_for_context(self.name, server_span)
         setattr(context, self.name, context_attr)
+
+    def on_child_span_created(self, span):
+        if isinstance(span, LocalSpan):
```
bsimpson63: is `span` here the child span? if so can it be called `child_span`?

ckwang8128: 👍

baseplate/diagnostics/tracing.py
```diff
@@ -223,6 +228,43 @@ def record(self):
         self.recorder.send(self)
 
 
+class TraceLocalSpanObserver(TraceSpanObserver):
```
bsimpson63: If in the future we want to create a new type of observer how do we decide whether to make a completely new class like this vs. just updating `on_child_span_created` like in `MetricsServerSpanObserver`? Is this separate class just needed here because traces/spans are so coupled?

ckwang8128: This separate class is needed because the serialization for tracing these local spans is different from the base spans (which I think you're alluding to with your second question). In general you'll want to follow the `on_child_span_created` path. These observers are strictly for new types of spans, and only if there's some weird serialization that's needed for tracing.

baseplate/context/sqlalchemy.py
ckwang8128: Ah, I see your point. Definitely easy to shoot yourself in the foot if you don't know that local spans create another session. Better to stick to one request -> one session.

ckwang8128:
:haircut: getting there!

baseplate/context/__init__.py
```diff
@@ -27,7 +27,10 @@ def my_handler(self, context):
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from ..core import BaseplateObserver
+from ..core import (
+    LocalSpan,
+    BaseplateObserver
```
spladug: nit: trailing comma please (or just put 'em on one line which is fine if it's short enough)

baseplate/core.py
```diff
@@ -317,19 +319,65 @@ def __exit__(self, exc_type, value, traceback):
...
+        The child span can either be a local span representing an in-request
+        operation or a span representing an outbound service call.
+
+        In a server, a local span represents the time spent within a
+        local component performing an operation or set of operations.
+        The local component is some grouping of business logic,
+        which is then split up into operations which could each be wrapped
+        in local spans.
+
+         :param str name: Name to identify the operation this span
```
spladug: nit: i don't think these are supposed to be extra-indented (per pep-257)

PR #76, opened 2017-01-27
Force string coercion for binary annotation values

ckwang8128:
Zipkin doesn't like non-string/boolean values for binary annotations, per 
https://github.com/openzipkin/zipkin/issues/1372
and changes introduced in 
https://github.com/openzipkin/zipkin/pull/1269
https://github.com/openzipkin/zipkin/commit/3b8d56ac1b817729517e32c7a3d311b7ceaa2fb7#diff-b603d520f1c771c817602413216dd6c3L191

Based on discussions in the above, the general approach is to coerce numerical values to strings. I'm choosing to do this directly at the baseplate tracing serialization level as a defensive measure. 

:eyeglasses: @bsimpson63

### DIFF
--- a/baseplate/diagnostics/tracing.py
+++ b/baseplate/diagnostics/tracing.py
@@ -140,6 +140,10 @@ class TraceSpanObserver(SpanObserver):
         self.record()
 
     def on_set_tag(self, key, value):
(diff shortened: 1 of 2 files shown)


bsimpson63:
🐟

PR #77, opened 2017-02-03
Handle Cassandra Statement queries in statement tag

pacejackson:
Currently we get errors like this when you use a Statement class rather than a string for the query and you have tracing enabled:

```
TypeError: <SimpleStatement query="SELECT * FROM table_name", consistency=Not Set> is not JSON serializable
```

### DIFF
--- a/baseplate/_compat.py
+++ b/baseplate/_compat.py
@@ -14,11 +14,13 @@ if sys.version_info.major == 3:  # pragma: nocover
     import queue
     from io import BytesIO
     range = range
+    string_types = str,
 else:  # pragma: nocover
     import ConfigParser as configparser
     import Queue as queue
     from cStringIO import StringIO as BytesIO
     range = xrange
+    string_types = basestring,
 
 
 __all__ = [
@@ -26,4 +28,5 @@ __all__ = [
     "queue",
     "BytesIO",
     "range",
+    "string_types",
 ]
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -4,9 +4,11 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 from cassandra.cluster import Cluster, _NOT_SET
+from cassandra.query import SimpleStatement, PreparedStatement, BoundStatement
 
 from . import ContextFactory
 from .. import config
+from .._compat import string_types
 
 
 def cluster_from_config(app_config, prefix="cassandra.", **kwargs):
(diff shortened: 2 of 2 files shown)


pacejackson:
👓  @spladug

baseplate/context/cassandra.py
```diff
@@ -3,7 +3,11 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import six
+
 from cassandra.cluster import Cluster, _NOT_SET
+from cassandra.query import SimpleStatement, PreparedStatement, BoundStatement
+from cassandra.query import Statement as CassandraStatement
```
pacejackson: :doh: I'll delete this.

baseplate/context/cassandra.py
```diff
@@ -3,7 +3,10 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import six
```
spladug: I think this is the first time we've explicitly used `six` in baseplate. Either it should get added into setup.py etc. or we should just add the relevant bits to the `_compat` module which is sorta a mini six. I slightly lean to the latter since it's one fewer thing to worry about.

baseplate/context/cassandra.py
```diff
@@ -93,8 +94,10 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
         # TODO: include custom payload
         if isinstance(query, six.string_types):
             span.set_tag("statement", query)
-        elif isinstance(query, SimpleStatement):
+        elif isinstance(query, SimpleStatement, PreparedStatement):
```
spladug: I think this is invalid syntax: `isinstance` only takes two params though the latter can be a tuple of types.

pacejackson:
:haircut_man: addressed feedback and cleaned up the commits.

baseplate/_compat.py
```diff
@@ -14,16 +14,19 @@
...
     range = xrange
+    string_types = basestring
 
 
 __all__ = [
     "configparser",
     "queue",
     "BytesIO",
     "range",
+    "string_types"
```
spladug: nit: trailing comma please

baseplate/_compat.py
```diff
@@ -14,16 +14,19 @@
     import queue
     from io import BytesIO
     range = range
+    string_types = str
```
spladug: should these be tuples of types?

pacejackson: They could be, they aren't in `six`

https://github.com/benjaminp/six/blob/master/six.py#L41

pacejackson:
:haircut_man:

baseplate/_compat.py
spladug: Yes they are, the comma at the end makes it into a tuple.

```
$ python2
>>> six.string_types
(<type 'basestring'>,)

$ python3
>>> six.string_types
(<class 'str'>,)

```

pacejackson:
:haircut_man:

pacejackson:
To test this, I used this version of baseplate to run the gateway locally.  Without these changes, I got the expected error when trying to flush a span.  With these changes, I did not get this error.  I did get an error  when trying to flush spans, but it was about not being able to reach the trace endpoint which is expected since the endpoint specified in my dev config does not exist. :)

PR #79, opened 2017-02-27
tracing: Optionally disable trace logging

bsimpson63:
Rename NullRecorder to LoggingRecorder and make it a noop recorder.
LoggingRecorder is used by default when no tracing endpoint is
specified and will write trace details to the logger.

On dev installs of r2 there's no tracing endpoint specified so the application logs are dominated by the traces. This change will let services (or r2) decide whether to log traces when calling `configure_tracing()`.

👓 @ckwang8128 @spladug

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -164,7 +164,8 @@ class Baseplate(object):
                           max_span_queue_size=50000,
                           num_span_workers=5,
                           span_batch_interval=0.5,
-                          sample_rate=0.1):
+                          sample_rate=0.1,
+                          log_if_unconfigured=True):
         """Collect and send span information for request tracing.
 
         :param str service_name: The name for the service this observer
@@ -187,6 +188,7 @@ class Baseplate(object):
                 num_span_workers=num_span_workers,
                 span_batch_interval=span_batch_interval,
                 sample_rate=sample_rate,
+                log_if_unconfigured=log_if_unconfigured,
             )
         )
 
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -177,6 +177,8 @@ class Baseplate(object):
         :param int num_span_workers: number of worker threads for span processing.
         :param float span_batch_interval: wait time for span processing in seconds.
         :param float sample_rate: percentage of unsampled requests to record traces for.
+        :param bool log_if_unconfigured: flag to write traces to logger if no
+            tracing_endpoint is specified.
         """
         from .diagnostics.tracing import TraceBaseplateObserver
         self.register(
--- a/baseplate/diagnostics/tracing.py
+++ b/baseplate/diagnostics/tracing.py
@@ -68,7 +68,8 @@ class TraceBaseplateObserver(BaseplateObserver):
                  num_span_workers=5,
                  span_batch_interval=0.5,
                  num_conns=100,
-                 sample_rate=0.1):
+                 sample_rate=0.1,
+                 log_if_unconfigured=True):
         self.service_name = service_name
         self.hostname = socket.gethostbyname(socket.gethostname())
         self.sample_rate = sample_rate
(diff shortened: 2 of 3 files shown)


baseplate/core.py
```diff
@@ -164,7 +164,8 @@ def configure_tracing(self, service_name,
                           max_span_queue_size=50000,
                           num_span_workers=5,
                           span_batch_interval=0.5,
-                          sample_rate=0.1):
+                          sample_rate=0.1,
+                          log_if_unconfigured=True):
```
ckwang8128: could you add this to the docstring above?

ckwang8128:
:nail_care: thanks for adding this option! small documentation request

bsimpson63:
💇

ckwang8128:
missed the test failure in first pass, I think the error message there is pretty clear, just need to fix the assertion with the new recorder:

```
FAIL: test_null_recorder_setup (tests.unit.diagnostics.tracing_tests.TraceObserverTests)
----------------------------------------------------------------------
  File "/home/travis/build/reddit/baseplate/tests/unit/diagnostics/tracing_tests.py", line 28, in test_null_recorder_setup
    self.assertEqual(type(baseplate_observer.recorder), NullRecorder)
AssertionError: <class 'baseplate.diagnostics.tracing.LoggingRecorder'> != <class 'baseplate.diagnostics.tracing.NullRecorder'>
```

baseplate/diagnostics/tracing.py
```diff
@@ -337,6 +344,22 @@ def flush_func(self, spans):
             self.logger.debug("Span recording: %s", span)
 
 
+class NullRecorder(BaseBatchRecorder):
```
spladug: can this have a docstring please?

baseplate/core.py
```diff
@@ -187,6 +190,7 @@ def configure_tracing(self, service_name,
                 num_span_workers=num_span_workers,
                 span_batch_interval=span_batch_interval,
                 sample_rate=sample_rate,
+                log_if_unconfigured=log_if_unconfigured,
```
spladug: Do we think that we'd want this to change at the whole application level (i.e. hard code the value) or do we want to make it something that can change in config? The latter feels more right to me.

bsimpson63: I was thinking about that, but then I convinced myself that any application using this stuff would be manually calling `configure_tracing`, at which point it has to handle the config itself. Is that not right?

spladug: Yeah, I wonder if it should work like the statsd stuff where there's a function that standardizes config parsing and returns an object that can be passed to configure_tracing rather than having individual config options on configure_tracing. That lets the configuration parsing be standardized but still gives the application the ability to get at the raw object if it needs it (for e.g. non-server setups). 

```python
tracing_client = make_tracing_client(config)
...
baseplate.configure_tracing(tracing_client)
```

ckwang8128: I don't feel strongly either way; I hadn't anticipated so many options being needed so I can definitely see the case for wrapping the config parsing into an object. Both seem to afford the same flexibility in configuration. This would break existing client integrations though right?

spladug: Yeah, it would. But now's the time before more services pick it up if we want to do it.

bsimpson63: I just took a quick stab at this. I think reworking this will require refactoring `TraceBaseplateObserver` to take a client, and doing that basically ripples throughout all the existing tracing stuff.

ckwang8128: > But now's the time before more services pick it up if we want to do it.

Definitely. I'm happy to help migrate existing internal clients!

ckwang8128:
:fish: I've filed https://reddit.atlassian.net/browse/IO-1629 to standardize a config object for tracing and then do existing client migration. Adding this for now would be a temporary stopgap that alleviates current issues with the debug logging causing pains in staging.

bsimpson63:
c1f56aeb6492f5af024062406398849bc81f380f

PR #80, opened 2017-03-09
Add percentage config helper and fix docs

### LINKED ISSUE #78: Add percent config validator
Things like sample rates are really common in configs, it'd be nice to make them more human readable:

```ini
[app:main]
rate = 37%
```

```python
>>> cfg = config.parse_config(raw_config, {
...    "rate": config.Percent,
... })
>>> print(cfg.rate)
.32
```

spladug:
This closes #78 and fixes a small documentation thing I noticed.

:eyeglasses: @ckwang8128 @krishnan-chandra

### DIFF
--- a/baseplate/config.py
+++ b/baseplate/config.py
@@ -47,6 +47,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     ...     },
     ...     "some_file": config.File(mode="r"),
     ...     "optional": config.Optional(config.Integer, default=9001),
+    ...     "sample_rate": config.Percent,
     ... })
 
     >>> print(cfg.simple)
@@ -61,6 +62,9 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     >>> cfg.some_file.read()
     'cool'
     >>> cfg.some_file.close()
+
+    >>> cfg.sample_rate
+    0.371
 
 .. testcleanup::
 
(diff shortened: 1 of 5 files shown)


ckwang8128:
:fish:

baseplate/config.py
```diff
@@ -235,6 +239,25 @@ def Timespan(text):
...
+    returns a float in the range [0.0, 1.0].
+
+    """
+
+    if not text.endswith("%"):
+        raise ValueError("the value is not a percentage")
+
+    percentage = float(text[:-1]) / 100.
+
+    if not 0 <= percentage <= 1:
```
krishnan-chandra: 🔕 

Would you want to do this validation before or after doing the float division? I don't think it'll make a material difference but it might be easier to understand if the user can't supply an argument like "100.01%" and have it work.

spladug: Outside of funky float stuff, I think this should cover the same case:

```python
>>> baseplate.config.Percent("100.1%")
  File "<stdin>", line 1, in <module>
  File "baseplate/config.py", line 256, in Percent
    raise ValueError("percentage is out of valid range")
ValueError: percentage is out of valid range
>>> baseplate.config.Percent("100.01%")
  File "<stdin>", line 1, in <module>
  File "baseplate/config.py", line 256, in Percent
    raise ValueError("percentage is out of valid range")
ValueError: percentage is out of valid range
>>> baseplate.config.Percent("100.0000000000001%")
  File "<stdin>", line 1, in <module>
  File "baseplate/config.py", line 256, in Percent
    raise ValueError("percentage is out of valid range")
ValueError: percentage is out of valid range

```

Issue #83, opened 2017-04-20
config.Optional always returns the default for Boolean values


bsimpson63:
https://github.com/reddit/baseplate/blob/master/baseplate/config.py#L307

If the value from the config is `False` this check is falsy and the default will be returned.

spladug:
That check is working on the raw string, not the already-parsed version, so truthiness should just be if the string is empty or not. AFAICT this does work as intended with booleans:

```python
>>> Optional(Boolean, default=None)("")
>>> Optional(Boolean, default=None)("True")
True
>>> Optional(Boolean, default=None)("False")
False
```

Do you have an example of it failing?

bsimpson63:
Arg. I was accidentally editing files in the debian directory so the config changes I expected weren't happening. Sorry for the false alarm.

PR #84, opened 2017-04-21
Add memcache serialization/deserialization methods

bsimpson63:
The serde methods are based on those found in pymemcache, but with
added features to make them compatible with pylibmc.

We can use this stuff in listing service, thing service, and r2. I'm not sure that baseplate.context.memcache.lib is the correct place to put the serde methods. Let me know if you think they should go somewhere else (maybe a base level module in baseplate?).

👓 @keithblaha @spladug


baseplate/context/memcache/__init__.py
```diff
@@ -5,11 +5,12 @@
...
 from pymemcache.client.base import PooledClient
 
-from . import ContextFactory
-from .. import config
+from .lib import make_memcache_serializer, memcache_deserializer
+from baseplate.context import ContextFactory
+from baseplate import config
 
 
-def pool_from_config(app_config, prefix="memcache.", **kwargs):
```
spladug: are there no other options that might need to get passed through? it be nice to keep the **kwargs if there are other options (and perhaps even if there aren't so some config can be hard-coded when desired)

baseplate/context/memcache/__init__.py
```diff
@@ -5,11 +5,12 @@
 
 from pymemcache.client.base import PooledClient
 
-from . import ContextFactory
-from .. import config
+from .lib import make_memcache_serializer, memcache_deserializer
+from baseplate.context import ContextFactory
```
spladug: can you keep these explicit-relative please just for consistency?

baseplate/context/memcache/__init__.py
```diff
@@ -41,19 +42,30 @@ def pool_from_config(app_config, prefix="memcache.", **kwargs):
             "max_pool_size": config.Optional(config.Integer, default=None),
             "connect_timeout": config.Optional(config.Float, default=None),
             "timeout": config.Optional(config.Float, default=None),
+            "min_compress_length": config.Optional(config.Integer, default=0),
+            "compress_level": config.Optional(config.Integer, default=0),
+            # TODO: bug in Optional prevents this from being unset
```
spladug: ?

baseplate/context/memcache/__init__.py
```diff
@@ -5,11 +5,12 @@
...
 
-from . import ContextFactory
-from .. import config
+from .lib import make_memcache_serializer, memcache_deserializer
+from baseplate.context import ContextFactory
+from baseplate import config
 
 
-def pool_from_config(app_config, prefix="memcache.", **kwargs):
+def pool_from_config(app_config, prefix="memcache."):
```
spladug: the docstring needs updating.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
+from six import string_types
```
spladug: there's a `_compat` module in the root of baseplate for this stuff. six is not a dependency right now.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+import logging
+import zlib
+
+
+try:
+    import cPickle as pickle
```
spladug: you can grab all these imports safely from the compat module without try/except shenanigans.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+    from cStringIO import StringIO
+except ImportError:
+    try:
+        from StringIO import StringIO
+    except ImportError:
+        # python3
+        from io import StringIO
+
+
+FLAG_PICKLE = 1 << 0
```
spladug: perhaps these would be good as a proper Enum?

(see: https://docs.python.org/3/library/enum.html . a backwards compat shim is imported on py2 in _compat)

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+    elif isinstance(value, long):
+        flags |= FLAG_LONG
+        value = "%d" % value
+    else:
+        flags |= FLAG_PICKLE
+        output = StringIO()
+
+        # use protocol 2 which is the highest value supported by python2
+        pickler = pickle.Pickler(output, protocol=2)
+        pickler.dump(value)
```
spladug: why not `dumps`?

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+    if flags & FLAG_INTEGER:
+        return int(value)
+
+    if flags & FLAG_LONG:
+        return long(value)
+
+    if flags & FLAG_PICKLE:
+        try:
+            buf = StringIO(value)
+            unpickler = pickle.Unpickler(buf)
```
spladug: likewise why not loads?

baseplate/context/memcache/__init__.py
spladug: probably also good to note somewhere in the module docs that we implement serde for you.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+    # an integer a python3 process will still be able to deserialize it.
+
+
+def python_memcache_serializer(key, value):
+    flags = 0
+
+    if isinstance(value, string_types):
+        pass
+    elif isinstance(value, int):
+        flags |= FLAG_INTEGER
```
spladug: why are we ORing when we're never actually combining flags? might as well just assign to be clearer?

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+            unpickler = pickle.Unpickler(buf)
+            return unpickler.load()
+        except Exception:
+            logging.info('Pickle error', exc_info=True)
+            return None
+
+    return value
+
+
+# These serde methods are compatible with pylibmc's implementation.
```
spladug: this comment seems out of place?

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+
+# extra bit flag from pylibmc, just grabbing the one we need which pymemcache's
+# serde.py doesn't have (currently on 1.3.2, but later versions do have this)
+FLAG_ZLIB = 1 << 3
```
spladug: should this live with the other flags?

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def memcache_deserializer(key, value, flags):
+    if flags & FLAG_ZLIB:
+        value = zlib.decompress(value)
+    return python_memcache_deserializer(key, value, flags)
+
+
+def make_memcache_serializer(min_compress_length=0, compress_level=1):
```
spladug: why do we have this two-layer system but then not expose any of it to the user? if we're just gonna force serialization on them, we should collapse these two layers to simplify.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+# These serde methods are compatible with pylibmc's implementation.
+
+# extra bit flag from pylibmc, just grabbing the one we need which pymemcache's
+# serde.py doesn't have (currently on 1.3.2, but later versions do have this)
+FLAG_ZLIB = 1 << 3
+
+
+def memcache_deserializer(key, value, flags):
```
spladug: nit: these functions should have more verby names

baseplate/context/memcache/lib.py
spladug: a module docstring would be nice too. since this is sorta hidden internal stuff we don't need to sphinx it up though.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,108 @@
...
+    # set by python2 processes. If a python3 process writes a large value as
+    # an integer a python3 process will still be able to deserialize it.
+
+
+def python_memcache_serializer(key, value):
+    flags = 0
+
+    if isinstance(value, string_types):
+        pass
+    elif isinstance(value, int):
```
keithblaha: don't know python 2/3 compat issues well, but would this mean that a python3 process would serialize a `long` with the `int` flag, and a python 2 process would try to deserialize that out as an `int` instead? does that matter?

tests/unit/context/memcache_tests.py
```diff
@@ -51,3 +54,153 @@ def test_alternate_prefix(self):
...
+
+    def test_deserialize_str(self):
+        value = memcache_lib.python_memcache_deserializer(
+            key="key", value="val", flags=0)
+        self.assertEqual(value, "val")
+
+    def test_deserialize_int(self):
+        value = memcache_lib.python_memcache_deserializer(
+            key="key", value="100", flags=memcache_lib.FLAG_INTEGER)
+        self.assertEqual(value, 100)
```
keithblaha: type check could be useful too, since `long(100) == int(100)`

baseplate/context/memcache/__init__.py
keithblaha: we make use of kwargs to set serde right now, maybe keeping it in and letting services override this serde would be nice: https://github.com/reddit/reddit-service-listing/blob/a9668dea8ee28e5e1dbb9fbdf7ca52a77e33c994/listing/__init__.py#L113

keithblaha:
Thanks for looping me in- overall this is really useful and it'd be nice to have for listing service. I'm thinking about if we want to enforce that services must use this exact serde or not since there is no way provided to override- and I think that we should, although I think it will be used relatively rarely.

baseplate/context/memcache/__init__.py
bsimpson63: just a note that we need to fix the bug in `config.Optional` before doing an optional boolean config.

baseplate/context/memcache/__init__.py
bsimpson63: Yeah that sounds reasonable, I'll make that change.

baseplate/context/memcache/__init__.py
spladug: oh! i hadn't seen #83.

baseplate/context/memcache/lib.py
bsimpson63: Yeah, a long value stored by python3 would be stored with the `int` flag. From the python2 docs:

> Plain integers (also just called integers) are implemented using long in C, which gives them at least 32 bits of precision (sys.maxint is always set to the maximum plain integer value for the current platform, the minimum value is -sys.maxint - 1). Long integers have unlimited precision.

I guess to be safe we could do a try/except, or just always cast to `long` when deserializing. What do you think?

baseplate/context/memcache/lib.py
bsimpson63: I could see it either way. Doing the OR indicates that's something we might be doing with these flags in general (as we do later on once we compress).

baseplate/context/memcache/lib.py
bsimpson63: This part is copied form pymemcache, and I think `Pickler` might only have `dump()` https://docs.python.org/2/library/pickle.html#pickle.Pickler

baseplate/context/memcache/lib.py
bsimpson63: same reason I guess https://docs.python.org/2/library/pickle.html#pickle.Unpickler

baseplate/context/memcache/lib.py
spladug: sure, but it could be `pickle.dumps(value, protocol=2)` and save the whole dance?

baseplate/context/memcache/lib.py
bsimpson63: I don't feel strongly, but I feel there's some value in keeping the serialization and compression separate.

baseplate/context/memcache/lib.py
bsimpson63: Seems like this isn't a problem actually:
```
>>> import sys
>>> sys.maxint
9223372036854775807
>>> int(sys.maxint + 100)
9223372036854775907L
```
python2 will automatically use `long` if the value is too big.

baseplate/context/memcache/lib.py
bsimpson63: Actually I think you're right, gonna just assign.

baseplate/context/memcache/lib.py
bsimpson63: 👍

bsimpson63:
💇

spladug:
CI's failing. For the "imported but not used" in _compat, you can just add them to `__all__`.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,111 @@
...
+
+Memcached can only store strings, so to store arbitrary objects we need to
+serialize them to strings and be able to deserialize them back to their
+original form. General purpose serialization and deserialization can be
+achieved with pickle_and_compress() and decompress_and_unpickle().
+
+"""
+
+
+FLAG_PICKLE = 1 << 0
```
spladug: I think a comment got lost in the shuffle here, can this be done with the py3.4 style Enum stuff? (there's a shim in _compat for it).

bsimpson63: I don't see that in _compat

spladug: Sorry, misspoke, it's just `import enum` either way but the backwords-compatibility package `enum34` will be present for py2.

https://github.com/reddit/baseplate/blob/master/baseplate/events/queue.py#L24

bsimpson63: Hmm, if I make it an `Enum`, then I'd need to do like `flags |= Flags.ZLIB.value` to get the actual integer value. What do you think about doing:

```
class Flags(object):
    """Memcached client flags

    Flags are an arbitrary 16-bit unsigned integer that the memcache server
    stores along with the data and sends back when the item is retrieved.
    Clients may use this as a bit field to store data-specific information;
    this field is opaque to the server.

    """

    PICKLE = 1 << 0
    INTEGER = 1 << 1
    LONG = 1 << 2
    ZLIB = 1 << 3
```

bsimpson63:
> Did you have any thoughts on pickle/legacy annotation here? I'd really like to discourage greenfield services from using pickle if possible.

I could add a note to the docstring. I think the best thing to do would be to provide an alternative set of serde methods that don't use `pickle`.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,226 @@
...
+    elif flags == Flags.INTEGER:
+        # python3 doesn't have a long integer type, so all integers are written
+        # with Flags.INTEGER. This means that a value written by a python3
+        # client can exceed the maximum integer value (sys.maxint). This
+        # appears to be ok--python2 will automatically convert to long if the
+        # value is too large.
+        return int(serialized)
+    elif flags == Flags.LONG:
+        return long(serialized)
+    else:
```
keithblaha: maybe this should be `elif flags == Flags.JSON`, just in case, and the else clause could either throw an error, or log error and return serialized?

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            return json.loads(serialized)
+        except Exception:
```
keithblaha: This should probably just catch `ValueError` right? Probably don't want to catch any system errors, e.g. `MemoryError`

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,226 @@
...
+            serialized = value
+            flags = 0
+        elif isinstance(value, int):
+            serialized = "%d" % value
+            flags = Flags.INTEGER
+        elif isinstance(value, long):
+            serialized = "%d" % value
+            flags = Flags.LONG
+        else:
+            serialized = json.dumps(value)
```
keithblaha: This can throw `TypeError` if value isn't serializable to JSON, I think expecting callers to handle this is totally reasonable however maybe it should be noted in docstring above that they are expected to possibly handle those errors

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,226 @@
...
+    elif flags == PickleFlags.INTEGER:
+        # python3 doesn't have a long integer type, so all integers are written
+        # with PickleFlags.INTEGER. This means that a value written by a python3
+        # client can exceed the maximum integer value (sys.maxint). This
+        # appears to be ok--python2 will automatically convert to long if the
+        # value is too large.
+        return int(serialized)
+    elif flags == PickleFlags.LONG:
+        return long(serialized)
+    else:
```
keithblaha: same thing as https://github.com/reddit/baseplate/pull/84/files#r113067590

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            return pickle.loads(serialized)
+        except Exception:
```
keithblaha: same thing as https://github.com/reddit/baseplate/pull/84/files#r113067601

baseplate/context/memcache/lib.py
bsimpson63: Yeah, that sounds good.

baseplate/context/memcache/lib.py
bsimpson63: The docs suggest that lots of exceptions could be raised here https://docs.python.org/2/library/pickle.html#pickle.UnpicklingError. I don't feel strongly either, but it's easier and more comprehensive to just catch all exceptions here.

bsimpson63:
💇

baseplate/context/memcache/__init__.py
```diff
@@ -31,6 +32,17 @@ def pool_from_config(app_config, prefix="memcache.", **kwargs):
         socket connected to memcache. Defaults to the underlying socket default
         timeout.
 
+    :param dict app_config: the config dictionary
+    :param str prefix: prefix for config keys
+    :param func serializer: function to serialize values to strings suitable
+        for being stored in memcached. An example is
+        ``baseplate.context.memcache.lib.make_memcache_serializer``
```
spladug: I think you can link these functions with `:py:func:...` for a little more sphinx juice.

baseplate/context/memcache/__init__.py
```diff
@@ -41,19 +53,22 @@ def pool_from_config(app_config, prefix="memcache.", **kwargs):
             "max_pool_size": config.Optional(config.Integer, default=None),
             "connect_timeout": config.Optional(config.Float, default=None),
             "timeout": config.Optional(config.Float, default=None),
+            # TODO: bug in Optional prevents this from being unset
```
spladug: can you link #83 here (or fix it? :)

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,233 @@
+import logging
```
spladug: ah, oops, the `from __future__` things all the other files have should be in here too. it keeps things more py3-friendly.

i think the module docstring goes at the very very top too.

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,233 @@
...
+        return serialized
+
+
+def make_pickle_and_compress_fn(min_compress_length=0, compress_level=1):
+    """Create a serialization method compatible with decompress_and_unpickle().
+
+    The resulting method is a chain of pickling and zlib compression.
+
+    This serializer is compatible with pylibmc.
+
```
spladug: can you add a `.. warning:: Please don't friggin use this` kind of thing here too please?

baseplate/context/memcache/lib.py
spladug: i should get a lint in CI for this.

baseplate/context/memcache/lib.py
spladug: :no_bell: Another option re: serialization, is thrift. Can even turn thrift objects into JSON if desired:

```python
transport = TMemoryBuffer()
protocol = TJSONProtocol(transport)
some_object_to_serialize.write(protocol)
return transport.getvalue()
```

baseplate/context/memcache/lib.py
```diff
@@ -0,0 +1,233 @@
...
+import zlib
+
+from ..._compat import (
+    long,
+    pickle,
+    string_types,
+)
+
+
+"""Memcache serialization/deserialization (and compression) helper methods.
```
spladug: if you want this module to get a page in the docs, you'll need to add a new or update an existing `.rst` in the `doc/` subtree.

baseplate/context/memcache/lib.py
bsimpson63: Ooh, that'd be slick. I'll play around with that.

bsimpson63:
Any thoughts on where the serde methods should live? Seems fine in baseplate/context/memcache/lib.py but I could also see putting them in like baseplate/utils/memcache.py or something more general like that.

spladug:
`baseplate/context/memcache/serde.py` ?

baseplate/context/memcache/lib.py
bsimpson63: I'm working on doing this within the thing service, but I don't think it's feasible to build it into this layer because the deserialization needs to know the thrift class and we can't pass that into the deserializing method. It's not too bad to do it in the application rather than in the memcache client.

spladug: > deserialization needs to know the thrift class

ah, yeah, you're totally right. my bad.

baseplate/context/memcache/__init__.py
bsimpson63: False alarm, this actually works as expected.

bsimpson63:
💇 I can't tell if my build failure is because of spellchecker or something else. Any idea how to fix that?

spladug:
Yup, it's the spellchecker. For any spelling errors: fix 'em! For legit but unrecognized words: add 'em to the dictionary https://github.com/reddit/baseplate/blob/master/docs/words.txt

bsimpson63:
Build is passing now. I couldn't really figure out where some of the spelling errors were coming from, the output was like:

```preparing documents... done
baseplate/context/index py seplate/context/index                                
baseplate/context/index func 
writing output... [100%] quickstart                                             
Spelling checker messages written to /home/vagrant/baseplate/build/spelling/output.txt
build finished with problems.
make: *** [tests] Error 1
```

and "py" and "func" only existed as ":py:func:" within a docstring.

bsimpson63:
Merged as https://github.com/reddit/baseplate/commit/7cc47cdbe14fe83e4c76549332e775d956005430

PR #85, opened 2017-04-27
Cqlmapper support

pacejackson:


baseplate/context/cqlmapper.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from cqlmapper.connection import Connection as CQLConnection
```
spladug: I always get kinda confused by these import-and-rename things. How about just doing `import cqlmapper.connection` and using `return cqlmapper.connection.Connection` if you want it scoped or just go straight up unmangled `Connection`.

baseplate/context/cqlmapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+from .cassandra import CassandraContextFactory
+
+
+class CQLMapperContextFactory(CassandraContextFactory):
```
spladug: would you be OK just putting this in the `cassandra` module so it's all in one place? that might mean putting the import in the method call so it's not required for non-cqlmapper clients, but that's OK.

pacejackson:
screen shot 2017-04-27 at 5 27 12 pm

pacejackson:
:haircut_man:

PR #88, opened 2017-05-07
Consistently use RawConfigParser; disable interpolation

spladug:
We're not currently using interpolation anywhere and this frees up the %
character for use in the Percent config parser and reduces suprise.

### DIFF
--- a/baseplate/config.py
+++ b/baseplate/config.py
@@ -22,7 +22,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     from baseplate import config
     from baseplate._compat import configparser
     from tempfile import NamedTemporaryFile
-    config_parser = configparser.ConfigParser()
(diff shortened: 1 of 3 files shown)


ckwang8128:
:fish: Code looks good; I wasn't sure where the best place for a comment was (or maybe it should just be with each of the three changes) but I think one would be good so we don't forget why we did this

spladug:
Is the commit message sufficient recording of rationale? Or maybe the "main" one in server/__init__.py needs a comment?

ckwang8128:
I'm worried it'll get overridden in the future, a comment block has more immediacy and staying power; the main one sounds good!

spladug:
:haircut_man:

ckwang8128:
:fish:

spladug:
99f5a609f007bbedac71cd280b229314633d6584

PR #89, opened 2017-05-12
Vault

spladug:
This adds support for secrets stored in Hashicorp Vault. What _is_ here should be heavily documented in the diff itself, so I'll defer to that rather than writing more here in the PR body. This is all rather much, so please do tear it apart and let me know if anything is confusing, misleading, dangerous, insecure, etc.

Open questions:

* monitoring of fetcher daemon: The daemon will run as an appcommon::daemon which means we'll get log shipping and QA checks for free. There's not any facility for alerting on sudden failure down the line though. 

Future work:

* [HVAC](https://github.com/ianunruh/hvac) integration. The integration should take the vault_token from this stuff and allow full in-app access to vault's more interesting features (like the transit backend). All with diagnostics instrumentation, of course!
* [Database credentials](https://www.vaultproject.io/docs/secrets/databases/index.html): The database helpers should get integration with the Vault stuff so that apps can automatically fetch rotating DB credentials.

### DIFF
--- a/baseplate/crypto.py
+++ b/baseplate/crypto.py
@@ -1,5 +1,32 @@
 """Utilities for common cryptographic operations.
 
+.. testsetup::
+
+    import datetime
+    from baseplate.crypto import make_signature, validate_signature
+    from baseplate.secrets import SecretsStore
+
+    secrets = SecretsStore("docs/secrets.json")
+
+.. testcode::
+
+    message = "Hello, world!"
+
+    secret = secrets.get_versioned("some_signing_key")
+    signature = make_signature(
+        secret, message, max_age=datetime.timedelta(days=1))
+
+    try:
+        validate_signature(secret, message, signature)
+    except SignatureError:
+        print("Oh no, it was invalid!")
+    else:
+        print("Message was valid!")
+
+.. testoutput::
+
+    Message was valid!
+
 
 """
 from __future__ import absolute_import
@@ -14,6 +41,9 @@ import hashlib
 import hmac
 import struct
 import time
+
+from baseplate._utils import warn_deprecated
+from baseplate.secrets import VersionedSecret
 
 
 if hasattr(hmac, "compare_digest"):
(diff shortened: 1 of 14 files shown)


baseplate/secrets/fetcher.py
```diff
@@ -0,0 +1,282 @@
...
+        "expiration": "2017-05-05T09:42.143:35Z",
+        "secrets": {
+            "secret/one": {...},
+            "secret/two": {...},
+            "secret/three": {...}
+        },
+        "vault_token": "9da4241c-3460-11e7-84ac-0e9f9d32522f"
+    }
+
+The vault_token can be used for direct communication with Vault using the
```
m104: Is the vault_token going to be per-instance, app, or role? In other words, can I piggy back off of this token to store cubbyhole data that is accessible across the app?

spladug: Per instance and rotating periodically.

m104: great, thanks!

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+
+    """
+
+    def __init__(self, path):
+        self._path = path
+        self._expiration = datetime.datetime.min
+        self._secrets = None
+        self._vault_token = None
+
+    def _load_if_expired(self):
```
spladug: Should we force a reload if the mtime is higher than when we last read the file even if not expired yet? That would mean extra `stat` calls, but that should be in-cache most of the time. The upside would be that we are sure we get the latest copy if someone restarts the fetcher (e.g. for new configs). Alternatively, we punt and say "if you really need new data, send an app restart".

ckwang8128: I think it makes sense to force a reload if we notice "newness"; a big part of this is abstracting the process of secret discovery from the app, needing to force a restart for cases like new configs feels antithetical to that goal. The only case where I'd be against is, like you pointed out, if there's major performance hits from the extra `stat`s.

ckwang8128:
@:fish: from a code and structure perspective. Non-blocking question: How does secret revocation look like in this pipeline, specifically with respect to the VersionedSecret implementation here (this is from the perspective of someone who doesn't know a ton about Vault details)? Will the `previous` attribute on the secret just be expected to change/be removed if a secret is explicitly removed and replaced?

spladug:
> How does secret revocation look like in this pipeline, specifically with respect to the VersionedSecret implementation here ... Will the previous attribute on the secret just be expected to change/be removed if a secret is explicitly removed and replaced?

Regular rotation would be like 

* read secret
* `new_value["previous"] = old_value["current"]`
* `new_value["current"] = old_value["next"]`
* `new_value["next"] = generate_new_secret()`
* write secret.

Revocation would depend on how seriously we need to revoke: in a hard cutover situation, it'd just be replacing the dictionary with `current = generate_new_secret()` and let the other two be null.

ckwang8128:
SGTM, just making sure I understand the process!

baseplate/crypto.py
```diff
@@ -92,89 +122,97 @@ class SignatureInfo(_SignatureInfo):
...
-            message, max_age=datetime.timedelta(days=1))
+    """
+    version = 1
+    expiration = int(time.time() + max_age.total_seconds())
+    header = _HEADER_FORMAT.pack(version, expiration)
+    digest = _compute_digest(secret.current, header, message)
+    return base64.urlsafe_b64encode(header + digest)
+
+
+def validate_signature(secret, message, signature):
```
JordanMilne: Probably best saved for another PR, but it'd be nice to have a wrapper like this around cryptography.io's [`MultiFernet`](https://cryptography.io/en/latest/fernet/#cryptography.fernet.MultiFernet) that can accept a `VersionedSecret`

baseplate/crypto.py
```diff
@@ -92,89 +122,97 @@ class SignatureInfo(_SignatureInfo):
...
+    :param datetime.timedelta max_age: The amount of time in the future
+        the signature will be valid for.
+    :return: An encoded signature.
 
-        signer = MessageSigner(b"supersecret")
-        signature = signer.make_signature(
-            message, max_age=datetime.timedelta(days=1))
+    """
+    version = 1
+    expiration = int(time.time() + max_age.total_seconds())
```
spladug: Mayyyybe, but if so I'd want you to have to opt into it really noisily. Trying to make it hard to do the wrong thing in the helpers that go into this lib. Do you have any use cases in mind? (And perhaps let's keep that for another PR since this is just reworking the helpers to use versioned secrets?)

baseplate/crypto.py
spladug: Yeah! Would love to get more helpers in that take advantage of the versioned things.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def __str__(self):  # pragma: nocover
+        return "{}: {}".format(self.path, self.message)
+
+
+_VersionedSecret = collections.namedtuple(
+    "VersionedSecret", "previous current next")
+
+
+class VersionedSecret(_VersionedSecret):
```
spladug: Negative. This is purely our convention. I'd like to get a cron job that rotates keys going but that'd be vault-side so apps don't even need to know it exists as long as they can handle versioned secrets OK.

baseplate/secrets/fetcher.py
```diff
@@ -0,0 +1,282 @@
...
+        response.raise_for_status()
+        payload = response.json()
+        return payload["data"], ttl_to_time(payload["lease_duration"])
+
+
+def octal_integer(text):
+    return int(text, base=8)
+
+
+def user_name(text):
```
spladug: Hm, yeah, it looks a bit weird when doing it like this but I'm going for the declarative style of the rest of the config helpers. Maybe I should capitalize like they do? or just ditch it?

baseplate/crypto.py
spladug: Also thanks for linking that. It makes me feel better to see their implementation of multi-version on the read path is the same. I was a bit worried there'd be some timing attack concern I wasn't thinking about.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+        :rtype: :py:class:`VersionedSecret`
+
+        """
+        secret_attributes = self.get_raw(path)
+
+        if secret_attributes.get("type") != "versioned":
+            raise CorruptSecretError(path, "secret does not have type=versioned")
+
+        previous = secret_attributes.get("previous")
+        next = secret_attributes.get("next")
```
spladug: Yes, yucky, thanks. Fixed.

baseplate/secrets/fetcher.py
spladug: Renamed them to be clearer.

spladug:
(rebased to deal with merge conflicts. changes since first review pass are still in the lower-case fixup commits)

baseplate/secrets/store.py
spladug: OK, done!

spladug:
:haircut_man:

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+    cache or store the values returned by this class's methods but rather get
+    them from this class each time you need them. The secrets are served from
+    memory so there's little performance impact to doing so and you will be
+    sure to always have the current version in the face of key rotation etc.
+
+    """
+
+    def __init__(self, path):
+        self._path = path
+        self._expiration = datetime.datetime.min
```
dellis23: Wow, TIL about `datetime.datetime.min`

baseplate/secrets/store.py
dellis23: Obviously late to this, but yeah, I like the idea of checking `mtime` but I wonder if this would get funky with a service that starts to need to check secrets many thousands of times per second.  We could do a check at most once per second or something like that.  I wonder if that's unrealistic though.  Would secrets mostly be grabbed for external calls and it be unlikely that we'd be grabbing them that frequently?  I'd bet it'd be cached in the kernel and would be a cheap call though.  Just thinking out loud.  Probably all pre-mature optimization.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+        self._secrets = None
+        self._vault_token = None
+
+    def _load_if_expired(self):
+        if self._expiration < datetime.datetime.utcnow():
```
dellis23: I haven't seen the later commit yet, but I wonder if, since we're presumably doing the mtime method, we even need to also do it this way!  Would make the code cleaner at least.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+
+            with open(self._path) as f:
+                raw_data = json.load(f)
+
+            self._expiration = datetime.datetime.strptime(
+                raw_data["expiration"], ISO_FORMAT)
+            self._vault_token = raw_data["vault_token"]
+            self._secrets = raw_data["secrets"]
+
+    def make_object_for_context(self, name, server_span):  # pragma: nocover
```
dellis23: What's up with this?

baseplate/secrets/store.py
spladug: Hmm, that's a good point. There's no real need to keep doing the expiration check if the mtime is being checked: even if the secrets _are_ expired, if the mtime hasn't changed we're not gonna magically get new secrets. I'll tear the expiration stuff out. Thanks!

baseplate/secrets/store.py
spladug: It lets you do

```
baseplate = Baseplate()
secrets = SecretsStore(...)
baseplate.add_to_context("secrets", secrets)
```

and then you get the secrets store on your context object. Similar to how the event queues work.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+        Versioned secrets are a convention of key/value pairs in the raw secret
+        payload. The following keys are significant:
+
+        ``type``
+            This must always be ``versioned`` for this method.
+        ``current``, ``next``, and ``previous``
+            The raw secret value's versions. ``current`` is the "active"
+            version, which is used for new creation/signing operations.
+            ``previous`` and ``next`` are only used for validation (e.g.
+            checking signatures) to ensure continuity when keys rotate. Both
```
dellis23: Hmm, maybe you can explain this to me in person, but I don't understand what the usefulness of keeping the previous version is if it's not actually valid for doing the operation (i.e. I'm assuming you mean it instantly expires when rotation happens.  If the client for whatever reason hasn't gotten that update yet, does the previous token then expire as well?

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+            ``previous`` and ``next`` are optional.
+        ``encoding``
+            (Optional) If present, how to decode the values from how they are
+            encoded at rest (Only ``base64`` currently supported).
```
dellis23: super super nit: that capital `Only`.

baseplate/secrets/store.py
spladug: Check out how it's used in the message signer for when the previous/next values come into play.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+        if secret_attributes.get("type") != "versioned":
+            raise CorruptSecretError(path, "secret does not have type=versioned")
+
+        previous = secret_attributes.get("previous")
+        next = secret_attributes.get("next")
+
+        try:
+            current = secret_attributes["current"]
+        except KeyError as exc:
+            raise CorruptSecretError(path, "secret does not have %s" % exc)
```
dellis23: I could see this error being confusing `secret does not have current`.  Maybe explicitly specifying that it's the key missing there that caused the error to be raised?

baseplate/secrets/store.py
spladug: Done.

baseplate/secrets/store.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        return cls(
+            previous=None,
+            current=value,
+            next=None,
+        )
+
+
+def _decode_secret(path, encoding, value):
+    if encoding == "identity":
```
dellis23: Why is this called `identity`?  Is it used for something special that makes it tend to not be random bytes that need to be encoded?  Also, I've noticed it is the default if no encoding is specified, so it almost feels like just a default non-encoding?  I guess I'm just not sure what it's for.

baseplate/secrets/store.py
spladug: Added a docstring.

baseplate/secrets/store.py
spladug: it's called big-O notation for a reason! 

(fixed)

baseplate/secrets/store.py
spladug: Yup, the identity encoding. Like multiplying by 1.

> (algebra, computing) Any function which maps all elements of its domain to themselves.

docs/baseplate/secrets.rst
```diff
@@ -0,0 +1,41 @@
...
+--------------
+
+The secret fetcher is a sidecar that is run as a single daemon on each server.
+It authenticates to Vault as the server itself and gets appropriate policies
+for access to secrets accordingly. Once authenticated, it fetches a given list
+of secrets from Vault and stores all of the data in a local file. It will
+automatically re-fetch secrets as their leases expire, ensuring that key
+rotation happens on schedule.
+
+Because this is a sidecar, individual application processes don't need to talk
```
dellis23: Cool design btw.

baseplate/secrets/store.py
spladug: GP on the exception message being unhelpful. And whoops, that weird parameterization was from an earlier version. Cleaned up.

baseplate/secrets/store.py
dellis23: Ahh yes, thanks.  Makes sense.

baseplate/secrets/store.py
```diff
@@ -244,6 +241,18 @@ def get_versioned(self, path):
             next=next_value and _decode_secret(path, encoding, next_value),
         )
 
+    def make_object_for_context(self, name, server_span):  # pragma: nocover
+        """Return an object that can be added to the context object.
```
dellis23: Thanks!

dellis23:
🐟 Thanks for the overview

ckwang8128:
:fish:

PR #90, opened 2017-05-14
Sentry

spladug:
This:

* Fixes the Thrift integration's capturing of unexpected exceptions into the `server_span.finish()` stuff and adds tests to ensure everything works properly. I did this instead of blocking the project any further on the FBThrift changes (though I'd still like to do those at some point).
* Adds support for Python Raven based Sentry error reporting.
* Fixes a couple of intersphinx issues.


baseplate/__init__.py
```diff
@@ -93,8 +96,90 @@ def make_tracing_client(raw_config, log_if_unconfigured=True):
     )
 
 
+def crash_reporter_from_config(raw_config, module_name):
```
spladug: I'm not at all convinced of taking the module name here. The name is maybe a little nicer than a straight file path, since potentially this function could do fancy stuff down the line with the module name and the `include_paths` directive. Is there a better/nicer way to do this though?

Makefile
```diff
@@ -17,6 +17,13 @@ $(THRIFT_BUILDDIR)/baseplate/thrift/baseplate.thrift_buildstamp: baseplate/thrif
 	cp -r $(THRIFT_BUILDDIR)/$</baseplate/thrift/ baseplate/
 	touch $@
 
+$(THRIFT_BUILDDIR)/tests/integration/test.thrift_buildstamp: tests/integration/test.thrift
+	@echo SPECIAL $< $@
```
dellis23: Is some of this a makefile specific thing, or do i just know even less bash than I thought I did? (particularly `@echo`)

spladug: All make-isms. [@ suppresses echoing the "recipe"](https://www.gnu.org/software/make/manual/html_node/Echoing.html#index-_0040-_0028in-recipes_0029). The $< and $@ are [make automatic variables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html).

dellis23: Neat, thanks.

baseplate/__init__.py
dellis23: Maybe it's because I haven't dealt with sentry a lot, but what is ``module_name`` being used for here?  Is it just what will be reported back in the error message?  Or will it be used to generate extra data like the surrounding lines for display on the web interface?  Should ``__name__`` always be used as the argument?

baseplate/__init__.py
```diff
@@ -93,8 +96,90 @@ def make_tracing_client(raw_config, log_if_unconfigured=True):
...
+        Comma-delimited list of module prefixes to ignore when discovering
+        where an error came from.
+    ``sentry.include_paths`` (optional)
+        Comma-delimited list of paths to include for consideration when
+        drilling down to an exception.
+    ``sentry.ignore_exceptions`` (optional)
+        Comma-delimited list of fully qualified names of exception classes
+        (potentially with * globs) to not report.
+    ``sentry.sample_rate`` (optional)
+        Percentage of crashes to report. (e.g. 37%)
```
dellis23: Is this just an int?  Or should it be a decimal?

ckwang8128: This is an optional baseplate.config.Percent (https://reddit.github.io/baseplate/baseplate/config.html#baseplate.config.Percent), which can take a decimal/float percentage string.

baseplate/core.py
```diff
@@ -190,6 +190,19 @@ def configure_tracing(self, tracing_client, *args, **kwargs):
 
         self.register(TraceBaseplateObserver(tracing_client))
 
+    def configure_crash_reporting(self, client):  # pragma: nocover
```
ckwang8128: would `error` be a better word instead of `crash` here and throughout? `crash` seems a bit off of the main language used for this kind of thing ("error", "exception").

ckwang8128:
:nail_care: looks good, just a few questions!

baseplate/__init__.py
dellis23: Yeah, scrolling down showed that, but I don't think it'd be super obvious just by reading the docstring.

dellis23:
🐟 Some docs questions, other than that, looks great!  This'll be super useful.

baseplate/diagnostics/sentry.py
```diff
@@ -0,0 +1,48 @@
...
+class SentryServerSpanObserver(ServerSpanObserver):
+    def __init__(self, raven):
+        self.raven = raven
+
+    def on_start(self):
+        self.raven.context.activate()
+
+    def on_set_tag(self, key, value):
+        if key.startswith("http"):
+            self.raven.http_context({key[len("http."):]: value})
```
spladug: Yup!

baseplate/__init__.py
spladug: Right now it's being used to figure out where the git repo is for doing `raven.fetch_git_sha` on for the `release` tag in Sentry. I used `__name__` instead of something else since I thought it might be more future-proofed for something like (from OP):

>  since potentially this function could do fancy stuff down the line with the module name and the include_paths directive

baseplate/__init__.py
spladug: I added quotes around the 37% to indicate that's the literal syntax. Does that help at all?

baseplate/core.py
spladug: Fair enough. I just wanted to distinguish between handleable errors and fatal errors.

baseplate/__init__.py
dellis23: Yup, that's exactly what I was thinking.

spladug:
:haircut_man:

dellis23:
🐟

baseplate/__init__.py
ckwang8128: I think it's okay to keep as is, I can't think of anything in particular that makes this significantly cleaner.

ckwang8128:
:fish:

spladug:
2b67217f67dfc9845d7a25527e45f6388e332478...c3032e3e3e9ebad2d3e4a5c8a419438037ea4766

Issue #73, opened 2017-01-18
Automate documentation building and publishing to gh-pages branch here


spladug:

spladug:
Docs are moved over to RTD now and build automatically!

PR #92, opened 2017-05-16
Switch to ReadTheDocs for documentation

spladug:
This allows the build process to be automated without giving
write-access credentials to CI. Additionally, it lets us have multiple
versions of docs live for when multiple releases are around.

Also a quick place I missed while doing the sentry stuff that I noticed
while doing this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,9 +16,9 @@ including:
 -  and various helper libraries like a Thrift client pool
 
 Check out the `quick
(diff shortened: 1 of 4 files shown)


spladug:
I chose "stable" over "latest" because that'll always point to the current latest _release_ rather than current latest _commit_.

spladug:
This'll take care of #73

ckwang8128:
:fish:

PR #94, opened 2017-05-23
Add trace ID to sentry context

spladug:
This allows correlation of errors with logs and other tools, and might
be helpful when looking at how errors propagate across the system.


dellis23:
🐟

Issue #100, opened 2017-05-30
Fix race condition in Cassandra span reporting


spladug:
The Cassandra driver is asynchronous under the hood. It runs IO in a reactor on another thread (or in the native reactor if already async, e.g. gevent). It provides an API that returns Futures. You can either attach callbacks to the future to get notified of completion/failure or call `result()` on the future to block the current thread until the result comes in. Callbacks are called on the IO thread or in the registering thread if the future had already completed at time of registration. The `result()` API works by blocking on a `threading.Event` object. When the operation finishes inside the IO thread, [the future signals the `Event` and then calls each of the callbacks in turn](https://github.com/datastax/python-driver/blob/fda90ccefa72f529493451b7afa7eeab4845471b/cassandra/cluster.py#L3746-L3750) in that order.

Baseplate's Cassandra instrumentation intercepts Futures and attaches callbacks to them before returning them to the application. This is how we mark spans finished. The `execute_async()` API works this way by default and the `execute()` API is just `execute_async()` with an immediate call to `result()` on the returned future.

Because the IO thread signals the event which unleashes the main thread's `result()` before doing the callbacks, there's a race condition where the main thread finishes up its work and finishes the server span before the callbacks and therefor the child span get to finish.

The impact of this is that a child span might finish after its parent span sometimes. For most observers this should be fine. The metrics observer will lose that child span's metric though because the batch for the server span will have already shipped. This means undercounting Cassandra calls and timings.

spladug:
I opened a speculative PR (see above) against the upstream driver and got some feedback/suggestions. Not sure the best way to fix this right now. It's also causing us issues on r2 when we have the worker threads try to send metrics.

spladug:
Fixed by the venerable @bsimpson63 in d366da31a95b69e2509f24c5b8538630ce0d6923

Issue #99, opened 2017-05-29
SQLAlchemy instrumentation doesn't handle local spans


spladug:
It's currently using a ServerSpan level threadlocal internally to pass state around which doesn't work with local spans. This means that the parentage of client spans made ostensibly inside local spans incorrectly points at the server span instead.

PR #101, opened 2017-05-30
Clean up span error reporting and add a bunch of integration tests

spladug:
The goal here was to get more consistent about error spans. IO-1783 asked for a counter for success/failure rates and [the OpenTracing semantic conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md) call for an `error=true` tag on spans that didn't succeed. 

While getting that set up, I wanted to make it easier to get right when writing integrations, so I tried to centralize it into the metrics and tracing observer respectively. That made it clear that we weren't calling `finish()` with proper `exc_info` in several of the observers. 

Sooo, I fixed things up and took a crack at #54 and wrote a bunch of integration tests for the instrumentation libraries to ensure that we're doing it right from now on. I also found a couple of other bugs in the process, but they're a bit gnarly so they're just ticketed up as #99 and #100 for now.

Note: the Cassandra integration tests aren't running on Travis right now. AFAICT that's because the Cassandra service doesn't install/run right on the Trusty image we're using for Travis. That's a bummer, but at least the tests skip properly.


tests/integration/__init__.py
```diff
@@ -1,6 +1,27 @@
+import socket
+import unittest
+
 from baseplate.core import BaseplateObserver, SpanObserver
 
 
+def skip_if_server_unavailable(name, port):
```
bsimpson63: nice

tests/integration/cassandra_tests.py
```diff
@@ -0,0 +1,105 @@
...
+import logging
+import time
+import unittest
+
+try:
+    from cassandra import InvalidRequest
+    from cassandra.cluster import Cluster
+    from cassandra.query import dict_factory, named_tuple_factory
+except ImportError:
+    raise unittest.SkipTest("cassandra-driver is not installed")
```
bsimpson63: Might not be any way around it, but are you concerned that people could unintentionally miss errors because of this? Could `cassandra-driver` be a required dependency?

tests/integration/memcache_tests.py
```diff
@@ -0,0 +1,59 @@
...
+from baseplate.core import Baseplate
+
+from . import TestBaseplateObserver, skip_if_server_unavailable
+from .. import mock
+
+
+skip_if_server_unavailable("memcached", 11211)
+
+
+class RedisIntegrationTests(unittest.TestCase):
```
bsimpson63: redis

baseplate/diagnostics/metrics.py
```diff
@@ -30,31 +30,53 @@ def __init__(self, client):
...
+        self.base_name = "clients." + span.name
+        self.timer = batch.timer(self.base_name)
+
+    def on_start(self):
+        self.timer.start()
+
+    def on_finish(self, exc_info):
+        self.timer.stop()
+        suffix = "success" if not exc_info else "failure"
+        self.batch.counter(self.base_name + "." + suffix).increment()
```
bsimpson63: I forget how the namespaces work in spans. For calls to other baseplate services will the resulting metric include the service name and endpoint, or just the service name, or something else?

bsimpson63:
💅

baseplate/diagnostics/metrics.py
spladug: Yup, it'll include it. The span names are up to the various integrations but some examples are:

* [`server.count_multi`](https://github.com/reddit/baseplate/blob/master/baseplate/integration/thrift/__init__.py#L75)
* [`clients.activity_service.count_multi`](https://github.com/reddit/baseplate/blob/master/baseplate/context/thrift.py#L91)
* [`clients.memcache.get`](https://github.com/reddit/baseplate/blob/master/baseplate/context/memcache/__init__.py#L114)

tests/integration/memcache_tests.py
spladug: whoops, fixed. thanks.

tests/integration/cassandra_tests.py
spladug: Yeah, I'm not sure what the right tack to take here is. This protects us in weird build environments like launchpad but at the cost of potentially hiding errors. The skips are at least loudly announced in the output of test run.

The dev environment (as set up by vagrant+puppet) will have a C* and all the relevant libs so things shouldn't get missed if people are running tests locally. I am a bit concerned that I haven't got C* running in Travis though.

ckwang8128:
this is awesome, thanks for improving the integration testing so much! :fish:

.travis.yml
```diff
@@ -15,16 +19,16 @@ virtualenv:
...
+    - curl -L https://debian.datastax.com/debian/repo_key | sudo apt-key add -
     - sudo apt-get update -q
-    - sudo apt-get install -y python-fbthrift python3-fbthrift fbthrift-compiler python-sphinx python-sphinxcontrib.spelling python-alabaster python-posix-ipc python3-posix-ipc python-enum34 python-flake8 pylint python-redis python-gevent python-pymemcache python3-pymemcache
-
-install:
-    - pip install -e ".[pyramid]"
-    - pip install webtest coverage
+    - sudo apt-get install -y cassandra=2.2.8 python python3 fbthrift-compiler make pylint python-flake8 python3-cassandra python3-coverage python3-cqlmapper python3-fbthrift python3-gevent python3-nose python3-posix-ipc python3-pymemcache python3-redis python3-requests python3-setuptools python3-sqlalchemy python3-webtest python-coverage python-cqlmapper python-fbthrift python-gevent python-nose python-posix-ipc python-pymemcache python-pyramid python-redis python-requests python-setuptools python-sqlalchemy python-webtest python-alabaster python-sphinx python-sphinxcontrib.spelling python-enum34 python-mock
```
pacejackson: 🔕  I know in you can break this one line apart like

```
    - sudo apt-get install -y cassandra=2.2.8 
        python python3 
        fbthrift-compiler 
        make 
        pylint 
        python-flake8 
        ...
```

If that works with Travis it might be worth doing to make it easier to keep track of updates to this.

tests/integration/cassandra_tests.py
spladug: Fixed C* in travis.

spladug:
:haircut_man: cool, cassandra's working now and i split across lines to make things cleaner in travis.yml (i'd love to dedupe that with puppet if anyone's got suggestions) thanks for looking!

bsimpson63:
🐟

PR #104, opened 2017-06-01
WeightedLottery should allow samples as long as the source list

ketralnis:
This is equivalent to a full weighted shuffle

👓  @spladug

### DIFF
--- a/baseplate/random.py
+++ b/baseplate/random.py
@@ -77,7 +77,7 @@ class WeightedLottery(object):
         :param int sample_size: The number of items to sample from the lottery.
 
         """
-        if not 0 <= sample_size < len(self.items):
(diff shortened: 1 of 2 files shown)


ketralnis:
Hrm yeah, and I'll be doing this on fairly large lists (~1k to ~25k). I probably need to approach a little differently

spladug:
Could do this trick? https://github.com/python/cpython/blob/master/Lib/random.py#L304-L309

spladug:
@ketralnis should this be closed out?

ketralnis:
Yeah, I think it's a good idea to close. If people bump up against it they're probably going to have the same efficiency trouble with it I did so working transparently is probably worse

PR #105, opened 2017-06-06
HVAC client instrumentation

spladug:
This adds HVAC instrumentation to Baseplate. The HVAC client is created in each server span using the secrets store to fetch a current Vault token. The client does share a connection pool, so TLS setup should be amortized. This comes with integration tests and docs too.


baseplate/context/hvac.py
```diff
@@ -0,0 +1,119 @@
...
+class InstrumentedHvacClient(hvac.Client):
+    def __init__(self, url, token, timeout, session, context_name, server_span):
+        self.context_name = context_name
+        self.server_span = server_span
+
+        super(InstrumentedHvacClient, self).__init__(
+            url=url, token=token, timeout=timeout, session=session)
+
+    # this ugliness is us undoing the name mangling that __request turns into
+    # inside python. this feels very dirty.
```
pacejackson: What if instead of instrumenting the hvac client itself, we instrumented the Session object?

spladug: I love it. Will do.

spladug: So I gave this a shot and it does two nice things: it gets rid of some ugly hacks like this one, and it gets us proper span tagging on failure status codes. However, it means that we don't know if the response was considered a failure by hvac without duplicating hvac's error code (check status code, generating the right exception for tagging, etc). So, I'm leaning towards keeping the current implementation right now. What do you think?

spladug:
:haircut_man:

baseplate/context/hvac.py
pacejackson: I think keeping it like this is fine, it's ugly but it's also not something we're not _really_ going to be interacting with this.

tests/integration/hvac_tests.py
```diff
@@ -0,0 +1,71 @@
...
+from baseplate.core import Baseplate
+
+from . import TestBaseplateObserver, skip_if_server_unavailable
+from .. import mock
+
+
+skip_if_server_unavailable("vault", 8200)
+
+
+class CassandraTests(unittest.TestCase):
```
pacejackson: CassandraTests

spladug:
:haircut_man: i moved the url from config to the output of the fetcher. squashed it all together, sorry.

baseplate/secrets/fetcher.py
```diff
@@ -267,8 +270,13 @@ def main():
...
 
             json.dump({
-                "vault_token": client.token,
                 "secrets": secrets,
+                "vault": {
+                    "token": client.token,
+                    "url": cfg.vault.url,
+                },
+                # this is here for backwards compatibility
+                "vault_token": client.token,
```
pacejackson: What does leaving this here actually do?  The implementation doesn't seem to check this if ['vault']['token'] is not present.

baseplate/secrets/store.py
```diff
@@ -255,6 +240,33 @@ def get_versioned(self, path):
...
+        Vault role. This is only necessary if talking directly to Vault.
+
+        :rtype: :py:class:`str`
+
+        .. seealso:: The :py:mod:`baseplate.context.hvac` module provides
+            integration with HVAC, a Vault client.
+
+        """
+        self._load_if_needed()
+        return self._data["vault"]["token"]
```
pacejackson: should this be checking `self._data['vault_token']` if this raises a `KeyError`?

baseplate/secrets/fetcher.py
spladug: I've updated the comment to explain my intent a little more. Does that make sense?

baseplate/secrets/store.py
spladug: I wasn't planning on it, instead going for the update path explained in the other comment, but could be convinced it'd be worthwhile.

spladug:
:haircut_man:

baseplate/secrets/store.py
pacejackson: I think that's fine... the only thing it could break this way is dev environments which can be fixed pretty easily

spladug:
f48c8cf070d881942f9ce30e560d99b9dd9e1dce...89e1cd70d2e7fdcbb271c84ad053b78a24d83203

PR #106, opened 2017-06-09
Polish the documentation some

spladug:
The goal here was to wipe out the custom wordlist for the spellchecker
and then see what was actually out there. A few things came out of this:

* Put bits of code into ``code blocks`` more consistently so we don't
  need to put e.g. module names into the dictionary.
* Choose consistent capitalizations for things like gevent (based on
  what upstream calls itself).
* Reword things that were using awkward, abbreviated, or made up words.
* Fix broken reStructuredText syntax that the spellchecker picked up
  as words.


baseplate/context/memcache/__init__.py
```diff
@@ -96,10 +95,10 @@ class MonitoredMemcacheConnection(PooledClient):
     """Memcache connection that collects diagnostic information.
 
     This connection acts like a
-    `PooledClient <https://pymemcache.readthedocs.io/en/latest/apidoc/pymemcache.client.base.html#pymemcache.client.base.PooledClient>`_
-    except that operations are wrapped with diagnostic collection. Some methods
-    may not yet be wrapped with monitoring, please request assistance if any
-    needed methods are not being monitored.
+    :py:class:`~pymemcache.client.base.PooledClient` except that operations are
+    wrapped with diagnostic collection. Some methods may not yet be wrapped
+    with monitoring, please request assistance if any needed methods are not
```
gtaylor: Period instead of comma between monitoring

baseplate/context/memcache/lib.py
```diff
@@ -44,13 +44,16 @@ class Flags(object):
 
 
 def decompress_and_load(key, serialized, flags):
-    """Deserialization method compatible with dump_and_compress().
+    """Modern deserialization method.
```
gtaylor: I'm lacking some context, but what makes this Modern? What is the un-modern equivalent? Something with memcache protocol/clients?

baseplate/context/memcache/lib.py
```diff
@@ -152,16 +159,17 @@ class PickleFlags(object):
 def decompress_and_unpickle(key, serialized, flags):
     """Deserialization method compatible with pickle_and_compress().
 
+    .. warning:: Please don't use this! Unless you really need backwards
+        compatibility with already stored picked values you should use the JSON
```
gtaylor: This might take some wrangling to read more clearly, unless you had another way to phrase this.

> already-stored, picked values

I may have mis-read the intent with the above suggestion. Big take-away is that this warning is ambiguous to someone who isn't familiar with the plumbing!

baseplate/context/memcache/lib.py
```diff
@@ -191,22 +199,25 @@ def decompress_and_unpickle(key, serialized, flags):
 
 
 def make_pickle_and_compress_fn(min_compress_length=0, compress_level=1):
-    """Create a serialization method compatible with decompress_and_unpickle().
+    """Create a legacy serializer.
```
gtaylor: This looks like the older counterpart tp the "modern" stuff I asked about above. What constitutes legacy vs modern? Is it work linking to an explanation, or are these familiar concepts for better-versed memcache users?

docs/baseplate/context/index.rst
```diff
@@ -20,7 +20,7 @@ DIY: The Factory
 ----------------
 
 If a library you want isn't supported here, it can be added to your own
-application by subclassing :py:class:`~baseplate.context.ContextFactory`.
+application by implementing :py:class:`~baseplate.context.ContextFactory`.
```
gtaylor: Implementing as in ContextFactory is some kind of interface? sub-classing/inherting made sense to me if we're just talking about inheriting and using a utility class, unless we are expected to override an implement a specific set of methods.

docs/baseplate/context/sqlalchemy.rst
```diff
@@ -1,12 +1,21 @@
...
 ---------------------
 
-.. autofunction:: sqlalchemy.engine_from_config
+.. function:: sqlalchemy.engine_from_config(configuration, prefix='sqlalchemy.', **kwargs)
+
+   Make an engine from a configuration dictionary.
+
+   The keys useful to :py:func:`~sqlalchemy.engine_from_config` should be
+   prefixed, e.g.  ``sqlalchemy.url`` etc. The ``prefix`` argument specifies
+   the prefix used to filter keys.  Each key is mapped to a corresponding
```
gtaylor: nit: double-space after period

docs/baseplate/context/sqlalchemy.rst
```diff
@@ -1,12 +1,21 @@
...
 Configuration Parsing
 ---------------------
 
-.. autofunction:: sqlalchemy.engine_from_config
+.. function:: sqlalchemy.engine_from_config(configuration, prefix='sqlalchemy.', **kwargs)
+
+   Make an engine from a configuration dictionary.
+
+   The keys useful to :py:func:`~sqlalchemy.engine_from_config` should be
+   prefixed, e.g.  ``sqlalchemy.url`` etc. The ``prefix`` argument specifies
```
gtaylor: nit: double-space after period

docs/baseplate/message_queue.rst
```diff
@@ -10,11 +10,11 @@ This module provides a thin wrapper around POSIX Message queues.
 
    There are also various limits on the sizes of queues:
 
-   * The ``msgqueue`` rlimit limits the amount of space the user can use on
+   * The ``msgqueue`` ``rlimit`` limits the amount of space the user can use on
      message queues.
-   * The ``fs.mqueue.msg_max`` and ``fs.mqueue.msgsize_max`` sysctls limit the
-     maximum number of messages and the maximum size of each message respectively
-     which a queue can be configured to have.
+   * The ``fs.mqueue.msg_max`` and ``fs.mqueue.msgsize_max`` ``sysctls`` limit
```
gtaylor: nit: ``sysctls`` may not be worth monospacing, since the command name is ``sysctl``. It's _possible_ that this could make a reader pause for a second to ponder intent.

docs/baseplate/message_queue.rst
```diff
@@ -10,11 +10,11 @@ This module provides a thin wrapper around POSIX Message queues.
...
+     the maximum number of messages and the maximum size of each message
+     respectively which a queue can be configured to have.
```
gtaylor: As the sentence is currently structured, I'd suggest removing "respectively". You were nice and specific to open the sentence, so the interruption of flow is probably not worth the mention of explicit pairing.

baseplate/__init__.py
```diff
@@ -18,13 +18,13 @@ def metrics_client_from_config(raw_config):
...
     ``metrics.namespace``
-        The root key to namespace all metrics in this application under.
+        The root key to prefix all metrics in this application with.
     ``metrics.endpoint``
         A ``host:port`` pair, e.g. ``localhost:2014``. If an empty string, a
         client that discards all metrics will be returned.
 
-    :param dict raw_config: The app configuration which should have settings
-        for the metrics client.
+    :param dict raw_config: The application configuration which should have
```
gtaylor: There are quite a few parameters documented similarly ("The application configuration which should have <X>") so I'll just be lazy and comment in one place for all.

I'm not 100% clear on how separate application configuration and settings for the metrics client (or the other <X>'s) are. I am interpreting this as "The application configuration, including settings for the metrics client".

Are the settings for the metrics client part of the application configuration, or are we combining separate things into one dict? Ditto for the other comments that are similar to this one.

baseplate/integration/pyramid.py
```diff
@@ -55,8 +55,9 @@ class BaseplateConfigurator(object):
 
     :param baseplate.core.Baseplate baseplate: The Baseplate instance for your
         application.
-    :param bool trust_trace_headers: Should this app trust trace headers from
-        the client? If ``False``, trace IDs will be generated for each request.
+    :param bool trust_trace_headers: Should this application trust trace
+        headers from the client? If ``False``, trace IDs will be generated for
```
jyoullreddit: Maybe explain the True case? 
"If True, trace headers from the client are used. If False, trace IDs will be generated for each request."
As with Greg's comments generally - if this already makes sense to someone more acquainted with the relevant libraries then maybe it doesn't need the more explicit comment. Given my unfamiliarity, I had to read through it a couple of times to understand what happens here.

baseplate/metrics.py
```diff
@@ -313,7 +313,7 @@ def replace(self, new_value):
 def make_client(namespace, endpoint):
     """Return a configured client.
 
-    :param str namespace: The root key to namespace all metrics under.
+    :param str namespace: The root key to prefix all metrics with.
```
jyoullreddit: You've translated "namespace" to "prefix" here from the function spec to the comment about what it is.
Should the parameter also be called "prefix"? I think we could either "namespace" a set of things or "prefix" some labels, but with this change you're doing both. "Namespace" implies something deeper than just attaching a prefix to labels.

baseplate/context/memcache/__init__.py
spladug: fixed

docs/baseplate/context/index.rst
spladug: It's an interface to implement. The base class has no implementation.

docs/baseplate/context/sqlalchemy.rst
spladug: fixed.

docs/baseplate/context/sqlalchemy.rst
spladug: fixed.

docs/baseplate/message_queue.rst
spladug: ok, un-coded.

docs/baseplate/message_queue.rst
spladug: done.

baseplate/__init__.py
spladug: The main function of a baseplate application gets the application configuration passed in as a dictionary. That's what this is supposed to be. It's basically just "all of the raw config values from the INI file." Does that make sense? or should I document it somewhere?

baseplate/integration/pyramid.py
spladug: :+1: done

baseplate/metrics.py
spladug: The prefix is how we implement a namespace, given the hierarchical nature of graphite metrics. Calling it a prefix is a practical explanation of effect from the abstract name. I'm not sure I see what the deeper meaning of namespace is here.

baseplate/context/memcache/lib.py
spladug: Agreed that it was obtuse. I've tried to clean it up with a bit more explicit definition of when you'd want the "legacy" style and I'm just making the "modern" style describe itself as _the_ way to do serde. How's this look now?

spladug:
Thanks for all the feedback. Really appreciate having things that aren't clear pointed out!

baseplate/integration/pyramid.py
```diff
@@ -56,8 +56,9 @@ class BaseplateConfigurator(object):
     :param baseplate.core.Baseplate baseplate: The Baseplate instance for your
         application.
     :param bool trust_trace_headers: Should this application trust trace
-        headers from the client? If ``False``, trace IDs will be generated for
-        each request.
+        headers from the client? If ``True``, trace headers in inbound requests
```
gtaylor: Awesome

baseplate/context/memcache/lib.py
```diff
@@ -157,11 +157,11 @@ class PickleFlags(object):
 
 
 def decompress_and_unpickle(key, serialized, flags):
-    """Deserialization method compatible with pickle_and_compress().
+    """Deserialize data stored by ``pylibmc``.
```
gtaylor: This is great

spladug:
:haircut_man:

gtaylor:
:fish:

PR #107, opened 2017-06-15
Remove gauge increment/decrement

spladug:
This seems like an anti-feature of gauges since we can't trust that the
"current" value is stable in the face of statsd restarts.


ckwang8128:
:fish: I don't think this is being used anywhere by baseplate-instrumented services right?

spladug:
Nope, not used as tallier doesn't currently support it.

PR #108, opened 2017-06-21
Factor FileWatcher out of ServiceInventory and SecretsStore

spladug:
This is some common code for watching a file, parsing it, and re-loading
it when it changes. We'll be wanting to use this in other situations
like sidecar-fetched live configuration as well so it's useful to factor
it out now.

I'm torn, this module has 97% coverage from tests of other users within baseplate, but isn't tested at all itself. Bad? OK?


spladug:
It looks like there's an issue in CI from the trusty base image in travis changing. I'll get it figured out.

fruchtose:
> I'm torn, this module has 97% coverage from tests of other users within baseplate, but isn't tested at all itself. Bad? OK?

Would it be possible to change the tests of the consumers to apply to `FileWatcher` instead? Then, you could test the consumers to ensure that `FileWatcher` functionality is being run.

baseplate/service_discovery.py
```diff
@@ -133,9 +113,13 @@ def get_backend(self):
...
+            inventory = self._filewatcher.get_data()
+        except WatchedFileNotAvailableError:
+            inventory = None
+
+        # pylint: disable=maybe-no-member
+        if not inventory or not inventory.backends:
             raise NoBackendsAvailableError
 
-        return self.lottery.pick()
+        return inventory.lottery.pick()
```
pacejackson: 🔕  This is so much nicer than what it was doing >.<

pacejackson:
> > I'm torn, this module has 97% coverage from tests of other users within baseplate, but isn't tested at all itself. Bad? OK?

> Would it be possible to change the tests of the consumers to apply to `FileWatcher` instead?

I like this idea.

spladug:
Ooookay, I'll not be lazy. Thanks for keeping me honest. :)

spladug:
I did like having the original tests around to prove the functionality still worked for those use cases.

fruchtose:
> I did like having the original tests around to prove the functionality still worked for those use cases.

If you have tests which show that `FileWatcher` API calls are being run, then you will still have that proof. Just indirectly. 😉

spladug:
Yeah, for sure, I just wanted to make sure I didn't break existing users of the library.

spladug:
> More of a question than any specific call to action: Are there any cases where we'd prefer to watch filesystem events instead of compare mtimes? For example, with watchdog?

I'm definitely interested in taking advantage of inotify down the line, and having all this code factored into one place makes it way easier to do that.

spladug:
:haircut_man: OK, tests are factored out as well.

spladug:
7334448ebb1d099e001b91cbb1a0b9dff5cda088

ckwang8128:
:fish:

PR #109, opened 2017-06-23
Add URL validation utilities

JordanMilne:
Add utilities useful for determining if a URL is safe to use in an anchor or redirect.

`is_web_safe_url()` is mostly copied from r2, minus the wonky `UrlParser` api.

### DIFF
--- a/baseplate/web.py
+++ b/baseplate/web.py
@@ -73,6 +73,8 @@ def is_same_domain(host, patterns):
     for pattern in patterns:
         pattern = pattern.lower()
         matches = (
+            # Use `[:1]` rather than `[0]` so we don't throw an `IndexError`
+            # when given a pattern of `""`
             pattern[:1] == '.' and (host.endswith(pattern) or host == pattern[1:]) or
             pattern == host
         )
--- a/baseplate/web.py
+++ b/baseplate/web.py
@@ -1,0 +1,87 @@
+"""Utilities useful for frontend web services"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import re
+
+from ._compat import urlparse, string_types
+
+# Characters that might cause parsing differences in different implementations
+# Spaces only seem to cause parsing differences when occurring directly before
+# the scheme
+URL_PROBLEMATIC_RE = re.compile(
+    r'(\A\x20|[\x00-\x19\u1680\u180E\u2000-\u2029\u205f\u3000\\])',
+    re.UNICODE
+)
+WEB_SAFE_SCHEMES = {"http", "https"}
+
+
+def is_web_safe_url(url):
+    """Determine if this URL could cause issues with different parsers"""
+
+    # There's no valid reason for this, and just serves to confuse UAs
+    # and urllib2.
+    if url.startswith("///"):
+        return False
+
+    # Reject any URLs that contain characters known to cause parsing
+    # differences between parser implementations
+    if re.search(URL_PROBLEMATIC_RE, url):
+            return False
+
+    parsed = urlparse(url)
+
+    # Double-checking the above
+    if not parsed.hostname and parsed.path.startswith("//"):
+        return False
+
+    # A host-relative link with a scheme like `https:/baz` or `https:?quux`
+    if parsed.scheme and not parsed.hostname:
+        return False
+
+    # Credentials in the netloc?
+    if "@" in parsed.netloc:
+        return False
+
+    # `javascript://www.example.com/%0D%Aalert(1)` is not safe, obviously
+    if parsed.scheme and parsed.scheme.lower() not in WEB_SAFE_SCHEMES:
+        return False
+
+    return True
+
+
+def is_same_domain(host, patterns):
+    """
+    Return ``True`` if the host is either an exact match or a match
+    to the wildcard pattern.
+    Any pattern beginning with a period matches a domain and all of its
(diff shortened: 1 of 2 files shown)


baseplate/web.py
```diff
@@ -0,0 +1,87 @@
...
+    if not host:
+        host = ""
+
+    if isinstance(patterns, string_types):
+        patterns = (patterns.lower(),)
+
+    for pattern in patterns:
+        pattern = pattern.lower()
+        matches = (
+            pattern[:1] == '.' and (host.endswith(pattern) or host == pattern[1:]) or
```
pacejackson: I imagine you're doing `pattern[:1]` rather than `pattern[0]` because the former works even with 0 length strings?  Could you leave a comment about that?  It took me a minute to figure out what was going on.

tests/unit/web_tests.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_iterable_patterns(self):
+        patterns = {"foo.com", "bar.com"}
+        self.assertTrue(web.is_same_domain("foo.com", patterns))
+        self.assertTrue(web.is_same_domain("bar.com", patterns))
+        self.assertFalse(web.is_same_domain("baz.com", patterns))
+
+    def test_subdomain_matching(self):
+        self.assertTrue(web.is_same_domain("foo.com", ".foo.com"))
+        self.assertTrue(web.is_same_domain("bar.foo.com", ".foo.com"))
+        self.assertFalse(web.is_same_domain("bar.foo.com", "foo.com"))
```
pacejackson: Can you add a test for the case where `patterns="bar.foo.com"` and `patterns=".bar.foo.com"`?

baseplate/web.py
JordanMilne: That's right, it was throwing when given a pattern of `""`, I added a comment to clarify

tests/unit/web_tests.py
JordanMilne: added

JordanMilne:
:haircut:

spladug:
We'll want to get this into the docs. To do that, you need to add a file to the `docs/` subdirectory.

spladug:
General question for documentation: why would anyone use these functions? are applications intended to use them directly? if so, when?

baseplate/web.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # `javascript://www.example.com/%0D%Aalert(1)` is not safe, obviously
+    if parsed.scheme and parsed.scheme.lower() not in WEB_SAFE_SCHEMES:
+        return False
+
+    return True
+
+
+def is_same_domain(host, patterns):
+    """
```
spladug: can you put a one line summary of the purpose of this function up here please?

baseplate/web.py
```diff
@@ -0,0 +1,89 @@
...
+    Return ``True`` if the host is either an exact match or a match
+    to the wildcard pattern.
+    Any pattern beginning with a period matches a domain and all of its
+    subdomains. (e.g. ``.example.com`` matches ``example.com`` and
+    ``foo.example.com``). Anything else is an exact string match.
+    An empty pattern is considered equivalent to "no domain".
```
spladug: should get some sphinxy param documentation going

baseplate/web.py
```diff
@@ -0,0 +1,89 @@
...
+from __future__ import unicode_literals
+
+import re
+
+from ._compat import urlparse, string_types
+
+# Characters that might cause parsing differences in different implementations
+# Spaces only seem to cause parsing differences when occurring directly before
+# the scheme
+URL_PROBLEMATIC_RE = re.compile(
```
spladug: maybe something like PROBLEMATIC_URL_CHARACTERS?

baseplate/web.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # There's no valid reason for this, and just serves to confuse UAs
+    # and urllib2.
+    if url.startswith("///"):
+        return False
+
+    # Reject any URLs that contain characters known to cause parsing
+    # differences between parser implementations
+    if re.search(URL_PROBLEMATIC_RE, url):
+            return False
```
spladug: weird indent?

baseplate/web.py
```diff
@@ -0,0 +1,89 @@
...
+# the scheme
+URL_PROBLEMATIC_RE = re.compile(
+    r'(\A\x20|[\x00-\x19\u1680\u180E\u2000-\u2029\u205f\u3000\\])',
+    re.UNICODE
+)
+WEB_SAFE_SCHEMES = {"http", "https"}
+
+
+def is_web_safe_url(url):
+    """Determine if this URL could cause issues with different parsers"""
```
spladug: can you add some param docs here too please?

JordanMilne:
>General question for documentation: why would anyone use these functions? are applications intended to use them directly? if so, when?

They would be plumbed into whatever input validation framework the app ended up using, or called directly if the app isn't using one. 

For example, in r2 world `is_safe_redirect_url()` would be called internally by `VDestination` with `allowed_bases={".reddit.com, ""}`. `is_web_safe_url()` would be used to implement a check like `VUrl()` where you want to know if a URL is a plain-old web link and safe to use in anchors.

spladug:
> They would be plumbed into whatever input validation framework the app ended up using

OK, can we add some helpers to integrate with the common case (Pyramid + whatever)?

JordanMilne:
I'll roll this up into the bigger input validation PR

PR #110, opened 2017-06-26
Test cleanup

spladug:
This does some maintenance on the test suite to improve coverage reporting and fix how we launch the suite so it's easier to use.

Rather than `nocover`ing the `__str__` implementations on exceptions, the messages are now pre-formatted during construction time and the base class `Exception`'s implementation of  `__str__` is used. This means we're sure the formatting doesn't crash without having to write tests that are tied to the specific output.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,13 @@
 [run]
 branch = 1
 omit = */baseplate/thrift/*.py
+
+[report]
+exclude_lines =
+    # re-enable the built-in pragma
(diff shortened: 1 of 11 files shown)


ckwang8128:
:fish: the commit messages = 💯

spladug:
8ff8dbd5aba72f6f7dd93e99fd0293ae658381ce...71dd0c09c06fd2913b70755e818bb82635633390

PR #111, opened 2017-06-26
Add DictOf config parser

spladug:
This is useful for less static configuration, stuff that's closer to
data than basic configuration, which will likely be much more common
when we get live configuration abilities in Baseplate.

See the included docs for an example of usage.

I'm concerned about this parser being confusing since it takes multiple
keys instead of a single one. Maybe there's a better name than `DictOf`?
Maybe it'd be better to do a more complex syntax like requiring the key
in the config dict to be something like `"key.*": config.DictOf(...),`?

### DIFF
--- a/baseplate/config.py
+++ b/baseplate/config.py
@@ -17,7 +17,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
 .. highlight:: py
 
-.. testsetup::
+.. testsetup:: overview
 
     from baseplate import config
     from baseplate._compat import configparser
@@ -30,7 +30,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     tempfile.flush()
     config_parser.set("app:main", "some_file", tempfile.name)
 
-.. doctest::
+.. doctest:: overview
 
     >>> raw_config = dict(config_parser.items("app:main"))
 
@@ -70,7 +70,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     >>> print(cfg.interval)
     0:00:30
 
-.. testcleanup::
+.. testcleanup:: overview
 
     tempfile.close()
 
@@ -83,6 +83,7 @@ from __future__ import unicode_literals
 import base64
 import collections
 import datetime
+import re
 import socket
 
 
@@ -196,8 +197,6 @@ def File(mode="r"):
 
     :param str mode: an optional string that specifies the mode in
         which the file is opened.
-
-
 
     """
     def open_file(text):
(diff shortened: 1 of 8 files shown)


baseplate/config.py
```diff
@@ -335,30 +334,125 @@ def __init__(self):
...
-
-            try:
-                parsed[key] = parser(raw_value)
-            except Exception as e:
-                raise ConfigurationError(key, e)
-        elif isinstance(parser_or_spec, dict):
-            subspec = parser_or_spec
-            parsed[key] = _parse_config_section(config, subspec, root=key_path)
+    @classmethod
+    def from_spec(self, spec):
```
pacejackson: This doesn't actually seem to care about the fact that it's a part of the `Parser` class.  Would it be neater to pull it out into a function that like `parser_from_spec`?

pacejackson: If you do want to keep it a classmethod, could you rename `self` to `cls`?  Also, should the subclasses be calling `self.from_spec` rather than `Parser.from_spec`?

tests/unit/config_tests.py
```diff
@@ -236,6 +236,41 @@ def test_fallback_option_works(self):
...
+            "my_key.a": "1",
+            "my_key.b": "2",
+        })
+        self.assertEqual(result, {"a": 1, "b": 2})
+
+    def test_vector_children(self):
+        parser = config.DictOf({"a": config.Integer, "b": config.String})
+        result = parser.parse("my_key", {
+            "my_key.first.a": "1",
+            "my_key.first.b": "test",
```
pacejackson: what happens if you add `"my_key.first.c"`?  I assume it just gets ignored?  It might be good to make that a part of the test.

baseplate/config.py
```diff
@@ -335,30 +334,125 @@ def __init__(self):
...
+            return self.callable(raw_value)
+        except Exception as exc:
+            raise ConfigurationError(key_path, exc)
+
+
+class DictOf(Parser):
+    """A group of options of a given type.
+
+    This is useful for providing data to the application without the
+    application having to know ahead of time all of the possible keys.
```
pacejackson: I feel like the current docstring implies that you can only do `DictOf(config.Type)`. Could you update it to include the `DictOf(spec_dict)` usage mode as well?

baseplate/config.py
spladug: Whoops, good point. I turned it into a `staticmethod` and removed the unused first parameter altogether.

As for why it's on the class at all: I think it adds a bit of semantic connection to put it there following the static factory method pattern. There are a [couple](https://github.com/reddit/baseplate/blob/master/baseplate/core.py#L82-L123) [other](https://github.com/reddit/baseplate/blob/master/baseplate/secrets/store.py#L77-L90) places in baseplate I've done the same. (I think I've reintroduced this to my vocabulary via Rust, so if we feel it's not very pythonic I can stop doing it.)

tests/unit/config_tests.py
spladug: Sure, added. Technically that's testing the spec subparser though :X

baseplate/config.py
spladug: roger, how's that?

spladug:
:haircut_man:

baseplate/config.py
```diff
@@ -333,30 +332,125 @@ def __init__(self):
...
+        raw_value = raw_config.get(key_path, "")
+
+        try:
+            return self.callable(raw_value)
+        except Exception as exc:
+            raise ConfigurationError(key_path, exc)
+
+
+class DictOf(Parser):
+    """A group of options of a given type.
```
dellis23: I guess I'll have to see how it comes out when the docs are generated, but I didn't really get what was going on here until I read `config_dictof_example.ini`.  Maybe it'll be obvious at that point though.

baseplate/config.py
```diff
@@ -333,30 +332,125 @@ def __init__(self):
...
+        """Parse and return the relevant info for a given key.
+
+        :param str key_path: The key this parser is looking for.
+        :param dict raw_config: The full raw configuration dictionary.
+
+        """
+        raise NotImplementedError
+
+
+class SpecParser(Parser):
```
dellis23: Do we document what this looks like anywhere?  Or is a static specification essentially just a non-`DictOf`?

spladug: yup, it's everything else in the docs. the "original" config parsing.

baseplate/config.py
spladug: screenshot from 2017-06-27 10 30 34

spladug:
:haircut_man: thought of an untested and broken case in the shower

baseplate/config.py
dellis23: Oh nice, thank you!

PR #112, opened 2017-07-13
Add experiments and feature flags

pacejackson:
This update ports the feature flag and experiment code from [r2](https://github.com/reddit/reddit/tree/master/r2/r2/config/feature) and their tests into baseplate.  Rather than just doing a straight port, this change attempts to:

1. Define a new, more extensible config format
2. Support for different experiment "providers" that can have different config formats and bucketing logic.
3. Formally separate feature flags from experiments
4. Remove un-used bits of code

### DIFF
--- a/baseplate/diagnostics/metrics.py
+++ b/baseplate/diagnostics/metrics.py
@@ -80,3 +80,7 @@ class MetricsClientSpanObserver(SpanObserver):
         self.timer.stop()
         suffix = "success" if not exc_info else "failure"
         self.batch.counter(self.base_name + "." + suffix).increment()
+
+    def on_log(self, name, payload):
+        if name == "error.object":
+            self.batch.counter.increment("errors.%s" % payload.__class__.__name__)
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -1,0 +1,205 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import json
+import logging
+
+from .providers import parse_experiment
+from .. import config
+from .._compat import iteritems
+from ..context import ContextFactory
+from ..events import Event, EventTooLargeError, EventQueueFullError
+from ..file_watcher import FileWatcher, WatchedFileNotAvailableError
+
+
+logger = logging.getLogger(__name__)
+
+
+class ExperimentsContextFactory(ContextFactory):
+    """Experiment client context factory
+
+    This factory will attach a new :py:class:`baseplate.experiments.Experiments`
+    to an attribute on the :term:`context object`.
+    """
+    def __init__(self, path, event_queue):
+        """ExperimentContextFactory constructor
+
+        :param str path: Path to the experiment config file.
+        :param baseplate.events.EventQueue event_queue: Event queue used for
+            logging bucketing events to the event pipeline.  You can set this
+            to None to disable bucketing events entirely.
+        """
+        self._filewatcher = FileWatcher(path, json.load)
+        self._event_queue = event_queue
+
+    def make_object_for_context(self, name, server_span):
+        return Experiments(
+            config_watcher=self._filewatcher,
+            event_queue=self._event_queue,
+            server_span=server_span,
+            context_name=name,
+        )
+
+
+class Experiments(object):
+    """Access to experiments with automatic refresh when changed.
+
+    This experiments client allows access to the experiments cached on disk
+    by the experiment config fetcher daemon.  It will automatically reload
+    the cache when changed.  This client also handles logging bucketing events
+    to the event pipeline when it is determined that the request is part of an
+    active variant.
+    """
+
+    def __init__(self, config_watcher, event_queue, server_span, context_name):
+        self._config_watcher = config_watcher
+        self._event_queue = event_queue
+        self._span = server_span
+        self._context_name = context_name
+        self._already_bucketed = set()
+
+    def _get_config(self, name):
+        try:
+            config_data = self._config_watcher.get_data()
+            return config_data[name]
+        except WatchedFileNotAvailableError as exc:
+            logger.warning("Experiment config unavailable: %s", str(exc))
+            return
+        except KeyError:
+            logger.warning(
+                "Experiment <%r> not found in experiment config",
+                name,
+            )
+            return
+        except TypeError as exc:
+            logger.warning("Could not load experiment config: %s", str(exc))
+            return
+
+    def variant(self, name, bucketing_event_override=None,
+                extra_event_fields=None, **kwargs):
+        """Which variant, if any, is active.
+
+        If a variant is active, a bucketing event will be logged to the event
+        pipeline unless any one of the following conditions are met:
+
+        1. bucketing_event_override is set to False.
+        2. The experiment specified by "name" explicitly disables bucketing
+           events.
+        3. We have already logged a bucketing event for the value specified by
+           experiment.get_unique_id(\*\*kwargs) within the current
+           request.
+
+        Since checking the status an experiment will fire a bucketing event, it
+        is best to only check the variant when you are making the decision that
+        will expose the experiment to the user.  If you absolutely must check
+        the status of an experiment before you are sure that the experiment
+        will be exposed to the user, you can use `bucketing_event_override` to
+        disabled bucketing events for that check.
+
+        :param str name: Name of the experiment you want to run.
+        :param bool bucketing_event_override: (Optional) Set if you need to
+            override the default behavior for sending bucketing events.  This
+            parameter should be set sparingly as it breaks the assumption that
+            if you check a variant multiple times in a request, it will fire
+            one and only one bucketing event.  If set to True, will always log
+            bucketing events unless the experiment explicitly disables them,
+            even if a bucketing event has already been logged or the variant is
+            `None`.  If set to False, will never send a bucketing event. If set
+            to None, no override will be applied.  Set to None by default.
+        :param dict extra_event_fields: (Optional) Any extra fields you want to
+            add to the bucketing event.
+        :param kwargs:  Arguments that will be passed to experiment.variant to
+            determine bucketing, targeting, and overrides.
+
+        :rtype: :py:class:`str`
+        :return: Variant name if a variant is active, None otherwise.
+        """
+        experiment_config = self._get_config(name)
+        if not experiment_config:
+            return None
+
+        experiment = parse_experiment(experiment_config)
+        variant = experiment.variant(**kwargs)
+
+        do_log = True
+
+        if variant is None:
+            do_log = False
+
+        bucketing_id = experiment.get_unique_id(**kwargs)
+
+        if bucketing_id and bucketing_id in self._already_bucketed:
+            do_log = False
+
+        if bucketing_event_override is not None:
+            do_log = bool(bucketing_event_override)
+
+        do_log = do_log and experiment.should_log_bucketing()
+
+        if do_log:
+            self._log_bucketing_event(experiment, variant, extra_event_fields)
+            if bucketing_id:
+                self._already_bucketed.add(bucketing_id)
+
+        return variant
+
+    def _log_bucketing_event(self, experiment, variant, extra_event_fields):
+        if not self._event_queue:
+            return
+
+        extra_event_fields = extra_event_fields or {}
+
+        event_type = experiment.get_event_type()
+        if not event_type:
+            return
+
+        event = Event("bucketing_events", event_type)
+        for field, value in iteritems(extra_event_fields):
+            event.set_field(field, value)
+
+        event.set_field("variant", variant)
+        event.set_field("experiment_id", experiment.id)
+        event.set_field("experiment_name", experiment.name)
+        event.set_field("owner", experiment.owner)
+
+        span_name = "{}.{}".format(self._context_name, "events")
+        with self._span.make_child(span_name) as child_span:
+            try:
+                self._event_queue.put(event)
+            except EventTooLargeError as exc:
+                logger.warning(
+                    "The event payload was too large for the event queue."
+                )
+                child_span.set_tag("error", True)
+                child_span.log("error.object", exc)
+            except EventQueueFullError as exc:
+                logger.warning("The event queue is full.")
+                child_span.set_tag("error", True)
+                child_span.log("error.object", exc)
+
+
+def experiments_client_from_config(app_config, event_queue):
+    """Configure and return an :py:class:`ExperimentsContextFactory` object.
+
+    This expects one configuration option:
+
+    ``experiments.path``
+        The path to the experiment config file generated by the experiment
+        config fetcher daemon.
+
+    :param dict raw_config: The application configuration which should have
+        settings for the experiments client.
+    :param baseplate.events.EventQueue event_queue: The EventQueue to be used
+        to log bucketing events.
+    :rtype: :py:class:`ExperimentsContextFactory`
+
+    """
+    cfg = config.parse_config(app_config, {
+        "experiments": {
+            "path": config.Optional(config.String, default="/var/local/experiments.json"),
+        },
+    })
+    # pylint: disable=maybe-no-member
+    return ExperimentsContextFactory(cfg.experiments.path, event_queue)
--- a/baseplate/experiments/providers/__init__.py
+++ b/baseplate/experiments/providers/__init__.py
@@ -1,0 +1,97 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+
+from datetime import datetime
+
+from .feature_flag import FeatureFlag
+from .forced_variant import ForcedVariantExperiment
+from .r2 import R2Experiment
+
+logger = logging.getLogger(__name__)
+
+
+ISO_DATE_FMT = "%Y-%d-%m"
+
+
+def parse_experiment(config):
+    """Factory method that parses an experiment config dict and returns an
+    appropriate Experiment class.
+
+    The config dict is expected to have the following values:
+
+        * **id**: Integer experiment ID, should be unique for each experiment.
+        * **name**: String experiment name, should be unique for each
+          experiment.
+        * **owner**: The group or individual that owns this experiment.
+        * **expires**: Date when this experiment expires in ISO format
+          (YYYY-MM-DD).  The experiment will expire at 00:00 UTC on the day
+          after the specified date.  Once an experiment is expired, it is
+          considered disabled.
+        * **type**: String specifying the type of experiment to run.  If this
+          value is not recognized, the experiment will be considered disabled.
+        * **experiment**: The experiment config dict for the specific type of
+          experiment.  The format of this is determined by the specific
+          experiment type.
+        * **enabled**:  (Optional) If set to False, the experiment will be
+          disabled and calls to experiment.variant will always return None and
+          will not log bucketing events to the event pipeline. Defaults to
+          True.
+        * **global_override**: (Optional) If this is set, calls to
+          experiment.variant will always return the override value and will not
+          log bucketing events to the event pipeline.
+
+    :param dict config: Configuration dict for the experiment you wish to run.
+    :rtype: :py:class:`baseplate.experiments.providers.base.Experiment`
+    :return: A subclass of :py:class:`Experiment` for the given experiment
+        type.
+    """
+    experiment_type = config["type"].lower()
+    experiment_id = config["id"]
+    assert isinstance(experiment_id, int)
+    name = config["name"]
+    owner = config.get("owner")
+    experiment_config = config["experiment"]
+    expiration = datetime.strptime(config["expires"], ISO_DATE_FMT).date()
+
+    if datetime.utcnow().date() > expiration:
+        return ForcedVariantExperiment(None)
+
+    enabled = config.get("enabled", True)
+    if not enabled:
+        return ForcedVariantExperiment(None)
+
+    if "global_override" in config:
+        # We want to check if "global_override" is in config rather than
+        # checking config.get("global_override") because global_override = None
+        # is a valid setting.
+        override = config.get("global_override")
+        return ForcedVariantExperiment(override)
+
+    if experiment_type == "r2":
+        return R2Experiment.from_dict(
+            id=experiment_id,
+            name=name,
+            owner=owner,
+            config=experiment_config,
+        )
+    elif experiment_type == "feature_flag":
+        return FeatureFlag.from_dict(
+            id=experiment_id,
+            name=name,
+            owner=owner,
+            config=experiment_config,
+        )
+    else:
+        logger.warning(
+            "Found an experiment <%s:%s> with an unknown experiment type <%s> "
+            "that is owned by <%s>. Please clean up.",
+            experiment_id,
+            name,
+            experiment_type,
+            owner,
+        )
+        return ForcedVariantExperiment(None)
--- a/baseplate/experiments/providers/base.py
+++ b/baseplate/experiments/providers/base.py
@@ -1,0 +1,47 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class Experiment(object):
+    """Base interface for experiment objects. """
+
+    def get_unique_id(self, **kwargs):
+        """Generate a unique ID for this experiment with the given inputs.
+
+        Used to determine if a bucketing ivent has alread been fired for a
+        given experiment and bucketing value pair.  Returns None by default.
+        If None is returned, we will not mark that a bucketing event has been
+        logged even if we do log a bucketing event. The kwargs should be the
+        same values passed to the call to Experiment.variant.  If your
+        experiment does log bucketing events, you must implement this function.
+
+        :rtype: :py:class:`str` or None
+        """
+        if self.should_log_bucketing():
+            raise NotImplementedError
+        return None
+
+    def get_event_type(self):
+        if self.should_log_bucketing():
+            raise NotImplementedError
+        return None
+
+    def variant(self, **kwargs):
+        """Determine which variant, if any, of this experiment is active.
+
+        All arguments needed for bucketing, targeting, and variant overrides
+        should be passed in as kwargs.  The parameter names are determined by
+        the specific implementation of the Experiment interface.
+
+        :rtype: :py:class:`str`
+        :returns: The name of the enabled variant as a string if any variant is
+        enabled.  If no variant is enabled, return None.
+        """
+        raise NotImplementedError
+
+    def should_log_bucketing(self):
+        """Should this experiment log bucketing events to the event pipeline.
+        """
+        raise NotImplementedError
--- a/baseplate/experiments/providers/feature_flag.py
+++ b/baseplate/experiments/providers/feature_flag.py
@@ -1,0 +1,62 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from .r2 import R2Experiment
+
+
+class FeatureFlag(R2Experiment):
+    """An experiment with a single variant "active".
+
+    Does not log bucketing events to the event pipeline.  Use this type of
+    experiment if you just want to control access to a feature but do not want
+    to run an actual experiment.  Some examples for when you would want to use
+    a FeatureFlag are:
+
+    1. Slowly rolling out a new feature to a % of users
+    2. Restricting a new feature to certain subreddits
+
+    The config dict is expected to have the following values:
+
+        * **variants**: dict mapping variant names to their sizes. Variant
+          sizes are expressed as numeric percentages rather than a fraction of
+          1 (that is, 1.5 means 1.5%, not 150%).  For a feature flag, you can
+          only specify a single variant named "active".
+        * **targeting**: (Optional) dict that maps the names of targeting
+          parameters to lists of valid values.  When determining the variant
+          of an experiment, the targeting parameters you want to use are passed
+          in as keyword arguments to the call to experiment.variant.
+        * **overrides**: (Optional) dict that maps override parameters to
+          dictionaries mapping values to the variant name you want to override
+          the variant to. When determining the variant of an experiment, the
+          override parameters you want to use are passed in as keyword
+          arguments to the call to experiment.variant.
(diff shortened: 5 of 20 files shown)


pacejackson:
:haircut_man: 
1. Added expiration to experiments and feature flags
2. You now pass in targeting/bucketing values with kwargs directly rather than passing in a `SessionContext`
3. Removed the notion of a `SessionContext` and `TargetingParams`

pacejackson:
Note: docs haven't been updated to reflect the new changes yet, gonna go back to docs soon

baseplate/experiments/providers/legacy.py
```diff
@@ -0,0 +1,233 @@
...
+        if content_id is None:
+            return None
+
+        # If we've restricted the experiment to certain page types, make sure
+        # the request is for one of those
+        if (self.content_flags.get('subreddit_only', False) and
+                content_type != 'subreddit'):
+            return None
+
+        if (self.content_flags.get('link_only', False) and
```
spladug: just to play devil's advocate, if we add a completely new content type we'd have to update this library too?

baseplate/features/feature.py
```diff
@@ -0,0 +1,300 @@
+
```
spladug: nit: a lot of these files have a weird extra blank at the top :grimacing:

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.warning(
+                "Experiment <%r> not found in experiment config",
+                name,
+            )
+            return
+        except TypeError:
+            logger.warning("Could not load experiment config: %r", name)
+            return
+
+    def variant(self, name, **args):
```
spladug: can this (and other similar ** usages) be named `kwargs` for consistency please?

baseplate/experiments/providers/base.py
```diff
@@ -0,0 +1,24 @@
+
```
spladug: nit: please make sure all the new files have the `from __future__` stuff at the top

spladug: (i should figure out how to get that into lint/ci stuff)

baseplate/features/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+Content = namedtuple("Content", ["id", "type"])
+
+
+class FeatureFlagsContextFactory(ContextFactory):
+    """ ContextFactory for FeatureFlags.
+
+    This factory will attach a :py:class:`baseplate.features.FeatureFlags`
+    object on the :term:`context object` that will use the config file
+    specified by the given path.  Uses a :py:class:`baseplate.file_watcher.FileWatcher`
```
spladug: probably best not to document how the internals work beyond what you want to promise forever and ever

baseplate/experiments/providers/legacy.py
```diff
@@ -0,0 +1,233 @@
...
+import hashlib
+
+from .base import ExperimentInterface
+from ..._compat import long, iteritems
+
+
+logger = logging.getLogger(__name__)
+
+
+class LegacyExperiment(ExperimentInterface):
```
spladug: can we call this something more descriptive of when it'd be preferred over an alternative (which we don't really have yet right?) maybe R2Experiment? 

(bringing this up since i got similar feedback recently: "legacy" got rightfully called out as vague and not informative when doing docs for the memcache serialization formats.)

baseplate/features/feature.py
```diff
@@ -0,0 +1,300 @@
...
+            return False
+        if percentage <= 0:
+            return False
+        if percentage >= 100:
+            return True
+        bucket = self._calculate_bucket(user_id)
+        scaled_percent = bucket / (self.num_buckets / 100)
+        if scaled_percent < percentage:
+            return True
+
```
spladug: otherwise we return none?

baseplate/experiments/providers/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        return ExperimentManager(
+            id=experiment_id,
+            name=name,
+            owner=owner,
+            experiment=ForcedVariantExperiment(None),
+        )
+
+    if "feature" in config:
+        feature_flag = feature_flag_from_config(config["feature"])
```
pacejackson: I've gone back and forth on this. The problem I've ran into is that features just return True if any of their targeting parameters are met while experiments use targeting to determine if they should run an experiment. 

I've thought about it some more I have an idea about how to use experiments to get this behavior. 

1. Make `targeting` generic. The keys in the `targeting` dict are potential `variant` kwarg values. The values are the "targeted" values. They can be either a single value or a list of values.
2. Add a notion of `overrides`. This would be a dict mapping string keys to dict. The keys would be potention kwargs. The inner dicts would map kwarg values to variants. 

A "feature_flag" experiment has one variant "active" and uses `overrides` rather than `targeting` for most of the values.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        if name not in self._cache:
```
pacejackson: That is a good point.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            self._cache[name] = self._bucket(name, **args)
+        return self._cache[name]
+
+    def _bucket(self, name, **args):
+        config = self._get_config(name)
+        if not config:
+            return None
+
+        experiment = experiment_from_config(config)
```
pacejackson: Doesn't seem to take long:

```python
In [7]: %timeit experiment_from_config(config)
100000 loops, best of 3: 10.5 µs per loop
```

spladug: excellent

spladug: thanks for checking!

baseplate/features/feature.py
```diff
@@ -0,0 +1,300 @@
...
+    """ A "basic" feature flag.  Provides targeting against discrete values and
+    a percentage of users based on ID.
+    """
+
+    def __init__(self, id, name, owner, seed, percent_logged_in,
+                 percent_logged_out, targeting):
+        self.id = id
+        self.name = name
+        self.owner = owner
+        self.num_buckets = 1000
```
pacejackson: My plan was to eventually have it configurable with the experiment config, but I wasn't sure if I wanted to open that up to start.

baseplate/experiments/providers/base.py
```diff
@@ -0,0 +1,24 @@
...
+class ExperimentInterface(object):
```
pacejackson: My thinking behind creating `Interface` classes for these is that the goal of this system is to be able to support different experiment providers that are constructed from a single factory method so I wanted to try to provide a basic interface that you could rely on.  If you don't like having this kind of thing around, I'm fine with removing it or waiting to add it until we _actually_ have different experiment types.

baseplate/features/feature.py
pacejackson: Updated to return False rather than None.

baseplate/experiments/providers/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+
+def experiment_from_config(config):
+    experiment_type = config["type"].lower()
+    experiment_id = config["id"]
+    name = config["name"]
+    owner = config.get("owner")
+    experiment_config = config["experiment"]
+    expiration = config["expires"]
+    if int(time.time()) > expiration:
+        logger.warning(
```
pacejackson: I definitely think that enforcement should be in the "experiments service" (maybe we wont even distribute experiment configs after they expire?).  I'm not opposed to taking out these warnings but I don think we want to have guards like this against out-dated experiments or experiments that we don't understand.

baseplate/experiments/providers/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+import time
+
+from .forced_variant import ForcedVariantExperiment
+from .legacy import LegacyExperiment
+from ...features.feature import feature_flag_from_config
+
+logger = logging.getLogger(__name__)
+
+
+class ExperimentManager(object):
```
pacejackson: I might just call this "Experiment".  I'm also considering rolling targeting back into experiments and treating feature flags as a type of experiment, in which case this might go away.

baseplate/experiments/providers/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    def event_params(self):
+        event_params = {
+            "experiment_id": self.id,
+            "experiment_name": self.name,
+            "experiment_owner": self.owner,
+        }
+        return event_params
+
+
+def experiment_from_config(config):
```
pacejackson: I think I might go with `parse_experiment` and `Experiment.from_dict`.  While I like the _sound_ of `from_json` I don't like to use that unless I am actually sending it a json string which I would not be in this case.

spladug: > I don't like to use that unless I am actually sending it a json string which I would not be in this case.

Agreed! Good call.

baseplate/experiments/providers/legacy.py
pacejackson: Looks like this got buried, but yes we would.  I'm re-working how we configure experiments so they just have generic "targeting" and "override" values and you can set what value you want to use for bucketing.  In this world, a "link-only, page" experiment looks like:
```
config = {
    "id": "1",
    "name": "test",
    "owner": "test",
    "type": "r2",
    "expires": int(time.time()) + THIRTY_DAYS_SEC,
    "experiment": {
        "bucket_val": "content_id",
        "targeting": {
            "content_type": ["link", "comment"],
        },
        "variants": {
            "larger": 5,
            "smaller": 10,
            "control_1": 10,
            "control_2": 10,
        },
    },
}
```

So it expects the call to `variant` to look like `experiment.variant(content_id="t5_id", content_type="subreddit", ...)

pacejackson:
:haircut_man:

1. I got most of the feedback but I still have a lot of cleanup
2. Got rid of most of the special targeting/override cases (`newer_than` is still special)
3. Rolled targeting/overrides into experiments
4. Implemented feature flags as a type of experiment

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+            logger.warning(
+                "Experiment <%r> not found in experiment config",
+                name,
+            )
+            return
+        except TypeError:
+            logger.warning("Could not load experiment config: %r", name)
+            return
+
+    def variant(self, name, bucket_event_override=None,
```
bsimpson63: 🔕 there are `variant()` methods in `Experiments `, `ExperimentManager`, and `ExperimentInterface`. Are they essentially all the same things, or could they be named differently?

pacejackson: I'm actually going to remove `ExperimentManager` next, I just wanted to change as little as possible when I refactoring the other parts.  I would say the difference between `Experiments.variant` and `ExperimentInterface.variant` is that `Experiments.variant` is more like `all_my_experiments[name].variant` where `all_my_experiments[name]` is an `ExperimentInterface` object.

baseplate/experiments/providers/base.py
spladug: Oh I'm totally on board with having an abstract base class, just suggesting the Interface suffix be dropped.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+                name,
+            )
+            return
+        except TypeError:
+            logger.warning("Could not load experiment config: %r", name)
+            return
+
+    def variant(self, name, bucketing_event_override=None,
+                extra_event_params=None, **kwargs):
+        cache_key = "%s:%s" % (name, str(sorted(iteritems(kwargs))))
```
pacejackson: I'll note, I'm not enthusiastic about this cache key scheme.  I could just cache based on the value used for bucketing and the experiment name, which is what r2 does.  My only concern with that is r2 pulls all the targeting information out of the aether (`g` & `c`) but with this you can ask for a variant for the same user but with different targeting parameters and just caching by bucketing value would not capture that.

baseplate/experiments/providers/__init__.py
pacejackson: I've been thinking about these more and I think I'm going to cut out these warnings and the expiration enforcement from baseplate, I think we can build a better system to bug people about these things within the experiment service that we use to manage the configs.

pacejackson:
So I took out most of the warnings on config parsing, I left in warnings for:

1. Cannot find the file
2. Cannot find the experiment you asked for
3. Found the file but it wasn't json
4. The experiment is an unknown type

pacejackson:
Also, while the baseplate module still checks the "expiration" and "enabled" flags, I think a good practice for those would be to stop distributing them after a certain point, say 7 days after they expire or they are set to disabled?

pacejackson:
:haircut_man: 

Cleaned up some cruft and added documentation.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        self._config_watcher = config_watcher
+        self._event_queue = event_queue
+        self._already_bucketed = set()
+
+    def _get_config(self, name):
+        try:
+            config_data = self._config_watcher.get_data()
+            return config_data[name]
+        except WatchedFileNotAvailableError:
+            logger.warning("Experiment config file not found")
```
spladug: can you include the stringified exception as well? it carries the IOError that cause the file to be unavailable so it's easier to tell why it's unavailable

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+                extra_event_params=None, **kwargs):
+        """ Which variant, if any, is active for the experiment specified by
+        "name".  If a variant is active, a bucketing event will be logged to
+        the event pipeline unless any one of the following conditions are met:
+
+        1. bucketing_event_override is set to False.
+        2. The experiment specified by "name" explicitly disables bucketing
+           events.
+        3. We have already logged a bucketing event for the value specified by
+           experiment.event_cache_key(**kwargs) within the current session.
```
spladug: "session"? should this be "span" or "request" to make it more clear the duration of the debounce?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+                "Experiment <%r> not found in experiment config",
+                name,
+            )
+            return
+        except TypeError:
+            logger.warning("Could not load experiment config: %r", name)
+            return
+
+    def variant(self, name, bucketing_event_override=None,
+                extra_event_params=None, **kwargs):
```
spladug: would "fields" be better than "params" for the event stuff? since it's `Event.set_field` and `Event.get_field`

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        3. We have already logged a bucketing event for the value specified by
+           experiment.event_cache_key(**kwargs) within the current session.
+
+        :param str name: Name of the experiment you want to run.
+        :param bool bucketing_event_override: Optional value, defaults to None.
+            If set to True, will always log bucketing events unless the
+            experiment explicitly disables them.  If set to False, will never
+            send a bucketing event.  If set to None, no override will be
+            applied.
+        :param dict extra_event_params: Optional value.  Any extra values you
```
spladug: i've used `(Optional)` a few times for param docs, maybe that'd be more consistent? also fields instead of values here maybe?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+
+        if variant is None:
+            do_log = False
+
+        cache_key = experiment.event_cache_key(**kwargs)
+
+        if cache_key and cache_key in self._already_bucketed:
+            do_log = False
+
+        if bucketing_event_override is True:
```
spladug:
```python
if bucketing_event_override is not None:
    do_log = bucketing_event_override
```

?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        if not config:
+            return None
+
+        experiment = parse_experiment(config)
+        variant = experiment.variant(**kwargs)
+
+        do_log = True
+
+        if variant is None:
+            do_log = False
```
spladug: should it just return none here?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+
+        cache_key = experiment.event_cache_key(**kwargs)
```
spladug: we're not actually caching right? so this is more of a bucketing id or something used to debounce events?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        do_log = do_log and experiment.should_log_bucketing()
+
+        if do_log:
+            self._log_bucketing_event(experiment, variant, extra_event_params)
+            if cache_key:
+                self._already_bucketed.add(cache_key)
+
+        return variant
+
+    def _log_bucketing_event(self, experiment, variant, extra_event_params=None):
```
spladug: does `extra_event_params` need to take a default since it's only used internally once?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+            event.set_field(field, value)
+
+        event.set_field("variant", variant)
+        event.set_field("experiment_id", experiment.id)
+        event.set_field("experiment_name", experiment.name)
+        event.set_field("owner", experiment.owner)
+        try:
+            self._event_queue.put(event)
+        except EventTooLargeError:
+            logger.exception("That event was too large for event queue.")
```
spladug: "for event queue"?

should we send metrics on these failures or something?

baseplate/experiments/__init__.py
pacejackson: Would it be better to just "logging.exception" or would you prefer a stringified error?

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self._filewatcher = FileWatcher(path, json.load)
+        self._event_queue = event_queue
+
+    def make_object_for_context(self, name, server_span):
+        return Experiments(self._filewatcher, self._event_queue)
+
+
+class Experiments(object):
+    """ Access to experiments with automatic refresh when changed. Handles
```
spladug: nit: pep257 says the summary line should be just one line

baseplate/experiments/__init__.py
pacejackson: bucketing_id is a good name... I'll use that.

baseplate/experiments/__init__.py
spladug:
```python
except WatchedFileNotAvailableError as exc:
    logger.warning("Experiment config unavailable: %s", exc)
```

should be sufficient

baseplate/experiments/__init__.py
pacejackson: No, I think I just got in the habit of passing along kwargs as kwargs but I'm not really sure why, I'll change it.

baseplate/experiments/__init__.py
pacejackson: Not if we want to do what r2 currently does, if you set `bucketing_event_override=True` it will still send one even if `variant is None`.

```python
        should_send_bucketing_event = False
        if variant is None or bucketing_event_override is False:
            # Never try sending if there's no variant or override is False
            should_send_bucketing_event = False
        elif bucketing_event_override:
            # Always try sending if override is truthy
            should_send_bucketing_event = True
```

I personally have no attachment to that behavior and am fine with removing it.

baseplate/experiments/__init__.py
pacejackson: I'll reword that.

I think it  would make sense to do that.  This would be the first place we would add metrics like that in one of these `ContextFactory` generated methods.  My first thought would be to pass in the `metrics_client` to the `ContextFactory` but maybe it makes more sense to use span tags and `on_set_tag`?  Do you have a preference for one over the other?

pacejackson: From looking at the tracing spec, it seems like using `log` with `error.kind` makes more sense for this.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+        event.set_field("experiment_name", experiment.name)
+        event.set_field("owner", experiment.owner)
+        try:
+            self._event_queue.put(event)
+        except EventTooLargeError:
+            logger.exception(
+                "The event payload was too large for the event queue."
+            )
+            if self._server_span:
+                self._server_span.log("error.kind", "events.too_large")
```
spladug: hmm... i don't think that's right. it implies that the whole server span errored when it'd actually be more accurate to say some child span about publishing events was what errored. it would suck to mark an otherwise perfectly successful request as erroring because of this.

baseplate/experiments/__init__.py
spladug: I'm OK not being tied to r2 behaviour, but "override" does seem to imply it's totally overridden; good point. Seems reasonable to stick to that.

baseplate/experiments/providers/base.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+class Experiment(object):
+    """ Base interface for experiment objects. """
```
spladug: nit: why do some of these docstrings have a space at the beginning?

baseplate/experiments/providers/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def bucketing_event_id(self, **kwargs):
+        """ Key used to track if a bucketing event has been fired for the
```
spladug: nit: there should be a summary line first

baseplate/experiments/providers/feature_flag.py
```diff
@@ -0,0 +1,20 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from .r2 import R2Experiment
+
+
+class FeatureFlag(R2Experiment):
+    """ An experiment with a single variant "active" that does not log
+    bucketing events to the event pipeline.  Use this type of experiment if
+    you just want to controll access to a feature but do not want to run an
```
spladug: controll

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+    This factory will attach a new :py:class:`baseplate.experiments.Experiments`
+    to an attribute on the :term:`context object`.
+    """
+    def __init__(self, path, event_queue):
+        """ExperimentContextFactory constructor
+
+        :param str path: Path to the experiment config file.
+        :param baseplate.events.EventQueue event_queue: Event queue used for
+            logging bucketing events to the event pipeline.  You can set this
+            to None to disabled bucketing events entirely.
```
bsimpson63: disabled

pacejackson:
:haircut_man: 

Cleaned up documentation.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+        event.set_field("experiment_id", experiment.id)
+        event.set_field("experiment_name", experiment.name)
+        event.set_field("owner", experiment.owner)
+        event.set_field("request_id", self._span.trace_id)
+        span_name = "{}.{}".format(self._context_name, "events")
+        with self._span.make_child(span_name) as child_span:
+            try:
+                self._event_queue.put(event)
+            except EventTooLargeError as exc:
+                logger.exception(
```
spladug: do we actually want the full traceback pasted in this case? it's not a proper crash so it might be cleaner just to log the message and move on?

baseplate/diagnostics/metrics.py
```diff
@@ -80,3 +80,7 @@ def on_finish(self, exc_info):
         self.timer.stop()
         suffix = "success" if not exc_info else "failure"
         self.batch.counter(self.base_name + "." + suffix).increment()
+
+    def on_log(self, name, payload):
+        if name == "error.object":
+            self.batch.counter.increment("errors.%s" % payload.__name__.lower())
```
spladug: whatcha think of leaving it capitalized for readability?

baseplate/diagnostics/metrics.py
```diff
@@ -80,3 +80,7 @@ def on_finish(self, exc_info):
         self.timer.stop()
         suffix = "success" if not exc_info else "failure"
         self.batch.counter(self.base_name + "." + suffix).increment()
+
+    def on_log(self, name, payload):
```
spladug: neeeeeeaaat!

docs/words.txt
```diff
@@ -23,6 +23,8 @@ Redis
 serializable
 serialized
 serializer
+subreddit
```
spladug: seems reasonable to add these to our wordlist :thinking:

baseplate/experiments/__init__.py
pacejackson: Fair, I'll change that.

docs/words.txt
pacejackson: Is there a different place I should add them?   `subreddits` was actually triggering the spell check.

spladug: nono, this is totally right and fine. I was just laughing that we didn't have this already.

pacejackson:
I took a little time to see how things would look if we went with the approach I talked about in here:
https://github.com/reddit/baseplate/pull/112#issuecomment-316852752

Pros:

- The specific experiment providers don't care about logging.
- Smaller experiment constructors/`from_dict` method signatures.
- Putting the event logging into the experiment level meant teasing it out into an `EventLogger` class which makes it simpler to do different types of logging if you need to.

Cons:

- One benefit of having the providers care about logging is you can be sure that a FeatureFlag or ForcedVariant experiment will not log a bucketing event and it is part of the interface you have to implement so you have to figure that out up-front. With this approach, you push this logic into the config parser and you don't have an interface that forces you to consider if you want to send events which makes it easier to make a mistake.
- You end up caching the experiment objects per request but you have to build an experiment object before constructing it so caching really only serves to prevent duplicate bucketing events
- One of the thoughts behind this is it makes the interface more like PlanOut except that it doesn't quite.  If we did eventually decide to use a more advanced experiment framework, like PlanOut, it would probably make more sense to just do a separate integration and phase out the existing one.
- It doesn't quite fit in the way other parts of baseplate work where the instrumentation lives in the top level of the integration.


In general, it made some parts nicer and but left other parts lacking and did not seem to provide enough of a benefit.

spladug:
re-:fish: to be clear. thanks for the alternate approach. awesome stuff.

baseplate/experiments/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+        event_type = "bucket"
+        event = Event("bucketing_events", event_type)
+        for field, value in iteritems(extra_event_fields):
+            event.set_field(field, value)
+
+        event.set_field("variant", variant)
+        event.set_field("experiment_id", experiment.id)
+        event.set_field("experiment_name", experiment.name)
+        event.set_field("owner", experiment.owner)
+        event.set_field("request_id", self._span.trace_id)
```
pacejackson: I like the idea of including this, but I'm not sure the trace_id has sufficient entropy to be a unique request identifier.

spladug: 64 random bits? should be pretty good

pacejackson: Is it? It looks like r2 handles ~1 million requests/minute which gives us ~1.4 billion requests per day so after 5 days we'd have a 1/2 chance of having a collision.

spladug: Maybe combine it with the time?

pacejackson: I was thinking the same thing, maybe something we want to work into when we parse the trace headers and thread it through?

pacejackson:
:haircut_man: 

I made some changes that came up while working on the [r2 integration](https://github.com/reddit/reddit-public/pull/7284), mainly that oauth_client_id needs to be case sensitive, unlike the other parameters.  I'll note, I'm not crazy about the approach taken in here to accomplish that, but the alternatives aren't great either.  Here are the options I considered:

1. Make case sensitivity configurable.  Keeps the configuration flexible so experiment designers can add new case sensitive parameters without having to update baseplate and the caller does not need to worry about messing with the case of their inputs, we'll `lower` them if they are not case sensitive and preserve it if they are.  The downside is it adds another layer of complexity for what is currently a one-of issue.
2. Hard-code "oauth_client_id" as case sensitive.  Simpler but means that we need to update baseplate if we want to add more case sensitive parameters.
3. Handle case-sensitivity in the config generation layer.  In this approach, everything that is not case-sensitive we `lower` before creating the config and expect the caller to `lower` their inputs unless it is explicitly case sensitive.  While this makes baseplate cleaner, I think it puts too much of a burden on the engineer adding the experiment since they have to know that all of their string inputs should be `lower`-ed _except_ for these special cases.

baseplate/experiments/__init__.py
spladug: I guess the real question is how would we be using this and over what time period does it really have to be unique?

pacejackson: I was mainly thinking it would be used to:
1. See all the bucketing events for a single request.
2. Let us de-dup bucketing events for the same request that were fired by different services.

spladug:
:fish: seems reasonable. as a potential alternative for the case-sensitive stuff, what about using a different type for the parameter? e.g. instead of a string (which we always lower) we use a byte array?

baseplate/experiments/__init__.py
spladug: In the data stack? I think it'd be fine to just limit the dates being queried to ensure uniqueness then?

pacejackson:
> instead of a string (which we always lower) we use a byte array

I don't believe we can get byte arrays from JSON so we still need some way to communicate that this parameter is a special case.  Also, I'm not sure how we want to handle the input part of that, would the caller be responsible for knowing that this is a string or a byte array?

baseplate/experiments/providers/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+            "that is owned by <%s>. Please clean up.",
+            experiment_id,
+            name,
+            experiment_type,
+            owner,
+        )
+        return ForcedVariantExperiment(None)
+
+
+__all__ = ["parse_experiment"]
```
pacejackson: @spladug According to other comments, the `unicode_literals` import breaks this.  I'm not sure that we really need one or the other, my thought would be to just remove the `__all__`.  Do you have a strong preference?

spladug: removing it seems fine.

spladug:
Fair point on the use of types, I didn't think it through enough. Carry on.

PR #114, opened 2017-07-18
Add in alternative MessageQueue backed by redis

jdost:
This offers an alternative `MessageQueue` handler that takes a Redis
pool and uses it for backing of the queue storage as opposed to the
POSIX system used normally.

Pros:
* Increased limits for queue size
* Some mechanism to back the queue up to disk (configuration dependent)
* No maximum limit for messages and message size
* No restriction on queue name (even supports empty strings as a key)

Cons:
* Blocking timeout behavior for reads is more restrictive
  (it uses the [`BLPOP`](https://redis.io/commands/blpop) which only allows
  integers for the seconds based timeout)
* More overhead before a timeout is issued compared to the POSIX variant

### DIFF
--- a/baseplate/context/redis.py
+++ b/baseplate/context/redis.py
@@ -5,7 +5,6 @@ from __future__ import unicode_literals
 
 import redis
 import redis.client
-import pickle
 from math import ceil
 
 from . import ContextFactory
@@ -188,18 +187,15 @@ class MessageQueue(object):
         else:
             message = self.client.blpop(self.queue, timeout=timeout)
 
-            if message:  # BLPOP returns a tuple of (key, value) and we just want value
-                message = message[1]
-
         if not message:
             raise message_queue.TimedOutError
 
-        return pickle.loads(message)
+        return message[1] if isinstance(message, tuple) else message
 
     def put(self, message, timeout=None):
         """Add a message to the queue.
         """
-        return self.client.rpush(self.queue, pickle.dumps(message))
+        return self.client.rpush(self.queue, message)
 
     def unlink(self):
         """Not implemented for redis variant
--- a/baseplate/context/redis.py
+++ b/baseplate/context/redis.py
@@ -152,7 +152,7 @@ class MonitoredRedisPipeline(redis.client.StrictPipeline):
 
 
 class MessageQueue(object):
-    """A redis-backed variant of :py:class:`~baseplate.message_queue.MessageQueue`.
+    """A Redis-backed variant of :py:class:`~baseplate.message_queue.MessageQueue`.
 
     :param name: can be any string.
 
@@ -188,10 +188,13 @@ class MessageQueue(object):
         else:
             message = self.client.blpop(self.queue, timeout=timeout)
 
+            if message:
+                message = message[1]
+
         if not message:
             raise message_queue.TimedOutError
 
-        return message[1] if isinstance(message, tuple) else message
+        return message
 
     def put(self, message, timeout=None):
         """Add a message to the queue.
@@ -203,14 +206,14 @@ class MessageQueue(object):
         return self.client.rpush(self.queue, message)
 
     def unlink(self):
-        """Not implemented for redis variant
+        """Not implemented for Redis variant
         """
         pass
 
     def close(self):
         """Close queue when finished
 
-        Will delete the queue from the redis server (Note, can still enqueue
+        Will delete the queue from the Redis server (Note, can still enqueue
         and dequeue as the actions will recreate the queue)
         """
         self.client.delete(self.queue)
--- a/baseplate/context/redis.py
+++ b/baseplate/context/redis.py
@@ -5,9 +5,12 @@ from __future__ import unicode_literals
 
 import redis
 import redis.client
+from math import ceil
 
 from . import ContextFactory
 from .. import config
+
+from .. import message_queue
 
 
 def pool_from_config(app_config, prefix="redis.", **kwargs):
@@ -148,3 +151,57 @@ class MonitoredRedisPipeline(redis.client.StrictPipeline):
     def execute(self, **kwargs):
         with self.server_span.make_child(self.trace_name):
             return super(MonitoredRedisPipeline, self).execute(**kwargs)
+
+
+class MessageQueue(object):
+    """A redis-backed variant of ``message_queue.MessageQueue``.
+
+    ``name`` can be any string.
+
+    ``client`` should be a ``redis.ConnectionPool`` or
+    ``redis.BlockingConnectionPool`` from which a client connection can be
+    created from (preferrably generated from the ``pool_from_config`` helper).
+
+    """
+    def __init__(self, name, client):
+        self.queue = name
+        if isinstance(client, redis.BlockingConnectionPool) \
+                or isinstance(client, redis.ConnectionPool):
+            self.client = redis.Redis(connection_pool=client)
+        else:
+            self.client = client
+
+    def get(self, timeout=None):
+        """Read a message from the queue.
+
+        :param int timeout: If the queue is empty, the call will block up to
+            ``timeout`` seconds or forever if ``None``, if a float is given,
+            it will be rounded up to be an integer
+        :raises: :py:exc:`TimedOutError` The queue was empty for the allowed
+            duration of the call.
+
+        """
+        if isinstance(timeout, float):
+            timeout = int(ceil(timeout))
+        message = self.client.blpop(self.queue, timeout=timeout)
+        if not message:
+            raise message_queue.TimedOutError
+
+        return message[1]
+
+    def put(self, message, timeout=None):
+        """Add a message to the queue.
+        """
+        return self.client.rpush(self.queue, message)
+
+    def unlink(self):
+        """Not implemented for redis variant
+        """
+        pass
+
+    def close(self):
+        """Close queue when finished
+        Will delete the queue from the redis server (Note, can still enqueue
+        and dequeue as the actions will recreate the queue)
+        """
+        self.client.delete(self.queue)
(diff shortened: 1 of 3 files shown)


baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+
+        :param int timeout: If the queue is empty, the call will block up to
+            ``timeout`` seconds or forever if ``None``, if a float is given,
+            it will be rounded up to be an integer
+        :raises: :py:exc:`TimedOutError` The queue was empty for the allowed
+            duration of the call.
+
+        """
+        if isinstance(timeout, float):
+            timeout = int(ceil(timeout))
```
jdost: This is kind of hacky as Redis only supports full second values for timeouts.  I am using a ceiling (as opposed to rounding) because anything below `0.5` would round to `0` and either be indefinitely blocking or be immediate (not use the `BLPOP` command).  Thoughts?

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+
+        if timeout == 0:
+            message = self.client.lpop(self.queue)
+        else:
+            message = self.client.blpop(self.queue, timeout=timeout)
+
+        if not message:
+            raise message_queue.TimedOutError
+
+        return message[1] if isinstance(message, tuple) else message
```
jdost: This is a bit gross.  The two methods in the `redis-py` library return different formats.  `blpop` will return a tuple with the key/queue and value (or `None` if it times out) and `lpop` will just return the value (or `None` if it is empty).  An option would be normalizing to some format in the conditional branching logic above, I am indifferent to which approach is better...

tests/integration/redis_tests.py
```diff
@@ -68,3 +73,49 @@ def test_pipeline(self):
...
+
+    def test_get_timeout(self):
+        message_queue = MessageQueue(self.qname, self.pool)
+
+        with contextlib.closing(message_queue) as mq:
+            start = time.time()
+            with self.assertRaises(TimedOutError):
+                message_queue.get(timeout=1)
+            elapsed = time.time() - start
+            self.assertAlmostEqual(elapsed, 1.0, places=0)
```
jdost: The precision here is pretty weak as the overhead when reaching out to the network layer and back seems to add some additional overhead if we rely on the `blpop` command to do the waiting and blocking.  There may be faster alternatives if this overhead is a concern.

baseplate/context/redis.py
spladug: Seems pretty reasonable to me. Well documented.

baseplate/context/redis.py
spladug: Is there ever a situation where the returned message itself could be a tuple? That could cause this logic to trigger accidentally. It seems reasonable to me to do the unpacking in the lpop/blpop branches, even though it'll be a bit icky to be doing the None check twice.

tests/integration/redis_tests.py
spladug: I'm concerned this could lead to spurious test failures. Can we test this via mocking? or just decide it's not worth testing redis's behaviour for this?

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
     def execute(self, **kwargs):
         with self.server_span.make_child(self.trace_name):
             return super(MonitoredRedisPipeline, self).execute(**kwargs)
+
+
+class MessageQueue(object):
+    """A redis-backed variant of ``message_queue.MessageQueue``.
```
spladug: doc nit: if you do `:py:class:`~baseplate.message_queue.MessageQueue`` sphinx will link the class and cut off the module prefixes to be shorter.

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+
+    ``name`` can be any string.
```
spladug: doc nit: can these be `:param`'d up for sphinx purposes?

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+
+    ``client`` should be a ``redis.ConnectionPool`` or
```
spladug: doc nit: same deal w/ ``monospace`` stuff vs. :py:class: for sphinx's purposes

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+    ``redis.BlockingConnectionPool`` from which a client connection can be
+    created from (preferrably generated from the ``pool_from_config`` helper).
+
+    """
+    def __init__(self, name, client):
+        self.queue = name
+        if isinstance(client, redis.BlockingConnectionPool) \
+                or isinstance(client, redis.ConnectionPool):
```
spladug: nit: pep8 says the binary operator (or) should be on the preceding line.

baseplate/context/redis.py
```diff
@@ -148,3 +151,62 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+        return self.client.rpush(self.queue, message)
+
+    def unlink(self):
+        """Not implemented for redis variant
+        """
+        pass
+
+    def close(self):
+        """Close queue when finished
+        Will delete the queue from the redis server (Note, can still enqueue
```
spladug: pep257 nit: blank line between the summary line and first graf of docs please

baseplate/context/redis.py
```diff
@@ -5,10 +5,13 @@
 
 import redis
 import redis.client
+from math import ceil
 
 from . import ContextFactory
 from .. import config
 
+from .. import message_queue
```
spladug: nit: this should be grouped with the imports above

baseplate/context/redis.py
jdost: I will change it, but it is this same way for the `message_queue::MessageQueue` docstring (which is where I copied the style from).

baseplate/context/redis.py
jdost: I tested this in a REPL: ```
>>> from redis import Redis
>>> r_client = Redis()
>>> r_client.rpush("testq", ("foo", "bar"))
1L
>>> val = r_client.lpop("testq")
>>> val
"('foo', 'bar')"
>>> type(val)
<type 'str'>
```

File: baseplate/context/redis.py (hunk 5 above)
Comment (spladug) [inline baseplate/context/redis.py]: roger, i'll fix that one up. been getting better at sphinx over time.

File: tests/integration/redis_tests.py (hunk 3 above)
Comment (jdost) [inline tests/integration/redis_tests.py]: Yeah, I did run into some failures for the equivalent test in the POSIX variant for precision issues on some cases.  It likely could be due to abstraction layers (I was running them inside a container) but it may be worth being aware of.

File: baseplate/context/redis.py (hunk 2 above)
Comment (jdost) [inline baseplate/context/redis.py]: Looking into this further, how redis stores the lists being used with the `LPOP` and `RPUSH` commands, *all* values are going to be cast as strings.  An option (but probably with some impact) would be translating the base message with a `pickle.dumps` and `pickle.loads` to preserve the rich datatype, but would then cause the inconsistency noted above, in which case cleanup would need to be done in the conditional block.

**Comment by jdost** [comment]
:haircut:

Cleaned up some of the style and sphinx recommendations and added pickling around queue operations to keep native datatypes.

File: baseplate/context/redis.py [hunk 10]
```diff
@@ -148,3 +150,66 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+        """
+        if isinstance(timeout, float):
+            timeout = int(ceil(timeout))
+
+        if timeout == 0:
+            message = self.client.lpop(self.queue)
+        else:
+            message = self.client.blpop(self.queue, timeout=timeout)
+
+            if message:  # BLPOP returns a tuple of (key, value) and we just want value
```
spladug: :+1: nice comment too.

baseplate/context/redis.py
```diff
@@ -148,3 +150,66 @@ def __init__(self, trace_name, server_span, connection_pool,
     def execute(self, **kwargs):
         with self.server_span.make_child(self.trace_name):
             return super(MonitoredRedisPipeline, self).execute(**kwargs)
+
+
+class MessageQueue(object):
+    """A redis-backed variant of :py:class:`~baseplate.message_queue.MessageQueue`.
+
+    :param ``name`` can be any string.
```
spladug: I don't think this syntax is right, see below. (Did it work in local test?)

baseplate/context/redis.py
```diff
@@ -148,3 +150,66 @@ def __init__(self, trace_name, server_span, connection_pool,
...
+        else:
+            message = self.client.blpop(self.queue, timeout=timeout)
+
+            if message:  # BLPOP returns a tuple of (key, value) and we just want value
+                message = message[1]
+
+        if not message:
+            raise message_queue.TimedOutError
+
+        return pickle.loads(message)
```
spladug: I'd really prefer to avoid pickle if we can. It's magical, hard to use with other languages involved, and has security complications. Maybe JSON if we really need to serialize? parameterize the serializer on this class? or just make it explicit that the client of this class is to handle serialization? (I think the posix one is all strings too right?)

jdost:
:haircut:

Reverted the `pickle` stuff and made some documentation changes.  Fun fact: if you don't add the new class to the docs definition, they never get included (or tested) in the docs steps.

Note: for some reason the travis build is failing because it insists on running a bad SHA that I force pushed over where I did not have `redis` capitalized.

baseplate/context/redis.py
spladug: aw, i rather liked how you'd moved this into the branches above before in the pickle version

spladug:
Since this proved out in the collectors, shall we merge it?

PR #117, opened 2017-07-25
Make FileWatcher use cached data when the underlying file goes bad

spladug:
This is a safety net so that if the underlying file becomes completely
corrupt the application can keep on working with stale data.

### DIFF
--- a/baseplate/file_watcher.py
+++ b/baseplate/file_watcher.py
@@ -82,19 +82,24 @@ class FileWatcher(object):
         the freshest data.
 
         """
+
         try:
-            file_changed = self._mtime < os.path.getmtime(self._path)
-        except OSError:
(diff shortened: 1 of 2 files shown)


spladug:
Good question. I did some spelunking:

[`os.path.getmtime`](https://docs.python.org/2/library/os.path.html#os.path.getmtime) says:

> The return value is a number giving the number of seconds since the epoch.

So we should be fine since the epoch is defined in UTC and doesn't do daylight savings. There is some weirdness with leap seconds, but we've generally used smearing in production and worst case we'd reload the file once or miss one update that happened right on the leap second.

This writeup seems to jive: http://www.madore.org/~david/computers/unix-leap-seconds.html#the-spec

PR #118, opened 2017-08-01
Add live data watcher and writer tools

spladug:
This adds the live data watcher sidecar daemon and some tooling to make it easier to write data into ZooKeeper from CI etc. I'm going to stay quiet in this pull request body and hope the code, docstrings, and sphinx docs explain what's going on here sufficiently. If not, please don't hesitate to say what's confusing or poorly explained!


baseplate/live_data/zookeeper.py
```diff
@@ -0,0 +1,69 @@
...
+        cfg.hosts,
+        timeout=cfg.timeout.total_seconds(),
+        auth_data=auth_data,
+        read_only=read_only,
+        connection_retry=dict(
+            max_tries=-1,  # keep reconnecting forever
+            delay=0.1,  # initial delay
+            backoff=2,  # exponential backoff
+            max_jitter=1,  # maximum amount to jitter sleeptimes
+            max_delay=60,  # never wait longer than this
```
gtaylor: I'm looking at http://kazoo.readthedocs.io/en/latest/_modules/kazoo/retry.html?highlight=max_delay

Does this mean that our backoff tops off at 60 seconds per retry?

Also, does this mean that we'll retry forever (max_tries=-1), topping out at one attempt per minute?

I ask for fear of causing a build backlog if we were to run this under Drone. We'd eventually hit the configurable build timeout, but the default is something like 30 minutes.

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+    while True:
+        time.sleep(HEARTBEAT_INTERVAL)
+
+        if zookeeper.connected:
+            for node in nodes:
+                try:
+                    logger.debug("Heartbeating %s", node.dest)
+                    os.utime(node.dest, None)
+                except OSError as exc:
+                    if exc.errno != errno.ENOENT:
```
rram: In what cases do we want this to be silently ignored? When the file's raison d'être is a sentinel value?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+                        logger.info("%s: could not heartbeat: %s",
+                                    node.dest, exc)
+
+
+def main():
+    arg_parser = argparse.ArgumentParser(
```
rram: nifty

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+        time.sleep(HEARTBEAT_INTERVAL)
+
+        if zookeeper.connected:
+            for node in nodes:
+                try:
+                    logger.debug("Heartbeating %s", node.dest)
+                    os.utime(node.dest, None)
+                except OSError as exc:
+                    if exc.errno != errno.ENOENT:
+                        logger.info("%s: could not heartbeat: %s",
```
rram: bump severity to warning?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+    arg_parser.add_argument("--debug", default=False, action="store_true",
+        help="enable debug logging")
+    args = arg_parser.parse_args()
+
+    if args.debug:
+        level = logging.DEBUG
+    else:
+        level = logging.INFO
+    logging.basicConfig(level=level, format="%(message)s")
+    logging.getLogger("kazoo").setLevel(logging.WARNING)
```
rram: I have a slight preference for a comment here

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+    nodes = cfg.nodes.values()
+
+    secrets = secrets_store_from_config(watcher_config)
+    zookeeper = zookeeper_client_from_config(
+        secrets, watcher_config, read_only=True)
+    zookeeper.start()
+    try:
+        watch_zookeeper_nodes(zookeeper, nodes)
+    finally:
+        zookeeper.stop()
```
rram: This is solely to prevent kazoo from issuing that warning on exit, right?

baseplate/live_data/writer.py
```diff
@@ -0,0 +1,119 @@
...
+from .._compat import configparser
+from ..secrets import secrets_store_from_config
+from .zookeeper import zookeeper_client_from_config
+
+
+logger = logging.getLogger(__name__)
+
+
+class WriterError(Exception):
+    exit_status = 0
```
rram: an error giving a zero exit status?!

baseplate/live_data/zookeeper.py
```diff
@@ -0,0 +1,69 @@
...
+        credentials = secrets.get_simple(path)
+        auth_data.append(("digest", credentials.decode("utf8")))
+
+    return KazooClient(
+        cfg.hosts,
+        timeout=cfg.timeout.total_seconds(),
+        auth_data=auth_data,
+        read_only=read_only,
+        connection_retry=dict(
+            max_tries=-1,  # keep reconnecting forever
```
jdost: Do we want this to *always* be infinite or would an optional override be preferable for the CLI helpers?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+
+def watch_zookeeper_nodes(zookeeper, nodes):
+    for node in nodes:
+        watcher = NodeWatcher(node.source, node.dest)
+        zookeeper.DataWatch(node.source, watcher.on_change)
+
+    while True:
+        time.sleep(HEARTBEAT_INTERVAL)
+
+        if zookeeper.connected:
```
spladug: Kazoo does a lot of logging on its own (quite verbosely in fact). With debug turned off, this is what logging looks like for a watcher where I suddenly shut down zookeeper for a little while then bring it back:

```
ubuntu@baseplate:~/src/baseplate$ python -m baseplate.live_data.watcher live.ini
Updating 'alpha.json'
Removing 'beta.json'; watched node deleted.
Updating 'alpha.json'
Connection dropped: socket connection broken
Transition to CONNECTING
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection error: Connection refused
Connection dropped: socket connection broken
Session has expired
Updating 'alpha.json'
```

baseplate/live_data/writer.py
rram: Is it because NoChangesError is a zero? I think I'd rather have the inverted logic. Errors should default to "bad" things unless we explicitly say the error doesn't matter.

baseplate/live_data/watcher.py
gtaylor: Ah, that certainly gets the point across.

baseplate/live_data/zookeeper.py
spladug: > Does this mean that our backoff tops off at 60 seconds per retry?
> Also, does this mean that we'll retry forever (max_tries=-1), topping out at one attempt per minute?

Yup.

> I ask for fear of causing a build backlog if we were to run this under Drone. We'd eventually hit the configurable build timeout, but the default is something like 30 minutes.

That's a good point. The cases I'm thinking of using this in Drone would be pretty important to block on though, so I'd like it to fail loudly but not just let the user move on. Does that seem reasonable?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+HEARTBEAT_INTERVAL = 60
+
+
+class NodeWatcher(object):
+    def __init__(self, source, dest):
+        self.source = source
```
pacejackson: What is this actually used for?

baseplate/live_data/zookeeper.py
gtaylor: I think that's very reasonable. As long as you sys.exit with non-zero, the build will fail loudly.

There is nothing scientific to this suggestion, but choosing a timeout that keeps overall build duration under 10 minutes would be ideal. Failing faster < 3 minutes would be preferred, but we can definitely work with sub-10 minutes.

baseplate/live_data/writer.py
```diff
@@ -0,0 +1,119 @@
...
+def write_file_to_zookeeper(zookeeper, source_file, dest_path):
+    logger.info("Writing to %s in ZooKeeper...", dest_path)
+
+    try:
+        current_data, stat = zookeeper.get(dest_path)
+        current_version = stat.version
+    except NoNodeError:
+        raise NodeDoesNotExistError
+
+    new_data = source_file.read()
```
rram: do you want to catch the error if you fail to read the file?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+        zookeeper.DataWatch(node.source, watcher.on_change)
+
+    while True:
+        time.sleep(HEARTBEAT_INTERVAL)
+
+        if zookeeper.connected:
+            for node in nodes:
+                try:
+                    logger.debug("Heartbeating %s", node.dest)
+                    os.utime(node.dest, None)
```
pacejackson: Won't this cause us to re-parse this data even if it hasn't changed after every heartbeat if we are using a FileWatcher?  Is that ok?

baseplate/live_data/zookeeper.py
spladug: OK! I did some more reading/testing and added a comment that explains my findings and thought processes. How does that look?

baseplate/live_data/zookeeper.py
spladug: I added a comment explaining my rationale, does that seem reasonable?

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+                logger.info("Removing %r; watched node deleted.", self.dest)
+                os.unlink(self.dest)
+            except OSError as exc:
+                logger.debug("%s: couldn't unlink: %s", self.dest, exc)
+            return
+
+        logger.info("Updating %r", self.dest)
+        with open(self.dest + ".tmp", "wb") as tmpfile:
+            tmpfile.write(data)
+        os.rename(self.dest + ".tmp", self.dest)
```
spladug: Added a comment, how's that?

baseplate/live_data/watcher.py
spladug: Yeah, it cleans up the background thread/greenlet.

baseplate/live_data/writer.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if current_data == new_data:
+        raise NoChangesError
```
spladug: Agreed, good point. I've made this a true/false return and simplified the exceptions a bit.

baseplate/live_data/writer.py
spladug: Good point. I've gotten rid of NoChangesError as Greg suggested and simplified the exceptions like you suggest.

baseplate/live_data/writer.py
spladug: I don't think there's anything useful I can do with the exception once caught, so probably not.

baseplate/live_data/watcher.py
spladug: Nothing, fixed!

spladug: Fixed, thanks.

baseplate/live_data/__init__.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+#from __future__ import unicode_literals This breaks __all__ on PY2
```
spladug: I do want to keep the comment so it's obvious why that `__future__` setting isn't being used here. Given that page, I agree that we should re-evaluate using that flag at all, but would like to keep that for a separate change.

baseplate/live_data/watcher.py
spladug: done

baseplate/live_data/watcher.py
spladug: I just didn't want the heartbeat to complain when the file was removed by the watcher when the node ceased to exist. In that case, it'd be an erroneous failure message.

baseplate/live_data/watcher.py
```diff
@@ -0,0 +1,106 @@
...
+            # the data node does not exist
+            try:
+                logger.info("Removing %r; watched node deleted.", self.dest)
+                os.unlink(self.dest)
+            except OSError as exc:
+                logger.debug("%s: couldn't unlink: %s", self.dest, exc)
+            return
+
+        logger.info("Updating %r", self.dest)
+        with open(self.dest + ".tmp", "wb") as tmpfile:
```
spladug: You thinking of the s3get thing the other day? I'm not using `tempfile.NamedTemporaryFile` here so it's not ever getting `0600` set on it. _That said_, it might be nice to allow permissions to be set like the secret fetcher does.

baseplate/live_data/watcher.py
spladug: I've a comment elsewhere that (hopefully) explains this, but you're right that it's not obvious here. I've added one here as well to point you to the answer. How's that look / sound?

baseplate/live_data/watcher.py
spladug: Agh! You're right. Let me think about this for a bit.

pacejackson: Maybe it makes sense to just also have an empty `node.dest + '.heartbeat'` file that this updates?

baseplate/live_data/writer.py
rram: sorry, I mean you catch and throw the other exceptions to generate a pretty error message and exit output. But this one will just do the regular stack trace bail. Cool if you don't want to change it, but it is slightly different behavior from the rest.

baseplate/live_data/watcher.py
spladug: Yeah, that's the way I'm leaning.

baseplate/live_data/watcher.py
rram: Yeah. I was just assuming that whatever could be here would be a secret and so we'd want to make sure no one is snooping. Also wanted to be wary of naming attacks since the filename is static.

baseplate/live_data/writer.py
spladug: We'll already get a nice error message from argparse if the file couldn't be opened at all, so this would only really come up if something's weird about the file once it's opened. I'm not super worried about that case.

baseplate/live_data/watcher.py
gtaylor: The comment on line 58 helps me (not super stronk with ZK) a good deal, at least.

gtaylor:
As with most things, we may still need to iterate in reaction to feedback, but I think this is ready to merge as a v1!

baseplate/live_data/watcher.py
rram: IRLed. I think he agreed with me but the point is also moot.

baseplate/live_data/writer.py
rram: I should have known argparse would do the error message there

baseplate/live_data/watcher.py
spladug: I think I'd like to keep it as is. I've upped the time between heartbeats to 5 minutes to lessen the load, but I'm concerned that using a separate heartbeat file would allow issues like fouled permissions on the watched file to escape monitoring. We really want to know that the watched file is getting updated, not some other canary file.

baseplate/live_data/watcher.py
spladug: I dropped the ENOENT check :)

rram:
Oh and you might want to add a comment about the reparsing and the not using a separate heartbeat file.

spladug:
Roger, to-do list for before I haircut:

* [ ] comment on why heartbeats are like they are
* [ ] file permissions configuration?
* [ ] comment about kazoo logging configuration

docs/baseplate/live_data.rst
```diff
@@ -0,0 +1,89 @@
...
+
+   nodes.a.source = /a/node/in/zookeeper
+   nodes.a.dest = /var/local/file-on-disk
+
+   nodes.b.source = /another/node/in/zookeeper
+   nodes.b.dest = /var/local/another-file
+
+Each of the defined ``nodes`` will be watched by the daemon.
+
+The watcher daemon will touch the ``mtime`` of the local files once a minute to
```
pacejackson: five minutes now

baseplate/live_data/watcher.py
spladug: done

docs/baseplate/live_data.rst
spladug: fixed, thanks

baseplate/live_data/watcher.py
spladug: There shouldn't be any secrets coming from this stuff (hopefully) since that should be the domain of the secrets fetcher and vault. But, it's always possible!

spladug:
:haircut_man: Thanks for all the feedback so far, it's definitely greatly improved because of it!

tests/unit/config_tests.py
```diff
@@ -180,6 +184,26 @@ def test_invalid_percentage(self):
...
 
 
+class UnixUserTests(unittest.TestCase):
+    def test_valid_user(self):
+        result = config.UnixUser("root")
+        self.assertEqual(result, 0)
+
+    def test_invalid_user(self):
+        with self.assertRaises(ValueError):
+            config.UnixUser("fhqwhgads")
```
rram: 🔕 maybe something more obviously bad like `!`

tests/unit/config_tests.py
```diff
@@ -180,6 +184,26 @@ def test_invalid_percentage(self):
...
+
+
+class UnixGroupTests(unittest.TestCase):
+    def test_valid_group(self):
+        result = config.UnixGroup("root")
+        self.assertEqual(result, 0)
+
+    def test_invalid_group(self):
+        with self.assertRaises(ValueError):
+            config.UnixGroup("fhqwhgads")
```
rram: 🔕 see above

tests/unit/config_tests.py
spladug: aw, but i said _come on, fhqwghads_ i said come on fhqwghads!

spladug:
b66be68bfc3be7b95f7cb313303d6054d510b4c0...d61a48df2024e05d7293129f7050a9680885c3b0

tests/unit/config_tests.py
fruchtose: https://soundcloud.com/dj-arique/sets/daft-punk-vs-strong-bad-random

PR #120, opened 2017-08-17
Authentication context propagation

jdost:
Adds wrappers and handlers for retrieving propagated authentication information from upstream requests into an easy to use interface and standard system for further downstream propagation.  Overall the process is to convert the JWT payload received from the authentication service into an `AuthenticationContext` which can do validation and extraction, then store this is the current context and downstream requests via the built in thrift context will pull this value out of the current context and store the necessary information in a standard way for downstream services to easily and automatically retrieve.

There are some assumptions that may need to be refined, such as the way that the public RSA key for the authentication service is stored and retrieved, currently it is assumed (and hardcoded) to be in a `SecretStore` object under the simple key of `jwt/authentication/secret`.  Also, it is currently dependent on the originating service of the authentication to correctly package and define the information in the ongoing context for retrieval in the context handlers.  Any HTTP requests to downstream services will require the correct header be defined by hand (as there is no baseplate based HTTP client).

### DIFF
--- a/baseplate/context/thrift.py
+++ b/baseplate/context/thrift.py
@@ -108,7 +108,7 @@ class PooledClientProxy(object):
                                 and span.context.authentication.token:
                             prot.trans.set_header(
                                 "Authentication",
-                                str(span.context.authentication)
+                                span.context.authentication.token
                             )
                         client = self.client_cls(prot)
                         method = getattr(client, name)
--- a/baseplate/context/thrift.py
+++ b/baseplate/context/thrift.py
@@ -104,6 +104,12 @@ class PooledClientProxy(object):
                             prot.trans.set_header("Sampled", sampled)
                         if span.flags:
                             prot.trans.set_header("Flags", str(span.flags))
+                        if hasattr(span.context, "authentication") \
+                                and span.context.authentication.token:
+                            prot.trans.set_header(
+                                "Authentication",
+                                str(span.context.authentication)
+                            )
                         client = self.client_cls(prot)
                         method = getattr(client, name)
                         return method(*args, **kwargs)
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -157,16 +157,17 @@ class AuthenticationContext(object):
         defined secrets store, where application secret lookups should be made
     """
     def __init__(self, token=None, secrets=None):
-        self.token = token
-        self.secrets = secrets
+        self._secret_store = secrets
+        self._token = token
         self._payload = None
         self._valid = None
+        self.defined = self._token is not None
 
     def _secret(self):
-        if not self.secrets:
+        if not self._secret_store:
             raise UndefinedSecretsException
 
-        return self.secrets.get_simple("jwt/authentication/secret")
+        return self._secret_store.get_simple("jwt/authentication/secret")
 
     def attach_context(self, context):
         """Attach this authentication wrapper to the provided context
@@ -189,16 +190,16 @@ class AuthenticationContext(object):
         """
         if self._valid is not None:
             return self._valid
-        elif self.token is None:
-            return None
+        elif not self.defined:
+            return False
 
         try:
-            self._payload = jwt.decode(self.token, self._secret(),
+            self._payload = jwt.decode(self._token, self._secret(),
                                        algorithms='RS256')
             self._valid = True
-        except jwt.ExpiredSignatureError:
+        except jwt.ExpiredSignatureError:  # when the token has expired
             self._valid = False
-        except jwt.DecodeError:
+        except jwt.DecodeError:  # When the token is malformed
             self._valid = False
 
         return self._valid
@@ -228,12 +229,12 @@ class AuthenticationContext(object):
             :py:class:`SecretsStore` has not been bound to the context handling
             class (means that the authentication payload could not be decrypted
             and the user_id returned)
-        :raises: :py:class:`UndefinedAuthenticationError` if there was no
+        :raises: :py:class:`WithheldAuthenticationError` if there was no
             authentication token defined for the current context
 
         """
-        if self.token is None:
-            raise UndefinedAuthenticationError
+        if not self.defined:
+            raise WithheldAuthenticationError
 
         return self.payload.get("sub", None)
 
@@ -251,14 +252,14 @@ class UndefinedSecretsException(Exception):
             "No SecretsStore defined for Authentication token parsing.")
 
 
-class UndefinedAuthenticationError(Exception):
+class WithheldAuthenticationError(Exception):
     """Error raised when attempting to read from an unset authentication token
 
     Occurs either because of a badly instantiated context or missing header
     coming from upstream requests.
     """
     def __init__(self):
-        super(UndefinedAuthenticationError, self).__init__(
+        super(WithheldAuthenticationError, self).__init__(
             "No Authentication token provided for this context.")
 
 
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -125,8 +125,7 @@ class TraceInfo(_TraceInfo):
 
 
 class AuthenticationContextFactory(object):
-    """Building factory for :py:class:`AuthenticationContext` objects, handles
-    context dependency for the secrets store.
+    """Factory for consistent AuthenticationContext creation.
 
     This factory should be passed into the constructor of any upstream
     :doc:`integrations <integration/index>` so that it is aware of the
@@ -140,9 +139,9 @@ class AuthenticationContextFactory(object):
         self.secrets = secrets_store
 
     def make_context(self, token):
-        """:py:class:`AuthenticationContext` builder
+        """Builds :py:class:`AuthenticationContext` using stored values
 
-        :param token: JWT value originating from the Authentication service
+        :param token: token value originating from the Authentication service
             either directly or from an upstream service
         :rtype: :py:class:`AuthenticationContext`
         """
@@ -152,7 +151,7 @@ class AuthenticationContextFactory(object):
 class AuthenticationContext(object):
     """Wrapper for the contextual authentication information
 
-    :param str token: the JWT value for the currently propagated authentication
+    :param str token: the value for the currently propagated authentication
         context
     :param baseplate.secrets.SecretsStore secrets_store: the application's
         defined secrets store, where application secret lookups should be made
@@ -238,29 +237,29 @@ class AuthenticationContext(object):
 
         return self.payload.get("sub", None)
 
-    def __str__(self):
-        return self.token if self.token else ""
-
 
 class UndefinedSecretsException(Exception):
-    """Exception raised when an :py:class:`AuthenticationContext` attempts to
-    parse an authentication JWT payload without having a
-    :py:class:`SecretsStore` defined to provide the necessary secrets values to
-    correctly decrypt and parse the token.
+    """Exception raised when no SecretsStore is defined during token parsing
+
+    Occurs in the :py:class:`AuthenticationContext` object when it attempts to
+    parse an authentication payload without having a :py:class:`SecretsStore`
+    defined to provide the necessary secrets values to correctly decrypt and
+    parse the token.
     """
     def __init__(self):
         super(UndefinedSecretsException, self).__init__(
-            "No SecretsStore defined.")
+            "No SecretsStore defined for Authentication token parsing.")
 
 
 class UndefinedAuthenticationError(Exception):
-    """Error raised when attempting to read from an authentication token that
-    is not defined, either because of a badly instantiated context or missing
-    header coming from upstream requests.
+    """Error raised when attempting to read from an unset authentication token
+
+    Occurs either because of a badly instantiated context or missing header
+    coming from upstream requests.
     """
     def __init__(self):
         super(UndefinedAuthenticationError, self).__init__(
-            "No Authentication JWT token provided for this context.")
+            "No Authentication token provided for this context.")
 
 
 class Baseplate(object):
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 
 import collections
 import random
+import jwt
 
 from .integration.wrapped_context import WrappedRequestContext
 from ._utils import warn_deprecated
@@ -121,6 +122,123 @@ class TraceInfo(_TraceInfo):
                 raise ValueError("invalid flags value")
 
         return cls(trace_id, parent_id, span_id, sampled, flags)
+
+
+class AuthenticationContextFactory(object):
+    """Building factory for :py:class:`AuthenticationContext` objects, handles
+    context dependency for the secrets store.
+
+    This factory should be passed into the constructor of any upstream
+    :doc:`integrations <integration/index>` so that it is aware of the
+    application's secret store and can create the authentication context for
+    the incoming requests.
+
+    :param baseplate.secrets.SecretsStore secrets_store: the application's
+        defined secrets store, where application secret lookups should be made
+    """
+    def __init__(self, secrets_store=None):
+        self.secrets = secrets_store
+
+    def make_context(self, token):
+        """:py:class:`AuthenticationContext` builder
+
+        :param token: JWT value originating from the Authentication service
+            either directly or from an upstream service
+        :rtype: :py:class:`AuthenticationContext`
+        """
+        return AuthenticationContext(token=token, secrets=self.secrets)
+
+
+class AuthenticationContext(object):
+    """Wrapper for the contextual authentication information
+
+    :param str token: the JWT value for the currently propagated authentication
+        context
+    :param baseplate.secrets.SecretsStore secrets_store: the application's
+        defined secrets store, where application secret lookups should be made
+    """
+    def __init__(self, token=None, secrets=None):
+        self.token = token
+        self.secrets = secrets
+        self.payload = {}
+        self._valid = None
+
+    def _secret(self):
+        if not self.secrets:
+            raise UndefinedSecretsException
+
+        return self.secrets.get_simple("jwt/authentication/secret")
+
+    @property
+    def valid(self):
+        """Validity of the current authentication context token
+
+        :type: bool
+        :raises: :py:class:`UndefinedSecretsException` if the
+            :py:class:`SecretsStore` has not been bound to the context handling
+            class (means that the authentication payload could not be decrypted
+            and validated)
+
+        """
+        if self._valid is not None:
+            return self._valid
+        elif self.token is None:
+            return None
+
+        try:
+            self.payload = jwt.decode(self.token, self._secret(),
+                                      algorithms='RS256')
+            self._valid = True
+        except jwt.ExpiredSignatureError:
+            self._valid = False
+        except jwt.DecodeError:
+            self._valid = False
+
+        return self._valid
+
+    @property
+    def account_id(self):
+        """Authenticated account_id for the current authenticated context
+
+        :type: account_id string or None if context authentication is invalid
+        :raises: :py:class:`UndefinedSecretsException` if the
+            :py:class:`SecretsStore` has not been bound to the context handling
+            class (means that the authentication payload could not be decrypted
+            and the user_id returned)
+        :raises: :py:class:`UndefinedAuthenticationError` if there was no
+            authentication token defined for the current context
+
+        """
+        if self.token is None:
+            raise UndefinedAuthenticationError
+        if not self.valid:
+            return None
+
+        return self.payload.get("sub", None)
+
+    def __str__(self):
+        return self.token if self.token else ""
+
+
+class UndefinedSecretsException(Exception):
+    """Exception raised when an :py:class:`AuthenticationContext` attempts to
+    parse an authentication JWT payload without having a
+    :py:class:`SecretsStore` defined to provide the necessary secrets values to
+    correctly decrypt and parse the token.
+    """
+    def __init__(self):
+        super(UndefinedSecretsException, self).__init__(
+            "No SecretsStore defined.")
+
+
+class UndefinedAuthenticationError(Exception):
+    """Error raised when attempting to read from an authentication token that
+    is not defined, either because of a badly instantiated context or missing
+    header coming from upstream requests.
+    """
+    def __init__(self):
+        super(UndefinedAuthenticationError, self).__init__(
+            "No Authentication JWT token provided for this context.")
 
 
 class Baseplate(object):
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -160,7 +160,7 @@ class AuthenticationContext(object):
     def __init__(self, token=None, secrets=None):
         self.token = token
         self.secrets = secrets
-        self.payload = {}
+        self._payload = None
         self._valid = None
 
     def _secret(self):
@@ -168,6 +168,14 @@ class AuthenticationContext(object):
             raise UndefinedSecretsException
 
         return self.secrets.get_simple("jwt/authentication/secret")
+
+    def attach_context(self, context):
+        """Attach this authentication wrapper to the provided context
+
+        :param context: request context to attach this authentication to
+
+        """
+        context.authentication = self
 
     @property
     def valid(self):
@@ -186,8 +194,8 @@ class AuthenticationContext(object):
             return None
 
         try:
-            self.payload = jwt.decode(self.token, self._secret(),
-                                      algorithms='RS256')
+            self._payload = jwt.decode(self.token, self._secret(),
+                                       algorithms='RS256')
             self._valid = True
         except jwt.ExpiredSignatureError:
             self._valid = False
@@ -195,6 +203,22 @@ class AuthenticationContext(object):
             self._valid = False
 
         return self._valid
+
+    @property
+    def payload(self):
+        """Decrypted payload of the authentication token.
+
+        :type: dict
+        :raises: :py:class:`UndefinedSecretsException` if the
+            :py:class:`SecretsStore` has not been bound to the context handling
+            class (means that the authentication payload could not be decrypted
+            and the user_id returned)
+
+        """
+        if not self.valid:
+            return {}
+
+        return self._payload
 
     @property
     def account_id(self):
@@ -211,8 +235,6 @@ class AuthenticationContext(object):
         """
         if self.token is None:
             raise UndefinedAuthenticationError
-        if not self.valid:
-            return None
 
         return self.payload.get("sub", None)
 
--- a/baseplate/integration/pyramid.py
+++ b/baseplate/integration/pyramid.py
@@ -29,7 +29,10 @@ import sys
 import pyramid.events
 import pyramid.tweens
 
-from ..core import TraceInfo
+from ..core import (
+    TraceInfo,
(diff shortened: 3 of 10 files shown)


baseplate/core.py
```diff
@@ -123,6 +124,123 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        self.token = token
+        self.secrets = secrets
+        self.payload = {}
+        self._valid = None
+
+    def _secret(self):
+        if not self.secrets:
+            raise UndefinedSecretsException
+
+        return self.secrets.get_simple("jwt/authentication/secret")
```
jdost: This could have some more defined error checking, currently the error that the `SecretsStore` throws if the simple key lookup fails is going to pass through here uncaught, which may be preferable or may cause issues.

baseplate/integration/pyramid.py
```diff
@@ -69,9 +77,12 @@ class BaseplateConfigurator(object):
     """
 
     # TODO: remove the default on trust_trace_headers once apps are updated
-    def __init__(self, baseplate, trust_trace_headers=False):
+    def __init__(self, baseplate, trust_trace_headers=False,
+                 auth_factory=None):
         self.baseplate = baseplate
         self.trust_trace_headers = trust_trace_headers
+        self.auth_factory = auth_factory if auth_factory \
+            else AuthenticationContextFactory()
```
jdost: This is mostly to gracefully handle any downstream services that are not setup to parse and handle authentication tokens from having issues propagating the headers further.  If this conditional is reached, attempting to read the token will cause errors due to undefined secrets in the handler.

jdost:
Also, this currently will fail for `python3` tests as we do not have a `python3-cryptography` package, which the `pyjwt` plugin requires to do the RSA based tokens.

baseplate/core.py
```diff
@@ -123,6 +124,123 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+            class (means that the authentication payload could not be decrypted
+            and the user_id returned)
+        :raises: :py:class:`UndefinedAuthenticationError` if there was no
+            authentication token defined for the current context
+
+        """
+        if self.token is None:
+            raise UndefinedAuthenticationError
+        if not self.valid:
+            return None
```
pacejackson: Relying on making these calls before accessing `self.payload` seems like a bug waiting to happen.  Maybe it makes more sense to have a `payload` `@property` method that checks these?

```python
@property
def payload(self):
    if self.token is None:
        raise UndefinedAuthenticationError
    if not self.valid:
        return {}
    return self._payload
```

docs/baseplate/core.rst
```diff
@@ -58,6 +58,37 @@ integrations <integration/index>`.
...
+--------------
+
+If the application is dealing with authentication either directly from the
+authentication service or from upstream services, there are some support interfaces
+available for easily working with passing context down or picking up passed down
+contexts.
+
+When receiving an initial authentication token from the authentication service, the
+application should properly store the token in an
+:py:class:`~baseplate.core.AuthenticationContext` object attached to the current
```
pacejackson: Should we add a helper function to do this so developers don't have to worry about setting it to the exact variable name that we want?

```python
def attach_authentication_context(context, authentication_context):
    context.authentication = authentication_context
```

tests/integration/pyramid_tests.py
```diff
@@ -51,6 +60,22 @@ def setUp(self):
...
+                    "value": self.TOKEN_SECRET,
+                },
+            },
+            "vault": {
+                "token": "test",
+                "url": "http://vault.example.com:8200/",
+            }
+        }
+        secrets = store.SecretsStore("/secrets")
+        secrets._filewatcher = mock_filewatcher
```
pacejackson: Rather than mocking out `secrets._filewatcher` it seems cleaner to me to mock out the SecretStore interface.  `AuthenticationContext` cares about the SecretStore but mocking out the FileWatcher makes these tests concerned about the inner workings of the SecretStore object rather than the interface it provides.

docs/baseplate/core.rst
jdost: I had thought about adding some classmethod to do the construction and binding, but also wanted to give something more explicit for how the binding looks.  I will ponder this some more...

tests/integration/pyramid_tests.py
jdost: I just copied this pattern from how the `get_simple` method on the secrets test was setup and run.  I didn't see any pattern for test helpers, but was going to try and consolidate secret creation for tests that need it into something shared rather than this ugly copy pasta that I used here.

baseplate/core.py
jdost: Originally I was wrapping the payload access entirely to allow for the error-ing (and had the no token error raise in the `valid` method) but wanted to make the conditions for this error to be more granular.  Only the `sub` field of the payload is being tested here as opposed to other (future) fields that may come down the line which I don't know if they should throw errors until we define how they should be handled in a richer format (scopes and origin service have both been discussed) and I would probably think those would have different conditions for errors.  I am kind of torn with this being adaptable for future proofing versus this being strict to avoid contextual missteps in the future.

pacejackson: I waffled back and forth on this too while I was thinking about the comment.  Maybe just renaming `self.payload` to `self._payload` is a nice medium where it communicates to developers that they shouldn't be messing with it and we can leave the checks the way they are?

baseplate/core.py
```diff
@@ -123,6 +124,123 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
         return cls(trace_id, parent_id, span_id, sampled, flags)
 
 
+class AuthenticationContextFactory(object):
```
jdost: I was afraid of the misunderstanding of what `Authentication` would mean for this.  The token conveys only a basic sense of authentication and the existence of the object does not necessarily correlate to having authentication, it just encapsulates the actual checking of the current authentication.  I don't know if the actual names `AuthenticationContext{,Factory}` make sense but I would prefer that whatever we call it, the wrapper and factory share the same basis rather than have an `[A]Factory` produce not an `[A]`...

baseplate/core.py
jdost: Yeah, I am not super familiar with how much we would dislike it if new fields were being added to the token and applications would need to stall for smart wrappers to access the new fields.  I can see doing what you had but not raise the `UndefinedAuthenticationError` to at least encapsulate the validation check but not attempt to do any more, it would also enforce the `payload` property to be read only, which is a plus.

baseplate/integration/pyramid.py
jdost: Yeah that is probably better, I was not sure if I wanted to do more for the conditional check on the param but that is probably YAGNI...

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+            self._valid = True
+        except jwt.ExpiredSignatureError:
+            self._valid = False
+        except jwt.DecodeError:
+            self._valid = False
+
+        return self._valid
+
+    @property
+    def payload(self):
```
spladug: Perhaps we should just not allow this so that people can't rely on internals of the auth token?

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+            class (means that the authentication payload could not be decrypted
+            and the user_id returned)
+        :raises: :py:class:`UndefinedAuthenticationError` if there was no
+            authentication token defined for the current context
+
+        """
+        if self.token is None:
+            raise UndefinedAuthenticationError
+
+        return self.payload.get("sub", None)
```
spladug: This seems dangerous since an application could get the account_id without ever having checked validity?

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+
+    def __str__(self):
```
spladug: Maybe a `__repr__` instead that just shows some tidbits about the object? I worry about putting user info into logs too much.

docs/baseplate/core.rst
```diff
@@ -58,6 +58,38 @@ integrations <integration/index>`.
...
+context::
+
+    authentication_token = retrieve_token_from_authentication_service()
+    authentication_context = AuthenticationContext(authentication_token, secrets_store)
+    authentcation_context.attach_context(context)
```
spladug: `authentcation`

setup.py
```diff
@@ -8,6 +8,7 @@
 install_requires = [
     "requests",
     "posix_ipc",
+    "pyjwt",
```
spladug: just to confirm: this ends up in the debian package dependencies right?

you'll probably also need to update `docs/requirements.txt` for doc building to work right still (shitty, sorry, i know :()

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+    """Wrapper for the contextual authentication information
+
+    :param str token: the JWT value for the currently propagated authentication
+        context
+    :param baseplate.secrets.SecretsStore secrets_store: the application's
+        defined secrets store, where application secret lookups should be made
+    """
+    def __init__(self, token=None, secrets=None):
+        self.token = token
+        self.secrets = secrets
```
spladug: can this be a private as well? no need for downstream users to know we've got a copy of the secrets store around (they should be using it from their own contexts for law of demeter etc.)

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+            either directly or from an upstream service
+        :rtype: :py:class:`AuthenticationContext`
+        """
+        return AuthenticationContext(token=token, secrets=self.secrets)
+
+
+class AuthenticationContext(object):
+    """Wrapper for the contextual authentication information
+
+    :param str token: the JWT value for the currently propagated authentication
```
spladug: There are a few places in the docs that explicitly mention that the token is a JWT. I think it'd be safer to continue describing it as an opaque authentication token as the clients should never be touching the JWT envelope themselves.

spladug: Maybe just don't bother documenting the parameters on this class at all since the (internal) factory should be the only one instantiating it?

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        context
+    :param baseplate.secrets.SecretsStore secrets_store: the application's
+        defined secrets store, where application secret lookups should be made
+    """
+    def __init__(self, token=None, secrets=None):
```
spladug: Why defaults here if the factory's instantiating it?

baseplate/core.py
spladug: Seems reasonable to allow the original exception up here. It's a fatal condition.

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        if self._valid is not None:
+            return self._valid
+        elif self.token is None:
+            return None
+
+        try:
+            self._payload = jwt.decode(self.token, self._secret(),
+                                       algorithms='RS256')
+            self._valid = True
+        except jwt.ExpiredSignatureError:
```
spladug: Looks like they all inherit from `InvalidTokenError` if you wanna reduce branches here: https://github.com/jpadilla/pyjwt/blob/master/jwt/exceptions.py

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        :type: bool
+        :raises: :py:class:`UndefinedSecretsException` if the
+            :py:class:`SecretsStore` has not been bound to the context handling
+            class (means that the authentication payload could not be decrypted
+            and validated)
+
+        """
+        if self._valid is not None:
+            return self._valid
+        elif self.token is None:
```
spladug: I feel like this should either just return `False` or raise an exception. It's kinda weird to have a three-state boolean.

docs/baseplate/core.rst
spladug: I like the way you've done this with the `attach_context` method.

spladug:
> Also, this currently will fail for python3 tests as we do not have a python3-cryptography package, which the pyjwt plugin requires to do the RSA based tokens.

Can you mark the relevant test `xfail` when PY3 please? Then the suite will pass but call out the expected failure.

pacejackson:
Do we want PY3 to fail though? We currently have python 3 baseplate services running in prod.

baseplate/context/thrift.py
```diff
@@ -104,6 +104,12 @@ def _call_thrift_method(self, name, *args, **kwargs):
                             prot.trans.set_header("Sampled", sampled)
                         if span.flags:
                             prot.trans.set_header("Flags", str(span.flags))
+                        if hasattr(span.context, "authentication") \
+                                and span.context.authentication.token:
+                            prot.trans.set_header(
+                                "Authentication",
+                                str(span.context.authentication)
```
spladug: It feels a bit weird to explicitly check if `.token` is available and then use `__str__` here. Maybe do one or the other? (I lean towards just using `.token` explicitly both places)

baseplate/core.py
spladug: :dizzy_face: I think I've ended up thinking this is good as is but totally see the confusion potential of `ContextFactory`

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
         return cls(trace_id, parent_id, span_id, sampled, flags)
 
 
+class AuthenticationContextFactory(object):
+    """Building factory for :py:class:`AuthenticationContext` objects, handles
```
spladug: "building factory"?

pep-257 nit: the first line of a docstring should be short enough to fit on one line.

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+    the incoming requests.
+
+    :param baseplate.secrets.SecretsStore secrets_store: the application's
+        defined secrets store, where application secret lookups should be made
+    """
+    def __init__(self, secrets_store=None):
+        self.secrets = secrets_store
+
+    def make_context(self, token):
+        """:py:class:`AuthenticationContext` builder
```
spladug: pep-257 nit:

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

jdost:
I have a stashed change to the `.travis.yml` file that adds all of the package dependencies, so I can add this but comment out the missing package and update the tests.  Will do it in a bit...

Python3 itself will not fail, it just doesn't have the package dependencies built (you can still install using `pip`) to have the travis instance set up in the standard way we are using (`python3-*` debian packages).

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        if self.token is None:
+            raise UndefinedAuthenticationError
+
+        return self.payload.get("sub", None)
+
+    def __str__(self):
+        return self.token if self.token else ""
+
+
+class UndefinedSecretsException(Exception):
```
spladug: "undefined" reads a bit weirdly to me here: how about `AuthenticationSecretsNotConfiguredException` ? (very long... hrm)

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+    """Exception raised when an :py:class:`AuthenticationContext` attempts to
+    parse an authentication JWT payload without having a
+    :py:class:`SecretsStore` defined to provide the necessary secrets values to
+    correctly decrypt and parse the token.
+    """
+    def __init__(self):
+        super(UndefinedSecretsException, self).__init__(
+            "No SecretsStore defined.")
```
spladug: might want to give a specific "defined for authentication context" or something. these exception strings sometimes end up devoid of context when doing single-line logging.

baseplate/core.py
```diff
@@ -123,6 +124,145 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+
+
+class UndefinedAuthenticationError(Exception):
```
spladug: "undefined" again: how about `NoAuthenticationContextError` ?

docs/baseplate/core.rst
spladug: (could turn this into a doctest if you want to be sure it works now and in the future)

docs/baseplate/core.rst
```diff
@@ -58,6 +58,38 @@ integrations <integration/index>`.
 .. autoclass:: TraceInfo
    :members: from_upstream
 
+Authentication
+--------------
+
+If the application is dealing with authentication either directly from the
```
spladug: "dealing with authentication" maybe call out why you would want to deal with authn, like "if the application wants to check the authenticated user for a given request to authorize data access..."

spladug:
@pacejackson py3 wouldn't fail, it just wouldn't be able to validate authn tokens until we update cryptography. Hoping to punt that work until the self-serve packaging stuff happens later this quarter and lets us skip a debianization step :)

baseplate/core.py
jdost: Not always.  For the originating service, it would explicitly calling to the authentication service and would be handling the raw payload from there.  We could write up helpers to handle talking to this helper which would handle this, but until we do, the token needs to be touched in some way.

baseplate/core.py
jdost: The `self.payload` property calls it, so it will raise it there.  If we want to not expose the payload itself, then the check would move back into here.

baseplate/context/thrift.py
jdost: That makes sense.  Originally I was trying to have the token conversion work via the string casting of the object, but could not get it to handle unset tokens cleanly.  I was hoping to have the case of an unset token handled internal to the wrapper rather than here so that the hand rolled HTTP implementations would not be too hairy.

baseplate/core.py
jdost: I didn't know how we would like to handle complete unauthenticated calls.  Because in that ideal, the authentication is not really valid or invalid, it just doesn't exist.  I am unsure if it is guaranteed that all calls into a service will have authentication attached and how services should handle if they are not, but a lack of authentication sounds like a different failure scenario than an expired authentication.

spladug: Is there any time you'd want to call `valid` and distinguish those though?

baseplate/core.py
jdost: I was not *super* confident we wanted to eat all of them.  I was trying to get an idea of various failure scenarios and try to only explicitly catch them (versus something like the token getting corrupted or other wicked things) and wanted to only handle the things I had seen (even though `DecodeError` shouldn't happen for correctly designed applications, I saw it when trying to use very short RSA keys).

baseplate/core.py
spladug: The token as an object does, yes, but not its guts.

baseplate/core.py
jdost: yeah, I always envisioned this error being pretty show stopping and mostly seen on startup or in development.

docs/baseplate/core.rst
jdost: The problem is that I have a "magical" method for showing token retrieval and undefined `secrets_store` and `context` variables used... :(

docs/baseplate/core.rst
pacejackson: agreed

docs/baseplate/core.rst
spladug: You can define a mock method in the test setup block.

baseplate/core.py
jdost: The comment was mostly in reference to your second comment (about not even documenting because only the factory is building it).  It is also a concern for another subsequent comment about double defaulting.

baseplate/core.py
jdost: I could see having another attribute on the object that is just `self.defined = self.token is not None` and using that instead and then just having this be
```
elif not self.defined:
    return False
```
🤔

baseplate/core.py
jdost: As we discussed before I made this change, I am fearful of new fields being added to to the token in the authentication service response and it requiring applications wait for a new baseplate release and upgrade in order to access them.  Ideally all of the attributes will have intelligent property wrappers to access them, but having a hard limitation feels more inhibiting for developers than beneficial.

baseplate/core.py
jdost: I am going to drop this, it was meant to be a wrapper for logic when "casting" the object into the header field, but too much logic was living outside of this to give it any real value.

baseplate/core.py
jdost: Agree it is not a good name, but `NoAuthenticationContextError` being raised *from* a method on an `AuthenticationContext` object seems super confusing.  I will ponder this name and see if I can find a good one.

setup.py
jdost: This would also need to add some other packages (`cryptography`) and I would need to confirm it works with the debian package builder...

baseplate/core.py
spladug: Agh, you're right, that's way more confusing! `NoAuthenticationTokenError`?

baseplate/core.py
spladug: Ahh interesting, that's a reasonable argument. Maybe comment that others are explicitly uncaught and why?

baseplate/core.py
spladug: On the flip side, I'm worried that applications would end up tied to the implementation of the token and it'd be really hard to change.

baseplate/core.py
spladug: Ah! Right, sorry. I got confused by `payload` vs `token`; kinda not dissimilar conceptually in this context.

baseplate/core.py
jdost: The optimist in me would like to think that it would make it to baseplate and be implemented using the methods there sooner rather than later...

baseplate/core.py
jdost: Probably, I left it accessible because it doesn't feel too scary to expose and I can't see much of an argument about what exposing it could hurt, I may spend some time and rework this and try to refactor how the location of the public key is defined (rather than having it hard-coded as it is now).

tests/unit/core_tests.py
```diff
@@ -14,11 +15,21 @@
...
+from baseplate.file_watcher import FileWatcher
+from baseplate.secrets import store
 
 from .. import mock
 
+cryptography_installed = True
+try:
+    import cryptography
+except:
+    cryptography_installed = False
```
jdost: This is ugly and I don't like it very much, but that fact may be a feature rather than a bug and will incentivize me to clean this up sooner rather than later.

tests/unit/core_tests.py
```diff
@@ -226,3 +237,55 @@ def test_from_upstream_handles_no_sampled_or_flags(self):
         span = TraceInfo.from_upstream(1, 2, 3, None, None)
         self.assertIsNone(span.sampled)
         self.assertIsNone(span.flags)
+
+
+class AuthenticationContextTests(unittest.TestCase):
+    VALID_TOKEN = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0X3VzZXJfaWQiLCJleHAiOjQ2NTY1OTM0NTV9.Q8bz2qccFOHLTQ6H3MPdjSh7wDkRQtbBuBwGMzNRKjDFSkCoVF5kiwhBUdwbW8UXO5iZn4Bh7oKdj69lIEOATUxFBblU8Do05EfjECXLYGdbr6ClNmldrB8SsdAtQYQ4Ud-70Z8_75QvkqX_TY5OA4asGJZwH9MC7oHey47-38I"
+    EXPIRED_TOKEN = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0X3VzZXJfaWQiLCJleHAiOjE1MDI5OTM3NTN9.OPBIxaOEx0hnnB_wrfCuqfSIeP0a1abNdoZ2KejXReKeETQathr-PW2GqAhjGcUdCG3rXK8ezFKXdlB65kloqNdQii5b3qaJ5PDIdMNxY0Oi7TAqH86oog_umm7G-_p4MPPhRjxUm6Qp85-EaJUgyv26BUKSYY7-KyySjnrmP8g"
+    TOKEN_SECRET = "-----BEGIN PUBLIC KEY-----\nMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC0Kd3qYtc6zI5tj3iKBux70BhE\nZLLJ7fAKNBUO7h9FCwUcYku+SFigzNOu3AAYt3seNgxl+cvMR2+SNwsa605J9D1v\n9eGmpcITQi85SeJnfR7LJUMu7RieY5wEl0RyuwnSkX3Gkv0+hZISC/XYcWEYolIi\n8725u7u/8HRtUeHoLwIDAQAB\n-----END PUBLIC KEY-----"
```
jdost: Just to note, these values were generated using a private key I spun up locally (and this is the public key) but it is not used anywhere (it is a super insecure 1024 no-passphrase keypair).  The "valid" token has a TTL of about 100 years (if we are still here when this expires, I will have no issue knowing that we have become immortal) and the expired token expired like 1 second after I generated it.

baseplate/core.py
jdost: Yeah, I actually used the wrong one in a few places during development.  I have thought about using something like `decrypted_token` but don't really like using such juicy words as `decrypted` in attribute names.  Ideally nothing outside of the wrapper accesses these attributes but we all know there will be some situation where it will make sense...

jdost:
:haircut:

Cleaned up some naming to try and convey the overall scenario instead of literal issues, removed most of the references to tokens being JWT outside of the very internal decoding logic, cleaned up some of the testing to allow for CI to pass, other misc things...

jdost:
`UndefinedSecrets` makes sense to me only in that if it comes up, it will be in initial rollout or implementation/development.  So it is not super helpful in terms of execution, but should not get through to execution unless someone is deploying untested code.

`WithheldAuthentication` is not my favorite, but without making some long name (`NotProvidedAuthentication` was what I was going with but `withheld` is similar in meaning to `not provided` which is how I got there) I am unsure of better options...

spladug:
Yeah, `UndefinedSecrets` is probably fine.

I think my qualm with `Withheld` is that it implies someone (or something) deliberately didn't provide something it had. What about `MissingAuthentication`?

PR #122, opened 2017-08-21
Experiment framework fixes and updates

pacejackson:
# Fixes

1. Allow non-string override values.
2. Handle `newer_than` correctly.  Should only check `newer_than` as a part of checking the `logged_id` targeting value.
3. Check overrides before checking the bucket_val.
4. Update the bucketing logic to better match what we do in r2.

# Updates

1. Remove `get_event_type` from experiments.
2. Cache the parsed `Experiment` objects in the `Experiments` client object.
3. Add tests for the fixes.

### DIFF
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -132,8 +132,9 @@ class Experiments(object):
         if experiment is None:
             return None
 
-        experiment = parse_experiment(experiment_config)
-        variant = experiment.variant(**kwargs)
+        span_name = "{}.{}".format(self._context_name, "variant")
+        with self._span.make_child(span_name, local=True, component_name=name):
+            variant = experiment.variant(**kwargs)
 
         do_log = True
 
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -163,11 +163,7 @@ class Experiments(object):
 
         extra_event_fields = extra_event_fields or {}
 
-        event_type = experiment.get_event_type()
-        if not event_type:
-            return
-
-        event = Event("bucketing_events", event_type)
+        event = Event("bucketing_events", "bucket")
         for field, value in iteritems(extra_event_fields):
             event.set_field(field, value)
 
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -59,6 +59,7 @@ class Experiments(object):
         self._span = server_span
         self._context_name = context_name
         self._already_bucketed = set()
+        self._experiment_cache = {}
 
     def _get_config(self, name):
         try:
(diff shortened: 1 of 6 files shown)


baseplate/experiments/__init__.py
```diff
@@ -77,6 +78,16 @@ def _get_config(self, name):
             logger.warning("Could not load experiment config: %s", str(exc))
             return
 
+    def _get_experiment(self, name):
+        if name not in self._experiment_cache:
```
spladug: Does something invalidate this if the underlying file changes? This'll cache away use of the filewatcher right?

pacejackson: No, nothing invalidates it so it does cache away the use of the filewatcher.  My thinking was you get a new `Experiments` object for each request so you're only caching away the file watcher for a single request.

pacejackson: Maybe that's a dangerous assumption to make though?  FWIW I've been thinking about the caching a bit more and I'm not really sure that we _need_ it.  It only shaves 20 ~ 30 microseconds off the `variant` time.  That's not nothing, especially in cases like the listing endpoints in r2 where we do over 1k feature checks, but even in those cases it's generally not a huge difference.  We have to do 3333 ~ 5000 feature checks before that adds up to 100 milliseconds.

bsimpson63: > My thinking was you get a new Experiments object for each request so you're only caching away the file watcher for a single request.

This is probably safe for now, but when we have long running processes (crons, queue consumers) in baseplate services this will mess them up.

If we have spans around this stuff we can get a good idea of the real life cost of excessive/unbatched feature checks.

bsimpson63:
💅

baseplate/experiments/__init__.py
spladug: The client should be re-fetching the experiment object in each new span context (i.e. each item fetched from the queue) since the user / data context has changed so I think maybe it's fine to keep the caching?

pacejackson:
:haircut_man: 

I kept the caching in and added a local span around checking the variant of an experiment.

baseplate/experiments/__init__.py
```diff
@@ -116,32 +126,36 @@ def variant(self, name, bucketing_event_override=None,
...
-        if not experiment_config:
+        experiment = self._get_experiment(name)
+
+        if experiment is None:
             return None
 
-        experiment = parse_experiment(experiment_config)
-        variant = experiment.variant(**kwargs)
+        span_name = "{}.{}".format(self._context_name, "variant")
+        with self._span.make_child(span_name, local=True, component=name):
```
pacejackson: I'll note, I'm not really sure that I'm naming this local span properly, in this case the span will look like `context_name.variant` and the value for LOCAL_COMPONENT will be `experiment_name`

pacejackson:
:haircut_man: 

Fixed a typo.

PR #123, opened 2017-08-29
Add support for v2 thrift-based events

spladug:
I've updated the docs too, so that should explain everything here. If something's confusing or not sufficiently explained in here please let me know and I'll update the code.

### DIFF
--- a/baseplate/events/queue.py
+++ b/baseplate/events/queue.py
@@ -22,6 +22,8 @@ import time
 import uuid
 
 from enum import Enum
+from thrift.util import Serializer
+from thrift.protocol.TJSONProtocol import TJSONProtocolFactory
 
 from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from baseplate.context import ContextFactory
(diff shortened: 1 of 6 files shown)


docs/baseplate/events.rst
```diff
@@ -6,6 +6,46 @@
...
+      ...
+
+
+   def my_handler(request):
+      event = Event(
+         source="baseplate",
+         action="test",
+         noun="baseplate",
+         client_timestamp=time.time() * 1000,
+         uuid=str(uuid.uuid4()),
```
pacejackson: 🔕 One nice thing about the legacy format is you don't have to worry about setting fields like `uuid` and `client_timestamp` that you want on every event in a specific format.  Is there some way we can work that into this?  Or does it make more sense to just have a snippet that we can tell people "if you want to use v2 events, here's something to help you get started"?

spladug: Yeah, I agree. That's definitely a downside here. This makes me a lot more sympathetic to distributing thriftfiles as pre-built python packages so we can add convenience wrappers to the generated code. Could also add some helpers to baseplate that just expect the right struct to be passed in or have a base schema that's public. :/

baseplate/events/queue.py
```diff
@@ -147,15 +149,46 @@ def __init__(self):
...
+        which publisher should send the events which can be useful for routing
+        to different event pipelines (prod/test/v2 etc.).
+    :param callable event_serializer: A callable that takes an event object
+        and returns serialized bytes ready to send on the wire. See below for
+        options.
+
+    """
 
-    def __init__(self, name):
+    def __init__(self, name, event_serializer=serialize_v1_event):
```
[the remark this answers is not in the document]
spladug: It's hard to get to the config section for the publishers from inside an app as baseplate's already taken just one section's info.

PR #125, opened 2017-09-13
Request context

pacejackson:
Provide a framework for passing context data about requests made to edge services between baseplate services.

1. Create Thrift structs to model the request context data.
2. Automatically initialize the `EdgeRequestContext` object from Thrift/HTTP headers.
3. Automatically attach the `Edge-Request` header to thrift requests when using the `PooledClientProxy` client object.
4. Integrate the new `User` and `EdgeRequestContext` objects into the Experiment framework.

NOTE:
This update depends on new fields being added to the AuthenticationContext token.  Currently I have stub functions for these new fields.

### DIFF
--- a/baseplate/context/thrift.py
+++ b/baseplate/context/thrift.py
@@ -106,11 +106,18 @@ class PooledClientProxy(object):
                             prot.trans.set_header("Flags", str(span.flags))
 
                         try:
-                            auth_token = span.context.authentication.token
+                            context_headers = span.context.request_context.header_values()
                         except AttributeError:
                             pass
                         else:
-                            prot.trans.set_header("Authentication", auth_token)
+                            prot.trans.set_header(
+                                "Authentication",
+                                context_headers.get("Authentication"),
+                            )
+                            prot.trans.set_header(
+                                "Edge-Request",
+                                context_headers.get("Edge-Request"),
+                            )
 
                         client = self.client_cls(prot)
                         method = getattr(client, name)
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -4,11 +4,21 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import collections
+import logging
 import random
+
 import jwt
+from thrift.protocol.TBinaryProtocol import TBinaryProtocolAcceleratedFactory
+from thrift.util import Serializer
 
 from .integration.wrapped_context import WrappedRequestContext
+from .thrift.ttypes import Loid as TLoid
+from .thrift.ttypes import Request as TRequest
+from .thrift.ttypes import Session as TSession
 from ._utils import warn_deprecated
+
+
+logger = logging.getLogger(__name__)
 
 
 class BaseplateObserver(object):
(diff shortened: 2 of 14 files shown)


baseplate/core.py
```diff
@@ -238,6 +244,156 @@ class (means that the authentication payload could not be decrypted
...
+            "cookie_created": self.cookie_created_ms,
+        }
+
+    @property
+    def id(self):
+        if self.authentication_context:
+            try:
+                return self.authentication_context.account_id
+            except WithheldAuthenticationError:
+                pass
```
pacejackson: Should this and `roles` catch the `WithheldAuthenticationError`s or let those bubble up?  After thinking about it some more, I think I'm leaning towards having them bubble up rather than catching them.

baseplate/core.py
```diff
@@ -238,6 +244,156 @@ class (means that the authentication payload could not be decrypted
...
+
+    @property
+    def oauth_client_id(self):
+        if not self.defined:
+            raise WithheldAuthenticationError
+
+        return self.payload.get("client_id", None)
+
+    @property
+    def oauth_client_type(self):
```
jdost: These will probably need docstrings as this object is generated in the sphinx docs.

baseplate/core.py
```diff
@@ -238,6 +244,156 @@ class (means that the authentication payload could not be decrypted
...
+        if not self.defined:
+            raise WithheldAuthenticationError
+
+        return self.payload.get("client_type", None)
```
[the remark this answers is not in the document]
pacejackson: Yep, I made an IO ticket earlier this sprint for adding that to the JWT payload.

baseplate/context/thrift.py
```diff
@@ -106,11 +106,18 @@ def _call_thrift_method(self, name, *args, **kwargs):
...
                             pass
                         else:
-                            prot.trans.set_header("Authentication", auth_token)
+                            prot.trans.set_header(
+                                "Authentication",
+                                context_headers["Authentication"],
+                            )
+                            prot.trans.set_header(
+                                "Request",
+                                context_headers["Edge-Request"],
```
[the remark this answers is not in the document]
pacejackson: [Yes](https://github.com/reddit/baseplate/pull/125/files#diff-e23a2f0193db987b0509d66964675524R327) _but_ it's probably better to switch these to `get`s to guard against changes to that.

baseplate/core.py
spladug: Yes, bubble.

pacejackson: Same question about the OauthClient stuff?

pacejackson:
:haircut_man:

baseplate/core.py
```diff
@@ -263,6 +323,232 @@ def __init__(self):
...
+            self._t_request.session = TSession()
+            if self._header:
+                try:
+                    Serializer.deserialize(
+                        self._HEADER_PROTOCOL_FACTORY,
+                        self._header,
+                        self._t_request,
+                    )
+                except Exception:
+                    pass
```
spladug: This should probably debug log.

baseplate/thrift/baseplate.thrift
```diff
@@ -18,3 +18,18 @@ service BaseplateService {
     */
     bool is_healthy(),
 }
+
+
+struct Loid {
```
spladug: can these have docstrings please?

docs/words.txt
```diff
@@ -18,8 +19,11 @@ localhost
 memcache
 Memcache
 memcached
+oauth
```
spladug: should be capitalized to OAuth

docs/words.txt
```diff
@@ -18,8 +19,11 @@ localhost
...
 Queing
 Redis
+Reddit
+reddit
```
spladug: we're technically not supposed to have any lower case reddits anymore... though I kinda love 'em still.

docs/baseplate/core.rst
```diff
@@ -58,6 +58,43 @@ integrations <integration/index>`.
...
+some support interfaces available for easily working with the request context data,
+passing context down, or picking up passed down contexts.
+
+If the application is an edge service that is directly serving reddit client
+requests, then the application should procure an
+:py:class:`~baseplate.core.AuthenticationContext` object and then store it along
+with the other context data in an :py:class:`~baseplate.core.EdgeRequestContext`
+object attached to the current context::
+
+    authentication_token = retrieve_token_from_authentication_service()
```
spladug: should the following section about authentication token hydration be subsumed into this or otherwise updated?

docs/baseplate/core.rst
```diff
@@ -58,6 +58,43 @@ integrations <integration/index>`.
 .. autoclass:: TraceInfo
    :members: from_upstream
 
+Edge Request Context
+--------------------
+
+If the application is directly receiving requests from a front-end clients or if it
```
spladug: this whole paragraph is kinda confusing/unnecessary given the next one covers similar stuff. how about giving a bit more concrete of examples of edge request context so people know this section's what they want: "If you want to know more detail about the original request from a user, such as URL or the kind of client, ..."

docs/baseplate/core.rst
pacejackson: 🤔 
Maybe it makes sense to deprecate attaching the authentication_context directly to the request object and just run everything through the `EdgeRequestContext`?

pacejackson:
:haircut_man:

baseplate/core.py
```diff
@@ -263,6 +324,231 @@ def __init__(self):
...
+            authentication token defined for the current context
+
+        """
+        return self.authentication_context.oauth_client_id
+
+    @property
+    def type(self):
+        """Authenticated client type for the current client
+
+        :type: string or None if context authentication is invalid
```
jdost: Because this *should* be one of a limited group of values, would it be better to translate the return value to some enum?

baseplate/core.py
```diff
@@ -151,6 +161,9 @@ def make_context(self, token):
 class AuthenticationContext(object):
     """Wrapper for the contextual authentication information
 
+    In general, this object should not be used directly but will rather be
+    wrapped by an :py:class:`EdgeRequestContext` object.
```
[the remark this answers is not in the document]
pacejackson: I think so, I like that the logic around the authentication token is contained in it's own object rather than making `EdgeRequestContext` into an uber-object or splitting the logic between the different classes that care about authentication.

baseplate/core.py
```diff
@@ -169,14 +182,6 @@ def _secret(self):
...
         return self._secret_store.get_simple("jwt/authentication/secret")
 
-    def attach_context(self, context):
-        """Attach this authentication wrapper to the provided context
-
-        :param context: request context to attach this authentication to
-
-        """
-        context.authentication = self
-
```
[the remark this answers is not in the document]
pacejackson: How about I put the helper method back in with a `warn_deprecated`?  Since the header is forwarded from the `context.request_context` attribute, I would prefer to discourage use of the `AuthenticationContext.attach_context` method.  If no one is currently using `context.authentication` I would rather just remove it to avoid any confusion or mis-use.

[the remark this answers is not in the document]
pacejackson: I re-added this code, I think we should consider removing it at some point though.

baseplate/core.py
pacejackson: I think we want this to be strings rather than Enums, I also think that the list of values should just live where the token is generated rather than in Baseplate so we can add new values without having to update Baseplate to support them.

pacejackson:
:haircut_man:

docs/baseplate/core.rst
```diff
@@ -167,6 +133,57 @@ statsd :py:meth:`~!baseplate.core.SpanObserver.on_finish`.
 
 .. _convenience_methods:
 
+Edge Request Context
+--------------------
+
+The :py:class:`~baseplate.core.EdgeRequestContext` provides an interface into both
```
spladug: this is a great user-focused explanation!

baseplate/core.py
spladug: We just need to be _very_ careful. A similar situation in r2 was "subreddit type". At first it was just public, then public/private, then public/private/restricted, then public/private/restricted/gold, then public/private/restricted/gold/employee. Lots of checks just did stuff like "if not private" and assumed public. Is there any way we can discourage that sort of future-unproofed check?

baseplate/core.py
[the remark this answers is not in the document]
spladug: warn_deprecated sounds fine. We're technically at a 0 major version still so breaking changes are "OK" but I really like to avoid them when at all possible because it's more friendly to open source and makes it easier to upgrade production services.

spladug:
:fish: in broad strokes. probably a few little details to hammer out but overall this looks great.

baseplate/core.py
pacejackson: I don't think we can completely _stop_ that sort of thing but we could consider giving this interface rather that just returning the type. 

```python
def is_type(self, *client_types):
    return self.auth_context.oauth_client_type in client_types
```

pacejackson:
:haircut_man: 
1. Updated docstring of `AuthenticationContext.attach_context` to note that it is deprecated and that attaching the `AuthenticationContext` directly will not forward the auth token.
2. Changed `OauthClient.type` property to `OauthClient.is_type` function.  This is to encourage developers to check that the client type is an allowed type.

pacejackson:
:haircut_man: 

I cleaned up the commits and made a couple of small updates (two of the updates got squashed down so I will link to the lines in the diff where they happened)

1. Rename some event fields to match the names that we currently use for bucketing events.  This is done when we log the event rather than in the `event_fields` functions because they are not really the names we want to use going forward.  This is one of the updates that got squashed, you can see the change starting here https://github.com/reddit/baseplate/pull/125/files#diff-63712203d887896d022bc211be536eb7R189
2. Add a test to validate the bucketing event payload.  This is the other update that got squashed, you can see the change here: https://github.com/reddit/baseplate/pull/125/files#diff-4efb8f285f57a2678fe92ee4e6413c54R44
3. Add some validation to `EdgeRequestContext.create` to verify that the `loid_id` value is in the fullname format and update/add tests for this.  These are the last three commits.

Sorry for the last bit of changes 😅

baseplate/core.py
```diff
@@ -170,7 +183,14 @@ def _secret(self):
         return self._secret_store.get_simple("jwt/authentication/secret")
 
     def attach_context(self, context):
-        """Attach this authentication wrapper to the provided context
+        """DEPRECATED - Attach this authentication wrapper to the given context
```
spladug: you can use `.. deprecated::` instead: http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated

docs/baseplate/core.rst
```diff
@@ -167,6 +133,57 @@ statsd :py:meth:`~!baseplate.core.SpanObserver.on_finish`.
...
+authentication and context information about the original request from a user.  For edge
+services, it provides helpers to create the initial object and serialize the context
+information into the appropriate headers.  Once this object is created and attached to
+the context, Baseplate will automatically forward the headers to upstream services so they
```
spladug: nit: we forward the information to downstream services (the request flows down to them)

PR #126, opened 2017-09-25
Add some new experiment fields

pacejackson:
Add some new fields:

1. version - a string to be passed along with bucketing events
2. start_ts - epoch timestamp float of the time to start the experiment
3. stop_ts - epoch timestamp float of the time to stop the experiment

Deprecate some existing fields:

1. expired - replaced with start_ts and stop_ts.

### DIFF
--- a/baseplate/experiments/providers/__init__.py
+++ b/baseplate/experiments/providers/__init__.py
@@ -4,12 +4,14 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import logging
+import time
 
 from datetime import datetime
 
 from .feature_flag import FeatureFlag
 from .forced_variant import ForcedVariantExperiment
 from .r2 import R2Experiment
+from ..._utils import warn_deprecated
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


baseplate/experiments/providers/__init__.py
```diff
@@ -54,16 +60,42 @@ def parse_experiment(config):
...
+        start_ts = config["start_ts"]
+        stop_ts = config["stop_ts"]
+    elif "expires" in config:
+        warn_deprecated(
+            "The 'expires' field is in experiment %s deprecated, you should "
+            "use 'start_ts' and 'stop_ts'." % name
+        )
+        start_ts = time.time()
+        expires = datetime.strptime(config["expires"], ISO_DATE_FMT)
+        epoch = datetime(1970, 1, 1)
```
phil303: does this need to be adjusted for time offset?

pacejackson: No, they're both timezone naive:

```python
In [1]: import pytz

In [2]: from datetime import datetime

In [3]: ISO_DATE_FMT = "%Y-%m-%d"

In [4]: DATE_EXPIRES = "2100-01-01"

In [5]: expires = datetime.strptime(DATE_EXPIRES, ISO_DATE_FMT)

In [6]: epoch = datetime(year=1970, day=1, month=1)

In [8]: expires_utc = expires.replace(tzinfo=pytz.utc)

In [12]: epoch_utc = epoch.replace(tzinfo=pytz.utc)

In [13]: (expires_utc - epoch_utc).total_seconds()
Out[13]: 4102444800.0

In [14]: (expires - epoch).total_seconds()
Out[14]: 4102444800.0

In [15]: (expires_utc - epoch_utc).total_seconds() == (expires - epoch).total_seconds()
Out[15]: True
```

baseplate/experiments/providers/__init__.py
```diff
@@ -27,10 +29,14 @@ def parse_experiment(config):
         * **name**: String experiment name, should be unique for each
           experiment.
         * **owner**: The group or individual that owns this experiment.
-        * **expires**: Date when this experiment expires in ISO format
-          (YYYY-MM-DD).  The experiment will expire at 00:00 UTC on the day
-          after the specified date.  Once an experiment is expired, it is
-          considered disabled.
+        * **version**: String to identify the specific version of the
+          experiment.
+        * **start_ts**: Epoch timestamp float of date and time when you want the
```
spladug: "A float of seconds since the epoch" would be more descriptive and then you don't need to dictionary up jargon! :)

PR #127, opened 2017-09-28
memcache quality of life improvements

bsimpson63:
👓 @spladug @pacejackson

### DIFF
--- a/baseplate/context/memcache/__init__.py
+++ b/baseplate/context/memcache/__init__.py
@@ -9,9 +9,40 @@ from ...context import ContextFactory
 from ... import config
 
 
+class OptInNoReplyClient(PooledClient):
+    """For writes override the default value of noreply to False.
+
+    noreply can be surprising and possibly dangerous because the client doesn't
+    wait for a response from the server and just returns the equivalent of a
+    success response. No exceptions occur in this mode.
+
+    Callers can still use noreply but they must explicitly opt in on each
+    write operation.
+
+    """
+
+    def set(self, key, value, expire=0, noreply=False):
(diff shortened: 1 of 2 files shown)


bsimpson63:
Ah, actually the version of pymemcache we're on (1.3.2) doesn't support default_noreply and instead all the operations default to `noreply=True`. The argument was added in version 1.3.3 https://github.com/pinterest/pymemcache/commit/ca9d579b722198de6188695afb63aa26689ecc71

bsimpson63:
💇 We could either upgrade pymemcache or make a simple wrapper around the client. I don't want to deal with packaging since there's no ubuntu package to backport.

pacejackson:
How do you feel about making the wrapper take the extra constructor argument and do what the updated client does rather than just making the default value `False`?

baseplate/context/memcache/__init__.py
```diff
@@ -9,9 +9,40 @@
...
+        return super(OptInNoReplyClient, self).delete_many(keys, noreply)
+
+    def incr(self, key, value, noreply=False):
+        return super(OptInNoReplyClient, self).incr(key, value, noreply)
+
+
 def pool_from_config(app_config, prefix="memcache.", serializer=None,
                      deserializer=None):
-    """Make a PooledClient from a configuration dictionary.
+    """Make a OptInNoReplyClient from a configuration dictionary.
```
pacejackson: "Make an

spladug:
It's a pure python package, I don't think packaging will be that hard. It feels pretty unpleasant to have to add a bunch of stuff here just because we're using an out-of-date library. :nail_care:

bsimpson63:
this is old, closing

PR #128, opened 2017-09-28
Request context fixes

pacejackson:


tests/unit/context/thrift_tests.py
```diff
@@ -84,3 +84,49 @@ def test_call_proxy_method(self, mock_enumerate):
...
+        def mock_set_header(header, val):
+            headers[header] = val
+
+        prot = self.mock_pool.connection().__enter__()
+        prot.trans.set_header = mock_set_header
+        child_span = self.mock_server_span.make_child().__enter__()
+        child_span.context.request_context.header_values.return_value = {
+            'Authentication': None,
+            'Edge-Request': None,
+        }
```
jdost: nit: you have almost the same chunk here and above, maybe make a helper method ?

baseplate/core.py
```diff
@@ -523,6 +520,9 @@ def create(cls, authentication_context=None, loid_id=None,
         :param str session_id: (Optional) ID for the current session cookie.
 
         """
+        from .thrift.ttypes import Loid as TLoid
```
spladug: could you put a comment above these explaining why so it's not cleaned up in a lint later?

PR #131, opened 2017-10-29
Change baseplate tween position

phil303:
Currently, the behavior of the baseplate/pyramid integration
is to handle all exceptions raised. This includes those used as
control flow and intended to be caught by an pyramid exception
view.

This is due to the ordering of the tweens. Here's the current
ordering of tweens, with pyramid's nomenclature in parens:

1. Request ingress (`tweens.INGRESS`)
2. Exception view (`tweens.EXCVIEW`)
3. baseplate tween
4. App handler code (`tweens.MAIN`)

Each tween wraps the one beneath it. Since baseplate is wrapping
the handler code, exceptions get caught by it, reraised, and
then caught by the exception view tween.

This is problematic because it has the effect of calling
`finish` on the server span too early, which triggers all the
observers' `on_finish` callbacks. When we have exceptions or
metrics in the exception view code, they're no longer implicitly
handled by baseplate.

To fix this, we just move the tween up above the exception view.
With this change, all code written by us is covered by the
baseplate tween's try/catch. If an exception view handles a
control flow exception, it doesn't make it up to the baseplate
layer unnecessarily. The final order ends up being:

1. Request ingress (`tweens.INGRESS`)
2. baseplate tween
3. Exception view (`tweens.EXCVIEW`)
4. App handler code (`tweens.MAIN`)

### DIFF
--- a/baseplate/integration/pyramid.py
+++ b/baseplate/integration/pyramid.py
@@ -17,6 +17,15 @@ An abbreviated example of it in use::
 
         return configurator.make_wsgi_app()
 
+.. warning::
+
+    Because of how Baseplate instruments Pyramid, you should not make an
+    `exception view`_ that handles the base :py:exc:`Exception` as this will
+    prevent Baseplate from seeing the unhandled error and reporting it
+    appropriately.
+
+    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html#exception-views
+
 """
 
 from __future__ import absolute_import
@@ -138,8 +147,19 @@ class BaseplateConfigurator(object):
 
     def includeme(self, config):
         config.add_subscriber(self._on_new_request, pyramid.events.ContextFound)
+
+        # Position of the tween is important. We need it to cover all code
+        # that can written in the app. This means that it should be above
+        # (wrap) both the exception view handling tween (EXCVIEW) and the
+        # request handling code (MAIN).
+        #
+        # The final order ends up being:
+        # 1. Request ingress (tweens.INGRESS)
+        # 2. baseplate tween
+        # 3. Exception view handler (tweens.EXCVIEW)
+        # 4. App handler code (tweens.MAIN)
         config.add_tween("baseplate.integration.pyramid._make_baseplate_tween",
-                         under=pyramid.tweens.EXCVIEW)
+                         over=pyramid.tweens.EXCVIEW)
 
         # the pyramid "scripting context" (e.g. pshell) sets up a
         # psuedo-request environment but does not call NewRequest. it does,
--- a/tests/integration/pyramid_tests.py
+++ b/tests/integration/pyramid_tests.py
@@ -33,10 +33,37 @@ class TestException(Exception):
     pass
 
 
+class ControlFlowException(Exception):
+    pass
+
+
+class ControlFlowException2(Exception):
+    pass
+
+
+class ExceptionViewException(Exception):
+    pass
+
+
 def example_application(request):
     if "error" in request.params:
         raise TestException("this is a test")
+
+    if "control_flow_exception" in request.params:
+        raise ControlFlowException()
+
+    if "exception_view_exception" in request.params:
+        raise ControlFlowException2()
+
     return {"test": "success"}
+
+
+def render_exception_view(request):
+    return
+
+
+def render_bad_exception_view(request):
+    raise ExceptionViewException()
 
 
 def local_tracing_within_context(request):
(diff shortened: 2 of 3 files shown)


phil303:
cc @bsimpson63

phil303:
This could use a test but I'll wait on the general go-ahead on approach.

pacejackson:
I think this change means that we can’t have “unhandled” error views like https://github.com/reddit/reddit-service-gateway/blob/e88c24a900d9c1d0d963aac4e10586975106a547/reddit_service_gateway/request_methods.py#L41 if we want those exceptions to be seen by baseplate for things like built in metrics or logging.  I think that would cause pyramid to catch all exceptions so baseplate would not seee them.

phil303:
Ah! Yep, thanks, I hadn't thought of that use case. 

I think the benefits of doing the tweens in this order outweigh being able to return a nice 500 error, which is arguably not that valuable. As an example, in that same block of code you posted, there is a [metric](https://github.com/reddit/reddit-service-gateway/blob/e88c24a900d9c1d0d963aac4e10586975106a547/reddit_service_gateway/request_methods.py#L43) being incremented that is not being sent at all (I can't find it in graphite to prove it though, but that's likely for that same reason.) That's the case for all of those exception views. And to me that's not the writer's fault, because that's totally unexpected behavior.

That being said, I'm guessing you're just pointing out the current behavior has dependencies and we need to be careful here.

pacejackson:
> I'm guessing you're just pointing out the current behavior has dependencies and we need to be careful here.

Yep! Depending on that the error page looks like, we might want to consider being able to configure baseplate to do something specific in that case. Overall, I think this is a good idea

ckwang8128:
Thanks for the detailed writeup on this. I agree that this is a good idea. Could you put a condensed (or not condensed) version of what the ordering is and why as a comment where you made the change? I think a generic 50* error return isn't valuable, and definitely can lead to bad debugging rabbit holes. 

Definitely want some tests for this as well, I think it's safe to go ahead with those. I'd still like to wait for Neil to get back and get eyes on this.

tests/integration/pyramid_tests.py
```diff
@@ -33,12 +33,39 @@ class TestException(Exception):
     pass
 
 
+class ControlFlowException(Exception):
+    pass
+
+
+class ControlFlowException2(Exception):
```
phil303: Not the best names but was running out of ideas. Definitely happy to take suggestions.

phil303:
Thanks @ckwang8128. I agree, Neil should have the final call. I've updated with tests and added that comment in the meantime.

ckwang8128:
nice! :nail_care: to indicate things look great right now.

phil303:
@spladug Good point, yes it would. I can't think of a way to way to catch that programmatically. For a docs addition, would throwing a warning down somewhere around [here](https://github.com/reddit/baseplate/blob/970b287a61eb2b6d7824b67789a03073c61dac6d/baseplate/integration/pyramid.py#L19) work?

spladug:
Yeah, that seems like a reasonable place to put it. Just explaining what specifically doesn't work and potentially documenting a workaround if any exists.

phil303:
Alright, checked in with all clients that might be affected by this (only android, ios, and web clients of gateway right now as far as I can tell) and switching to a plain 500 response over the json one won't be a problem for any of them. I'm going to make a PR to gateway to remove that and then move forward with this PR.

phil303:
:haircut: Added a warning into the docstring (not sure how to test it's correctness though). There's not an obvious workaround that I can think of unfortunately - you can't re-raise in the exception handler since you have to return a response object.

ckwang8128:
You can generate the docs locally and visit the build HTML:
`sphinx-build -M html docs/ build/` or `make docs` should do it. That'll generate documentation in `build/` I think. 

Pointing out needing `request.trace.finish` seems reasonable to me. 

:fish:

baseplate/integration/pyramid.py
```diff
@@ -17,6 +17,11 @@ def make_app(app_config):
 
         return configurator.make_wsgi_app()
 
+.. warning::
+
+    You should never set an exception handler for the base ``Exception``. This
```
spladug: I think this needs something a bit more concrete. Questions someone might think while reading this for the first time: What's an exception handler? (they appear to be called "exception views" in pyramid docs) what's a baseplate observer? (i.e. what does that actually mean to me as an application dev)

Maybe:

```
Because of how Baseplate instruments Pyramid, you should not 
make an `exception view`_ that handles the base :py:exc:`Exception`
as this will prevent Baseplate from seeing the unhandled error
and reporting it appropriately.

.. exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html
```

phil303: Oh yea, I like that a lot more. Thanks for the catch.

spladug: take my rST syntax with a grain of salt though :)

phil303: Ha yea I did :P

phil303:
:haircut:

phil303:
Awesome! Thanks everyone.

ckwang8128:
:fish:

Issue #132, opened 2017-11-01
Support kubernetes auth backend for vault


praxist:
From my understanding, we currently only authenticate to vault using the EC2 instance identity document. For the future when we run applications on kubernetes, this will not be sufficient as many arbitrary applications may run on the same instance in different pods.

The kubernetes auth backend is used by providing the role and a JWT mounted within kubernetes pods at `/run/secrets/kubernetes.io/serviceaccount/token` which is used to authenticate as a service account. Therefore, we simply just need to read the contents of this file and authenticate at a designated mount point for the kubernetes auth backend in order to get this working. 

Here is some sample code using the hvac library:
```
import hvac

TOKEN_FILE = '/run/secrets/kubernetes.io/serviceaccount/token'
VAULT_URL = 'xxxx' 
AUTH_ENDPOINT = '/v1/auth/kubernetes/login'

with open(TOKEN_FILE, 'r') as f:
    token = f.read()

role = 'xxxx'

params = {
  'jwt': token,
  'role': role
}

client = hvac.Client(url=VAULT_URL)
client.auth(AUTH_ENDPOINT, json=params)
```

I'm still getting familiar with the code, but it seems we may want to build this functionality into [VaultClientFactory](https://github.com/reddit/baseplate/blob/master/baseplate/secrets/fetcher.py#L117) rather than use hvac for more flexibility.

praxist:
👓  @spladug @gtaylor

gtaylor:
Will `AUTH_ENDPOINT` always be static, or would it differ if there are multiple Kubernetes auth backends set up on Vault?

IE: '/v1/auth/kubernetes/some-cluster/login'

praxist:
Just as you said - the auth endpoint will change if the vault backend is mounted elsewhere, like in the example of having multiple clusters.

gtaylor:
This all makes sense to me. I don't have a ton of useful advice on the baseplate implementation side, but Andrew Boyle and Adam Collins over on Platform have used the current baseplate+Vault setup enough to potentially have some opinions until Neil gets back.

spladug:
72ba4b023553fd7e71eb1173f3aeeb88b3a2981f

PR #134, opened 2017-11-08
Add support for Vault kubernetes auth backend

praxist:
related issue: https://github.com/reddit/baseplate/issues/132

In `VaultClientFactory`, `_make_client` now delegates the logic of fetching the client token to the appropriate auth backend method. To prevent breakage of downstream services, the original aws_ec2 authentication method is kept as default.

Services wishing to utilize the kubernetes auth backend will configure an ini like so:
```
[secret-fetcher]
vault.url = https://vault.url:8200
vault.role = some-service-role
vault.auth_type = kubernetes
vault.auth_path = /v1/auth/kubernetes/login
```

### DIFF
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -128,7 +128,7 @@ class VaultClientFactory(object):
 
     def _make_client(self):
         """Obtain a client token from an auth backend and return a Vault client with it."""
-        client_token, lease_duration = self.auth_type()
+        client_token, lease_duration = self.auth_type(self)
 
         return VaultClient(
             self.session,
@@ -173,7 +173,7 @@ class VaultClientFactory(object):
 
         logger.debug("Obtaining Vault token via kubernetes auth.")
         response = self.session.post(
-            urljoin(self.base_url, "v1/%s/login" % self.mount_point),
+            urljoin(self.base_url, "v1/auth/%s/login" % self.mount_point),
             json=login_data,
         )
         response.raise_for_status()
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -6,6 +6,8 @@ The file should contain a section looking like:
     [secret-fetcher]
     vault.url = https://vault.example.com:8200/
     vault.role = my-server-role
+    vault.auth_type = {aws_ec2,kubernetes}
+    vault.auth_path = /v1/auth/{aws-ec2,kubernetes}/login
 
     output.path = /var/local/secrets.json
     output.owner = www-data
@@ -18,7 +20,7 @@ The file should contain a section looking like:
         secret/three,
 
 where each secret is a path to look up in Vault. The daemon authenticates with
-Vault as the EC2 server it is running on. Vault can map that context to
+Vault using the auth backend designated by `auth_type`. Vault can map that context to
 appropriate policies accordingly.
 
 The secrets will be read from Vault and written to output.path as a JSON file
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -112,12 +112,12 @@ def ttl_to_time(ttl):
 class VaultClientFactory(object):
     """Factory that makes authenticated clients."""
     def __init__(self, base_url, role, auth_type, auth_path):
-        self.base_url  = base_url
-        self.role      = role
+        self.base_url = base_url
+        self.role = role
         self.auth_type = auth_type
         self.auth_path = auth_path
-        self.session   = requests.Session()
-        self.client    = None
+        self.session = requests.Session()
+        self.client = None
 
     def _make_client(self):
         """Obtain a client token from an auth backend and return a Vault client with it."""
@@ -155,9 +155,9 @@ class VaultClientFactory(object):
 
         """
         try:
-            with open(K8S_SERVICE_ACCOUNT_TOKEN_FILE , "r") as f:
+            with open(K8S_SERVICE_ACCOUNT_TOKEN_FILE, "r") as f:
                 token = f.read()
-        except IOError as exc:
+        except IOError:
             logger.debug("Could not read Kubernetes token file '%s'",
                          K8S_SERVICE_ACCOUNT_TOKEN_FILE)
             raise
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -6,8 +6,8 @@ The file should contain a section looking like:
     [secret-fetcher]
     vault.url = https://vault.example.com:8200/
     vault.role = my-server-role
-    vault.auth_type = {aws_ec2,kubernetes}
-    vault.auth_path = /v1/auth/{aws-ec2,kubernetes}/login
+    vault.auth_type = {aws,kubernetes}
+    vault.mount_point = {aws-ec2,kubernetes}
 
     output.path = /var/local/secrets.json
     output.owner = www-data
@@ -20,8 +20,15 @@ The file should contain a section looking like:
         secret/three,
 
 where each secret is a path to look up in Vault. The daemon authenticates with
-Vault using the auth backend designated by `auth_type`. Vault can map that context to
-appropriate policies accordingly.
+Vault as a role using a token obtained from an auth backend designated by `auth_type`.
+
+Currently supported auth types:
+    - aws: uses an AWS-signed instance identity document from the instance
+    metadata API
+    - kubernetes: uses a JWT mounted within a pod associated with a service account
+
+Upon authenticating with this token, the Vault client then gets access based
+upon the policies mapped to the role.
 
 The secrets will be read from Vault and written to output.path as a JSON file
 with the following structure:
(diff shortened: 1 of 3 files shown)


baseplate/secrets/fetcher.py
```diff
@@ -108,16 +111,76 @@ def ttl_to_time(ttl):
...
 
     def _make_client(self):
-        """Authenticate with Vault and return a client containing that token.
+        """Obtain a client token from an auth backend and return a Vault client with it."""
+        if self.auth_type == "aws_ec2":
+            client_token, lease_duration = self._vault_aws_ec2_auth()
+        elif self.auth_type == "kubernetes":
+            client_token, lease_duration = self._vault_kubernetes_auth()
+        else:
+            raise ValueError("Unsupported vault auth_type '%s' provided" % self.auth_type)
```
praxist: not entirely sure how we want to perform error handling here.

baseplate/secrets/fetcher.py
```diff
@@ -108,16 +111,73 @@ def ttl_to_time(ttl):
 
 class VaultClientFactory(object):
     """Factory that makes authenticated clients."""
-    def __init__(self, base_url, role):
+    def __init__(self, base_url, role, auth_type, auth_path):
```
gtaylor: If I understand correctly, this may break backward compatibility with things in the wild currently using VaultClientFactory. If that is the case, we should probably provide kwarg defaults to preserve the current EC2 auth default.

baseplate/secrets/fetcher.py
```diff
@@ -108,16 +111,73 @@ def ttl_to_time(ttl):
...
+                max_ttl=4h
+
+        See https://www.vaultproject.io/docs/auth/kubernetes.html for more info.
+
+        """
+        try:
+            with open(K8S_SERVICE_ACCOUNT_TOKEN_FILE, "r") as f:
+                token = f.read()
+        except IOError:
+            logger.debug("Could not read Kubernetes token file '%s'",
```
gtaylor: Should probably be a higher level than debug. Probably `error`, since this is probably concerning.

baseplate/secrets/fetcher.py
praxist: Couldn't find any reference to things using this, so we may be safe with just passing in the defaults through the configparser. No strong feelings towards either method though.

gtaylor: Sounds like a @spladug question, then. I will defer to his wisdom.

spladug: It's just an internal helper for this file. Should be safe.

baseplate/secrets/fetcher.py
```diff
@@ -108,16 +111,73 @@ def ttl_to_time(ttl):
...
+        response = self.session.post(
+            urljoin(self.base_url, self.auth_path),
+            json=login_data,
+        )
+        response.raise_for_status()
+        auth = response.json()["auth"]
+        return auth["client_token"], ttl_to_time(auth["lease_duration"])
+
+    def _vault_aws_ec2_auth(self):
+        """This authenticates with Vault as a specified role using its AWS EC2
```
spladug: [pep-257] asks for: "Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.". this line should keep being a summary like it was before rather than running into the rest of the first paragraph.

[pep-257]: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

baseplate/secrets/fetcher.py
```diff
@@ -108,16 +111,73 @@ def ttl_to_time(ttl):
...
-        This authenticates with Vault as a specified role using its AWS EC2
+        return VaultClient(
+            self.session,
+            self.base_url,
+            client_token,
+            lease_duration,
+        )
+
+    def _vault_kubernetes_auth(self):
+        """This authenticates with Vault as a specified role using its
```
spladug: same pep-257 request.

baseplate/secrets/fetcher.py
```diff
@@ -228,6 +280,8 @@ def main():
         "vault": {
             "url": config.String,
             "role": config.String,
+            "auth_type": config.Optional(config.String, default="aws_ec2"),
```
spladug: one of the goals of baseplate's config stuff is to make the app fail as fast as possible on configuration mistakes. to that end, getting more strict on the config parser here would allow us to ignore the possibility of configuration errors downstream (wrt your question above about error handling). i'd recommend something like:

```python
auth_types = {
    "aws": VaultClientFactory._vault_aws_ec2_auth,
    "kubernetes": VaultClientFactory._vault_kubernetes_auth,
}

...
config.Optional(config.OneOf(auth_types), default=auth_types["aws"]),
...
```

then you just pass that function along to the class and it does it from there

(:no_bell: just going with "aws" there since they've renamed the backend and it's shorter)

baseplate/secrets/fetcher.py
```diff
@@ -18,7 +20,7 @@
         secret/three,
 
 where each secret is a path to look up in Vault. The daemon authenticates with
-Vault as the EC2 server it is running on. Vault can map that context to
+Vault using the auth backend designated by `auth_type`. Vault can map that context to
```
spladug: there's a quick note of auth type above, but no list of the auth types with prose explanations of their purpose / configurables. it'd be nice to add that to this module docstring.

baseplate/secrets/fetcher.py
```diff
@@ -148,23 +208,15 @@ def _make_client(self):
         if nonce:
             login_data["nonce"] = nonce
 
-        logger.debug("Logging into Vault.")
+        logger.debug("Obtaining Vault token via aws_ec2 auth.")
         response = self.session.post(
-            urljoin(self.base_url, "/v1/auth/aws-ec2/login"),
+            urljoin(self.base_url, self.auth_path),
```
spladug: it feels a bit odd to have the client knowing vault API paths and passing them in directly. is it flexible enough to just parameterize the path based on mount point and let the user specify the mount point but not the whole v1/ ... /login part?

praxist: sure, we can just require mount_point.

baseplate/secrets/fetcher.py
praxist: It feels strange for things outside of `VaultClientFactory` to be aware of `_vault_aws_ec2_auth`. Where would be the best place to define auth types?

baseplate/secrets/fetcher.py
```diff
@@ -224,10 +282,18 @@ def main():
     parser.readfp(args.config_file)
     fetcher_config = dict(parser.items("secret-fetcher"))
 
+    auth_types = {
```
praxist: It feels strange for things outside of `VaultClientFactory` to be aware of `_vault_aws_ec2_auth`. Where might  be a better place to define auth types?

baseplate/secrets/fetcher.py
```diff
@@ -224,6 +224,14 @@ def _vault_aws_auth(self):
         store_nonce(auth["metadata"]["nonce"])
         return auth["client_token"], ttl_to_time(auth["lease_duration"])
 
+    @staticmethod
+    def auth_types():
```
praxist: @bsimpson63 @spladug thoughts on this implementation?

praxist:
💇‍♂️

baseplate/secrets/fetcher.py
```diff
@@ -228,6 +294,9 @@ def main():
         "vault": {
             "url": config.String,
             "role": config.String,
+            "auth_type": config.Optional(config.OneOf(**VaultClientFactory.auth_types()),
```
spladug: love it! much better encapsulated than my suggestion :)

baseplate/secrets/fetcher.py
```diff
@@ -108,16 +118,72 @@ def ttl_to_time(ttl):
...
+
+        return VaultClient(
+            self.session,
+            self.base_url,
+            client_token,
+            lease_duration,
+        )
+
+    def _vault_kubernetes_auth(self):
+        """Gets a client token from Vault through the Kubernetes auth backend.
```
spladug: sorry to be so picky, but for consistency pep-257 says it should be in the imperative i.e. "Get" vs "Gets".

baseplate/secrets/fetcher.py
```diff
@@ -18,8 +20,15 @@
         secret/three,
 
 where each secret is a path to look up in Vault. The daemon authenticates with
-Vault as the EC2 server it is running on. Vault can map that context to
-appropriate policies accordingly.
+Vault as a role using a token obtained from an auth backend designated by `auth_type`.
+
+Currently supported auth types:
+    - aws: uses an AWS-signed instance identity document from the instance
```
spladug: could you build the docs locally (`make docs` in the VM then go to `http://baseplate.local`) and make sure this list renders ok? reStructuredText can be a bit finnicky

baseplate/secrets/fetcher.py
praxist: no problem, consistency is good

praxist:
💇‍♂️

praxist:
72ba4b023553fd7e71eb1173f3aeeb88b3a2981f

PR #137, opened 2017-12-01
Emit an event to Pyramid after creating the ServerSpan

pacejackson:
Right now, there's not really a step between Baseplate running it's initialization on the Pyramid Request object and the service handling the request which makes it difficult to run code that you want to run before each request but after Baseplate has attached the client objects to the requests (say authenticating the request with a thrift service).

With this change, Baseplate will emit an event after it creates the ServerSpan for the request that services can add subscribers to in order to run code like this.

### DIFF
--- a/baseplate/integration/pyramid.py
+++ b/baseplate/integration/pyramid.py
@@ -17,6 +17,20 @@ An abbreviated example of it in use::
 
         return configurator.make_wsgi_app()
 
+Events
+------
+
+Baseplate will emit events at various stages of the request lifecycle that
+services can hook into.
+
+:py:class:`ServerSpanInitialized` - Fired after Baseplate initializes the
+ServerSpan for a new Request and before passing the Request to it's handler.
+Note, that Baseplate initializes the ServerSpan in response to a
+:py:class:`pyramid.events.ContextFound` event emitted by Pyramid so while we
+can guarantee what Baseplate has done when this event is emitted, we cannot
+guarantee that any other subscribers to :py:class:`pyramid.events.ContextFound`
+have been called or not.
+
 .. warning::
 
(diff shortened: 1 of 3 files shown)


baseplate/integration/pyramid.py
```diff
@@ -62,6 +62,18 @@ def baseplate_tween(request):
     return baseplate_tween
 
 
+class BaseplateEvent(object):
+
+    def __init__(self, request):
+        self.request = request
+
+
+class ServerSpanInitialized(BaseplateEvent):
```
spladug: could you add this to the pyramid integration docs with a note explaining when this is guaranteed to fire during request lifecycle please?

pacejackson:
:haircut_man:

pacejackson:
I wonder if it makes more sense to put the `Events` section after the `Warning` section in the docs?

screen shot 2017-12-01 at 2 29 25 pm

baseplate/integration/pyramid.py
```diff
@@ -62,6 +62,25 @@ def baseplate_tween(request):
...
+    """Event that Baseplate fires after creating the ServerSpan for a Request.
+
+    This event will be emitted before the Request is passed along to it's
```
spladug: its

pacejackson: Doh, I should have caught that:

https://youtu.be/yc2udEpyPpU?t=20

pacejackson:
FYI, it looks like this now:

screen shot 2017-12-01 at 3 32 21 pm

I like having the `Events` section after the `Configurator`, the only downside is I think it would be a little nicer if the `Events` section was a sub-section of the `Pyramid` section but there doesn't seem to be a way to do that without putting it in the module docstring which would place it before the `Configurator` docs.

PR #138, opened 2017-12-04
Add thrift build step to setup.py

ckwang8128:
In order to support Wheels builds of Baseplate, we need to
make sure that we compile Baseplate thrift definitions before
packaging. 

Building Baseplate strictly via `sdist` or other mechanisms
that rely on setup.py will lead to silent failures upon installation
for the baseplate.thrift module because there is no definition
in the setup.py build step for generating the included Baseplate
Thrift definitions in `baseplate/thrift/baseplate.thrift`. This adds
an override to build_py to generate Thrift files via the Makefile
and checks for a valid Thrift compiler. 

Note: I have very limited experience with properly extending setuptools stuff 
so if there's other patterns to follow here I'm happy to change things. I tested
this locally and got a working wheels build with these changes as well as the 
standard sdist that gets uploaded to PyPi. 

:eyeglasses: @spladug @pacejackson

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,9 @@
+import os
 import sys
 
+import setuptools.command.build_py
+import subprocess
+from distutils.spawn import find_executable
 from setuptools import setup, find_packages
 
 
@@ -40,6 +44,30 @@ extras_require = {
         "alabaster",
     ],
 }
+
+
+class BuildPyCommand(setuptools.command.build_py.build_py):
+    """Custom build command for packaging."""
+
+    def _make_thrift(self):
+        """Generate Baseplate Thrift definitions.
+        """
+        # Verify build dependencies are installed.
+        if not find_executable('make'):
+            print("'make' not found.")
+            exit(1)
+
+        if not find_executable('thrift1'):
+            print("Thrift compiler not found.")
+            exit(1)
+
+        make_cmd = ['/usr/bin/make', 'thrift']
+        subprocess.check_call(make_cmd)
+
+    def run(self):
+        self._make_thrift()
+        setuptools.command.build_py.build_py.run(self)
+
 
 setup(
     name="baseplate",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,6 +44,29 @@
...
 
+
+class BuildPyCommand(setuptools.command.build_py.build_py):
+    """Custom build command for packaging."""
+
+    def _make_thrift(self):
+        """Generate Baseplate Thrift definitions.
+        """
+        # Make sure thrift executable exists.
+        thrift_check_cmd = ['/usr/bin/which', 'thrift1']
```
spladug: why check for thrift1 but not make? any reason to not just run make and check status at the end without any up front checks?

ckwang8128: Ah, can and should do that as well! 

I just wanted to make the error message a bit more informative for each case; in the debian case you can define these dependencies so I wanted to try and echo that to an extent here.

setup.py
```diff
@@ -41,6 +44,41 @@
...
+        if not _executable_exists('thrift1'):
+            print("Thrift compiler not found.")
+            exit(1)
+
+        make_cmd = ['/usr/bin/make', 'thrift']
+        subprocess.check_call(make_cmd)
+
+    def run(self):
+        self._make_thrift()
+        setuptools.command.build_py.build_py.run(self)
```
pacejackson: Does this not support `super`?

ckwang8128: This was the example I followed from https://seasonofcode.com/posts/how-to-add-custom-build-steps-and-commands-to-setuppy.html (the very end). 

I don't believe this particular class supports super -> https://github.com/pypa/setuptools/blob/master/setuptools/command/build_py.py#L23 + https://github.com/python/cpython/blob/2.7/Lib/distutils/command/build_py.py#L16 

These don't appear to be `object`-based python classes.

ckwang8128:
:nail_care: added a check for `make` existence.

ckwang8128:
:haircut: added a check for `make` existence.

setup.py
```diff
@@ -41,6 +44,41 @@
     ],
 }
 
+
+def _executable_exists(executable_name):
```
spladug: :no_bell: you can use `distutils.spawn.find_executable` for this.

setup.py
pacejackson: Not surprised :)

PR #140, opened 2017-12-11
Restructure EdgeContext/Authentication and add thrift_pool_from_config

spladug:
Bonus commit comes first: adding a helper for turning config -> thrift pools. This should remove some boilerplate from most services that talk to thrift servers.

The majority of this work however is in the edgecontext/authentication stuff. There were some bugs that prevented it from being used currently, so I took the opportunity to restructure it for simplicity's sake and more thoroughly integrate the two concepts since they were only separate due to timing of landing. There are breaking API changes here, some blatant and some more subtle, but we've only one service attempting to use this right now so it'll be worth it.

I tested this by setting up a trio of basic test services that talk to eachother and carry the information around. It worked on py2 and py3. The test suite is also updated, of course.


baseplate/core.py
```diff
@@ -132,305 +136,177 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
-        self._payload = None
-        self._valid = None
-        self.defined = self._token is not None
-
-    def _secret(self):
-        if not self._secret_store:
-            raise UndefinedSecretsException
-
-        return self._secret_store.get_simple("jwt/authentication/secret")
+class AuthenticationToken(object):
```
bsimpson63: this is a nice interface

baseplate/integration/pyramid.py
```diff
@@ -118,17 +111,13 @@ def _on_new_request(self, event):
...
-                    authentication_context=authentication_context,
-                )
-                request_context.attach_context(request)
+                edge_payload = request.headers.get("X-Edge-Request", None)
+                if self.edge_context_factory:
+                    edge_context = self.edge_context_factory.from_upstream(
+                        edge_payload)
+                    edge_context.attach_context(request)
+                else:
+                    request.raw_request_context = edge_payload
```
bsimpson63: what's happening here?

spladug: added a comment to try and explain, does that help? would it be better to structure this differently so i don't need the same branching in both integrations? or not worth the added abstraction?

bsimpson63: ok I get it. I think it's fine

tests/integration/pyramid_tests.py
```diff
@@ -22,11 +23,17 @@
...
-    from pyramid.request import Request
 except ImportError:
     raise unittest.SkipTest("pyramid/webtest is not installed")
 
-from .. import mock
+from .. import (
+    mock,
+    AUTH_TOKEN_PUBLIC_KEY,
+    AUTH_TOKEN_VALID,
+    SERIALIZED_EDGECONTEXT_WITH_EXPIRED_AUTH,
```
bsimpson63: this isn't used

spladug: fixed

tests/integration/pyramid_tests.py
```diff
@@ -22,11 +23,17 @@
...
+    SERIALIZED_EDGECONTEXT_WITH_NO_AUTH,
+    SERIALIZED_EDGECONTEXT_WITH_VALID_AUTH,
```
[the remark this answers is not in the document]
spladug: fixed

baseplate/core.py
```diff
@@ -132,305 +136,177 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
 
         :type: set(string)
-        :raises: :py:class:`UndefinedSecretsException` if the
-            :py:class:`SecretsStore` has not been bound to the context handling
-            class (means that the authentication payload could not be decrypted
-            and the user_id returned)
-        :raises: :py:class:`WithheldAuthenticationError` if there was no
-            authentication token defined for the current context
+        :raises: :py:class:`NoAuthenticationError` if there was no
+            authentication token or it was invalid
```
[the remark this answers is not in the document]
spladug: added a `has_role`. since roles are used for security, I felt it'd be safer to not do a plural version since the interface there gets kinda confusing.

baseplate/core.py
```diff
@@ -132,305 +136,177 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
-        :raises: :py:class:`WithheldAuthenticationError` if there was no
-            authentication token defined for the current context
+        :raises: :py:class:`NoAuthenticationError` if there was no
+            authentication token, it was invalid, or the subject is not an
+            account.
 
         """
-        return self.authentication_context.account_id
+        subject = self.authentication_token.subject
+        if not subject.startswith("t2_"):
```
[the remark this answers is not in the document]
spladug: fixed the anon auth case. as for the logged in case and NoAuthenticationError, I think this is probably clearer since the user should always check is_logged_in first?

spladug:
:haircut_man:

spladug:
b7260b663b739e6b5968d93d8342f7b8ac7cc4c0...41e8939a89f1b77d03c0c2076038eb80be2e0d2c

PR #143, opened 2018-01-16
Add a histogram interface to Baseplate's metrics support

ckwang8128:
This adds a simple histogram interface to Baseplate metrics.
Our implementation sends simple values as timer stats to the StatsD
backend, expecting the backend to be configured to aggregate
and store histograms, e.g. [Statsite](https://github.com/statsite/statsite/blob/master/README.md#architecture) (Search "histograms"). 

Histograms on top of timers is currently
in both etsy's StatsD implementation as well as Statsite, and anyone trying to use
this with a backend that doesn't have histograms configured will just be sending arbitrary timer values. 

Even though this is just a simple wrapper around timers, I think the 
Histogram abstraction is worth adding. 

:eyeglasses: @spladug

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -138,6 +138,17 @@ class BaseClient(object):
         gauge_name = _metric_join(self.namespace, name.encode("ascii"))
         return Gauge(self.transport, gauge_name)
 
+    def histogram(self, name):
+        """Return a Histogram with the given name.
+
+        :param str name: The name the histogram should have.
+
+        :rtype: :py:class:`Histogram`
+
+        """
+        histogram_name = _metric_join(self.namespace, name.encode("ascii"))
+        return Histogram(self.transport, histogram_name)
+
 
 class Client(BaseClient):
     """A client for statsd."""
@@ -260,6 +271,31 @@ class Counter(object):
 
         """
         self.increment(delta=-delta, sample_rate=sample_rate)
+
+
+class Histogram(object):
+    """A bucketed distribution of integer values across a specific range.
+
+    Records data value counts across a configurable integer value range
+    with configurable buckets of value precision within that range.
+
+    Configuration of each histogram is managed by the backend service,
+    not by this interface. This implementation also depends on histograms
+    being supported by the StatsD backend. Specifically, the StatsD
+    backend must support the :code:`h` key, e.g. :code:`metric_name:320|h`.
+    """
+    def __init__(self, transport, name):
+        self.transport = transport
+        self.name = name
+
+    def add_sample(self, value):
+        """Add a new value to the histogram.
+
+        This records a new value to the histogram; the bucket it goes in
+        is determined by the backend service configurations.
+        """
+        serialized = self.name + (":{:g}|h".format(value).encode())
+        self.transport.send(serialized)
 
 
 class Gauge(object):
--- a/docs/baseplate/metrics.rst
+++ b/docs/baseplate/metrics.rst
@@ -29,3 +29,7 @@ Metrics
 .. autoclass:: Gauge()
    :members:
    :undoc-members:
+
+.. autoclass:: Histogram()
+   :members:
+   :undoc-members:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -38,7 +38,7 @@ separately from the rest of Baseplate.
    baseplate.file_watcher: Read files from disk as they change <baseplate/file_watcher>
    baseplate.live_data: Tools for centralized data that updates near instantly <baseplate/live_data>
    baseplate.message_queue: POSIX IPC Message Queues <baseplate/message_queue>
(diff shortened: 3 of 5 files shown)


baseplate/metrics.py
```diff
@@ -261,6 +272,24 @@ def decrement(self, delta=1, sample_rate=1.0):
         """
         self.increment(delta=-delta, sample_rate=sample_rate)
 
+class Histogram(object):
+    """Record of arbitrary values for distribution over time.
```
spladug: record of?

baseplate/metrics.py
```diff
@@ -261,6 +272,24 @@ def decrement(self, delta=1, sample_rate=1.0):
...
+        self.transport = transport
+        self.name = name
+
+    def log(self, value):
+        """Log a new value to the histogram.
+
+        This records a new value to the histogram; the bucket it goes in
+        is determined by the backend service configurations.
+        """
+        serialized = self.name + (":{:g}|ms".format(value).encode())
```
spladug: is it worth using the `h` alias just for explicitness?

baseplate/metrics.py
ckwang8128: Yeah, I'm kinda stuck on this wording. i wanted it to be a noun as this is a class, but this does read clunkily. Any ideas?

baseplate/metrics.py
ckwang8128: ah yeah, sure

baseplate/metrics.py
ckwang8128: I was thinking just "Distribution of arbitrary values over time" maybe? In my mind this was technically a record of said distribution.

ckwang8128: I'll change to `A record`

spladug: I'm wondering if the naming involved can take some hints from other histogram projects, e.g. https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/AbstractHistogram.java

spladug: (I totally misread this originally as the verb "record" so "record of" just sounded really off)

baseplate/metrics.py
```diff
@@ -261,6 +272,24 @@ def decrement(self, delta=1, sample_rate=1.0):
...
+    """Record of arbitrary values for distribution over time.
+
+    This implementation depends on histograms being created for
+    timer values in the supporting StatsD backend.
+    """
+    def __init__(self, transport, name):
+        self.transport = transport
+        self.name = name
+
+    def log(self, value):
```
spladug: maybe `add_sample` ?

baseplate/metrics.py
spladug: This interface is sorta RPCey so maybe we just treat this histogram object as the histogram as exists in statsite. Then the description is along the lines of just describing what a histogram in general is.

baseplate/metrics.py
ckwang8128: 👍

baseplate/metrics.py
ckwang8128: I like that.

ckwang8128:
:haircut: 

Reworked the class description and updated namings.

baseplate/metrics.py
```diff
@@ -261,6 +272,28 @@ def decrement(self, delta=1, sample_rate=1.0):
...
+        self.transport = transport
+        self.name = name
+
+    def add_sample(self, value):
+        """Add a new value to the histogram.
+
+        This records a new value to the histogram; the bucket it goes in
+        is determined by the backend service configurations.
+        """
+        serialized = self.name + (":{:h}|ms".format(value).encode())
```
spladug: Er, I thought the `ms` would turn into `h` not the python side stuff?

spladug:
Love the new docstrings

baseplate/metrics.py
ckwang8128: Ah whoops!

ckwang8128:
:haircut: I dumbed

spladug:
Not sure why the linter's not catching this, but there should be two blanks between the histogram class on either side.

ckwang8128:
:haircut:

Updated the index documentation for metrics.py as well as the class documentation.

baseplate/metrics.py
```diff
@@ -262,6 +273,30 @@ def decrement(self, delta=1, sample_rate=1.0):
...
 
 
+class Histogram(object):
+    """A bucketed distribution of integer values across a specific range.
+
+    Records data value counts across a configurable integer value range
+    with configurable buckets of value precision within that range.
+
+    Configuration of each histogram is managed by the backend service,
+    not by this interface. This implementation depends on histograms
```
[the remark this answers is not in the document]
ckwang8128: Good point, I'll be more clear that the backend service must support the `h` type in some form and leave the rest to the user to uncover.

PR #148, opened 2018-01-31
Enable E501, set max length to 100

### LINKED ISSUE #68: Enable the E501 (line length) flake8 check
This'll require choosing a max line length (we're generally OK up to 100 I think) and dealing with the three violators of that rule (mostly long URLs in comments).

cshoe:
Remove ignore of E501 in flake8 config and set max line length to 100.

Closes #68

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -592,8 +592,8 @@ class Baseplate(object):
         if not isinstance(tracing_client, TracingClient):
             warn_deprecated("Passing tracing configuration directly to "
                             "configure_tracing is deprecated in favor of "
-                            "using baseplate.make_tracing_client and passing "
-                            "the constructed client on.")
+                            "using baseplate.tracing_client_from_config and "
+                            "passing the constructed client on.")
             tracing_client = make_client(tracing_client, *args, **kwargs)
 
         self.register(TraceBaseplateObserver(tracing_client))
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -775,7 +775,9 @@ class LocalSpan(Span):
             span.component_name = component_name
             context_copy.shadow_context_attr('trace', span)
         else:
-            span = Span(self.trace_id, self.id, span_id, self.sampled, self.flags, name, self.context)
+            span = Span(
+                self.trace_id, self.id, span_id, self.sampled,
+                self.flags, name, self.context)
         for observer in self.observers:
             observer.on_child_span_created(span)
         return span
--- a/baseplate/integration/pyramid.py
+++ b/baseplate/integration/pyramid.py
@@ -20,11 +20,11 @@ An abbreviated example of it in use::
 .. warning::
 
     Because of how Baseplate instruments Pyramid, you should not make an
-    `exception view`_ that handles the base :py:exc:`Exception` as this will
-    prevent Baseplate from seeing the unhandled error and reporting it
-    appropriately.
+    `exception view`_ prevent Baseplate from seeing the unhandled error and
+    reporting it appropriately.
 
-    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html#exception-views  # noqa: E501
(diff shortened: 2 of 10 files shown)


baseplate/core.py
```diff
@@ -592,8 +592,8 @@ def configure_tracing(self, tracing_client, *args, **kwargs):
         if not isinstance(tracing_client, TracingClient):
             warn_deprecated("Passing tracing configuration directly to "
                             "configure_tracing is deprecated in favor of "
-                            "using baseplate.make_tracing_client and passing "
-                            "the constructed client on.")
+                            "using baseplate.tracing_client_from_config and "
```
spladug: Good change, but this is unrelated to the E501 thing. Can it be split out into its own commit please?

baseplate/integration/pyramid.py
```diff
@@ -24,7 +24,7 @@ def make_app(app_config):
     prevent Baseplate from seeing the unhandled error and reporting it
     appropriately.
 
-    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html#exception-views
+    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html#exception-views  # noqa: E501
```
spladug: does this show up in the docs once they're built?

baseplate/integration/pyramid.py
```diff
@@ -214,7 +214,7 @@ def pshell_setup(env):
         [pshell]
         setup = baseplate.integration.pyramid:pshell_setup
 
-    See http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html#extending-the-shell
+    See http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html#extending-the-shell  # noqa: E501
```
spladug: same question about docs

spladug:
Just to be clear: I just would prefer that in a separate commit. Same PR is totally fine!

cshoe:
Ah, ok. Thanks for getting a head of me.

spladug:
Huh, I like the re-request-review thing as a way to come back. Right now that doesn't do anything with harold though. For future reference you'll wanna use the haircut emoji. I might steal this idea though so we can have a fully native flow!

baseplate/integration/pyramid.py
```diff
@@ -20,11 +20,11 @@ def make_app(app_config):
...
 
     Because of how Baseplate instruments Pyramid, you should not make an
-    `exception view`_ that handles the base :py:exc:`Exception` as this will
-    prevent Baseplate from seeing the unhandled error and reporting it
-    appropriately.
+    `exception view`_ prevent Baseplate from seeing the unhandled error and
+    reporting it appropriately.
 
-    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/exceptions.html#exception-views
+    .. _exception view: https://docs.pylonsproject.org/projects/pyramid_cookbook/\
```
spladug: 🔕 Kinda a bummer we can't click the link in the source anymore :(

spladug:
🐟

spladug:
Thanks! Please squash into atomic commits when merging: https://github.com/reddit/styleguide/tree/master/general-style-guidelines#git-workflow

baseplate/integration/pyramid.py
cshoe: This was the only thing I could find that worked. There's no option in Sphinx to ignore instances of #noqa. I tried to find a different way of doing the notating the link but came up empty. I also looked for a way to tell flake8 to ignore it by adding something to a different line but nothing seemed to work.

spladug: :( does this suggestion of putting the # noqa after the docstring work? https://stackoverflow.com/questions/10739843/how-should-i-format-a-long-url-in-a-python-comment-and-still-be-pep8-compliant/39768096#39768096

cshoe: No, unfortunately.

PR #149, opened 2018-02-01
Add v2 bucketing events to experiments

Kaitaan:
This change updates the experiments framework to fire v2 bucketing
events when asking for the variant.


baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
                 inputs["user_id"] = user.id
                 inputs["user_roles"] = user.roles
+                inputs["logged_in"] = user.is_logged_in
             else:
                 inputs["user_id"] = user.loid
-            inputs["logged_in"] = user.is_logged_in
+                inputs["logged_in"] = False
+
+        #loid is numeric. Convert it to a base-36 string; Ideally, this will
+        #   be fixed before the caller sends it in, but this will catch
```
pacejackson: style nit: comments should always have one space after the `#`.

baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+
+        """ clean up user id. If not a baseplate user, we risk inconsistency."""
         if user:
             if user.is_logged_in:
                 inputs["user_id"] = user.id
                 inputs["user_roles"] = user.roles
+                inputs["logged_in"] = user.is_logged_in
             else:
                 inputs["user_id"] = user.loid
-            inputs["logged_in"] = user.is_logged_in
```
pacejackson: Why this change?

baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+                inputs["logged_in"] = False
+
+        #loid is numeric. Convert it to a base-36 string; Ideally, this will
```
pacejackson: `user.loid` should already be in the fullname format: https://github.com/reddit/baseplate/blob/master/baseplate/core.py#L419

baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+        if isinstance(inputs.get("user_id"), int):
+            if (inputs.get("user_id") < 0):
+                logger.warning("unable to bucket user with invalid user_id < 0")
+                return None
+            inputs["user_id"] = to36(inputs["user_id"])
+
+        # if it's not already a fullname, let's make it one (and strip leading
+        #    zeroes while we're at it. And lowercase it to be safe.)
+        if inputs.get("user_id") and not inputs["user_id"].startswith("t2_"):
+            inputs["user_id"] = "t2_" + inputs["user_id"].lstrip("0").lower()
```
pacejackson: I don't think we want to fix this, I think we want to raise an error if someone gives us the wrong format if we want to be pedantic.

baseplate/experiments/__init__.py
```diff
@@ -172,54 +194,65 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
-        if version:
-            event.set_field("version", version)
+        event = self._event_constructor(event_context_fields)
 
         span_name = "{}.{}".format(self._context_name, "events")
         with self._span.make_child(span_name) as child_span:
             try:
-                self._event_queue.put(event)
+                serialized = self._event_queue.put(event)
+                print(serialized)
```
pacejackson: Looks like you have a leftover print statement

baseplate/experiments/__init__.py
```diff
@@ -23,23 +23,26 @@ class ExperimentsContextFactory(ContextFactory):
...
         :param str path: Path to the experiment config file.
         :param baseplate.events.EventQueue event_queue: Event queue used for
             logging bucketing events to the event pipeline.  You can set this
-            to None to disable bucketing events entirely.
+            to None to disable bucketing events entirely. This queue should be
+            configured to use baseplate.events.serialize_v2_event
         """
         self._filewatcher = FileWatcher(path, json.load)
         self._event_queue = event_queue
+        self._event_constructor = event_constructor
```
pacejackson: This pattern of passing the constructor in seems pretty inelegant to me especially since you have to set it up in the exact format that you expect in Baseplate.  

What do you think about instead passing making an `EventLogger` interface that just has one method `def log(self, event_name, **inputs)`.  You can make a "debug" logger here that will just log everything to `logging.debug` but you could also make one that fires v2 events and give that to the `Experiments` client.  Then instead of calling `self._log_bucketing_event` you get rid of `_log_bucketing_event` and just call `self._event_logger.log('bucket', **inputs)`.

baseplate/experiments/__init__.py
Kaitaan: can we be certain that user.is_logged_in is always populated and never null?

pacejackson: Yes https://github.com/reddit/baseplate/blob/master/baseplate/core.py#L262

baseplate/experiments/__init__.py
Kaitaan: only if it's passed in via a baseplate user object. the loid as it exists in the r2 context is numeric, and if someone passes it in via `**kwargs`, I want to make sure we've got it formatted correctly.

baseplate/experiments/__init__.py
Kaitaan: fair point.

baseplate/experiments/__init__.py
Kaitaan: gah. Will fix

baseplate/experiments/__init__.py
pacejackson: I don't really think that we want to try and fix that in baseplate though, the framework should just assume that the input you passed in is what you want to use.  I think if we tried to "fix" this, we would actually rebucket every logged out experiment since r2 always used the full `loid.id` value, not the fullname format, for bucketing.

baseplate/experiments/__init__.py
Kaitaan: gotcha. I think this actual change was a carryover from my various modifications that I didn't "correct" since the new change isn't broken. Will fix.

baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
         """
+
         experiment = self._get_experiment(name)
 
         if experiment is None:
             return None
 
         inputs = dict(kwargs)
+
+        """ clean up user id. If not a baseplate user, we risk inconsistency."""
```
spladug: style nit: if it's a comment just write it as a comment rather than a string literal

baseplate/experiments/__init__.py
spladug: Agreed. Baseplate's style is usually to yell at you loudly if you do something wrong rather than silently fix it. Since this would be a problem only in r2 (and only until we get the stuff wired up like a baseplate app there) we can just put any value-massaging code in r2 and leave baseplate strict and sassy.

baseplate/experiments/__init__.py
```diff
@@ -134,25 +138,43 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+                inputs["logged_in"] = user.is_logged_in
             else:
                 inputs["user_id"] = user.loid
-            inputs["logged_in"] = user.is_logged_in
+                inputs["logged_in"] = False
+
+        #loid is numeric. Convert it to a base-36 string; Ideally, this will
+        #   be fixed before the caller sends it in, but this will catch
+        if isinstance(inputs.get("user_id"), int):
+            if (inputs.get("user_id") < 0):
```
spladug: I don't see this kind of "warn and null out" pattern going on much in the rest of this code. Is that the sort of behaviour we want to use here?

nit: at this point we know that `inputs["user_id"]` is an int so the `.get()` guard on dictionary access is unnecessary.

style nit: no parens around the condition in python

baseplate/experiments/__init__.py
```diff
@@ -172,54 +194,65 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
 
-    def _log_bucketing_event(self, experiment, variant, user,
-                             extra_event_fields):
-        if not self._event_queue:
-            return
-
-        extra_event_fields = extra_event_fields or {}
+    def _log_bucketing_event(self, experiment, variant, user, extra_event_fields):
+
+        """ Logs a v2 bucketing event.
```
spladug: style nit: [PEP-257](https://github.com/reddit/styleguide/blob/master/python/pep-0257.rst) says `The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".`

baseplate/experiments/__init__.py
```diff
@@ -172,54 +194,65 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+
+            :param dict experiment: An experiment config dict.
```
spladug: style nit: this all looks oddly indented. we usually just align with the open quote.

baseplate/experiments/__init__.py
```diff
@@ -172,54 +194,65 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+                - cookie_created_timestamp - epoch ms
+                - created_timestamp - epoch ms
+            :param dict extra_event_fields: addtional fields to be added. For
+                now, this expects and will process only the set of
+                user-related fields specified in the user parameter.
+        """
 
-        event = Event("bucketing_events", "bucket")
+        if not self._event_queue or not self._event_constructor:
+            logger.warning("missing event queue or bucketing constructor")
```
spladug: `warning` is a pretty high criticality of log message. is this something we expect to be a standard case e.g. in dev (and so should be lower level)? or should it be even stronger and raise an exception?

tests/unit/experiments/experiment_tests.py
```diff
@@ -22,11 +22,15 @@
 THIRTY_DAYS = timedelta(days=30).total_seconds()
 
 
+def event_constructor_for_use_in_tests(input_dict=None):
+    return input_dict
+
```
spladug: style nit: two blanks between top-level stuff

Kaitaan:
Let's try this again... the experiment constructor will now take a logger, the code for which will live elsewhere (first pass will be in r2, but ultimately, a v2 logger will live in the data-schemas repo) and will default to a debug logger.

:haircut_man:

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
...
         return Experiments(
             config_watcher=self._filewatcher,
-            event_queue=self._event_queue,
             server_span=server_span,
             context_name=name,
+            event_logger=self._event_logger
         )
 
 
+class V2DebugLogger(EventLogger):
```
pacejackson: I would just call this `DebugEventLogger` since it doesn't really care about v2 events.

baseplate/events/queue.py
```diff
@@ -170,6 +170,11 @@ def serialize_v2_event(event):
     return Serializer.serialize(_V2_PROTOCOL_FACTORY, event)
 
 
+class EventLogger(object):
+    def log(self, **kwargs):
+        pass
```
pacejackson: Maybe this should be called `EventLoggerInterface` and `raise NotImplementedError` rather than just `pass`ing.

baseplate/experiments/__init__.py
pacejackson: Also, I think it makes sense to put this in `events.py` rather than here since it's not specific to experiments.

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
...
+    def log(self, **kwargs):
+        inputs = dict(kwargs)
```
pacejackson: You don't really need to do this unless you need a copy for some reason, kwargs is already a dict.

baseplate/experiments/__init__.py
```diff
@@ -134,25 +143,28 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
         :return: Variant name if a variant is active, None otherwise.
         """
+
+        inputs = dict(kwargs)
         experiment = self._get_experiment(name)
 
         if experiment is None:
             return None
 
-        inputs = dict(kwargs)
```
pacejackson: I would leave this here, if the experiment doesn't exist, then you can skip copying the dict.

baseplate/experiments/__init__.py
```diff
@@ -171,66 +183,19 @@ def variant(self, name, user=None, bucketing_event_override=None,
 
         if do_log:
             assert bucketing_id
-            self._log_bucketing_event(experiment, variant, user,
-                                      extra_event_fields)
+            if self._event_logger:
```
pacejackson: You don't need this `if` anymore since you set it to the Debug logger if nothing is defined.

baseplate/events/queue.py
spladug: `raise NotImplementedError` would be nice for this abstract method

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
     This factory will attach a new :py:class:`baseplate.experiments.Experiments`
     to an attribute on the :term:`context object`.
     """
-    def __init__(self, path, event_queue):
+    def __init__(self, path, event_logger=None):
         """ExperimentContextFactory constructor
 
         :param str path: Path to the experiment config file.
         :param baseplate.events.EventQueue event_queue: Event queue used for
```
spladug: This param's docs need to be updated now.

baseplate/experiments/__init__.py
spladug: Is "V2" necessary in this name?

baseplate/events/queue.py
pacejackson: I _think_ we want `event_name` or just `name` to be an argument so you don't have to define a separate logger object for each event you want to send.

baseplate/events/queue.py
```diff
@@ -170,6 +170,11 @@ def serialize_v2_event(event):
     return Serializer.serialize(_V2_PROTOCOL_FACTORY, event)
 
 
+class EventLogger(object):
+    def log(self, **kwargs):
```
spladug: The interface could have explicit parameters right? How those parameters get used is up to implementation but what baseplate's gonna pass in is gonna stay constant? That seems more friendly to future development.

baseplate/experiments/__init__.py
```diff
@@ -171,66 +183,19 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
-            self._log_bucketing_event(experiment, variant, user,
-                                      extra_event_fields)
+            if self._event_logger:
+                self._event_logger.log(
+                    experiment=experiment,
+                    variant=variant,
+                    user_id=inputs.get('user_id'),
+                    logged_in=inputs.get('logged_in'),
+                    cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+                    app_name='r2'
```
pacejackson: We shouldn't have this value hardcoded in baseplate, we should pass this value to the `ContextFactory` which will pass it to the `Experiments` client.

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
...
-            to None to disable bucketing events entirely.
+            to None to disable bucketing events entirely. This queue should be
+            configured to use baseplate.events.serialize_v2_event
         """
         self._filewatcher = FileWatcher(path, json.load)
-        self._event_queue = event_queue
+        self._event_logger = event_logger
 
-    def make_object_for_context(self, name, server_span):
+    def make_object_for_context(self, name, server_span, event_logger=None):
```
spladug: I don't think that extra parameter's needed / used here.

tests/unit/experiments/experiment_tests.py
```diff
@@ -9,7 +9,7 @@
 from datetime import timedelta
 
 from baseplate.core import ServerSpan, User, AuthenticationToken
-from baseplate.events import EventQueue
+from baseplate.events import EventQueue, EventLogger
```
pacejackson: I don't think you need `EventQueue` anymore.

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
...
         return Experiments(
             config_watcher=self._filewatcher,
-            event_queue=self._event_queue,
             server_span=server_span,
             context_name=name,
+            event_logger=self._event_logger
```
spladug: nit: trailing comma please for future diff friendliness!

baseplate/experiments/__init__.py
```diff
@@ -23,26 +22,33 @@ class ExperimentsContextFactory(ContextFactory):
...
         )
 
 
+class V2DebugLogger(EventLogger):
+    def log(self, **kwargs):
+        inputs = dict(kwargs)
+        logger.debug("logger output: {}".format(inputs))
```
spladug: maybe "Would send event: "? that'd match https://github.com/reddit/baseplate/blob/master/baseplate/metrics.py#L73

baseplate/events/queue.py
spladug: I think it's fine for the ABC to just be EventLogger; it's functionally an interface but we don't put that word in the names in other places in baseplate.

baseplate/events/queue.py
pacejackson: There are probably some explicit parameters but different events don't have all the same inputs so it needs the `**kwargs`.

spladug: Sure, but this interface is just used for the experiments system to send events right? Maybe we should call the ABC `ExperimentEventLogger`?

pacejackson: Sure right now it is, but I think it's a really useful interface that we could use in other places as well and it seems like a waste to restrict it to experiments.

spladug: OK, but maybe we should worry about those cases only once they come up? Feels a bummer to overgeneralize without information.

baseplate/experiments/__init__.py
Kaitaan: agreed, but I'm going to disagree about putting in the ContextFactory. I think it makes sense to pass into the variant method, since it can be dynamic even in the context of a single Experiment instance.

I'm thinking if relevance, for instance, was trying out a new algo, they could pass along the app_name a user is accessing from for each variant call.

Make sense?

pacejackson: I guess I'm not even sure what `app_name` even means?  Would it really be that dynamic?  Where would the developer even get it from?

Kaitaan: the app_name is meant to pass in the context around what service/app is checking experimental variants. Might be a better name for it though.

The more I think about it, the more I think it doesn't belong here at all, and should be the responsibility of the EventLogger to populate, and/or as an optional field pulled from the extra_event_fields.

tests/unit/experiments/experiment_tests.py
Kaitaan: fixed

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/experiments/__init__.py
Kaitaan: fixed. Didn't realize it's a deep copy and seemed cleaner up above.

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/experiments/__init__.py
Kaitaan: whoops; missed it.

Also, thanks for explaining why. I knew we liked trailing commas, but I never really knew why. That makes so much sense.

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/events/queue.py
Kaitaan: I'm with Andrew on this; the code incoming today supports a very general logger that can take in "arbitrary" parameters (quoted, because it still relies on *some* formatting) and construct and send v2 events.

baseplate/experiments/__init__.py
Kaitaan: Nope. fixed

baseplate/experiments/__init__.py
Kaitaan: fixed

baseplate/events/queue.py
Kaitaan: fixed

Kaitaan: We won't need to define a separate logger for each event type; we can create a logger which is relatively dynamic, accepting "arbitrary" params to construct an event. That also doesn't preclude us from creating a logger which *expects* an event_type parameter to be passed along.
I think you'll see what I mean in an upcoming PR (feel free to block this one in the interim if you're really concerned)

baseplate/experiments/__init__.py
pacejackson: Slight correction, it's not _technically_ a deep copy since it doesn't recursively other containers within the dict:

```python
In [1]: x = {1: 'test', 2: ['a', 'b']}

In [2]: y = dict(x)

In [3]: x is y
Out[3]: False

In [4]: y[2] is x[2]
Out[4]: True

In [5]: from copy import deepcopy

In [6]: z = deepcopy(x)

In [7]: z[2] is x[2]
Out[7]: False

In [16]: y[2].append('c')

In [17]: y
Out[17]: {1: 'test', 2: ['a', 'b', 'c']}

In [18]: x
Out[18]: {1: 'test', 2: ['a', 'b', 'c']}

In [19]: z
Out[19]: {1: 'test', 2: ['a', 'b']}
```

Kaitaan:
:haircut_man: 

I'll sync up with the growth team per @pacejackson's comment

baseplate/experiments/__init__.py
```diff
@@ -183,15 +177,14 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
-                    cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
-                    app_name='r2'
-                )
+            self._event_logger.log(
+                experiment=experiment,
+                variant=variant,
+                user_id=inputs.get('user_id'),
+                logged_in=inputs.get('logged_in'),
+                cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+                app_name=inputs.get('app_name'),
```
pacejackson: I think you also want to pass in `extra_event_fields` in some way, unless we want to get rid of that?

baseplate/events/queue.py
```diff
@@ -170,6 +171,19 @@ def serialize_v2_event(event):
...
+        raise NotImplementedError
+
+
+class DebugLogger(EventLogger):
```
spladug: might be nice to call this `DebugEventLogger`, sorry for missing that in last pass

baseplate/events/queue.py
spladug: feels a bit magical to me, but ok

baseplate/experiments/__init__.py
```diff
@@ -171,66 +177,18 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
             assert bucketing_id
-            self._log_bucketing_event(experiment, variant, user,
-                                      extra_event_fields)
+            self._event_logger.log(
+                experiment=experiment,
+                variant=variant,
+                user_id=inputs.get('user_id'),
+                logged_in=inputs.get('logged_in'),
+                cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+                app_name=inputs.get('app_name'),
```
spladug: i like the idea of this just being something the logger knows since it'd be created once at startup. it's a bit weird that we could vary the app_name across experiment calls.

Kaitaan: My concern is around services that will serve multiple apps. If relevance decides to run an experiment that changes how listings are generated, that will apply to ios, android, web, redesign. The results of those experiments may differ significantly from app to the other, and analysis should be able to bear that out.

If we created this once at startup, we'd need to create a different Experiment for each app (because we'd need a different logger for each), or else use the service as the "app_name".

I'm willing to accept that the logger is meant to know it for now (and thus remove it from the call here), but I'm trying to think ahead to how this could be handled down the line.  Either way, it's an easy addition later on if it's needed.

baseplate/events/queue.py
Kaitaan: Makes sense. Will fix.

baseplate/experiments/__init__.py
spladug: Ah, I misunderstood the granularity of app_name. Sounds good.

PR #151, opened 2018-02-02
Add baseplate-shell command

bradengroom:


baseplate/server/__init__.py
```diff
@@ -197,3 +202,163 @@ def load_and_run_script():
...
+    parser.add_argument("config_file", type=argparse.FileType("r"),
+        help="path to a configuration file")
+
+    args = parser.parse_args(sys.argv[1:])
+    config = read_config(args.config_file, server_name=None, app_name=args.app_name)
+    configure_logging(config, args.debug)
+
+    locs = dict(__name__="pylons-admin")
+
+    is_http_service = pylons.config['pylons.package'] is not None
```
bradengroom: Is there a better way to do this?

baseplate/server/__init__.py
```diff
@@ -197,3 +202,163 @@ def load_and_run_script():
...
+        banner = _configure_thrift_service_shell(config, locs)
+
+    _run_shell(sys.argv[1:], banner, locs)
+
+
+def _configure_thrift_service_shell(config, locs):
+    app = make_app(config.app)
+    locs['app'] = app
+
+    baseplate = app._event_handler.baseplate
```
bradengroom: Is there a better way to do this?

baseplate/server/__init__.py
spladug: The only reddit service using pylons is r2 and that has `reddit-shell` already. New HTTP services use Pyramid which has `pshell`. We really only need a custom shell for Thrift services.

spladug:
Wooo! Thanks for taking a crack at this. I think we don't need the HTTP stuff at all (see comment above) which simplifies this a lot.

bradengroom:
Closing in favor of new PR https://github.com/reddit/baseplate/pull/161

PR #154, opened 2018-02-03
Coalesce counters in metrics batchs

### LINKED ISSUE #146: Coalesce counters in metrics batches
If we increment the same counter a number of times in the same request (e.g. success counters for a client call) we'll currently append a bunch of +1 lines to the metrics batch. It'd be much better to turn that into a single +N line to save space in metrics packets.

cshoe:
* Introduced ``BatchCounter`` class. The main difference between it and the
``Counter`` is that the transport ``send`` method is not called
everytime increment is called. The new class was introduced to maintain
the current ``counter`` API in the ``Client`` implementation.
* The ``Batch`` class now stores a reference to created ``BatchCounter``
instances. These are looked up by counter name and reused when possible.
* The ``send`` method is called on all created ``BatchCounter``
instances during ``Batch.flush``.

Fixes #146

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -51,6 +51,8 @@ import logging
 import socket
 import time
 
+from collections import defaultdict
+
 
 logger = logging.getLogger(__name__)
 
@@ -184,6 +186,7 @@ class Batch(BaseClient):
     def __init__(self, transport, namespace):
         self.transport = BufferedTransport(transport)
         self.namespace = namespace
+        self.counters = {}
 
     def __enter__(self):
         return self
@@ -193,7 +196,27 @@ class Batch(BaseClient):
 
     def flush(self):
         """Immediately send the batched metrics."""
+        for name, counter in self.counters.items():
+            counter.send()
         self.transport.flush()
+
+    def counter(self, name):
+        """Return a BatchCounter with the given name.
+
+        The sample rate is currently up to your application to enforce.
+
+        :param str name: The name the counter should have.
+
+        :rtype: :py:class:`Counter`
+
+        """
+        counter_name = _metric_join(self.namespace, name.encode("ascii"))
+        batch_counter = self.counters.get(counter_name)
+        if batch_counter is None:
+            batch_counter = BatchCounter(self.transport, counter_name)
+            self.counters[counter_name] = batch_counter
+
+        return batch_counter
 
 
 class Timer(object):
@@ -256,16 +279,7 @@ class Counter(object):
         :param float sample_rate: What rate this counter is sampled at. [0-1].
 
         """
-        parts = [
-            self.name + (":{:g}".format(delta).encode()),
-            b"c",
-        ]
-
-        if sample_rate and sample_rate != 1.0:
-            parts.append("@{:g}".format(sample_rate).encode())
-
-        serialized = b"|".join(parts)
-        self.transport.send(serialized)
+        self.send(delta, sample_rate)
 
     def decrement(self, delta=1, sample_rate=1.0):
         """Decrement the counter.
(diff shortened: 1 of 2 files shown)


baseplate/metrics.py
```diff
@@ -193,8 +194,31 @@ def __exit__(self, exc_type, value, traceback):
...
     def flush(self):
         """Immediately send the batched metrics."""
+        for name, counter in self.counters.items():
+            counter.send()
         self.transport.flush()
 
+    def counter(self, name):
+        """Return a BatchCounter with the given name.
+
+        BatchCounter instances are stored in the :term:`counters` attribute.
```
spladug: This is an internal implementation detail and probably shouldn't be exposed in the documentation.

baseplate/metrics.py
```diff
@@ -255,18 +279,71 @@ def increment(self, delta=1, sample_rate=1.0):
...
+        self._sample_rate = 1.0
+
+    def increment(self, delta=1, sample_rate=1.0):
+        """Increment the counter.
+
+        :param float delta: The amount to increase the counter by.
+        :param float sample_rate: What rate this counter is sampled at. [0-1].
+
+        """
+        if sample_rate != self._sample_rate:
```
spladug: Is this actually safe? This sets the sample rate to the most recently seen one without scaling previous values. So if I sent `increment(1, sample_rate=0.1)` and `increment(1, sample_rate=1)` it'd just send 2 @ 100% when it should really indicate 11 @ 100%. 

Should we internally normalize everything to a sample rate of 1? (That might mess up the derived metrics statsd makes?) Could we have multiple batches for each seen sample rate? Just raise an exception if the sample rate doesn't match the first seen?

baseplate/metrics.py
```diff
@@ -193,8 +194,31 @@ def __exit__(self, exc_type, value, traceback):
...
+        """Return a BatchCounter with the given name.
+
+        BatchCounter instances are stored in the :term:`counters` attribute.
+        They are reused on subsequent calls.
+
+        The sample rate is currently up to your application to enforce.
+
+        :param str name: The name the counter should have.
+
+        :rtype: :py:class:`BatchCounter`
```
spladug: This subclass isn't publicly documented (and that's good, I think, since it meets the LSP and is just an internal detail) and so the link doesn't work in docs. Maybe just say `:py:class:`Counter`` here? It's not a lie since a BatchCounter is a Counter! :)

baseplate/metrics.py
cshoe: Ooof, yeah.

> Should we internally normalize everything to a sample rate of 1?

This feels disingenuous to users in addition to your comment on the metrics.

> Just raise an exception if the sample rate doesn't match the first seen?

This feels brittle.

> Could we have multiple batches for each seen sample rate?

This seems to make the most sense. Multiple sample rates will mean more data sent to the stats backend but we save when we can *and* stay true to the backend API.

I think it's straight forward enough for ``BatchCounter`` to keep a map of sample rates to deltas and iterate through those in the ``send`` method.

``increment`` and ``send`` would look something like:

```python

def increment(self, delta=1, sample_rate=1.0):
    packet = self.packets.get(sample_rate)  # I think floats can be used as keys?
    if packet is None:
        self.packets[sample_rate] = delta
    else:
        self.packets[sample_rate] = self.packets[sample_rate] + delta

def send(self):
    for sample_rate, delta in self.packet.items()
        super(BatchCounter, self).send(delta, sample_rate)

```

Thoughts @spladug?

spladug: TBH in all our metrics usage here we've almost never touched sample rate so I'm not sure how much effort it's worth putting into it which is why I was suggesting just throwing an error and if it becomes a blocker down the line we can make it more complex and robust.

re: keying off sample rate, I always get a little worried using floats in a context like that since they're so fickle. Maybe it'd be better to serialize them to string like they'll be on the wire before using as a key? Maybe overly paranoid.

cshoe:
:haircut:

baseplate/metrics.py
```diff
@@ -255,18 +276,73 @@ def increment(self, delta=1, sample_rate=1.0):
...
+        self.packets = {}
+
+    def increment(self, delta=1, sample_rate=1.0):
+        """Increment the counter.
+
+        :param float delta: The amount to increase the counter by.
+        :param float sample_rate: What rate this counter is sampled at. [0-1].
+
+        """
+        stale_delta = self.packets.get(sample_rate)
```
spladug: a `collections.defaultdict(int)` would make this just `self.packets[sample_rate] += delta`

cshoe:
:haircut:

PR #156, opened 2018-02-06
EdgeRequestContextFactory: Add service property

bsimpson63:
For service tokens generated by authenticate_service.

Will be used for r2, messenger, others to communicate with thing service for non-request tasks.

👓 @spladug @pacejackson

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -237,6 +237,7 @@ _User = collections.namedtuple(
 _OAuthClient = collections.namedtuple(
     "_OAuthClient", ["authentication_token"])
 Session = collections.namedtuple("Session", ["id"])
+_Service = collections.namedtuple("_Service", ["authentication_token"])
 
 
 class User(_User):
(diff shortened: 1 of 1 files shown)


baseplate/core.py
```diff
@@ -506,6 +528,11 @@ def session(self):
         """:py:class:`~baseplate.core.Session` object for the current context"""
         return Session(id=self._t_request.session.id)
 
+    @cached_property
+    def service(self):
+        """:py:class:`~baseplate.core.Service` object for the current context"""
+        return Service(self.authentication_token)
+
```
bsimpson63: I'm planning on using this in thing service by first trying to get `request_context.user.id` and if that fails checking `request_context.service.name`. Does that sound reasonable?

baseplate/core.py
```diff
@@ -358,6 +359,27 @@ def event_fields(self):
...
+        """Authenticated Service name.
+
+        :type: name string or None if context authentication is invalid
+        :raises: :py:class:`NoAuthenticationError` if there was no
+            authentication token, it was invalid, or the subject is not a
+            servce.
+
+        """
+        subject = self.authentication_token.subject
+        if not (subject and subject.startswith("service/")):
```
pacejackson: Maybe set `"service/"` to a variable `service_prefix = "service/"`?

baseplate/core.py
spladug: Sounds OK to start out. If things get more complicated we might want to figure out how to turn the user context into application authorization more natively. I'd really love to hook into Pyramid's authorization stuff where possible.

bsimpson63:
@spladug should I merge this now, or wait until right before a new release?

spladug:
Now is fine!

PR #160, opened 2018-02-16
Add details to memcache tracing spans

cshoe:
This is an explicit > implicit approach. The wrapped methods of the
memcache pool are now fully redefined. This allows for the spans to be
tagged according to each method being called.

In the future, this should probably be moved "lower" into the client
class to capture just the interactions with the server.

### DIFF
--- a/baseplate/context/memcache/__init__.py
+++ b/baseplate/context/memcache/__init__.py
@@ -107,31 +107,154 @@ class MonitoredMemcacheConnection(PooledClient):
         self.server_span = server_span
         self.pooled_client = pooled_client
 
-    #pylint: disable=no-self-argument
-    def monitored(fn_name):
-        def proxy_with_instrumentation(self, *a, **kw):
-            trace_name = "{}.{}".format(self.context_name, fn_name)
-            with self.server_span.make_child(trace_name):
-                method = getattr(self.pooled_client, fn_name)
-                return method(*a, **kw)
-        return proxy_with_instrumentation
+    def close(self):
+        with self._make_span("close"):
+            return self.pooled_client.close()
 
-    close = monitored('close')
-    set = monitored('set')
-    set_many = monitored('set_many')
-    replace = monitored('replace')
-    append = monitored('append')
-    prepend = monitored('prepend')
-    cas = monitored('cas')
-    get = monitored('get')
-    get_many = monitored('get_many')
-    gets_many = monitored('gets_many')
-    delete = monitored('delete')
-    delete_many = monitored('delete_many')
-    add = monitored('add')
-    incr = monitored('incr')
-    decr = monitored('decr')
-    touch = monitored('touch')
-    stats = monitored('stats')
-    flush_all = monitored('flush_all')
-    quit = monitored('quit')
+    def set(self, key, value, expire=0, noreply=None):
+        with self._make_span("set") as span:
+            span.set_tag("key", key)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.set(
+                key, value, expire=expire, noreply=noreply)
+
+    def set_many(self, values, expire=0, noreply=None):
+        with self._make_span("set_many") as span:
+            span.set_tag("key_count", len(values))
+            span.set_tag("keys", make_keys_str(values.keys()))
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.set_many(
+                values, expire=expire, noreply=noreply)
+
+    def replace(self, key, value, expire=0, noreply=None):
+        with self._make_span("replace") as span:
+            span.set_tag("key", key)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.replace(
+                key, value, expire=expire, noreply=noreply)
+
+    def append(self, key, value, expire=0, noreply=None):
+        with self._make_span("append") as span:
+            span.set_tag("key", key)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.append(
+                key, value, expire=expire, noreply=noreply)
+
+    def prepend(self, key, value, expire=0, noreply=None):
+        with self._make_span("prepend") as span:
+            span.set_tag("key", key)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.prepend(
+                key, value, expire=expire, noreply=noreply)
+
+    def cas(self, key, value, cas, expire=0, noreply=None):
+        with self._make_span("cas") as span:
+            span.set_tag("key", key)
+            span.set_tag("cas", cas)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.cas(
+                key, value, cas, expire=expire, noreply=noreply)
+
+    def get(self, key, **kwargs):
+        with self._make_span("get") as span:
+            span.set_tag("key", key)
+            return self.pooled_client.get(key, **kwargs)
+
+    def get_many(self, keys):
+        with self._make_span("get_many") as span:
+            span.set_tag("key_count", len(keys))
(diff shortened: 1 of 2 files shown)


cshoe:
:glasses: @spladug

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,157 @@ def __init__(self, context_name, server_span, pooled_client):
...
+    def gets(self, key, **kwargs):
+        with self.server_span.make_child(self.context_name) as span:
+            span.set_tag("method", "gets")
+            span.set_tag("key", key)
+            return self.pooled_client.gets(key, **kwargs)
+
+    def gets_many(self, keys):
+        with self.server_span.make_child(self.context_name) as span:
+            span.set_tag("method", "gets_many")
+            span.set_tag("key_cnt", len(keys))
```
spladug: any reason not to just spell out "count"?

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,157 @@ def __init__(self, context_name, server_span, pooled_client):
...
+    def set_many(self, values, expire=0, noreply=None):
+        with self.server_span.make_child(self.context_name) as span:
+            span.set_tag("method", "set_many")
+            span.set_tag("key_cnt", len(values))
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.set_many(
+                values, expire=expire, noreply=noreply)
+
+    set_multi = set_many
```
spladug: Where'd this come from?

tests/integration/memcache_tests.py
```diff
@@ -41,7 +41,7 @@ def test_simple(self):
 
         server_span_observer = self.baseplate_observer.get_only_child()
         span_observer = server_span_observer.get_only_child()
-        self.assertEqual(span_observer.span.name, "memcache.get")
+        self.assertEqual(span_observer.span.name, "memcache")
```
spladug: Hm, why remove the method name from the span name?

 When viewing a trace in zipkin it's really useful to be able to see the basic details in the span name since that's what's visible in the tree of spans and you have to click through to an individual span to see its tags. 

(Ah, I see the discussion of this in the ticket. I think it'd be better to bring r2 up to spec with how baseplate names its spans rather than reduce baseplate to r2's level)

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,157 @@ def __init__(self, context_name, server_span, pooled_client):
...
+            span.set_tag("key", key)
+            return self.pooled_client.get(key, **kwargs)
+
+    def get_many(self, keys):
+        with self.server_span.make_child(self.context_name) as span:
+            span.set_tag("method", "get_many")
+            span.set_tag("key_cnt", len(keys))
+            return self.pooled_client.get_many(keys)
+
+    get_multi = get_many
```
spladug: Where'd this come from?

cshoe:
> I'm torn on the explicitness, it's also a little scary for chance of accidentally mismapping arguments. Maybe there should be more integration tests to ensure it all works?

I'll take another approach and see if I can implement something at the client level. There are good places for it in the ``_store_cmd`` and ``_fetch_cmd`` methods but I need to figure out how to get current context into a pooled client.

spladug:
I don't think this needed to be thrown out. Seems fine to move ahead with it with minor changes.

cshoe:
💇

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,147 @@ def __init__(self, context_name, server_span, pooled_client):
...
+            span.set_tag("value", value)
+            return self.pooled_client.replace(
+                key, value, expire=expire, noreply=noreply)
+
+    def append(self, key, value, expire=0, noreply=None):
+        with self._get_span("append") as span:
+            span.set_tag("key", key)
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            span.set_tag("value", value)
```
spladug: In the context of the meeting we had the other day, I think we should avoid sending `value`s to the tracing system as they're quite likely to have user-data in them. (And regardless may be large binary blobs)

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,147 @@ def __init__(self, context_name, server_span, pooled_client):
...
+                key, value, cas, expire=expire, noreply=noreply)
+
+    def get(self, key, **kwargs):
+        with self._get_span("get") as span:
+            span.set_tag("key", key)
+            return self.pooled_client.get(key, **kwargs)
+
+    def get_many(self, keys):
+        with self._get_span("get_many") as span:
+            span.set_tag("key_count", len(keys))
```
spladug: r2 is actually sending some key names which I believe has been helpful. I'm not sure if we should keep doing this in light of the same desire to avoid private info in tracing data, but it may be nice to have if we think it's safe enough. 

```python
def make_keys_str(keys):
    keys_str = ','.join((keys))
    if len(keys_str) > 100:
        return keys_str[:100] + "..."
    else:
        return keys_str
```

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,147 @@ def __init__(self, context_name, server_span, pooled_client):
...
+        with self._get_span("flush_all") as span:
+            span.set_tag("delay", delay)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.flush_all(delay=delay, noreply=noreply)
+
+    def quit(self):
+        with self._get_span("quit"):
+            return self.pooled_client.quit()
+
+    def _get_span(self, method_name):
```
spladug: nit: this feels more like `_make_span` than `_get_span` since we're creating a new one each time

cshoe:
💇

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,153 @@ def __init__(self, context_name, server_span, pooled_client):
...
+    def delete_many(self, keys, noreply=None):
+        with self._make_span("delete_many") as span:
+            span.set_tag("key_count", len(keys))
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.delete_many(keys, noreply=noreply)
+
+    def add(self, key, value, expire=0, noreply=None):
+        with self._make_span("add") as span:
+            span.set_tag("key", key)
+            span.set_tag("value", value)
```
spladug: looks like there's still `value` in a few places

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,153 @@ def __init__(self, context_name, server_span, pooled_client):
...
+            span.set_tag("expire", expire)
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.add(key, value, expire=expire, noreply=noreply)
+
+    def incr(self, key, value, noreply=False):
+        with self._make_span("incr") as span:
+            span.set_tag("key", key)
+            span.set_tag("value", value)
```
spladug: including here

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,153 @@ def __init__(self, context_name, server_span, pooled_client):
...
+            span.set_tag("noreply", noreply)
+            return self.pooled_client.incr(key, value, noreply=noreply)
+
+    def decr(self, key, value, noreply=False):
+        with self._make_span("decr") as span:
+            span.set_tag("key", key)
+            span.set_tag("value", value)
```
spladug: and here

baseplate/context/memcache/__init__.py
```diff
@@ -107,31 +107,153 @@ def __init__(self, context_name, server_span, pooled_client):
...
+        that corresponds to the current context name and called method.
+
+        """
+        trace_name = "{}.{}".format(self.context_name, method_name)
+        span = self.server_span.make_child(trace_name)
+        span.set_tag("method", method_name)
+        return span
+
+
+def make_keys_str(keys):
```
spladug: i don't see this function getting used anywhere? are the `key_cnt` tags supposed to be updated to use it?

cshoe:
💇

PR #161, opened 2018-02-20
Add command for opening a shell for Thrift services

bradengroom:
This adds a `tshell` command which will provide a way for Thrift services to open an interactive shell similar to `pshell`:

```
$ tshell example.ini
Baseplate Interactive Shell
Python 2.7.6 (default, Nov 23 2017, 15:49:48) 
[GCC 4.8.4]

Available Objects:

  app          This project's app instance
  context      The context for this shell instance's span
>>> 
```

Services may expose additional variables to the shell environment by providing a setup function in their `.ini` file.

config file:
```
...
[tshell]
setup = my_service:tshell_setup
```

setup function:
```
def tshell_setup(env, env_banner):
    from my_service import models
    env['models'] = models
    env_banner['models'] = 'Models module'
```

```
$ tshell example.ini
Baseplate Interactive Shell
Python 2.7.6 (default, Nov 23 2017, 15:49:48) 
[GCC 4.8.4]

Available Objects:

  app          This project's app instance
  context      The context for this shell instance's span
  models       Models module
>>> 
```


baseplate/server/__init__.py
```diff
@@ -197,3 +202,69 @@ def load_and_run_script():
...
+    parser.add_argument("--debug", action="store_true", default=False,
+        help="enable extra-verbose debug logging")
+    parser.add_argument("--app-name", default="main", metavar="NAME",
+        help="name of app to load from config_file (default: main)")
+    parser.add_argument("config_file", type=argparse.FileType("r"),
+        help="path to a configuration file")
+
+    args = parser.parse_args(sys.argv[1:])
+    config = read_config(args.config_file, server_name=None, app_name=args.app_name)
+    configure_logging(config, args.debug)
```
bradengroom: Should I delete this?

setup.py
```diff
@@ -89,6 +89,7 @@ def run(self):
     scripts=[
         "bin/baseplate-serve{:d}".format(sys.version_info.major),
         "bin/baseplate-script{:d}".format(sys.version_info.major),
+        "bin/tshell",
```
bradengroom: Do I need a `tshell2` and `tshell3` similar to the serve and script commands?

bradengroom:
@bsimpson63

bradengroom:
👓 @bsimpson63

baseplate/server/__init__.py
```diff
@@ -66,12 +67,16 @@ def read_config(config_file, server_name, app_name):
                      if server_name else None)
     app_config = dict(parser.items("app:" + app_name))
     has_logging_config = parser.has_section("loggers")
+    tshell_config = dict()
```
spladug: `None` might be a little more obvious here.

baseplate/server/__init__.py
spladug: We probably don't want fileConfig taking over, so it might be best to just manually set up a `logging.basicConfig(level=logging.INFO)` with a simple formatter or something? Not sure how much logging we'd want to see while trying to interact with the REPL.

baseplate/server/__init__.py
```diff
@@ -197,3 +202,69 @@ def load_and_run_script():
...
+    # generate banner text
+    banner = "Available Objects:\n"
+    for var in sorted(env_banner.keys()):
+        banner += '\n  %-12s %s' % (var, env_banner[var])
+
+    try:
+        # try to use IPython if possible
+        try:
+            try:
+                # 1.0 <= ipython
```
spladug: Would you be OK dropping the old IPython support? 

> IPython 1.0 was released on August 8, 2013 (http://ipython.org/news.html)

In general we should probably be writing the minimal version of what we need here rather than just copying pylons; they have many more cases (and much longer history) to deal with.

baseplate/server/__init__.py
```diff
@@ -197,3 +202,69 @@ def load_and_run_script():
...
+            shell = InteractiveShellEmbed(banner2=banner)
+        except ImportError:
+            # ipython < 0.11
+            from IPython.Shell import IPShellEmbed
+            shell = IPShellEmbed(argv=args)
+            shell.set_banner(shell.IP.BANNER + '\n\n' + banner)
+        shell(local_ns=env, global_ns={})
+    except ImportError:
+        import code
+        py_prefix = sys.platform.startswith('java') and 'J' or 'P'
```
spladug: haha let's drop this too. we're never gonna run in Jython.

setup.py
spladug: Yeah, I think that makes most sense for consistency.

Should this be `baseplate-` prefixed? It's almost a generic thrift shell but it does tie into baseplate's span stuff (as it should) so it's definitely custom.

bradengroom: > Should this be `baseplate-` prefixed?

Sounds reasonable. It also doesn't have to be called `tshell` if you have some other name that you would prefer.

baseplate/server/__init__.py
bradengroom: Sure thing. Yeah, this was copied from r2 which was copied from pylons.

bradengroom:
💇

bradengroom:
> One request: can you make it clearer in the argparse banner and the sphinx docs that tshell is just for thrift apps?

https://github.com/reddit/baseplate/pull/161/commits/8dabc5015560c97641bd88e389a8e13c397225c6

PR #162, opened 2018-02-21
Add baseplate.ratelimit

bsimpson63:
Upstreaming from [sendbird service](https://github.com/reddit/reddit-service-sendbird/blob/master/sendbird/rate_limit.py) and extended to support memcached.

### DIFF
--- a/baseplate/ratelimit.py
+++ b/baseplate/ratelimit.py
@@ -1,0 +1,114 @@
+"""Simple bucketed rate limits in memcached or redis.
+
+Break up time into fixed buckets and allow a maximum number of counts per
+buckets.
+
+Buckets are formed by dividing the 86400 seconds of the day into equal length
+intervals.
+
+Example:
+
+If an interval length of 60 seconds is used there are 1440 buckets, one for
+each minute. When the minute ticks over (e.g. from 12:31:59 to 12:32:00) the
+count is reset because we use a new bucket.
+
+"""
+
+from __future__ import division
+
+import datetime
+import pytz
+
+from pymemcache.client.base import PooledClient
+from redis import StrictRedis
+
+
+class WrappedMemcache(object):
+    def __init__(self, cache):
+        self.cache = cache
+
+    def incr_and_get(self, key, ttl):
+        self.cache.add(key, 0, expire=ttl)
+        count = self.cache.incr(key, 1)
+        return count
+
+
+class WrappedRedis(object):
+    def __init__(self, cache):
+        self.cache = cache
+
+    def incr_and_get(self, key, ttl):
+        with self.cache.pipeline('incr') as pipe:
+            pipe.incr(key, 1)
+            pipe.expire(key, time=ttl)
+            responses = pipe.execute()
+
+        # extract the first response from the pipeline, which is the value of
+        # the key that was incr'd
+        count = responses[0]
+        return count
+
+
+def get_current_interval(interval_seconds):
+    """Convert the current time to an interval.
+
+    The 86400 seconds in a day are broken into into buckets of length
+    ``interval_seconds``. The current interval is then the bucket that the
+    current time falls into.
+
+    """
+    assert interval_seconds <= 86400, 'maximum interval is 1 day'
+
+    now = datetime.datetime.now(pytz.UTC)
+    seconds = (now.hour * 3600) + (now.minute * 60) + now.second
+
+    # use floor division to get the current time's bucket
+    interval = seconds // interval_seconds
+    return interval
+
+
+def _incr_and_get(wrapped_cache, key, interval_seconds=60, cache_prefix=''):
+    interval = get_current_interval(interval_seconds)
+
+    if cache_prefix:
+        cache_key = '%s%s:%s' % (cache_prefix, interval, key)
+    else:
+        cache_key = '%s:%s' % (interval, key)
+
+    # set the key to expire in 2 intervals (1 interval would be enough but
+    # let's be a bit more careful)
+    ttl = 2 * interval_seconds
+    count = wrapped_cache.incr_and_get(cache_key, ttl)
+    return count
+
+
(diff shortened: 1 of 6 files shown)


docs/baseplate/ratelimit.rst
```diff
@@ -0,0 +1,22 @@
+``baseplate.ratelimit``
+=======================
+
+.. automodule:: baseplate.ratelimit
+
+Assuming your project uses MemcacheContextFactory to attach a pymemcache client
+to the context object (or the request object for HTTP projects), you can
```
pacejackson: Should this description also mention Redis  since you can use that here as well?

tests/unit/ratelimit_tests.py
```diff
@@ -0,0 +1,138 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import datetime
+import unittest
+
+from baseplate import ratelimit
+from pymemcache.client.base import PooledClient
+import pytz
```
pacejackson: style nit: I think this should be before any of the `from baseplate import ratelimit`

baseplate/ratelimit.py
```diff
@@ -0,0 +1,114 @@
...
+    :param cache:
+        pymemcache client instance
+        (:py:class:`pymemcache.client.base.PooledClient`) or
+        redis client instance (:py:class:`redis.StrictRedis`).
+        Typically in baseplate this is an instance of
+        :py:class:`baseplate.context.memcache.MonitoredMemcacheConnection` or
+        :py:class:`baseplate.context.redis.MonitoredRedisConnection`.
+    :param str key: The rate limit's identifier.
+    :param int interval_seconds: Bucket size in seconds.
+    :param str cache_prefix: (optional) The prefix to prepend to cache keys.
```
pacejackson: Should you add a note that baseplate does not add a separator between the prefix and the final key so you might want to include one in this argument?

tests/unit/ratelimit_tests.py
```diff
@@ -0,0 +1,138 @@
...
+            wrapped_cache=wrapped_cache,
+            key='key',
+            interval_seconds=10,
+        )
+
+        self.assertEqual(result, 3)
+        get_current_interval.assert_called_once_with(10)
+        wrapped_cache.incr_and_get.assert_called_once_with('23:key', 20)
+
+    @mock.patch('baseplate.ratelimit.datetime')
```
pacejackson: Does this actually work?  I've had issues with trying to patch this sort of thing in the past.

tests/unit/ratelimit_tests.py
bsimpson63: is the idea that `import` goes before `from`?

baseplate/ratelimit.py
bsimpson63: I think automatically adding a separator would be unexpected, no?

tests/unit/ratelimit_tests.py
bsimpson63: Yeah, it didn't work when I tried to patch datetime.datetime though.

docs/baseplate/ratelimit.rst
bsimpson63: will update

tests/unit/ratelimit_tests.py
pacejackson: yes

baseplate/ratelimit.py
pacejackson: I feel the opposite, but that might just be me :)

baseplate/ratelimit.py
```diff
@@ -0,0 +1,114 @@
...
+
+    The 86400 seconds in a day are broken into into buckets of length
+    ``interval_seconds``. The current interval is then the bucket that the
+    current time falls into.
+
+    """
+    assert interval_seconds <= 86400, 'maximum interval is 1 day'
+
+    now = datetime.datetime.now(pytz.UTC)
+    seconds = (now.hour * 3600) + (now.minute * 60) + now.second
```
spladug: Why not just something like `time.time() // interval`?  Then we can drop the dep on pytz and maybe the restriction on interval size?

docs/baseplate/ratelimit.rst
```diff
@@ -0,0 +1,22 @@
+``baseplate.ratelimit``
+=======================
+
+.. automodule:: baseplate.ratelimit
+
+Assuming your project uses MemcacheContextFactory or RedisContextFactor to
```
spladug: Factor

tests/unit/ratelimit_tests.py
spladug: pep8 also says local imports should come last, so the baseplate one should be in the final group

baseplate/ratelimit.py
```diff
@@ -0,0 +1,114 @@
...
+    interval = get_current_interval(interval_seconds)
+
+    if cache_prefix:
+        cache_key = '%s%s:%s' % (cache_prefix, interval, key)
+    else:
+        cache_key = '%s:%s' % (interval, key)
+
+    # set the key to expire in 2 intervals (1 interval would be enough but
+    # let's be a bit more careful)
+    ttl = 2 * interval_seconds
```
spladug: should this take into account how far into the current interval we are?

baseplate/ratelimit.py
```diff
@@ -0,0 +1,114 @@
...
+    count = wrapped_cache.incr_and_get(cache_key, ttl)
+    return count
+
+
+def incr_and_get(cache, key, interval_seconds=60, cache_prefix=''):
```
spladug: Is there ever a time we would want the interval to change from call to call? Would that even be safe to do? If not, maybe it makes more sense to create an object and carry the interval there? Maybe the cache_prefix should be carried over too? It'd be good to have examples of usage for those in the docs (prefix=login, key=user_id) or something.

cshoe:
🏃

PR #163, opened 2018-02-21
Add baseplate.lock

bsimpson63:
Upstreaming from [thing service](https://github.com/reddit/reddit-service-thing/blob/master/thing/lock.py) and extended to support redis.

### DIFF
--- a/baseplate/lock.py
+++ b/baseplate/lock.py
@@ -1,0 +1,162 @@
+"""Simple locks in memcached or redis."""
+from __future__ import absolute_import
+
+import os
+import random
+import socket
+import time
+
+from pymemcache.client.base import PooledClient
+from redis import StrictRedis
+
+
+class TimeoutExpired(Exception):
+    pass
+
+
+MIN_RETRY_WAIT = 0.1
+MAX_RETRY_WAIT = 1.0
+
+
+class WrappedMemcache(object):
+    def __init__(self, cache):
+        self.cache = cache
+
+    def add_if_unset(self, key, value, ttl):
+        did_add = self.cache.add(key, value, expire=ttl, noreply=False)
+        return did_add
+
+    def get(self, key):
+        return self.cache.get(key)
+
+    def delete(self, key):
+        self.cache.delete(key, noreply=False)
+
+
+class WrappedRedis(object):
+    def __init__(self, cache):
+        self.cache = cache
+
+    def add_if_unset(self, key, value, ttl):
+        did_add = self.cache.set(key, value, nx=True, ex=ttl)
+
+        # StrictRedis returns True for 'set', and None for 'not set'
+        return bool(did_add)
+
+    def get(self, key):
+        return self.cache.get(key)
+
+    def delete(self, key):
+        self.cache.delete(key)
+
+
+class CacheLock(object):
+    """A lock implemented on memcached or redis.
+
+    Use the cache operation to add a key only if it doesn't exist. If
+    the key is not present in the cache its value is set and the server
+    responds that it was set. If the key is already present in the cache its
+    value is not modified and the server responds that it was not set.
+
+    We use this operation to ensure that only one client can obtain the lock
+    at a time. The first client gets the "set" response and interprets that to
+    mean that the lock was acquired. Any subsequent clients will get the
+    "not set" response and interpret that to mean that some other process has
+    the lock. They can wait and try again, or raise an exception.
+
+    """
+
+    def __init__(self, wrapped_cache, key, hold_limit, wait_limit):
+        self.wrapped_cache = wrapped_cache
+        self.key = key
+        self.hold_limit = hold_limit
+        self.wait_limit = wait_limit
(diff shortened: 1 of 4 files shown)


baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+        while not did_add:
+            did_add = self.wrapped_cache.add_if_unset(
+                self.key, self.nonce, ttl=self.hold_limit)
+
+            if did_add:
+                self.have_lock = True
+                return
+
+            wait_so_far = time.time() - start
+            remaining_wait_budget = self.wait_limit - wait_so_far
```
pacejackson: Can we use baseplate's `retrying` stuff for this?

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+        automatically calls :py:meth:`release`.
+
+        :raises: :py:exc:`TimeoutExpired` if ``wait_limit`` is exceeded
+            attempting to get the lock.
+
+        .. _context manager protocol:
+            https://docs.python.org/3/reference/datamodel.html#context-managers
+
+        """
+        self.nonce = "%s.%s" % (socket.gethostname(), os.getpid())
```
pacejackson: Should we stick a timestamp in here as well?  I think any concurrent requests will all have the same nonce, that might be an issue when deleting if you have a case where your lock expired and another concurrent request in the same app process also acquired the lock and adding a timestamp _should_ address that.

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+                # sleep before attempting to get the lock again
+                sleep_time = min(
+                    random.uniform(MIN_RETRY_WAIT, MAX_RETRY_WAIT),
+                    remaining_wait_budget,
+                )
+                time.sleep(sleep_time)
+
+    def release(self):
+        """Release the lock."""
+        if self.wrapped_cache.get(self.key) == self.nonce:
```
pacejackson: Should this have an `assert self.nonce` to make sure you don't call `release` without first calling acquire?

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    def __init__(self, wrapped_cache, key, hold_limit, wait_limit):
+        self.wrapped_cache = wrapped_cache
+        self.key = key
+        self.hold_limit = hold_limit
+        self.wait_limit = wait_limit
+        self.have_lock = False
+
+    def acquire(self):
```
spladug: Rather than this object having multiple states to juggle, could `acquire()` return an object that has `release()` on it? That way you can't even try to do it the wrong way as the methods won't exist.

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+        return did_add
+
+    def get(self, key):
+        return self.cache.get(key)
+
+    def delete(self, key):
+        self.cache.delete(key, noreply=False)
+
+
+class WrappedRedis(object):
```
spladug: Should we do [more proper locking](https://redis.io/topics/distlock) in Redis? And if so, should we actually keep doing janky-locking in memcached?

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+            if did_add:
+                self.have_lock = True
+                return
+
+            wait_so_far = time.time() - start
+            remaining_wait_budget = self.wait_limit - wait_so_far
+            if remaining_wait_budget <= 0:
+                raise TimeoutExpired
+            else:
+                # sleep before attempting to get the lock again
```
spladug: probably worth explaining in comment the purpose of the jitter?

baseplate/lock.py
```diff
@@ -0,0 +1,162 @@
...
+        lock is deleted on successful exit, so this is mostly concerned with
+        making sure the lock doesn't persist too long if the delete fails. On
+        the other hand if the lifetime is too short the lock may expire (and
+        be aquireable by other processes) before its owner is finished with it.
+    :param int wait_limit: The maximum time (in seconds) to wait and retry if
+        the lock isn't available.
+
+    """
+
+    if isinstance(cache, PooledClient):
```
spladug: both of these are relying on the current class hierarchy in the context helpers for the monitored equivalents to continue getting automatically mapped appropriately. that's all internal to baseplate which is fine, but it'd be good to have a test covering that so that if those modules change we notice a test failure.

bsimpson63:
> Mostly wondering if we want to promote doing the old-style janky memcached locks in future applications.

Yeah, 🤷‍♀️  . In addition to the problems with our standard lock implementation, we might not want to encourage heavy use of locks because they should often be a last resort. I guess if a service really really needs locks they can implement it themselves and it'll be more of a "proceed at your own peril" situation rather than being blessed by the framework.

If there were a way to make an interface that could eventually be eventually switched to use redlock that could be worth exploring.

I don't fee strongly about this PR or any of the options for including locks (or not) in baseplate.

cshoe:
🏃

PR #164, opened 2018-02-21
IO 2979: Add details to Cassandra spans

cshoe:
Extract statement type, table name and keyspace name from CQL
statements.

👓  @spladug

### DIFF
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -3,6 +3,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import re
+import six
+
 from cassandra.cluster import Cluster, _NOT_SET
 from cassandra.query import SimpleStatement, PreparedStatement, BoundStatement
 
@@ -130,9 +133,8 @@ class CassandraSessionAdapter(object):
 
     def execute_async(self, query, parameters=None, timeout=_NOT_SET):
         trace_name = "{}.{}".format(self.context_name, "execute")
-        span = self.server_span.make_child(trace_name)
+        span = self._get_span(trace_name, query)
         span.start()
-        # TODO: include custom payload
         if isinstance(query, string_types):
             span.set_tag("statement", query)
         elif isinstance(query, (SimpleStatement, PreparedStatement)):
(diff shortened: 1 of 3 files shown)


baseplate/context/cassandra.py
```diff
@@ -150,6 +152,71 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
...
+
+        # TODO: include custom payload
+        span.set_tag("statement", cql_str)
+        for key, value in CQLMetadataExtractor.extract(cql_str).items():
+            span.set_tag(key, value)
+
+        return span
+
+
+class CQLMetadataExtractor(object):
```
spladug: :grimacing: I'm really really skeptical of regex parsing arbitrary CQL queries for this. Do we have any data on the performance impact of doing this?

Does extracting from the CQL like this actually help when we have the full statement in the tags anyway?

cshoe: > Does extracting from the CQL like this actually help when we have the full statement in the tags anyway?

I'm not sure of the desire behind the ticket. My guess is wanting the spans from baseplate to resemble those generated by r2. If people are searching Zipkin traces by column family tag names, this would be helpful.

I share your concern. By the time the query is in baseplate, it's already a string or SimpleStatement type. With that in mind, there's no structured way I can see to access things like key or column family without parsing the CQL statement.

spladug: > By the time the query is in baseplate, it's already a string or SimpleStatement type.

That's the external API. By the time it's written in an application it's like that. This is pretty similar to how we don't tag tables etc. in SQL because it'd just require writing a full query parser for it.

@bsimpson63 do you have any thoughts on what benefit extracting and tagging this stuff has? I understand r2 has it, but that's the thrift API and I'm just not sure how worthwhile this is.

Either way, if we want to move forward with this I'd like to see some performance testing gauge the impact, pick out any low-hanging fruit of improvements, and potentially inform the choice of ditching the idea, sampling, or going full-bore.

cshoe: I talked to @ckwang8128. The aim here was to allow developers to search for traces based on column family and statement type. We're going to see if we can't provide this another way. Possibly by making improving the query functionality of the trace store by moving to ES.

PR #166, opened 2018-02-22
Kombu publisher and consumer

bsimpson63:
Copied from sendbird service ([consumer](https://github.com/reddit/reddit-service-sendbird/blob/master/sendbird/amqp_consumers.py) and [publisher](https://github.com/reddit/reddit-service-sendbird/blob/master/sendbird/kombu.py))

The publisher stuff is pretty straightforward.

For the consumer stuff maybe we want to go in a different direction. @pacejackson I think you said you have another version of this that has health checks and maybe some other nice features. Is that something that we should use instead, or could we extend what is here to support health checks.

👓 @spladug @pacejackson


baseplate/context/kombu.py
```diff
@@ -0,0 +1,71 @@
...
+        self.span = span
+        self.connection = connection
+        self.exchange = exchange
+        self.producers = producers
+
+    def publish(self, body, routing_key=None, **kwargs):
+        """Publish a message to the routing_key.
+
+        :param str body: The message body.
+        :param str routing_key: The AMQP routing key to publish to. If there
```
spladug: One of the reasons Kombu was chosen over alternatives is that it supports [more than just AMQP](http://docs.celeryproject.org/projects/kombu/en/latest/userguide/connections.html#transport-comparison). Would it be possible to make the docs more generic so that they work in context of other transports as well?

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+        self.thread.start()
+
+    @contextlib.contextmanager
+    def get_message(self, timeout=None):
+        with self.get_batch(max_items=1, timeout=timeout) as (context, batch, span):
+            yield context, batch[0], span
+
+    @contextlib.contextmanager
+    def get_batch(self, max_items, timeout):
+        assert self.thread is not None, "you must start the consumer first"
```
spladug: Rather than requiring this object be in a specific state for some of its methods to be used, how about having a consumer factory that does what `start()` currently does and returns an object which has `get_message` / `get_batch` on it? That way the user is funnelled to doing it the right way by the types involved.

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+        for _ in range(max_items):
+            try:
+                item = self.work_queue.get(block=block, timeout=timeout)
+                batch.append(item)
+            except queue.Empty:
+                break
+
+        context = ConsumerContext()
+        with self.baseplate.make_server_span(context, self.queues[0].name) as span:
+            span.set_tag("message_bus.arrival", self.queues[0].routing_key)
```
spladug: Should we add any other tags from https://github.com/opentracing/specification/blob/master/semantic_conventions.md#message-bus ?

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+    queues = []
+    for routing_key in routing_keys:
+        queue = Queue(
+            name=queue_name,
+            exchange=exchange,
+            routing_key=routing_key,
+        )
+        queues.append(queue)
+
+    print("registering %s as a handler for %r" % (handler.__name__, queues))
```
spladug: From debug times? Or should this be turned into a proper `logging` call?

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+    kombu_consumer = KombuConsumer(baseplate, connection, exchange, queues)
+    kombu_consumer.start()
+
+    print("waiting for messages")
```
spladug: Same thing.

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def on_message(self, message):
+        self.work_queue.put(message)
+
+
+class ConsumerContext(object):
+    pass
+
+
+class KombuConsumer(object):
```
spladug: I think the direct interface could be useful to expose as well for more complex users. Can this be exposed and documented? (Modulo potential changes for splitting start/get_)

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if timeout == 0:
+            block = False
+        else:
+            block = True
+
+        batch = []
+        for _ in range(max_items):
+            try:
+                item = self.work_queue.get(block=block, timeout=timeout)
```
spladug: should `timeout` be a "per check" timeout? or maximum timeout? i.e. should it be decreasing as time progresses across multiple `work_queue.get()` calls?

baseplate/queue_consumer.py
bsimpson63: I've been using this so I can see in the consumer logs what is happening on startup. I can switch this to `logging`.

spladug: Sounds good, logging would be preferable since it lets the application control the output destination and format.

pacejackson:
The other version I have of this is actually pretty similar to "kombu.py" and now that I've worked with it more, I would rather just move to using this.  The system I have feels kinda bloated and I'm not sure that the healthcheck stuff is really worth it.  What it does is you run the consumer as `baseplate-serve` and it starts up a baseplate pyramid app to respond to health checks on the port you passed it.  I think we would be better off figuring out a lighter-weight way to do healthchecks for this.

puppet/modules/baseplate/manifests/init.pp
```diff
@@ -24,6 +24,8 @@
     'python3-hvac',
     'python3-jwt',
     'python3-kazoo',
+    # TODO: must be backported from xenial
+    'python3-kombu',
```
pacejackson: We _do_ have a wheel for this.

baseplate/queue_consumer.py
bsimpson63: One thing I wondered about here: the list of queues probably all have different routing keys. It might be possible to figure out the actual routing key by looking at the message, but each message in the batch might have a different routing key. Maybe instead it makes more sense to just use the queue name here or omit entirely?

baseplate/queue_consumer.py
bsimpson63: I'm not sure. In the original version of this that I copied from you there was a TODO comment about this. What is the thinking on using a variable timeout?

spladug: If the timeout=1 and max_items is 3 we could end up waiting 3 seconds (1 second per item) for a full batch when you might expect that you'd time out at 1 total second regardless.

It just occurred to me that we can use baseplate's retry stuff for this:

```python
for time_remaining in RetryPolicy.new(attempts=max_items, budget=timeout):
    try:
        item = self.work_queue.get(block=block, timeout=time_remaining)
        batch.append(item)
    except queue.Empty:
        break
```

baseplate/queue_consumer.py
spladug: Hrm, batches do sorta mess things up. I wonder if we should be specializing the single-item case better so it can have more detailed tags since it doesn't have to worry about this general case.

Probably easiest just to let the batch be tagged with some basic stuff like total number of items and the consumer name or something like that?

spladug:
> What happens when you deploy to a queue consumer process? Does it just kill the consumer and start a new one or does it try to let the current worker(s) finish but not accept new work?

Currently it'd be kill-and-start-anew. Maybe the consumer function should catch [SIGUSR2 and bail after the current item/batch is done](https://github.com/reddit-archive/reddit/blob/master/r2/r2/lib/einhorn.py#L74-L75)? We'd need to move prod consumers under einhorn but I've been meaning to do that for a while anyway.

baseplate/queue_consumer.py
bsimpson63: I tried doing something like this:

```
kombu_consumer_factory = KombuConsumerFactory(connection, exchange, queues)
baseplate.add_to_context(queue_name, kombu_consumer_factory)

while True:
    context = ConsumerContext()
    with baseplate.make_server_span(context, queue_name) as span:
        queue = getattr(context, queue_name)
        message = queue.get_message()
        handler(context, message.body, message)
        message.ack()
```

but this doesn't quite work because when `baseplate.make_server_span` is called it creates a KombuConsumer instance with a work_queue, and all the messages are put in that queue, but the next time through the loop we call `baseplate.make_server_span` which makes another instance of KombuConsumer with a different work_queue, but all the messages are in the original work_queue.

I can flip it around like:

```
kombu_consumer_factory = KombuConsumerFactory(connection, exchange, queues)
baseplate.add_to_context(queue_name, kombu_consumer_factory)

context = ConsumerContext()
with baseplate.make_server_span(context, queue_name) as span:
    queue = getattr(context, queue_name)
    while True:
        message = queue.get_message()
        handler(context, message.body, message)
        message.ack()
```

which works, but now we're using the same span to process many messages, which seems wrong (in zipkin this will look like a giant span that includes the processing of hundreds or thousands of messages).

What do you think?

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,127 @@
...
+
+class ConsumerContext(object):
+    pass
+
+
+class KombuConsumer(object):
+    def __init__(self, baseplate, connection, exchange, queues):
+        self.baseplate = baseplate
+        self.connection = connection
+        self.exchange = exchange
```
pacejackson: I don't think we're actually using this.

baseplate/queue_consumer.py
pacejackson: Would it be better to make `start` private and just call `_start` on the first call to `get_batch`?

Alternatively, if we prefer the factory route, I don't think we want to use a `ContextFactory`, just a regular `Factory` that `starts` the consumer before returning it, I think the idea is to break KombuConsumer into 2 classes like:

```python
class KombuConsumerFactory(object):
    def __init__(self, baseplate, connection, exchange, queues):
        ...

    def build_consumer(self):
        work_queue = queue.Queue()
        consumer = _ConsumerWorker(self.connection, self.queues, self.work_queue)
        thread = Thread(target=consumer.run)
        thread.name = "consumer message pump"
        thread.daemon = True
        thread.start()
        return KombuConsumer(self.baseplate, self.queues, work_queue, thread)

class KombuConsumer(object):
    def __init__(self, baseplate, queues, work_queue, thread):
        ...

    @contextlib.contextmanager
    def get_message(self, timeout=None):
        ...

    @contextlib.contextmanager
    def get_batch(self, max_items, timeout):
        ...
```

spladug: Ah yeah, sorry for the Factory/ContextFactory confusion. Just a normal factory is all I meant!

Splitting it up like that just means you can't misuse it and it makes things a bit more testable since the worker thread can be mocked out.

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Can be used directly, outside of standard baseplate context.
+
+    """
+    def __init__(self, worker, worker_thread):
+        self.worker = worker
+        self.worker_thread = worker_thread
+
+    @classmethod
+    def new(cls, connection, queues):
```
bsimpson63: I'm not sure if this is what you had in mind for the factory, but it does do a better job of setting things up and isolating the worker to make things testable.

bsimpson63:
💇

docs/baseplate/queue_consumer.rst
```diff
@@ -0,0 +1,46 @@
...
+        routing_keys=[
+            'link_created',
+            'link_deleted',
+            'link_updated',
+        ],
+        handler=process_links,
+    )
+
+
+This will create a queue named 'process_links_q' and bind the routing keys
```
spladug: if you want these to be monospace they need double-backticks

baseplate/queue_consumer.py
spladug: Yes, this is great! 

I think this method needs a docstring so it'll get picked up by sphinx. It's currently missing from the docs.

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,201 @@
...
+        worker_thread.name = "consumer message pump"
+        worker_thread.daemon = True
+        worker_thread.start()
+
+        return cls(worker, worker_thread)
+
+    def get_message(self):
+        """Return a single message."""
+        batch = self.get_batch(max_items=1, timeout=None)
+        return batch[0]
```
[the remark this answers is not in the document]
spladug: Sounds reasonable to me.

baseplate/queue_consumer.py
```diff
@@ -0,0 +1,201 @@
...
+            for no timeout.
+
+        """
+        if timeout == 0:
+            block = False
+        else:
+            block = True
+
+        batch = []
+        retry_policy = retry.RetryPolicy.new(attempts=max_items, budget=timeout)
```
spladug: Classes are generally imported directly in baseplate (bare functions are kept with their modules)

baseplate/context/kombu.py
```diff
@@ -0,0 +1,69 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from kombu.pools import Producers
+
+from baseplate.context import ContextFactory
+
+
```
spladug: do we want a `_from_config` function for kombu at all?

baseplate/queue_consumer.py
pacejackson: 👍 I'm a fan of factory classmethods like this.

baseplate/queue_consumer.py
bsimpson63: I had to do this to get the tests to play nice, but I'll just change my mock

bsimpson63:
💇

baseplate/context/kombu.py
```diff
@@ -0,0 +1,110 @@
...
+from kombu import Connection
+from kombu import Exchange
+from kombu.pools import Producers
+
+from baseplate import config
+from baseplate.context import ContextFactory
+
+
+def connection_from_config(app_config, prefix, **kwargs):
```
spladug: thanks for adding these! they need docstrings and to be in the .rst file though :(

bsimpson63:
💇

pacejackson:
Ack!  One thing I forgot about, it would be nice if you could use vault to get rabbitmq credentials rather than having to hard-code them into the connection url by passing `secrets` into `connection_from_config` and having an option to set the `credentials` vault path in the config.

bsimpson63:
rebasing, previous HEAD d3c7f959469b9764aaf13702441a3ecf684b464c

bsimpson63:
> I think that adding vault support for credentials would be super and I can show you an example if you like but it doesn't have to block this if others feel differently.

Can you link me to an example?

I think I'd like to merge this as is unless people feel strongly about adding vault support. I think adding vault support could be a project that touches rabbit, pg, and others.

bsimpson63:
for future reference https://github.com/reddit/reddit-service-mailroom/blob/cd9353557cb437175ac70c117917c2c09a4f0528/mailroom/rabbitmq/integration.py#L15

bsimpson63:
now that python3-kombu is available we can update the install and tests 1388498

PR #167, opened 2018-02-23
Close thrift connection after unexpected exception in thrift

spladug:
In cases where Thrift fails dealing with serialization/deserialization
(such as catastrophically wrong types being passed in) the connection
will end up in a bad state but we'll check it back into the pool to be
reused. This makes it so we close the connection in that case to prevent
it from causing issues for the next user.

(Found this while digging a bit into the "zero bytes" thing. I don't think this is the cause of most of our woes but is definitely a bug in itself)


bsimpson63:
🐟 cool

PR #169, opened 2018-03-22
Allow you to run the secret-fetcher as a script rather than a daemon

pacejackson:
While you generally want to run the secret fetcher as a daemon so it will pick up any updates, there are some situations where you only want to run it once.  This PR adds a `--once` argument to `secret_fetcher.py` that allows you to run it a single time and then exit.

### DIFF
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -283,9 +283,11 @@ def main():
     if args.debug:
         level = logging.DEBUG
     else:
-        level = logging.WARNING
-    logging.basicConfig(level=level)
+        level = logging.INFO
 
+    logging.basicConfig(
+        format='%(asctime)s:%(levelname)s:%(message)s',
+        level=level)
     parser = configparser.RawConfigParser()
     parser.readfp(args.config_file)
     fetcher_config = dict(parser.items("secret-fetcher"))
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -272,12 +272,45 @@ class VaultClient(object):
         return payload["data"], ttl_to_time(payload["lease_duration"])
 
 
+def fetch_secrets(cfg, client_factory):
+    logger.info("Fetching secrets.")
+    client = client_factory.get_client()
+    secrets = {}
+    soonest_expiration = client.token_expiration
+    for secret_name in cfg.secrets:
+        secrets[secret_name], expiration = client.get_secret(secret_name)
+        soonest_expiration = min(soonest_expiration, expiration)
+
+    with open(cfg.output.path + ".tmp", "w") as f:
+        os.fchown(f.fileno(), cfg.output.owner, cfg.output.group)
+        os.fchmod(f.fileno(), cfg.output.mode)
+
+        json.dump({
+            "secrets": secrets,
+            "vault": {
+                "token": client.token,
+                "url": cfg.vault.url,
+            },
+
+            # this is here to allow an upgrade path. the fetcher should
+            # be upgraded first followed by the application workers.
+            "vault_token": client.token,
+        }, f, indent=2, sort_keys=True)
+
+    # swap out the file contents atomically
+    os.rename(cfg.output.path + ".tmp", cfg.output.path)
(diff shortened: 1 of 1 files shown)


baseplate/secrets/fetcher.py
```diff
@@ -272,14 +272,46 @@ def get_secret(self, secret_name):
...
+    return soonest_expiration
+
+
 def main():
     arg_parser = argparse.ArgumentParser()
     arg_parser.add_argument("config_file", type=argparse.FileType("r"),
         help="path to a configuration file")
     arg_parser.add_argument("--debug", default=False, action="store_true",
         help="enable debug logging")
+    arg_parser.add_argument("--mode", type=str, default="daemon",
```
spladug: what do you think of just like `--once` or something?

baseplate/secrets/fetcher.py
```diff
@@ -308,42 +340,20 @@ def main():
...
-            }, f, indent=2, sort_keys=True)
-
-        # swap out the file contents atomically
-        os.rename(cfg.output.path + ".tmp", cfg.output.path)
-
-        time_til_expiration = soonest_expiration - datetime.datetime.utcnow()
-        time_to_sleep = time_til_expiration - VAULT_TOKEN_PREFETCH_TIME
-        time.sleep(max(int(time_to_sleep.total_seconds()), 1))
-
+    logger.info("Running secret fetcher in mode '%s'", args.mode)
```
spladug: nice :+1: the logs are _too_ silent right now

baseplate/secrets/fetcher.py
```diff
@@ -272,14 +272,46 @@ def get_secret(self, secret_name):
...
+            "vault": {
+                "token": client.token,
+                "url": cfg.vault.url,
+            },
+
+            # this is here to allow an upgrade path. the fetcher should
+            # be upgraded first followed by the application workers.
+            "vault_token": client.token,
+        }, f, indent=2, sort_keys=True)
+
```
spladug: worth adding a "fetched secrets" log output just to confirm it worked?

baseplate/secrets/fetcher.py
```diff
@@ -272,12 +272,45 @@ def get_secret(self, secret_name):
...
+    return soonest_expiration
+
+
 def main():
     arg_parser = argparse.ArgumentParser()
     arg_parser.add_argument("config_file", type=argparse.FileType("r"),
         help="path to a configuration file")
     arg_parser.add_argument("--debug", default=False, action="store_true",
         help="enable debug logging")
+    arg_parser.add_argument("--mode", type=str, default="daemon",
```
gtaylor: How would we feel about a single `--run-once` flag instead of mode? I feel like the latter may not be immediately evident at a glance, and could be confused with other things.

spladug: noice

gtaylor: beat me to it, @spladug

baseplate/secrets/fetcher.py
pacejackson: That sounds much better.

baseplate/secrets/fetcher.py
spladug: :highfive:

pacejackson:
💇‍♂️

baseplate/secrets/fetcher.py
```diff
@@ -318,8 +318,11 @@ def main():
         level = logging.DEBUG
     else:
         level = logging.INFO
-    logging.basicConfig(level=level)
 
+    logging.basicConfig(
```
spladug: syslog handles this in normal prod, does k8s require this?

gtaylor: as long as you're emitting to stdout/stderr, it'll be easily reachable (and shippable once we plug into Ricky's logging work)

spladug: sorry, i meant adding a timestamp to the log line

gtaylor: Whoops, misunderstood. It will not modify the log output. Once we have log shipping in place, the agent would be attaching that as structured metadata.

PR #171, opened 2018-04-03
Baseplate healthcheck rework

ckwang8128:
Rather than installing version-specific baseplate scripts, this moves to a central `baseplate-<script-name>` format so interfacing with these scripts can be the same across Python version installations. For Debian packages, we still include version-specific scripts to allow Python2-baseplate and Python3-baseplate to be installed on the same system. 

:eyeglasses: @spladug

### DIFF
--- a/bin/baseplate-healthcheck
+++ b/bin/baseplate-healthcheck
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+
+# Perform health-checks for a baseplate service.
+
+from baseplate.server.healthcheck import run_healthchecks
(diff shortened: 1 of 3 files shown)


ckwang8128:
Ahhh good call. I'll do that

ckwang8128:
I didn't bother adding documentation about the fact that baseplate-<script><2,3> are still available in the debian package. We've already been referring to the scripts without the version suffixes elsewhere. Let me know if I should add notes or something in the relevant sections.

PR #173, opened 2018-04-12
Move span publishing to sidecar

cshoe:
Followed the pattern established by the events publisher.

A new config section is expected following the same naming convention as
the events publisher: ``trace-publisher:<queue_name``.

In the main app config, an optional ``tracing.queue_name`` value is
introduced. When this value evaluates to truthy, the sidecar reporter is
used by the tracing client.

It looks like there's some work that needs to be done in the [Puppet
repo](https://github.com/reddit/puppet/tree/8bdc69b19b9eba9833909a0e4e4769cb9cf4b2e4/modules/baseplate)
to create queues and such. I'm still wrapping my head around that but
wanted to get this open for review.

👓 @spladug @ckwang8128


baseplate/diagnostics/tracing/__init__.py
```diff
@@ -67,14 +74,18 @@ def make_client(service_name, tracing_endpoint=None, max_span_queue_size=50000,
...
     :param int num_span_workers: number of worker threads for span processing.
     :param float span_batch_interval: wait time for span processing in seconds.
     :param float sample_rate: percentage of unsampled requests to record traces
         for.
     """
-    if tracing_endpoint:
+    if tracing_queue_name:
+        logger.info("Recording spans to queue %s", tracing_queue_name)
+        recorder = SidecarRecorder(tracing_queue_name)
+    elif tracing_endpoint:
```
spladug: Can we raise a deprecation warning in this case?

https://github.com/reddit/baseplate/blob/master/baseplate/_utils.py#L13-L21

baseplate/__init__.py
```diff
@@ -55,6 +55,8 @@ def tracing_client_from_config(raw_config, log_if_unconfigured=True):
         The name for the service this observer is registered to.
     ``tracing.endpoint`` (optional)
```
spladug: Should the in-app mode be marked as deprecated here? Or just flat out removed from docs but left functional?

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +505,42 @@ def flush_func(self, spans):
...
+            max_message_size=MAX_SIDECAR_MESSAGE_SIZE
+        )
+
+    def send(self, span):
+        # Don't raise exceptions from here. This is called in the
+        # request/response path and should finish cleanly.
+        serialized_str = json.dumps(span._serialize())
+        if len(serialized_str) > MAX_SIDECAR_MESSAGE_SIZE:
+            logger.error("Trace too big")
+        try:
```
spladug: nit: could you put a blank between the if and the try so they don't read as one big control structure please?

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +505,42 @@ def flush_func(self, spans):
...
+            self.queue.put(serialized_str, timeout=0)
+        except TimedOutError:
+            logger.error("Sidecar trace queueing timed out")
```
spladug: "timed out" is a bit confusing of an error message when the `timeout=0`. It's also useful to have a remediation action in the message. Maybe `Trace queue %s is full. Is the trace sidecar healthy?` (where %s is the queue name)

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +505,42 @@ def flush_func(self, spans):
...
+        else:
+            logger.debug("Trace queued")
```
spladug: If I'm running `baseplate-serve --debug` locally does this get printed out? If so, it's probably a bit spammy for regular devs.

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+                    logger.exception("HTTP Request failed. Response not available")
+            except IOError:
+                self.metrics.counter("error.io").increment()
+                logger.exception("HTTP Request failed")
+            else:
+                self.metrics.counter("sent").increment(payload.count)
+                return
+
+            tries += 1
+            if tries > self.retry_limit:
```
spladug: Maybe this could use https://baseplate.readthedocs.io/en/stable/baseplate/retry.html#baseplate.retry.RetryPolicy ?

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+        tries = 0
+        while True:
+            try:
+                with self.metrics.timer("post"):
+                    response = self.session.post(
+                        self.endpoint,
+                        data=payload.bytes,
+                        headers=headers,
+                        timeout=self.post_timeout,
+                        stream=False
```
spladug: nit: trailing comma please

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+
+        :param baseplate.events.publisher.SerializedBatch payload: Count and
+            payload to publish.
+        """
+        if not payload.count:
+            return
+
+        logger.info("Sending batch of %s traces", payload.count)
+        headers = {
+            "User-Agent": "baseplate-trace-publisher/1.0",
```
spladug: :+1:

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+    def publish(self, payload):
+        """Publish spans to Zipkin API.
+
+        :param baseplate.events.publisher.SerializedBatch payload: Count and
+            payload to publish.
+        """
+        if not payload.count:
+            return
+
+        logger.info("Sending batch of %s traces", payload.count)
```
spladug: %d?

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+    app_raw_config = dict(config_parser.items("app:{}".format(args.app_name)))
+    app_cfg = config.parse_config(app_raw_config, {
+        "tracing": {
+            "endpoint": config.Optional(config.Endpoint),
+            "queue_name": config.Optional(config.String)
+        }
+    })
+
+    # pylint: disable=maybe-no-member
+    assert args.queue_name == app_cfg.tracing.queue_name, \
```
spladug: Why have this on the command line and the config file if they must match?

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+
+    trace_queue = MessageQueue(
+        "/traces-" + args.queue_name,
+        max_messages=MAX_QUEUE_SIZE,
+        max_message_size=MAX_SPAN_SIZE,
+    )
+
+    # pylint: disable=maybe-no-member
+    inner_batch = TraceBatch(publisher_cfg.max_batch_size)
+    batcher = TimeLimitedBatch(inner_batch)
```
spladug: really cool to see this getting reused!

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+        "Queue name command line arg ({}) doesn't match tracing configuration ({})".format(
+            args.gueue_name, app_cfg.queue_name)
+
+    publisher_raw_cfg = dict(config_parser.items("tracing-publisher:" + args.queue_name))
+    publisher_cfg = config.parse_config(publisher_raw_cfg, {
+        "post_timeout": config.Optional(config.Integer, POST_TIMEOUT_DEFAULT),
+        "max_batch_size": config.Optional(config.Integer, MAX_BATCH_SIZE_DEFAULT),
+        "retry_delay": config.Optional(config.Integer, RETRY_DELAY_DEFAULT),
+        "retry_limit": config.Optional(config.Integer, RETRY_LIMIT_DEFAULT),
+        "zipkin_api_version": config.Optional(config.String, ZIPKIN_API_VERSION_DEFAULT),
```
spladug: Is there any advantage to breaking the URL components all out in the config instead of just asking for a full URL? I imagine if the API version were different we'd need to change code here anyway?

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+            payload to publish.
+        """
+        if not payload.count:
+            return
+
+        logger.info("Sending batch of %s traces", payload.count)
+        headers = {
+            "User-Agent": "baseplate-trace-publisher/1.0",
+            "Content-Type": "application/json",
+        }
```
spladug: Is there anything in Zipkin to allow us to tell it which service we are to aid in ratelimiting / diagnosis of bad actors?

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -0,0 +1,191 @@
+import argparse
+import logging
+import time
+
+import requests
+
+from . import MAX_SPAN_SIZE, MAX_QUEUE_SIZE
+from baseplate import config, metrics_client_from_config
+from baseplate._compat import configparser
+from baseplate.events.publisher import BatchFull, V1Batch, TimeLimitedBatch
```
spladug: we may want to move these to a more common location (`baseplate._utils`?) `V1Batch` is a pretty awkward name here, so it may be good to rename that to `RawJSONBatch` or something and then subclass it to `TraceBatch` and `V1Batch` in the respective clients?

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +505,42 @@ def flush_func(self, spans):
...
+    """Interface for recording spans to a POSIX message queue.
+
+    The SidecarRecorder deserializes spans to a string representation before
+    adding them to the queue.
+    """
+    def __init__(self, queue_name):
+        self.queue = MessageQueue(
+            "/traces-" + queue_name,
+            max_messages=MAX_SIDECAR_QUEUE_SIZE,
+            max_message_size=MAX_SIDECAR_MESSAGE_SIZE
```
spladug: nit: trailing comma please

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +505,42 @@ def flush_func(self, spans):
...
+        )
+
+    def send(self, span):
+        # Don't raise exceptions from here. This is called in the
+        # request/response path and should finish cleanly.
+        serialized_str = json.dumps(span._serialize())
+        if len(serialized_str) > MAX_SIDECAR_MESSAGE_SIZE:
+            logger.error("Trace too big")
```
spladug: maybe include the queue name, observed size, and limit?

spladug: also, what do we expect application devs to do about this? can we put remediation in the message?

baseplate/diagnostics/tracing/__init__.py
cshoe: This implies the plan is to fully deprecate and eventually remote the in-app approach. There's no mention of that in the ticket.  I just want to make sure that's the path we're taking.

spladug: That would be my inclination, yes.

baseplate/diagnostics/tracing/publisher.py
cshoe: Agreed. I thought about that but didn't want to introduce a refactor into these changes. If you think that's needed to get the rest of this in, I'll start on it.

baseplate/diagnostics/tracing/publisher.py
cshoe: > s there any advantage to breaking the URL components all out in the config instead of just asking for a full URL? 

I was trying to maintain some backwards compatibility with the in-app approach which only asks for an hostname and optional port.

> I imagine if the API version were different we'd need to change code here anyway?

The POST to the ``/spans`` endpoint is the same in both V1 and V2 so in theory, a config change is all that would be needed to use V2.

I think a move to a single ``zipkin_api_url`` (example value -- ``https://zipkin.snooguts.net:9411/api/v2``) is best. However, I was hesitant to introduce two different configuration paths while the in-app trace publishing is still supported.

baseplate/diagnostics/tracing/publisher.py
cshoe: Spans are tied to the service using the [``service_name`` config](https://github.com/reddit/baseplate/blob/master/baseplate/__init__.py#L80). I don't see a way to the publishing of a set of spans to the publisher.

baseplate/diagnostics/tracing/__init__.py
cshoe: Unless MAX_SIDECAR_MESSAGE_SIZE is turned into a configurable value, there's not really anything a dev can do.

I'm fine turning into a configurable but if we do, I propose we do the same to the event publisher in a separate PR.

baseplate/diagnostics/tracing/publisher.py
spladug: no prob, thanks!

baseplate/diagnostics/tracing/__init__.py
spladug: We probably want the limit to stay around -- just boosting it blindly is likely to cause issues. But usually a single span being super large would indicate that it has way too many tags right? So we could say something like "try sending fewer / smaller tags" ?

baseplate/diagnostics/tracing/publisher.py
spladug: Totally appreciated on trying to stay consistent. In this case, it's gonna be in a separate section of config even so it seems like a reasonable place to break backwards-compatibility/consistency if we're agreed that full URL is more future-flexible / easier to deal with.

cshoe:
💇

baseplate/_utils.py
```diff
@@ -42,3 +44,84 @@ def __get__(self, obj, type=None):
...
         setattr(obj, self.wrapped.__name__, ret)
         return ret
+
+
+SerializedBatch = collections.namedtuple("SerializedBatch", "count bytes")
+
+# maximum time (seconds) an event can sit around while we wait for more
+# messages to batch
+MAX_BATCH_AGE = 1
+# maximum size (in bytes) of a batch of events
```
spladug: these constants are event-specific, can they be in the events module only? (it makes sense for the base to not have any default since various users would have different needs)

(V1Batch/TraceBatch can specify the default for their use cases)

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -16,12 +16,14 @@
 import requests
 from requests.exceptions import RequestException
 
-from .._compat import queue
-from ..core import (
+from ...message_queue import MessageQueue, TimedOutError
```
spladug: :no_bell: i'm sorry for my sins. feel free to convert to absolute imports if you'd prefer.

baseplate/diagnostics/tracing/__init__.py
```diff
@@ -494,3 +507,46 @@ def flush_func(self, spans):
...
+
+
+MAX_SIDECAR_QUEUE_SIZE = 102400
+MAX_SIDECAR_MESSAGE_SIZE = 10000
+
+
+class SidecarRecorder(BaseBatchRecorder):
+    """Interface for recording spans to a POSIX message queue.
+
+    The SidecarRecorder deserializes spans to a string representation before
```
spladug: it serializes not deserializes

baseplate/diagnostics/tracing/publisher.py
spladug: bump

spladug:
it looks like a couple of threads got lost in the shuffle:

* https://github.com/reddit/baseplate/pull/173#discussion_r181272167
* https://github.com/reddit/baseplate/pull/173#discussion_r181143954
* https://github.com/reddit/baseplate/pull/173#discussion_r181143305

pacejackson:
🔕 Feel free to ignore this but since we can use either posix queues or redis as `MessageQueue`s does it make sense to make that pluggable in here so services can use one or the other?

baseplate/diagnostics/tracing/publisher.py
cshoe: I thought I responded to this...

I left the command line arg for queue name to make the interface consistent with the event publisher. I'll remove the command line arg, I suppose.

spladug:
We're not currently using redis queues for any event publishers, just for the collector->injector stage which is not in baseplate. It doesn't sound like a bad thing to do, but I don't think we'd use it soon and I don't wanna expand @cshoe's scope too much.

baseplate/diagnostics/tracing/publisher.py
spladug: Or remove it from the config to match the event publishers?

baseplate/diagnostics/tracing/__init__.py
spladug: Child spans? yeah that's fair. We ran into a similar situation where thing service was hitting the maximum packet size for metrics and the root cause was that there was no limit on batching for fetching messages and r2 was happily requesting 10k messages in one thrift call which had all sorts of other issues. So maybe just a general admonishment to investigate why the request is generating such a large trace?

baseplate/diagnostics/tracing/publisher.py
cshoe: Oh... I was using the [listing service config](https://github.com/reddit/puppet/blob/master/modules/reddit_listing_server/templates/listing.ini.erb#L29) as a reference. It has an ``events.queue_name`` but now taking a closer look at the code, it doesn't seem to be used anywhere.

cshoe:
💇

PR #174, opened 2018-04-13
Exposure events

Kaitaan:
- support for firing exposure events
- support for validating experiment names (to differentiate between when `variant` returns null because not bucketed, or "invalid" experiment name
- support for getting a list of all experiment names. This will be used to bucket a user for all (valid) experiments in order to return the full list of variants in a single call.
- fix weird bucketing issue with very small bucketing percentages on r2 experiments
- replace `event_queue` with `event_logger`

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -299,8 +299,8 @@ class User(_User):
 
         return {
             "user_id": user_id,
-            "user_logged_in": self.is_logged_in,
-            "cookie_created": self.cookie_created_ms,
+            "logged_in": self.is_logged_in,
+            "cookie_created_timestamp": self.cookie_created_ms,
         }
 
 
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -190,7 +190,7 @@ class Experiments(object):
         return variant
 
 
-def experiments_client_from_config(app_config, event_queue):
+def experiments_client_from_config(app_config, event_logger):
     """Configure and return an :py:class:`ExperimentsContextFactory` object.
 
     This expects one configuration option:
@@ -201,15 +201,15 @@ def experiments_client_from_config(app_config, event_queue):
 
     :param dict raw_config: The application configuration which should have
         settings for the experiments client.
-    :param baseplate.events.EventQueue event_queue: The EventQueue to be used
+    :param baseplate.events.EventQueue event_logger: The EventLogger to be used
         to log bucketing events.
     :rtype: :py:class:`ExperimentsContextFactory`
 
-    """
+    """ 
     cfg = config.parse_config(app_config, {
         "experiments": {
             "path": config.Optional(config.String, default="/var/local/experiments.json"),
         },
     })
     # pylint: disable=maybe-no-member
-    return ExperimentsContextFactory(cfg.experiments.path, event_queue)
+    return ExperimentsContextFactory(cfg.experiments.path, event_logger)
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -6,6 +6,8 @@ from __future__ import unicode_literals
 import json
 import logging
 
+from enum import Enum
+
 from .providers import parse_experiment
 from .. import config
 from ..context import ContextFactory
@@ -14,6 +16,11 @@ from ..file_watcher import FileWatcher, WatchedFileNotAvailableError
 
 
 logger = logging.getLogger(__name__)
+
+
+class EventType(Enum):
+    EXPOSE = 'expose'
+    BUCKET = 'choose'
 
 
 class ExperimentsContextFactory(ContextFactory):
@@ -91,7 +98,7 @@ class Experiments(object):
         return self._experiment_cache[name]
 
     def get_all_experiment_names(self):
-        """Returns a list of all valid experiment names from the
+        """Return a list of all valid experiment names from the
         configuration file.
 
         :rtype: :py:class:`list`
@@ -103,7 +110,7 @@ class Experiments(object):
         return experiment_names
 
     def is_valid_experiment(self, name):
-        """Returns true if the provided experiment name is a valid
+        """Return true if the provided experiment name is a valid
         experiment.
 
         :param str name: Name of the experiment you want to check.
@@ -115,7 +122,7 @@ class Experiments(object):
         return self._get_experiment(name) is not None
 
     def variant(self, name, user=None, bucketing_event_override=None,
-                extra_event_fields=None, **kwargs):
+                **kwargs):
         """Which variant, if any, is active.
 
         If a variant is active, a bucketing event will be logged to the event
@@ -152,10 +159,9 @@ class Experiments(object):
             If set to None, no override will be applied.  Set to None by
             default.  Note that setting bucketing_event_override to True has no
             effect, it will behave the same as when it is set to None.
-        :param dict extra_event_fields: (Optional) Any extra fields you want to
-            add to the bucketing event.
         :param kwargs:  Arguments that will be passed to experiment.variant to
-            determine bucketing, targeting, and overrides.
+            determine bucketing, targeting, and overrides. These values will also
+            be passed to the logger.
 
         :rtype: :py:class:`str`
         :return: Variant name if a variant is active, None otherwise.
@@ -169,13 +175,7 @@ class Experiments(object):
         inputs = dict(kwargs)
 
         if user:
-            if user.is_logged_in:
-                inputs["user_id"] = user.id
-                inputs["user_roles"] = user.roles
-            else:
-                inputs["user_id"] = user.loid
-            inputs["logged_in"] = user.is_logged_in
-            inputs['cookie_created_timestamp'] = user.event_fields().get('cookie_created')
+            inputs.update(user.event_fields())
 
         span_name = "{}.{}".format(self._context_name, "variant")
         with self._span.make_child(span_name, local=True, component_name=name):
@@ -208,10 +208,46 @@ class Experiments(object):
                 logged_in=inputs.get('logged_in'),
                 cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
                 app_name=inputs.get('app_name'),
+                event_type=EventType.BUCKET,
+                inputs=inputs,
             )
             self._already_bucketed.add(bucketing_id)
 
         return variant
+
+    def expose(self, experiment_name, variant_name, user=None, **kwargs):
+        """Log an event to indicate that a user has been exposed to an
+        experimental treatment.
+
+        :param str experiment_name: Name of the experiment that was exposed.
+        :param str variant_name: Name of the variant that was exposed.
+        :param baseplate.core.User user: (Optional) User object for the user
+            you want to check the experiment variant for. If unset, it is
+            expected that user_id and logged_in values will be set in the kwargs
+        :param kwargs: Additional arguments that will be passed to logger.
+
+        """
+
+        experiment = self._get_experiment(experiment_name)
+
+        if experiment is None:
+            return
+
+        inputs = dict(kwargs)
+
+        if user:
+            inputs.update(user.event_fields())
+
+        self._event_logger.log(
+            experiment=experiment,
+            variant=variant_name,
+            user_id=inputs.get('user_id'),
+            logged_in=inputs.get('logged_in'),
+            cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+            app_name=inputs.get('app_name'),
+            event_type=EventType.EXPOSE,
+            inputs=inputs,
+        )
 
 
 def experiments_client_from_config(app_config, event_logger):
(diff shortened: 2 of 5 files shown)


Kaitaan:
Let me know if this is too many things in one PR. I can break it out into 2 or 3 if necessary

baseplate/experiments/__init__.py
```diff
@@ -90,6 +90,18 @@ def _get_experiment(self, name):
             self._experiment_cache[name] = experiment
         return self._experiment_cache[name]
 
+    def _get_all_experiment_names(self):
```
pacejackson: It seems like this isn't used currently, can we leave it out until we actually need it?

baseplate/experiments/__init__.py
```diff
@@ -189,8 +201,36 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+            inputs['cookie_created_timestamp'] = user.event_fields().get('cookie_created')
+
+        self._event_logger.log(
+            experiment=experiment,
+            variant=variant_name,
+            user_id=inputs.get('user_id'),
+            logged_in=inputs.get('logged_in'),
+            cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+            app_name=inputs.get('app_name'),
+            exposed=True,
```
pacejackson: Controlling this with a boolean flag seems pretty inflexible, maybe we want to send a string value `event_type` that can be`'bucket'` or `'exposure'`?

baseplate/experiments/__init__.py
```diff
@@ -189,8 +201,36 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+        )
+
 
-def experiments_client_from_config(app_config, event_queue):
+def experiments_client_from_config(app_config, event_logger):
```
pacejackson: Is anything currently using this?  This update isn't backward-compatible so it could cause issues if something is using this already.

baseplate/experiments/__init__.py
```diff
@@ -90,6 +90,18 @@ def _get_experiment(self, name):
...
+        config = self._config_watcher.get_data()
+        exp_names = []
+        if config:
+            for name, cfg in config.items():
+                exp_names.append(name)
+
+        return exp_names
+
+    def is_valid_experiment(self, name):
+        return self._get_config(name) is not None
```
pacejackson: We should probably use `self._get_experiment` for this since the config could be invalid in some way.

baseplate/experiments/__init__.py
```diff
@@ -90,6 +90,18 @@ def _get_experiment(self, name):
...
+    def _get_all_experiment_names(self):
+        config = self._config_watcher.get_data()
+        exp_names = []
+        if config:
+            for name, cfg in config.items():
+                exp_names.append(name)
+
+        return exp_names
+
+    def is_valid_experiment(self, name):
```
pacejackson: 🔕 Rather than providing this, it might make more sense to have `variant` raise an exception if the experiment name does not map to a known experiment.  I'm not 100% sure about this as it has some downsides:

1. It means that code will randomly break when experiments get removed, in some ways this is _good_ because it forces you to clean up your code but it will likely break things.
2. It also means that we have to update the r2 code to account for this.
3. It means that you _have_ to put up your experiment config before you deploy your code that uses the experiment, this could be a pro or a con depending on how you look at it.

baseplate/experiments/__init__.py
```diff
@@ -189,8 +201,36 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+        inputs = dict(kwargs)
+
+        if user:
+            if user.is_logged_in:
+                inputs["user_id"] = user.id
+                inputs["user_roles"] = user.roles
+            else:
+                inputs["user_id"] = user.loid
+            inputs["logged_in"] = user.is_logged_in
+            inputs['cookie_created_timestamp'] = user.event_fields().get('cookie_created')
```
pacejackson: `user.event_fields` should probably already do all this work for you, can you put this work in there and just use `event_fields`?

baseplate/experiments/__init__.py
spladug: enum would be nicer than strings if it's a multi-option but finite set

baseplate/experiments/__init__.py
Kaitaan: I can move it to another PR if necessary, but I'm working on something that depends on this now.

baseplate/experiments/__init__.py
Kaitaan: updated

baseplate/experiments/__init__.py
Kaitaan: The only caller of this right now within the organization is messenger service, and it's using postitional parameters and sending in an EventLogger. This is not a functional change from the current state. The ExperimentsContextFactory already expects an EventLogger, not an EventQueue. It's just a param rename.

Though I did miss renaming it in the relevant test. Will update that.

baseplate/experiments/__init__.py
Kaitaan: > It means that code will randomly break when experiments get removed... 
This the primary reason I didn't go that route. Given that experiments can be disabled/removed from an external source (like an experiment config/setup tool), and that by people who are not the implementing engineers, I think it's just too risky to throw exceptions. I don't like the idea that someone could remove an experiment via the configuration system and break systems that haven't properly caught that exception.

> in some ways this is good because it forces you to clean up your code
I don't love breaking systems as a forcing function for a process. Cleaning up old code is a process problem, and probably shouldn't be solved with the threat of broken systems. 

> It means that you have to put up your experiment config before you deploy your code that uses the experiment
This is also not ideal. Experiments are deliberately run for a fixed time period, and having to coordinate the setup and deployment of experimental code to get close to your experiment window increases complication. I think it's better to allow the code to go out, then flip the experiment on when ready.

baseplate/experiments/__init__.py
Kaitaan: done.

Kaitaan:
💇‍♂️

baseplate/experiments/__init__.py
```diff
@@ -184,13 +197,36 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+            inputs.update(user.event_fields())
+
+        self._event_logger.log(
+            experiment=experiment,
+            variant=variant_name,
+            user_id=inputs.get('user_id'),
+            logged_in=inputs.get('logged_in'),
+            cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
+            app_name=inputs.get('app_name'),
+            event_type=EventType.EXPOSE,
```
pacejackson: Does _something_ need to be updated to account for this new field?

baseplate/experiments/__init__.py
pacejackson: That's fine, will there be a non-`_` method that uses this?  Or do we want to expose this directly?

baseplate/experiments/__init__.py
spladug: if this is meant to be a public function, it'll need a docstring so it shows up in docs.

baseplate/experiments/__init__.py
spladug: i'd suggest moving it to the other PR if that PR is what's going to need it. then it's able to change with the rest of that PR if needed.

baseplate/experiments/__init__.py
```diff
@@ -184,13 +197,36 @@ def variant(self, name, user=None, bucketing_event_override=None,
                 logged_in=inputs.get('logged_in'),
                 cookie_created_timestamp=inputs.get('cookie_created_timestamp'),
                 app_name=inputs.get('app_name'),
+                event_type=EventType.BUCKET,
             )
             self._already_bucketed.add(bucketing_id)
 
         return variant
 
+    def expose(self, experiment_name, variant_name, user=None, **kwargs):
```
spladug: this probably needs a docstring too?

baseplate/experiments/__init__.py
```diff
@@ -90,6 +97,18 @@ def _get_experiment(self, name):
             self._experiment_cache[name] = experiment
         return self._experiment_cache[name]
 
+    def _get_all_experiment_names(self):
+        config = self._config_watcher.get_data()
+        exp_names = []
+        if config:
+            for name, cfg in config.items():
```
spladug: this could be written `exp_names = list(config.keys())`

i'd also suggest not using an abbreviated variable name like that (per [pep-8](https://github.com/reddit/styleguide/blob/master/python/pep-0008.rst))

baseplate/experiments/__init__.py
Kaitaan: I think this should actually be public (not `_`), and will update accordingly. The "other PR" in this case is in a different repo (GraphQL).

baseplate/experiments/__init__.py
Kaitaan: Not in here. The EventLoggers that use this will need to be updated, but they're not in this repo.

baseplate/core.py
```diff
@@ -299,8 +299,8 @@ def event_fields(self):
 
         return {
             "user_id": user_id,
-            "user_logged_in": self.is_logged_in,
-            "cookie_created": self.cookie_created_ms,
+            "logged_in": self.is_logged_in,
+            "cookie_created_timestamp": self.cookie_created_ms,
```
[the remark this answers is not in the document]
Kaitaan: It's not required for this work, and adding it is more complex than just dropping it in (raises NoAuthenticationErrors) in tests. Solving that is outside the scope of the work in this PR.

baseplate/experiments/__init__.py
Kaitaan: added

baseplate/experiments/__init__.py
Kaitaan: added

baseplate/experiments/__init__.py
Kaitaan: fixed

Kaitaan:
💇

baseplate/experiments/__init__.py
```diff
@@ -90,6 +97,30 @@ def _get_experiment(self, name):
             self._experiment_cache[name] = experiment
         return self._experiment_cache[name]
 
+    def get_all_experiment_names(self):
+        """Returns a list of all valid experiment names from the
```
spladug: nit: [pep-257](https://github.com/reddit/styleguide/blob/master/python/pep-0257.rst) says these should be as a command like "Return a list" rather than "Returns a list". The summary (first line) should fit on one line if possible as well.

baseplate/experiments/__init__.py
```diff
@@ -90,6 +97,30 @@ def _get_experiment(self, name):
...
+        :rtype: :py:class:`list`
+        :return: List of all valid experiment names.
+        """
+        config = self._config_watcher.get_data()
+        experiment_names = list(config.keys())
+
+        return experiment_names
+
+    def is_valid_experiment(self, name):
+        """Returns true if the provided experiment name is a valid
```
spladug: same nit about tense

Kaitaan:
💇‍♂️ 

recreated all commits to be atomic, and also each commit is independent. These do not need (and probably should not be) squashed before merging.

baseplate/experiments/__init__.py
```diff
@@ -184,13 +210,54 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
             self._already_bucketed.add(bucketing_id)
 
         return variant
 
+    def expose(self, experiment_name, variant_name, user=None, **kwargs):
+        """Log an event to indicate that a user has been exposed to an
+        experimental treatment.
+
+        :param str experiment_name: Name of the experiment that was exposed.
+        :param str variantnt_name: Name of the variant that was exposed.
```
pacejackson: typo here `variantnt_name`

baseplate/experiments/__init__.py
```diff
@@ -184,13 +210,54 @@ def variant(self, name, user=None, bucketing_event_override=None,
...
+        :param baseplate.core.User user: (Optional) User object for the user
+            you want to check the experiment variant for.  If you set user,
+            the experiment parameters for that user ("user_id", "logged_in",
+            and "user_roles") will be extracted and added to the inputs to the
+            call to Experiment.variant.  The user's event_fields will also be
+            extracted and added to the bucketing event if one is  logged.  It
+            is recommended that you provide a value for user rather than
+            setting the user parameters manually in kwargs.
```
pacejackson: It looks like this is still the description for this argument in `variant`

baseplate/experiments/__init__.py
```diff
@@ -128,10 +159,11 @@ def variant(self, name, user=None, bucketing_event_override=None,
             If set to None, no override will be applied.  Set to None by
             default.  Note that setting bucketing_event_override to True has no
             effect, it will behave the same as when it is set to None.
-        :param dict extra_event_fields: (Optional) Any extra fields you want to
-            add to the bucketing event.
+        :param dict extra_event_fields: (Deprecated) Any extra fields you want to
+            add to the bucketing event. Not presently used.
```
pacejackson: 🔕  Since this method accepts arbitrary kwargs, you can just remove this if you're not using this anymore, there might be some reason not to that I'm not familiar with though?

Kaitaan:
💇‍♂️

PR #176, opened 2018-04-25
Simplify trace publisher config

cshoe:
While making changes to the baseplate puppet module I realized I made a
few mistakes in the final round of changes in trace sidecar PR. I wanted to set
them straight before the next version of baseplate is officially
released.

These changes include:

* Renaming the config section from "tracing-publisher" to "trace-publisher". This is more in line with the existing Event publisher config section.
* Remove all Zipkin scheme and API version settings. Instead rely on the ``zipkin_api_url`` config option.

👓 @spladug

### DIFF
--- a/baseplate/diagnostics/tracing/publisher.py
+++ b/baseplate/diagnostics/tracing/publisher.py
@@ -8,6 +8,7 @@ from baseplate import config, metrics_client_from_config
 from baseplate._compat import configparser
 from baseplate.message_queue import MessageQueue, TimedOutError
 from baseplate.retry import RetryPolicy
+from baseplate._compat import urlparse
 from baseplate._utils import BatchFull, RawJSONBatch, TimeLimitedBatch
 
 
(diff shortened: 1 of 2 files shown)


baseplate/diagnostics/tracing/publisher.py
```diff
@@ -44,16 +42,15 @@ def __init__(self,
...
             num_conns=5):
 
         adapter = requests.adapters.HTTPAdapter(pool_connections=num_conns,
                                                 pool_maxsize=num_conns)
+        parsed_url = urlparse(zipkin_api_url)
         self.session = requests.Session()
-        self.session.mount("{}://".format(zipkin_scheme), adapter)
-        self.endpoint = "{}/spans".format(zipkin_api_url.rstrip('/'), zipkin_api_version)
+        self.session.mount("{}://".format(parsed_url.scheme), adapter)
+        self.endpoint = "{}{}/spans".format(parsed_url.netloc, parsed_url.path)
```
spladug: should this have the scheme too?

cshoe: Yes, thank you. Added a simple test to check ``self.endpoint`` as well as the args to ``self.session.mount``.

cshoe:
💇

PR #178, opened 2018-05-17
Automatically cache prepared cassandra statements.

alienth:
Prepared statements are only useful if we can re-use them. This change
will cache prepared statements based upon the text of the query being
prepared, and automatically use that cache without preparing a new
statement if an identical statement was already prepared.

One potential gotcha here would be continual memory growth if someone was to prepare a huge number of unique queries. If we think that could be a problem, I can implement a capped LRU cache.

:eyeglasses: @spladug  @prthiaga

### DIFF
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -59,9 +59,11 @@ class CassandraContextFactory(ContextFactory):
     """
     def __init__(self, session):
         self.session = session
+        self.prepared_statements = {}
 
     def make_object_for_context(self, name, server_span):
-        return CassandraSessionAdapter(name, server_span, self.session)
+        return CassandraSessionAdapter(name, server_span, self.session,
+                                       self.prepared_statements)
 
 
 class CQLMapperContextFactory(CassandraContextFactory):
@@ -100,10 +102,11 @@ def _on_execute_failed(exc, span):
 
 
 class CassandraSessionAdapter(object):
-    def __init__(self, context_name, server_span, session):
+    def __init__(self, context_name, server_span, session, prepared_statements):
         self.context_name = context_name
         self.server_span = server_span
         self.session = session
+        self.prepared_statements = prepared_statements
 
     @property
     def cluster(self):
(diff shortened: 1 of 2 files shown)


baseplate/context/cassandra.py
```diff
@@ -148,8 +153,25 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
...
         trace_name = "{}.{}".format(self.context_name, "prepare")
+        query_hash = ""
+
         with self.server_span.make_child(trace_name) as span:
+            if cache:
+                query_hash = hashlib.md5(query.encode()).hexdigest()
+                if query_hash in self.prepared_cache:
+                    return self.prepared_cache[query_hash]
+
             span.set_tag("statement", query)
```
alienth: Should I add a span for the hash lookup? Not sure how I should place that in here.

baseplate/context/cassandra.py
```diff
@@ -148,8 +153,25 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
...
+        automatically cached and reused. The cache is keyed on the text of the
+        statement.
+        """
+
         trace_name = "{}.{}".format(self.context_name, "prepare")
+        query_hash = ""
+
         with self.server_span.make_child(trace_name) as span:
+            if cache:
+                query_hash = hashlib.md5(query.encode()).hexdigest()
```
spladug: I don't think it's necessary to do hashing ourselves (or use a cryptographic hash) here. Python's dictionaries will use the built-in `hash()` on the string when dealing with lookups so it's perfectly fine to just do `cache[query.encode()]`

baseplate/context/cassandra.py
spladug: That's probably too granular for an automatic metric.

baseplate/context/cassandra.py
```diff
@@ -59,9 +61,11 @@ class CassandraContextFactory(ContextFactory):
     """
     def __init__(self, session):
         self.session = session
+        self.prepared_cache = {}
```
spladug: maybe `statement_cache` or `prepared_statements`?

baseplate/context/cassandra.py
```diff
@@ -148,8 +153,25 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
         future.add_errback(_on_execute_failed, span)
         return future
 
-    def prepare(self, query):
+    def prepare(self, query, cache=True):
+        """Prepares a CQL statement.
+
+        :param bool cache: If set to True (default), prepared statements will be
```
spladug: when do you see this being set to false? if there are known specific situations it may be good to pop that into the docs.

baseplate/context/cassandra.py
```diff
@@ -148,8 +153,25 @@ def execute_async(self, query, parameters=None, timeout=_NOT_SET):
...
+        automatically cached and reused. The cache is keyed on the text of the
+        statement.
+        """
+
         trace_name = "{}.{}".format(self.context_name, "prepare")
+        query_hash = ""
+
         with self.server_span.make_child(trace_name) as span:
```
spladug: Do we actually want to generate a span when we skip the server round trip? I think maybe we only want to do `make_child` in the case where we're doing the full thing.

baseplate/context/cassandra.py
alienth: TIL

alienth:
@spladug hmmmm, how might I test populating the cache without a server or mock implementation of CQL?

alienth:
I guess I could have a mock adapter, or something?

spladug:
The integration test suite actually has a full Cassandra installed in both CI and local dev.

That said, it may be better for testing purposes to mock out the cql driver and assert that its prepare method only gets called once even with multiple calls to our prepare method.

alienth:
:haircut:

alienth:
:haircut:

alienth:
:haircut: Added the version cut.

alienth:
Travis build passed on an effectively identical HEAD: https://travis-ci.org/reddit/baseplate/builds/380350925

PR #179, opened 2018-05-17
Fix decode bug

cshoe:
Strings only need to be decoded when they are instances of ``bytes``.

Added tests to explicitly tests both bytes and strs.

### DIFF
--- a/baseplate/context/memcache/__init__.py
+++ b/baseplate/context/memcache/__init__.py
@@ -253,7 +253,7 @@ def make_keys_str(keys):
     """Make a string representation of an iterable of keys.
 
     """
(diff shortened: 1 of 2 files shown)


cshoe:
👓 @alienth @spladug

spladug:
This was merged as 3f030d643c8e4a7549d4a4ad2030b76d6ab0c62c.

@JordanMilne I don't think "this" affects anything else as this is specific to the memcached stuff. Things are definitely more particular in Python 3 about using the right string types in the right places (as they should be). If you've found a bug that's not being covered by the existing multi-python testing can you please submit that as a separate issue?

PR #180, opened 2018-05-18
Skip queue_consumer_tests if kombu not present.

alienth:
:eyeglasses: @spladug  @cshoe


alienth:
Verified tests succeeded with kombu uninstalled.

PR #183, opened 2018-06-13
Add a new thrift build command that works with multiple thrift files

snoost:
The original command to build Thrift, `ThriftBuildPyCommand`, is too smart that it renames the generated python module with the `_thrift` suffix. When you only have one thrift file it works fine, but the moment you need to include a different thrift file it creates a huge problem.

For instance, if thrift file `a` includes `sub_a`, in the *generated* python file by thrift compiler, it `import sub_a`, and yet we are generating a python module called `sub_a_thrift`, so the generated py files are effectively broken. 

This issue is observed across Reddit, leading to a ton of ad hoc solutions that plagued the organization.

Unfortunately it is not easy to just kill the `_thrift` suffix in the original command, as many teams are using `import xx_thrift.ttypes ` in their python code.  Changing the original command would lead to an incompatible upgrade when we bump up baseplate version.

This is not ideal but going forward, we should discourage people from using the original command and use the new one. It is developer's responsibility to avoid namespace collision, the framework does not need to be that smart. 

👓 @spladug @mattknox @ckwang8128

### DIFF
--- a/baseplate/integration/thrift/command.py
+++ b/baseplate/integration/thrift/command.py
@@ -55,6 +55,45 @@ class BuildThriftCommand(Command):
 
                 self.copy_tree(input_package, output_package)
 
+class NewBuildThriftCommand(Command):
+    description = "Generate Python code from Thrift IDL."
+    user_options = [
+        ("build-base=", "b", "base directory for build library"),
+    ]
+
+    def initialize_options(self):
+        self.build_base = None
+
+    def finalize_options(self):
+        self.set_undefined_options("build",
+            ("build_base", "build_base"),
+        )
+
+    def run(self):
+        if self.dry_run:
+            return
+
+        temp_dir = os.path.join(self.build_base, "thrift")
+        self.mkpath(temp_dir)
+
+        for package in self.distribution.packages:
+            package_dir = os.path.join(*package.split("."))
+
+            for thriftfile in glob.glob(os.path.join(package_dir, "*.thrift")):
+                subprocess.check_call([
+                    "thrift1",
+                    "-strict",
+                    "-gen", "py:utf8strings,slots,new_style",
+                    "-out", temp_dir,
+                    "-I", os.path.dirname(baseplate.thrift.__file__),
+                    thriftfile,
+                ])
+
+                module_name = os.path.splitext(os.path.basename(thriftfile))[0]
+                input_package = os.path.join(temp_dir, module_name)
+                output_package = os.path.join(package_dir, module_name)
+                self.copy_tree(input_package, output_package)
+
 
 class ThriftBuildPyCommand(build_py):
     def run(self):


spladug:
Is this a problem outside v2 schemas? (Feel free to PM me if you don't want to link internal things in this public repo)

snoost:
yes, the problem for us is to use a common ml.thrift file across multiple projects (listing, minsky). It is used both as RPC interface as storage format. 

I have already tried the hackish way: https://github.com/reddit/minsky/commit/c9dff7cd1a21b2a719a4122f2d7273f06589610f to Minsky, but then we need to do the same hack in listing as it is part of the shared RPC, and then I also need to do that for relevance_model_loader, which uses this thrift struct as persistence format. I hate to plague the repos more if we can stop it from the root.

snoost:
In general we should make assumptions that sharing multiple thrifts are inevitable cross the org. v2 event schema is only one of the first instances that we'd have more than one thrift files shared, but it will not be the last one. We are also likely to thrift beyond baseplate but in python (especially for consumers), so deviating from thrift standard complier convention in baseplate is in fact gonna be an exception down the road.

spladug:
So this is identical except not putting `_thrift` on the end of the generated module names, right? Seems pretty reasonable. I honestly don't even remember why I made it suffix like that.

I'm trying to think through what a transition to the new format looks like. The current flow looks like:

* Projects override their setup.py's `build_py` command (which is a subcommand of `build`) with `"build_py": ThriftBuildPyCommand,` in their setup.py.
* `ThriftBuildPyCommand` defined here in Baseplate calls `build_thrift` before calling the superclass's `build_py` command. This basically just injects `build_thrift` into `build` so it happens automatically when installing etc.
* `build_thrift` is defined in Baseplate's setup.py to map to `BuildThriftCommand` which is what we're reworking here.

So, I think we may need to do something a little different for the cleanest/smoothest transition:

* Modify the existing `BuildThriftCommand` to take a parameter which configures whether or not it suffixes `_thrift` to the built package. (This lets us not have to muck with setup.py registrations of what build_thrift maps to and also saves us from having to duplicate the whole class just to change one line)
* Add a new `ThriftBuildPyCommand` class (maybe `NonManglingThriftBuildPyCommand`?) here that calls `build_thrift -someflag` to trigger the non-name-mangling behaviour. 
* Update the cookiecutter to use the new `ThriftBuildPyCommand` so that new projects start out with the new behaviour.

Once this is done, existing projects will continue to work as they always have, new projects will get the new behaviour, and existing projects that want to migrate can just change the override of "build" in their setup.py when they're ready.

Let me know if you have any questions about what I meant with this brain dump.

:nail_care:

ckwang8128:
:nail_care: I think this will be on hold for a bit, clearing out of my queue until spl.adug's suggestion is addressed!

spladug:
I'm going to close this out as the baseplate thrift builder was never intended for non-service thriftfiles and making it support them is 1) out of scope, and 2) not necessary as new stuff like wheel-based thriftfile distribution is coming online.

PR #185, opened 2018-06-21
Add ignore_exception_cls arg to configure_error_reporting

fishy:
Some Python frameworks allow you to raise certain exceptions to indicate
certain errors, and the framework will handle that accordingly. Those type of
exceptions are raised intentionally and reporting them to sentry would be
noisy.

Add ignore_exception_cls arg to configure_error_reporting to enable the feature
of ignoring certain exceptions.

There are in general two approaches to this problem, one is passing in an
exception class and the other is passing in a function to test against
exc_info. The latter is more flexible (it can handle the case that there are
more than one exception classes to ignore that doesn't share the same
superclass), but the former is easier and can handle 90% of the cases.

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -625,7 +625,7 @@ class Baseplate(object):
 
         self.register(TraceBaseplateObserver(tracing_client))
 
-    def configure_error_reporting(self, client):
+    def configure_error_reporting(self, client, ignore_exception_cls=None):
         """Send reports for unexpected exceptions to the given client.
 
         This also adds a :py:class:`raven.Client` object to the ``sentry``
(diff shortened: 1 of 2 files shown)


pacejackson:
The sentry client can already do this: https://docs.sentry.io/clients/python/advanced/

pacejackson:
In fact, I think gateway already does this for `GatewayException`s

spladug:
How'd it work out? I'd much rather let the application control this via the raven object than manually doing it unless there's a compelling reason to do it in baseplate.

spladug:
:nail_care:

fishy:
We have some other issues so I'll keep it open for a few more days.

fishy:
I think we are good now.

PR #186, opened 2018-06-22
support new bucketing logic

Kaitaan:
add support for:
- single variant experiments (one control, one variant)
- multi variant experiments (multiple controls/variants)
- feature rollout (single variant)


mattknox:
probably should include @pacejackson, given his previous work and r2 knowledge.

baseplate/experiments/providers/__init__.py
```diff
@@ -87,11 +94,12 @@ def parse_experiment(config):
         )
         version = None
     now = time.time()
-    if now < start_ts or now > stop_ts:
-        return ForcedVariantExperiment(None)
 
     enabled = config.get("enabled", True)
-    if not enabled:
+    if now < start_ts or now > stop_ts:
```
mattknox: either or both of start_ts and stop_ts could be None, I think, so this might raise.

mattknox: I'd check for enabled first-a disabled experiment might lack one or both of start/stop.

baseplate/experiments/providers/__init__.py
```diff
@@ -119,6 +127,35 @@ def parse_experiment(config):
             version=version,
             config=experiment_config,
         )
+    elif experiment_type == "multi_variant":
+        return MultiVariantExperiment.from_dict(
```
mattknox: I'd probably assign the class to something like experiment_type and then return experiment_type.from_dict(...) to DRY this up

baseplate/experiments/providers/feature_flag.py
```diff
@@ -7,7 +7,9 @@
 
 
 class FeatureFlag(R2Experiment):
-    """An experiment with a single variant "active".
+    """An experiment with a single variant "active". This type of experiment
```
mattknox: might be useful to link to the design doc rationale for why.  (For example, I don't know)

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,70 @@
...
+logger = logging.getLogger(__name__)
+
+from .simple_experiment import SimpleExperiment
+
+
+class MultiVariantExperiment(SimpleExperiment):
+    """Basic experiment, handling more than two variants (typically two
+    controls, and multiple treatments).
+    This type of experiment does not guarantee that changing the variant size
+    will not rebucketing existing users.
```
mattknox: will not rebucket.

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,70 @@
...
+        :param variants -- a dictionary of
+                           <string:variant name>:<float:percentage> pairs.
+        :return string -- the variant name, or None if bucket doesn't fall into
+                          any of the variants
+        """
+
+        current_offset = 0.0
+
+        for variant in self.variants:
+            current_offset += variant['size'] * self.num_buckets
```
mattknox: I'm pretty sure this multiplication is a bug.  If not, it needs a comment.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+
+class SimpleExperiment(Experiment):
+    """Base class for simple experiments. Note that this class shares a lot of
+    code from the r2.py class, which is slated for deprecation and removal.
+    """
+
+    def make_seed(self, seed_data):
+        id = seed_data.get('id')
+        name = seed_data.get('name')
+        version = seed_data.get('version')
```
mattknox: version should NOT be used in the randomization.  Shuffle_version should be, although it's fine to exclude it for now if you want.

mattknox: oh, hmm.  Since this is legacy, maybe that's OK?  Still think it should not be, for consistency.

baseplate/experiments/providers/single_variant.py
```diff
@@ -0,0 +1,68 @@
...
+            stop_ts=stop_ts,
+            enabled=enabled,
+            config=config,
+        )
+
+    def _validate_variants(self, variants):
+
+        super(SingleVariantExperiment, self)._validate_variants(variants)
+
+        if len(variants) != 2:
```
mattknox: probably want to verify that the sizes work out too, right?

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,70 @@
...
+            stop_ts=stop_ts,
+            enabled=enabled,
+            config=config,
+        )
+
+    def _validate_variants(self, variants):
+
+        super(MultiVariantExperiment, self)._validate_variants(variants)
+
+        if len(variants) < 3:
```
mattknox: sum(sizes) < 1.0 . ?

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+    def _is_targeted(self, **kwargs):
+        """Check if user/etc is targeted for this experiment.
+
+        Advanced targeting functionality to be supported in the future.
+        For now, return True.
+        """
+        return True
+
+    def _is_enabled(self, **kwargs):
+        return self.start_ts
```
mattknox: also the enabled field ,right?

baseplate/experiments/providers/single_variant.py
```diff
@@ -0,0 +1,68 @@
...
+        with the same bucket and varaints will result in the same answer
+
+        :param bucket -- an integer bucket representation
+        :param variants -- a dictionary of
+                           <string:variant name>:<float:percentage> pairs.
+        :return string -- the variant name, or None if bucket doesn't fall into
+                          any of the variants
+        """
+
+        if bucket < (self.variants[0]["size"] * self.num_buckets):
```
mattknox: again, why * num_buckets?

baseplate/experiments/providers/feature_flag.py
Kaitaan: This repo is OS, so can't really link internal docs.

baseplate/experiments/providers/single_variant.py
Kaitaan: what do you mean by "sizes work out"?

baseplate/experiments/providers/multi_variant.py
Kaitaan: done by the super call

baseplate/experiments/providers/__init__.py
Kaitaan: if start_ts or stop_ts is None, there's an error that would be thrown a couple lines above (line 83).

Though, now that you mention it, if they're explicitly passed as None, that'll cause an issue. Fix inbound.

baseplate/experiments/providers/multi_variant.py
Kaitaan: not a bug. Can add documentation. Have to multiply the size of the variant (ex: `0.5` for 50%) by the number of potential buckets to determine whether the assigned bucket falls in the variant's bucket range.

baseplate/experiments/providers/single_variant.py
Kaitaan: See comment below about super call, if you're referring to sum(sizes) < 1.0

baseplate/experiments/providers/feature_flag.py
Kaitaan: Reason for deprecation? Just because it's redundant. I'd like to stop supporting the legacy bucketing altogether for simplicity if possible.

baseplate/experiments/providers/__init__.py
pacejackson: I believe that the initial intention was that it was always required.

baseplate/experiments/providers/feature_flag.py
pacejackson: I'm not opposed to deprecating this to try to simplify the config format/keep things more consistent.  The "legacy" bucketing doesn't really affect anything here since there's only one variant, it can be any size.

Kaitaan: My thinking is that it's basically a feature rollout, with 100%, but with specific targeting. That's going to be replaced with the feature_rollout option once targeting is in place (work in progress)

baseplate/experiments/providers/simple_experiment.py
Kaitaan: note that the "version" used in this case is whatever version is desired to be used for seed generation. I actually pass the "shuffle_version" value in here, since that's the specific type of version we want to use for making the seed. So it's actually doing the right thing, but I agree that it's actually a bit convoluted. Will clean this up a bit.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: couple things missing here. Fixed.

pacejackson:
It looks like your `.vagrant` stuff snuck into this PR, maybe we should add that to `.gitignore`?

baseplate/experiments/providers/__init__.py
```diff
@@ -103,20 +119,25 @@ def parse_experiment(config):
...
+    elif experiment_type in ['single_variant', 'multi_variant',
+                             'feature_rollout']:
+        return experiment_class.from_dict(
             id=experiment_id,
             name=name,
             owner=owner,
-            version=version,
+            start_ts=start_ts,
+            stop_ts=stop_ts,
+            enabled=enabled,
```
pacejackson: Why pass this stuff to the experiment?  This stuff should be handled by `ForcedVariantExperiment`-s rather than the other providers so you don't have to worry about it.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+
+        self.variants = config.get("variants", [])
+
+        seed_data = {"id": id, "name": name, "shuffle_version": self.shuffle_version}
+        self.seed = config.get("bucket_seed", self.make_seed(seed_data))
+
+        self._validate_variants(self.variants)
+
+    def _get_seed(self):
+        return self.seed
```
pacejackson: This seems like a weird pattern, the thing you are returning is "public" but this is "private", I think the way to do this would be something like:

```
@property
def seed(self):
    return self._seed
```

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+        self.enabled = enabled
+
+        self.bucket_val = config.get("bucket_val", "user_id")
+        self.version = config.get("experiment_version")
+        self.shuffle_version = config.get("shuffle_version")
+
+        self.variants = config.get("variants", [])
+
+        seed_data = {"id": id, "name": name, "shuffle_version": self.shuffle_version}
+        self.seed = config.get("bucket_seed", self.make_seed(seed_data))
```
pacejackson: I think the general preference is to do this parsing in the factory methods (i.e. `from_dict`) and pass everything into the constructor explicitly rather than having it done in the constructor.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+            return None
+
+        bucket = self._calculate_bucket(lower_kwargs[self.bucket_val])
+        return self._choose_variant(bucket)
+
+    def _check_overrides(self, **kwargs):
+        """Check if any of the kwargs override the variant. Functionality
+        to be built in the future. For now, overrides are not supported.
+        """
+        return None
```
pacejackson: IMO, we should leave code like this out if it's not being implemented, we can implement it when we need it rather than having code that doesn't actually do anything lying around until we do.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _is_targeted(self, **kwargs):
+        """Check if user/etc is targeted for this experiment.
+
+        Advanced targeting functionality to be supported in the future.
+        For now, return True.
+        """
+        return True
```
pacejackson: Same comment as about `_check_overrides`.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _is_enabled(self, **kwargs):
+        current_ts = time.time()
+
+        return (self.enabled and current_ts >= self.start_ts
```
pacejackson: I don't think the experiments should be checking this, this would be handled by the parser who would return a `ForcedVariantExperiment` if any of this is true.

baseplate/experiments/providers/single_variant.py
```diff
@@ -0,0 +1,68 @@
...
+                  enabled=True, **kwargs):
+        return cls(
+            id=id,
+            name=name,
+            owner=owner,
+            start_ts=start_ts,
+            stop_ts=stop_ts,
+            enabled=enabled,
+            config=config,
+        )
```
pacejackson: This seems like it's the same across the subclasses of `SimpleExperiment`, maybe we should just implement this there?

baseplate/experiments/providers/single_variant.py
```diff
@@ -0,0 +1,68 @@
...
+        if len(variants) != 2:
+            raise ValueError("Single Variant experiments expect only one "
+                "variant and one control.")
+
+    def _choose_variant(self, bucket):
+        """Deterministically choose a percentage-based variant. Every call
+        with the same bucket and varaints will result in the same answer
+
+        :param bucket -- an integer bucket representation
+        :param variants -- a dictionary of
```
pacejackson: This isn't a parameter.

baseplate/experiments/providers/single_variant.py
```diff
@@ -0,0 +1,68 @@
...
+                           <string:variant name>:<float:percentage> pairs.
+        :return string -- the variant name, or None if bucket doesn't fall into
+                          any of the variants
+        """
+
+        if bucket < int(self.variants[0]["size"] * self.num_buckets):
+            return self.variants[0]["name"]
+        elif bucket >= int(self.num_buckets
+                - (self.variants[1]["size"] * self.num_buckets)):
+            return self.variants[1]["name"]
```
pacejackson: I don't think we really need to implement this in every one of these subclasses, it seems like it's all the same logic, the only difference is how many variants are allowed.

baseplate/experiments/providers/__init__.py
```diff
@@ -19,6 +22,15 @@
 ISO_DATE_FMT = "%Y-%m-%d"
 
 
+type_class_map = {
+    'r2': R2Experiment,
+    'feature_flag': FeatureFlag,
+    'single_variant': SingleVariantExperiment,
+    'multi_variant': MultiVariantExperiment,
+    'feature_rollout': FeatureRollout,
```
pacejackson: maybe this type should be `feature_flag_v2`?  Establish a versioning pattern rather than finding new names for the same thing?

baseplate/experiments/providers/simple_experiment.py
Kaitaan: Is that a python thing, a Reddit thing, or a "this is how it was done in here previously, so please stay consistent" thing?  I realize that currently my `from_dict` method is pretty useless, but I'm just trying to determine the reasoning for needing that method at all is.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: Normally, I'd agree. In this case, however, I think there's value in including it for 2 reasons:
1) to indicate that this code is coming shortly (it's currently in-progress)
2) by including the non-functionality here, anyone using this can easily override targeting and overrides in subclasses.

baseplate/experiments/providers/__init__.py
Kaitaan: Setting aside how it was done in the old types, I disagree for a few reasons:
1) an experiment being disabled (or not started, or ended) doesn't make it a fundamentally different experiment. It just makes it disabled so that it doesn't bucket users. 
2) if we don't pass this information in, then we can't use it in the experiment. I can see cases where we want to enforce different restrictions based on the type of experiment. For instance, multivariant may be required to run for 2 weeks, whereas a multi-armed bandit may be required to run longer. 
3) If we don't pass this in, it makes it more complex to allow us to start an experiment in the future (or end one). Assume the following flow:
a) I create an experiment, set to start in an hour
b) a bucketing call is made, but the experiment hasn't started yet, so it gets created as a ForcedVariantExperiment. This experiment is then added to the experiment_cache
c) No further config changes happen for a day

Now it's going to be treated as a forced variant until the config has been re-parsed for some reason.

baseplate/experiments/providers/single_variant.py
Kaitaan: The logic is actually a bit different, at least for the single_variant. Single variant buckets one variant at the high end of the bucketing range, and one at the low end. This is to allow those variants to change sizes without rebucketing.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: See comment above

baseplate/experiments/providers/__init__.py
Kaitaan: The implication with a feature flag is "on/off (for a targeted group)" which isn't the case here. Even if targeted, there's no guarantee that a user will be bucketed in one or the other (unless the rollout percentage is 0 or 100)

baseplate/experiments/providers/simple_experiment.py
Kaitaan: See reply above

baseplate/experiments/providers/simple_experiment.py
Kaitaan: Fixed

baseplate/experiments/providers/single_variant.py
Kaitaan: Yeah, could do. Fixed

baseplate/experiments/providers/single_variant.py
Kaitaan: Fixed

baseplate/experiments/providers/__init__.py
pacejackson: I still think that the current way this is handled is the way to go, not implementing it in the experiment provider:

1. While true, now you have to implement this logic in all of your experiment providers, the reason for putting this in the parsing logic and using ForcedVariantExperiments is so you don't have to re-implement the same "disable" logic every time.
2. But we _don't_ need to use this right now, if we have some new type of experiment where we do need to use it in the future, we can pass it in/implement it there.
3. I'm not sure where an experiment would be cached for an hour, the only cache is local to the request, it's not shared between requests.

Kaitaan: 1) but that's fundamentally a part of the experiment. I'd be willing to agree that it could be in the base experiment class (I put it in the base class for a simple experiment for that reason), but the status, start, and end times of an experiment are features of an experiment, not features of an experiment parser.
2) You're right, in that the hypothetical situations are just that: hypothetical. I still think that the properties of an experiment (whether it's enabled, when it's meant to start and end) should be _available_ to the experiment, not used to infer the type of experiment (that's what the `experiment_type` is for)
3) I may be showing my ignorance on some things here, but is there anything inherent in the `baseplate.Experiments` class that restricts its scope to request level? When you request bucketing on an experiment that hasn't been seen yet, its config is parsed, and the experiment object is added to this dict: https://github.com/reddit/baseplate/blob/master/baseplate/experiments/__init__.py#L67
That's the cache I'm referring to. 

Generally speaking though, I think using an entirely new type of object for the sake of abstracting away the enabled/disabled/started/ended has more downsides than upsides.

baseplate/experiments/providers/simple_experiment.py
pacejackson: It's more of 3 thing, although that's not always consistent.  You can see other examples of this in other parts of baseplate such as https://github.com/reddit/baseplate/blob/69eba73ad7493e8c4822677b391346b5a8ee7c38/baseplate/queue_consumer.py#L113

pacejackson: IMO, my reason for preferring this is I like the constructor to make it explicit what the dependencies of that object are.  When you can just pass in a couple of things and build off of that, you have to dive into the constructor to try to figure that out.

Kaitaan:
:haircut:

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,208 @@
...
+        bucket = long(hashed.hexdigest(), 16) % self.num_buckets
+        return bucket
+
+    def _validate_variants(self, variants):
+        """ Validate that the variants, as provided, are valid for this type of experiment.
+            For example, ensure that variant percentages do not add to more than 100%.
+        """
+
+        if variants is None:
+            raise ValueError('Sum of all variants is greater than 100%')
```
Kaitaan: This error message is wrong. Will fix.

baseplate/experiments/providers/__init__.py
```diff
@@ -55,28 +67,30 @@ def parse_experiment(config):
...
-        epoch = datetime(1970, 1, 1)
-        stop_ts = (expires - epoch).total_seconds()
-    else:
-        raise ValueError(
-            "Invalid config for experiment %s, missing start_ts and/or "
-            "stop_ts." % name
-        )
+    start_ts = config.get("start_ts")
+    stop_ts = config.get("stop_ts")
+    if (start_ts is None or stop_ts is None):
```
spladug: nit: no parens around the if condition in python please!

baseplate/experiments/providers/__init__.py
```diff
@@ -55,28 +67,30 @@ def parse_experiment(config):
     :return: A subclass of :py:class:`Experiment` for the given experiment
         type.
     """
-    experiment_type = config["type"].lower()
-    experiment_id = config["id"]
+    experiment_type = config.get("type")
+    if experiment_type:
+        experiment_type = experiment_type.lower()
+    experiment_id = config.get("id")
```
spladug: why change from `["id"]` to `.get("id")`? with the assert on the next line all we've done is change the error from a fairly understandable `KeyError: id` to `AssertionError` in the case where the ID wasn't defined

baseplate/experiments/providers/__init__.py
```diff
@@ -103,20 +119,25 @@ def parse_experiment(config):
         override = config.get("global_override")
         return ForcedVariantExperiment(override)
 
-    if experiment_type == "r2":
-        return R2Experiment.from_dict(
+    experiment_class = type_class_map.get(experiment_type, ForcedVariantExperiment)
+
+    if experiment_type in ['r2', 'feature_flag']:
```
spladug: it feels really weird to have the type map and then still branch on types. it'd be cleaner to move this branching into factory functions so this part of the code doesn't continue to need to be updated any time new variants are added.

baseplate/experiments/providers/feature_flag.py
spladug: rather than overembiggening the summary line of the docstring, you can add a sphinx deprecation notice to the docs instead.

http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated

baseplate/experiments/providers/__init__.py
spladug: I'm not sure which chain to respond to here but I wanted to hop in. I'm seeing a lot of discussion of "override in subclasses" and "implement in base classes". In general, I'd really strongly like to avoid 1) implementation inheritance, and 2) deep hierarchies of objects.  This ends up being quite byzantine to follow (jumping back and forth between class definitions to follow a code path) and very brittle.  Rather, it'd be good to encapsulate specific pieces of logic in functions or discrete objects we bring together with composition. Duplication by putting into multiple classes is definitely a bummer, but so is putting it into a base class instead. Would it make sense to put this kind of shared logic into an object that only handles expiration logic and then it can be included in any experiment type that cares?

baseplate/experiments/providers/forced_variant.py
```diff
@@ -7,7 +7,8 @@
 
 
 class ForcedVariantExperiment(Experiment):
-    """An experiment that always returns a specified variant.
+    """An experiment that always returns a specified variant. This experiment
```
spladug: same note about sphinx deprecation warning.

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,54 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+from .simple_experiment import SimpleExperiment
+
+
+class MultiVariantExperiment(SimpleExperiment):
+    """Basic experiment, handling more than two variants (typically two
```
spladug: summary line is supposed to be separated from the rest and generally fit on one line.

we need to get this into CI, but can you run the `pep257` tool on this please?

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,54 @@
...
+    controls, and multiple treatments).
+    This type of experiment does not guarantee that changing the variant size
+    will not rebucket existing users.
+
+    Does log bucketing events.
```
spladug: sentence fragment

baseplate/experiments/providers/multi_variant.py
```diff
@@ -0,0 +1,54 @@
...
+
+        super(MultiVariantExperiment, self)._validate_variants(variants)
+
+        if len(variants) < 3:
+            raise ValueError("MultiVariant experiments expect two controls "
+                "and at least one variant.")
+
+    def _choose_variant(self, bucket):
+        """Deterministically choose a percentage-based variant. Every call
+        with the same bucket and varaints will result in the same answer.
```
spladug: varaints

baseplate/experiments/providers/r2.py
```diff
@@ -14,7 +14,9 @@
 
 
 class R2Experiment(Experiment):
-    """A "legacy", r2-style experiment.
+    """A "legacy", r2-style experiment. This experiment type is deprecated,
```
spladug: same note about sphinx deprecation thing

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,208 @@
...
+
+from .base import Experiment
+from ..._compat import long, iteritems
+
+
+logger = logging.getLogger(__name__)
+
+
+class SimpleExperiment(Experiment):
+    """Base class for simple experiments. Note that this class shares a lot of
```
spladug: the stuff about sharing code would be better for a comment since it's irrelvant to someone _using_ baseplate and just reading the docs but is important to someone working inside baseplate.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,208 @@
...
+        if variant is not None and variant in self.variants:
+            return variant
+
+        if self.bucket_val not in lower_kwargs:
+            logger.info(
+                "Must specify %s in call to variant for experiment %s.",
+                self.bucket_val,
+                self.name,
+            )
+            return None
```
spladug: nit: can you put a blank between these ifs please? when jammed together it reads from bird's eye view as an if/elif

baseplate/experiments/providers/simple_experiment.py
spladug: "coming shortly" is weird in fixed-release libraries though.

baseplate/experiments/providers/__init__.py
spladug: ah! I see @pacejackson had a much better suggested structure for this here: https://github.com/reddit/baseplate/pull/186/files#r198285273 :)

pacejackson: As far as (3) goes, the `Experiments` objects are created by the `ExperimentsContextFactory` when you get a new request/context so you get a new cache on each request.

baseplate/experiments/providers/__init__.py
Kaitaan: fixed. This used to be multi-line.

baseplate/experiments/providers/__init__.py
Kaitaan: Fair. Altered it to still raise a KeyError (let me know if you think there's a better exception type), but made the error message a bit more explicit.

baseplate/experiments/providers/__init__.py
Kaitaan: I've changed this a bit, so let me know if you still have objections here.  I'm not sure I'm entirely following what you're suggesting though; this, for all intents and purposes _is_ the factory function.

Keep in mind that the legacy types (r2, feature_flag, forced_variant) are slated for deprecation soon.

baseplate/experiments/providers/feature_flag.py
Kaitaan: fixed

baseplate/experiments/providers/forced_variant.py
Kaitaan: fixed

baseplate/experiments/providers/multi_variant.py
Kaitaan: This file has been removed.

baseplate/experiments/providers/multi_variant.py
Kaitaan: file removed

baseplate/experiments/providers/multi_variant.py
Kaitaan: file removed

baseplate/experiments/providers/r2.py
Kaitaan: fixed

baseplate/experiments/providers/simple_experiment.py
Kaitaan: removed

baseplate/experiments/providers/simple_experiment.py
Kaitaan: fixed

baseplate/experiments/providers/simple_experiment.py
Kaitaan: 2 to 1. Removed.

Kaitaan:
CI probably failed on the tests that don't run locally. Will fix them up.

Kaitaan:
that should do it.

💇‍♂️

Kaitaan:
💇‍♂️ 
ping

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class MultiVariantSet(VariantSet):
+
+    def __init__(self, variants, num_buckets=1000):
+        self._validate_variants(variants)
+        self.variants = variants
+        self.num_buckets = num_buckets
+
+    def __contains__(self, item):
```
pacejackson: 🔕 If you check this often, it might be a good idea to just store a `set` of all of the variant names and `return item in self._variant_names`.

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,52 @@
...
+
+        return False
+
+    def _validate_variants(self, variants):
+
+        if variants is None:
+            raise ValueError('No variants provided')
+
+        if len(variants) < 3:
+            raise ValueError("MultiVariant experiments expect two controls "
```
pacejackson: Should this also check that at least 2 of the variants are named `control_X`?

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,52 @@
+from .base import VariantSet
+
+
+class MultiVariantSet(VariantSet):
```
pacejackson: Can you add docstrings for each of these `VariantSet`s?

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(variant_counts["variant_1"], 259)
+        self.assertEqual(variant_counts["variant_2"], 259)
+        self.assertEqual(variant_counts["variant_3"], 259)
+        self.assertEqual(variant_counts["variant_4"], 259)
+        self.assertEqual(variant_counts[None], 1)
+
+    def test_contains(self):
+        variant_set = create_multi_variant_set()
+
+        self.assertTrue("variant_2" in variant_set)
```
pacejackson: 🔕  `self.assertIn`/`self.assertNotIn` work as well for this.

baseplate/experiments/providers/variant_sets/base.py
```diff
@@ -0,0 +1,16 @@
+
+
+class VariantSet:
```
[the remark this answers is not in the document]
Kaitaan: fixed

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,186 @@
...
+from __future__ import unicode_literals
+
+import logging
+import hashlib
+import time
+
+from .base import Experiment
+from ..._compat import long, iteritems
+
+from .variant_sets.single_variant_set import SingleVariantSet
```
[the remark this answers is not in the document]
Kaitaan: I also don't have any strong opinions here. I just put it here because a variantset is a core piece of a provider.

baseplate/experiments/providers/__init__.py
```diff
@@ -103,21 +123,25 @@ def parse_experiment(config):
         override = config.get("global_override")
         return ForcedVariantExperiment(override)
 
-    if experiment_type == "r2":
-        return R2Experiment.from_dict(
+    if experiment_type in legacy_type_class_map.keys():
+        experiment_class = legacy_type_class_map.get(experiment_type)
```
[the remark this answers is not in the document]
Kaitaan: changed

baseplate/experiments/providers/variant_sets/base.py
```diff
@@ -0,0 +1,16 @@
...
+        """Return true if the variant name provided exists
+        in this variant set.
+        """
+        raise NotImplementedError
+
+    def choose_variant(self, bucket):
+        raise NotImplementedError
+
+    def validate_variants(self):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
Kaitaan: removed

baseplate/experiments/providers/variant_sets/multi_variant_set.py
Kaitaan: done

baseplate/experiments/providers/variant_sets/multi_variant_set.py
Kaitaan: I don't think the system should enforce variant naming.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,186 @@
...
+        if kwargs.get(self.bucket_val):
+            return ":".join(
+                [self.name, self.bucket_val, str(kwargs[self.bucket_val])]
+            )
+        else:
+            return None
+
+    def should_log_bucketing(self):
+        """ Default to true. Override if logging of eligibility events not required. """
+        return True
```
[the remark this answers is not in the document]
Kaitaan: good catch. Given that it's not inherent to a variant though, I'm going to punt the responsibility to the config to determine whether it wants to log them or not, defaulting to true. Added a param on instantiation

baseplate/experiments/providers/__init__.py
```diff
@@ -103,21 +123,25 @@ def parse_experiment(config):
         override = config.get("global_override")
         return ForcedVariantExperiment(override)
 
-    if experiment_type == "r2":
-        return R2Experiment.from_dict(
+    if experiment_type in legacy_type_class_map.keys():
```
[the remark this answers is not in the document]
Kaitaan: TIL

baseplate/experiments/providers/variant_sets/multi_variant_set.py
pacejackson: Makes sense, should the error message not reference "control" then?  I feel like that implies that it is looking for variants called "control".

Kaitaan: Makes sense. Changed.

Kaitaan:
💇‍♂️

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,65 @@
...
+    MultiVariantSets are not designed to support changes in variant sizes without
+    rebucketing.
+    """
+
+    def __init__(self, variants, num_buckets=1000):
+        """ :param variants -- array of dicts, each containing the keys 'name'
+            and 'size'. Name is the variant name, and size is the fraction of
+            users to bucket into the corresponding variant. Sizes are expressed
+            as a floating point value between 0 and 1.
+        :param num_buckets -- the number of potential buckets that can be
```
pacejackson: Can you add the `type`s to these `:param` declarations?  Also, same formatting comments as the docstrings in `simple_experiment.py`

baseplate/experiments/providers/variant_sets/rollout_variant_set.py
```diff
@@ -0,0 +1,56 @@
...
+    def _validate_variants(self, variants):
+
+        if variants is None:
+            raise ValueError('No variants provided')
+
+        if len(variants) != 1:
+            raise ValueError("Rollout variant only supports one variant.")
+
+        size = variants[0].get('size')
+        if size is None or size < 0.0 or size > 1.0:
```
pacejackson: Since you're not doing math, the floating point stuff shouldn't come up but it still might be better to do the same `int(size * num_buckets)` trick here as well.

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -0,0 +1,331 @@
...
+            # Calculating the percentage difference instead of looking at the
+            # raw difference scales better as we change NUM_USERS.
+            percent_equal = float(actual) / expected
+            self.assertAlmostEqual(percent_equal, 1.0, delta=.10,
+                                   msg='bucket: %s' % bucket)
+
+    def test_variant_returns_none_if_out_of_time_window(self):
+        valid_cfg = get_simple_config()
+        experiment_valid = parse_experiment(valid_cfg)
+        experiment_valid._choose_variant = choose_variants_override
```
pacejackson: Rather than monkey patching the function like this, you should decorate the test with a `mock.patch` decorator like and set the `return_value` on the mock, i.e.:

```python
@mock.patch('baseplate.experiments.providers.simple_experiment._choose_variant')
def test_variant_returns_none_if_out_of_time_window(self, choose_variant_mock):
    choose_variant_mock.return_value = 'fake_variant'
```

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -0,0 +1,331 @@
...
+    def test_no_bucket_val(self):
+        experiment = create_simple_experiment()
+        no_user_id_provided_variant = experiment.variant(not_user_id="t2_1")
+        none_user_id_provided_variant = experiment.variant(not_user_id=None)
+
+        self.assertIs(no_user_id_provided_variant, None)
+        self.assertIs(none_user_id_provided_variant, None)
+
+    def test_experiment_disabled(self):
+        experiments_cfg = get_simple_config()
```
pacejackson: For this test, you might also want to change the variant sizes so the "enabled" one is at 100% to highlight that it will always be `None`.

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -0,0 +1,331 @@
...
+        experiments_cfg['experiment']['enabled'] = False
+
+        experiment = parse_experiment(experiments_cfg)
+
+        variant = experiment.variant(user_id="t2_1")
+        self.assertIs(variant, None)
+
+    def test_bucket_val(self):
+        cfg = {"id": 1,
```
pacejackson: formatting nit: can you also take this first key down to the next line?

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -0,0 +1,331 @@
...
+
+        experiment._choose_variant = choose_variants_override
+
+        variant_default_bucket_val = experiment.variant(user_id="t2_1")
+        variant_new_bucket_val = experiment.variant(new_bucket_val="some_value")
+
+        self.assertIs(variant_default_bucket_val, None)
+        self.assertIsNot(variant_new_bucket_val, None)
+
+    def test_change_shuffle_version_changes_bucketing(self):
```
pacejackson: You might want to add the `skipIf` decorator used in other tests to this one since it's running a lot of bucket checks.

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
```diff
@@ -0,0 +1,120 @@
...
+        variant_set = create_multi_variant_set()
+
+        variant_counts = {
+            "variant_1": 0,
+            "variant_2": 0,
+            "variant_3": 0,
+            None: 0,
+        }
+
+        for bucket in range(0,NUM_BUCKETS_DEF):
```
pacejackson: nit: space between `,` and `NUM_BUCKETS_DEF`

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
```diff
@@ -0,0 +1,120 @@
...
+
+from datetime import timedelta
+
+from baseplate._compat import range
+from baseplate.experiments.providers.variant_sets.multi_variant_set import MultiVariantSet
+
+logger = logging.getLogger(__name__)
+
+
+NUM_BUCKETS_DEF = 1000
```
pacejackson: If `DEF` supposed to be `DEFAULT`? Can you call it that so it's more clear?

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
```diff
@@ -0,0 +1,120 @@
...
+        {"name": "variant_1", "size": 0.25},
+        {"name": "variant_2", "size": 0.25},
+        {"name": "variant_3", "size": 0.25},
+    ]
+
+    return cfg
+
+def create_multi_variant_set():
+    cfg = generate_variant_config()
+    return MultiVariantSet(cfg)
```
pacejackson: IIRC you can pass in the number of buckets, you should probably pass it in here, maybe making it an argument to `create_multi_variant_set` with a default value of `NUM_BUCKETS_DEF` so you don't have to worry about keeping those in sync with the default values.

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        for bucket in range(0,NUM_BUCKETS_DEF):
+            variant = variant_set.choose_variant(bucket)
+            variant_counts[variant] += 1
+
+        self.assertEqual(len(variant_counts), 4)
+
+        for variant_count in variant_counts.values():
+            self.assertEqual(variant_count, 250)
```
pacejackson: Can you also add a test for the other extreme end of this, where you only have a single bucket i.e.:

```
[
    {'name': 'active', 'size': 0.001},
    {'name': 'control_1', 'size': 0},
    {'name': 'control_2', 'size': 0},
]
```

tests/unit/experiments/providers/variant_sets/rollout_variant_set_tests.py
```diff
@@ -0,0 +1,112 @@
...
+
+from datetime import timedelta
+
+from baseplate._compat import range
+from baseplate.experiments.providers.variant_sets.rollout_variant_set import RolloutVariantSet
+
+logger = logging.getLogger(__name__)
+
+
+NUM_BUCKETS_DEF = 1000
```
pacejackson: The comments in `multi_variant_set_tests.py` also apply here.

tests/unit/experiments/providers/variant_sets/single_variant_set_tests.py
```diff
@@ -0,0 +1,124 @@
...
+
+from datetime import timedelta
+
+from baseplate._compat import range
+from baseplate.experiments.providers.variant_sets.single_variant_set import SingleVariantSet
+
+logger = logging.getLogger(__name__)
+
+
+NUM_BUCKETS_DEF = 1000
```
pacejackson: The comments in `multi_variant_set_tests.py` also apply here.

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
Kaitaan: added

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
Kaitaan: fixed

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
Kaitaan: fixed

tests/unit/experiments/providers/variant_sets/multi_variant_set_tests.py
Kaitaan: fixed

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: it's only running a lot of bucketing checks if users keep getting bucketed into the same buckets, which shouldn't happen. It will end the loop as soon as a user winds up in two different buckets, which should happen very quickly (likely the first or second iteration of the loop.

I don't like skipping tests locally where not absolutely necessary.

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: fixed.

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: changed

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: added

baseplate/experiments/providers/variant_sets/rollout_variant_set.py
Kaitaan: Didn't change this since it's not necessary, and I want to do checks as close to the provided config as possible.

baseplate/experiments/providers/variant_sets/multi_variant_set.py
Kaitaan: fixed.

baseplate/experiments/providers/variant_sets/base.py
```diff
@@ -0,0 +1,13 @@
+
+
+class VariantSet(object):
+    """ Base interface for variant sets """
```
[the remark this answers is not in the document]
Kaitaan: added more explanation.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+            return None
+
+        bucket = self._calculate_bucket(lower_kwargs[self.bucket_val])
+        return self._choose_variant(bucket)
+
+    def _is_enabled(self, **kwargs):
+        current_ts = time.time()
+
+        return (self.enabled and current_ts >= self.start_ts
+                and current_ts < self.stop_ts)
```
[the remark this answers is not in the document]
Kaitaan: changed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+            return None
+
+    def should_log_bucketing(self):
+        """ Default to true. Override if logging of eligibility events not required. """
+        return self._log_bucketing
+
+    def variant(self, **kwargs):
+        lower_kwargs = {k.lower(): v for k, v in iteritems(kwargs)}
+
+        if not self._is_enabled():
```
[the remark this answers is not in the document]
Kaitaan: changed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+    def get_unique_id(self, **kwargs):
+        if kwargs.get(self.bucket_val):
+            return ":".join(
+                [self.name, self.bucket_val, str(kwargs[self.bucket_val])]
+            )
+        else:
+            return None
+
+    def should_log_bucketing(self):
+        """ Default to true. Override if logging of eligibility events not required. """
```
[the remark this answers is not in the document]
Kaitaan: changed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError('Experiment version must be provided.')
+
+        self.variant_set = variant_set
+
+        seed_data = {"id": id, "name": name, "shuffle_version": self.shuffle_version}
+        self._seed = bucket_seed or self.make_seed(seed_data)
+
+    @classmethod
+    def from_dict(cls, id, name, owner, start_ts, stop_ts, config,
+                  variant_type, enabled=True, **kwargs):
```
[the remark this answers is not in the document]
Kaitaan: removed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+        id = seed_data.get('id')
+        name = seed_data.get('name')
+        version = seed_data.get('shuffle_version')
+
+        return "{}.{}.{}".format(id, name, version)
+
+    def __init__(self, id, name, owner, start_ts, stop_ts, config,
+                 experiment_version, shuffle_version, variant_set,
+                 bucket_seed, bucket_val, enabled=True, log_bucketing=True,
+                 num_buckets=1000, **kwargs):
```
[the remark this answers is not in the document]
Kaitaan: removed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+            to the experimental configuration
+        :param int shuffle_version -- distinct from the experiment version, this
+            value is used in constructing the default bucketing seed value (if not
+            provided). When this value changes, rebucketing will occur.
+        :param list variants -- the list of variants for this experiment. This should
+            be provided as an array of dicts, each containing the keys 'name'
+          and 'size'. Name is the variant name, and size is the fraction of
+          users to bucket into the corresponding variant. Sizes are expressed
+          as a floating point value between 0 and 1.
+        :param bucket_seed -- if provided, this provides the seed for determining
```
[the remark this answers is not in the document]
Kaitaan: fixed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def __init__(self, id, name, owner, start_ts, stop_ts, config,
+                 experiment_version, shuffle_version, variant_set,
+                 bucket_seed, bucket_val, enabled=True, log_bucketing=True,
+                 num_buckets=1000, **kwargs):
+        """
+        :param int id -- the experiment id. This should be unique.
+        :param string name -- the human-readable name of the experiment.
+        :param string owner -- who is responsible for this experiement.
+        :param timestamp start_ts -- when this experiment is due to start.
```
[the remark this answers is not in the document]
Kaitaan: fixed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+        :param bucket_seed -- if provided, this provides the seed for determining
+            which bucket a variant request lands in. Providing a consistent
+            bucket_seed will ensure a user is bucketed consistently.
+        :param bool enabled -- whether or not this experiment is enabled.
+            disabling an experiment means all variant calls will return None
+        :param bool log_bucketing -- whether or not to log bucketing events
+        :param int num_buckets -- how many available buckets there are for
+            bucketing requests. This should match the num_buckets in the
+            provided VariantSet. The default value is 1000, which provides
+            a potential variant granularity of 0.1%
```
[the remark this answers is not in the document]
Kaitaan: fixed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+            value is used in constructing the default bucketing seed value (if not
+            provided). When this value changes, rebucketing will occur.
+        :param list variants -- the list of variants for this experiment. This should
+            be provided as an array of dicts, each containing the keys 'name'
+          and 'size'. Name is the variant name, and size is the fraction of
+          users to bucket into the corresponding variant. Sizes are expressed
+          as a floating point value between 0 and 1.
+        :param bucket_seed -- if provided, this provides the seed for determining
+            which bucket a variant request lands in. Providing a consistent
+            bucket_seed will ensure a user is bucketed consistently.
```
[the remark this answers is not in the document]
Kaitaan: fixed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+        version = seed_data.get('shuffle_version')
+
+        return "{}.{}.{}".format(id, name, version)
+
+    def __init__(self, id, name, owner, start_ts, stop_ts, config,
+                 experiment_version, shuffle_version, variant_set,
+                 bucket_seed, bucket_val, enabled=True, log_bucketing=True,
+                 num_buckets=1000, **kwargs):
+        """
+        :param int id -- the experiment id. This should be unique.
```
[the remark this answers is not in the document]
Kaitaan: fixed.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,199 @@
...
+variant_type_map = {
+    'single_variant': SingleVariantSet,
+    'multi_variant': MultiVariantSet,
+    'feature_rollout': RolloutVariantSet,
+}
+
+
+class SimpleExperiment(Experiment):
+
+    def make_seed(self, seed_data):
```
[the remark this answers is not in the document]
Kaitaan: Removed method altogether.

baseplate/experiments/providers/__init__.py
```diff
@@ -19,6 +21,19 @@
 ISO_DATE_FMT = "%Y-%m-%d"
 
 
+legacy_type_class_map = {
+    'r2': R2Experiment,
+    'feature_flag': FeatureFlag,
+}
+
+
+simple_type_class_list = [
```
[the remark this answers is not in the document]
Kaitaan: done.

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,65 @@
...
+            raise ValueError("MultiVariant experiments expect three or "
+                "more variants.")
+
+        total_size = 0.0
+        for variant in variants:
+            if variant.get('size') is None:
+                raise ValueError('Variant size not provided: {}'.format(variants))
+            total_size += variant.get('size')
+
+        if total_size > 1.0:
```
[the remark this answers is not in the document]
Kaitaan: updated.

Kaitaan:
💇‍♂️

baseplate/experiments/providers/feature_flag.py
```diff
@@ -7,7 +7,10 @@
 
 
 class FeatureFlag(R2Experiment):
-    """An experiment with a single variant "active".
+    """.. deprecated:: 0.27
+    Use FeatureRollout instead.
```
pacejackson: Maybe these comments should say `Use SimpleExperiment with X VariantStore`?

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def _is_enabled(self, **kwargs):
+        current_ts = time.time()
+
+        return (self.enabled and self.start_ts <= current_ts < self.stop_ts)
+
+    def _calculate_bucket(self, bucket_val):
+        """Sort something into one of self.num_buckets buckets.
+
+        :param bucket_val -- a string used for shifting the deterministic bucketing
```
pacejackson: This is missing the type and still has the ` -- ` rather than `:`.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -0,0 +1,196 @@
...
+                       algorithm.  In most cases, this will be an Account's
+                       _fullname.
+        :return int -- a bucket, 0 <= bucket < self.num_buckets
+        """
+        # Mix the experiment seed with the bucket_val so the same users don't
+        # get bucketed into the same bucket for each experiment.
```
pacejackson: `placed into the same bucket`?

baseplate/experiments/providers/variant_sets/base.py
```diff
@@ -0,0 +1,18 @@
+
+
+class VariantSet(object):
+    """ Base interface for variant sets.
+
+    A VariantSet contains a set of experimental variants, as well as
+    their distributions. It is used by experiments to track which
+    variant a check is meant to return.
```
pacejackson: `to track which bucket a variant is assigned to`?

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if len(variants) < 3:
+            raise ValueError("MultiVariant experiments expect three or "
+                "more variants.")
+
+        total_size = 0.0
+        for variant in variants:
+            if variant.get('size') is None:
+                raise ValueError('Variant size not provided: {}'.format(variants))
+            total_size += int(variant.get('size') * self.num_buckets)
```
pacejackson: I think it's better to make this a `@classmethod` and pass in `num_buckets` as an argument.  It's a little weird to have a function called in the constructor that expects some parts to be set up.  Another option that might be ok is to assign everything and then call `self._validate_variants` and use both `self.size` and `self.variants`.

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if total_size > self.num_buckets:
+            raise ValueError('Sum of all variants is greater than 100%')
+
+    def choose_variant(self, bucket):
+        """Deterministically choose a variant. Every call with the same bucket
+        on one instance will result in the same answer
+
+        :param bucket -- an integer bucket representation
```
pacejackson: Formatting and param type.

baseplate/experiments/providers/variant_sets/rollout_variant_set.py
```diff
@@ -0,0 +1,56 @@
...
+    users. Those users going from no treatment to the provided treatment
+    (or vice versa) are limited to the change in the provided treatment size.
+    For instance, going from 45% to 55% will result in only the new 10% of
+    users changing treatments. The initial 45% will not change. Conversely,
+    going from 55% to 45% will result in only 10% of users losing the
+    treatment.
+    """
+
+    def __init__(self, variants, num_buckets=1000):
+        """ :param variants -- array of dicts, each containing the keys 'name'
```
pacejackson: Formatting and param type.

baseplate/experiments/providers/variant_sets/rollout_variant_set.py
```diff
@@ -0,0 +1,56 @@
...
+
+        size = variants[0].get('size')
+        if size is None or size < 0.0 or size > 1.0:
+            raise ValueError('Variant size must be between 0 and 1')
+
+    def choose_variant(self, bucket):
+        """Deterministically choose a percentage-based variant. Every call
+        with the same bucket and varaints will result in the same answer.
+
+        :param bucket -- an integer bucket representation
```
pacejackson: Formatting and param type.

baseplate/experiments/providers/variant_sets/single_variant_set.py
```diff
@@ -0,0 +1,67 @@
...
+    This VariantSet allows adjusting the sizes of variants without
+    changing treatments, where possible. When not possible (eg:
+    switching from a 60/40 distribution to a 40/60 distribution),
+    this will minimize changing treatments (in the above case, only
+    those buckets between the 40th and 60th percentile of the bucketing
+    range will see a change in treatment).
+    """
+
+    def __init__(self, variants, num_buckets=1000):
+        """ :param variants -- array of dicts, each containing the keys 'name'
```
pacejackson: Formatting and param type.

baseplate/experiments/providers/variant_sets/single_variant_set.py
```diff
@@ -0,0 +1,67 @@
...
+        total_size = variants[0].get('size') + variants[1].get('size')
+
+        if total_size < 0.0 or total_size > 1.0:
+            raise ValueError('Sum of all variants must be between 0 and 1.')
+
+    def choose_variant(self, bucket):
+        """Deterministically choose a variant. Every call with the same bucket
+        on one instance will result in the same answer
+
+        :param bucket -- an integer bucket representation
```
pacejackson: Formatting and param type.

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -0,0 +1,356 @@
...
+            bucket2 = experiment_version_2._calculate_bucket(fullname)
+            # check that the bucketing changed at some point. Can't compare
+            # bucket1 to bucket2 inline because sometimes the user will fall
+            # into both buckets, and test will fail. When a user doesn't match,
+            # break out of loop
+            if bucket1 != bucket2:
+                bucketing_changed = True
+                break
+
+        self.assertTrue(bucketing_changed)
```
pacejackson: nit: trailing newline

baseplate/experiments/providers/variant_sets/multi_variant_set.py
Kaitaan: I like the latter approach. Calling validate first is a remnant of the older setup, where there was some processing on the variants themselves.

baseplate/experiments/providers/__init__.py
```diff
@@ -55,28 +70,31 @@ def parse_experiment(config):
...
-    experiment_type = config["type"].lower()
-    experiment_id = config["id"]
-    assert isinstance(experiment_id, int)
-    name = config["name"]
+    experiment_type = config.get("type")
+    if experiment_type:
+        experiment_type = experiment_type.lower()
+    experiment_id = config.get("id")
+    if not isinstance(experiment_id, int):
+        raise KeyError("Integer id must be provided for experiment.")
```
pacejackson: It seems weird to raise a `KeyError` if experiment_id happens to be a string or something like that, `KeyError` only really makes sense if `"id"` isn't in the config so `experiment_id` would be `None`.

baseplate/experiments/providers/variant_sets/multi_variant_set.py
```diff
@@ -0,0 +1,66 @@
...
+    def _validate_variants(self):
+
+        if self.variants is None:
+            raise ValueError('No variants provided')
+
+        if len(self.variants) < 3:
+            raise ValueError("MultiVariant experiments expect three or "
+                "more variants.")
+
+        total_size = 0.0
```
pacejackson: I think you want to make this an `total_size = 0` so it doesn't get turned into a float.

baseplate/experiments/providers/variant_sets/rollout_variant_set.py
```diff
@@ -0,0 +1,59 @@
...
+        if len(variants) != 1:
+            raise ValueError("Rollout variant only supports one variant.")
+
+        size = variants[0].get('size')
+        if size is None or size < 0.0 or size > 1.0:
+            raise ValueError('Variant size must be between 0 and 1')
+
+    def choose_variant(self, bucket):
+        """Deterministically choose a percentage-based variant. Every call
+        with the same bucket and varaints will result in the same answer.
```
pacejackson: `varaints`

baseplate/experiments/providers/__init__.py
spladug: `TypeError` would be the more appropriate built-in exception

baseplate/experiments/providers/__init__.py
spladug: this doesn't appear fixed in the current diff?

baseplate/experiments/providers/__init__.py
```diff
@@ -55,28 +70,31 @@ def parse_experiment(config):
...
-        raise ValueError(
-            "Invalid config for experiment %s, missing start_ts and/or "
-            "stop_ts." % name
-        )
+    start_ts = config.get("start_ts")
+    stop_ts = config.get("stop_ts")
+    if (start_ts is None or stop_ts is None):
+        if "expires" in config:
+            warn_deprecated(
+                "The 'expires' field is in experiment %s deprecated, you should "
```
spladug: I think that "is" needs to move after the `%s`

baseplate/experiments/providers/feature_flag.py
```diff
@@ -7,7 +7,10 @@
 
 
 class FeatureFlag(R2Experiment):
-    """An experiment with a single variant "active".
+    """.. deprecated:: 0.27
```
spladug: the deprecated thing should go near the bottom of the docstring, not replace the summary line. you can run `make docs` in the vm and navigate to `http://baseplate.local/html` to see what the output of sphinx looks like to make sure it all makes sense.

baseplate/experiments/providers/simple_experiment.py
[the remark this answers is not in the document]
spladug: seems reasonable to pop it up a level if it itself is not a provider

baseplate/experiments/providers/feature_flag.py
Kaitaan: Would it make sense to have it at the top of the docstring, but not replace the summary line? Seems to me the _first_ thing you'd want to know when reading a block of docs is that the class is deprecated?

Kaitaan:
💇‍♂️

Kaitaan:
Don't merge this just yet. I realized I didn't run local dev tests since overhauling. I'll update/delete this comment as soon as I've done so.

PR #188, opened 2018-06-29
187 B3 Headers Support

### LINKED ISSUE #187: Support for `X-B3-*` trace headers.
[B3 propagation](https://github.com/openzipkin/b3-propagation) bas become the standard when using HTTP. Baseplate currently supports different headers for propagating trace data ([Thrift ref](https://github.com/reddit/baseplate/blob/master/baseplate/integration/thrift/__init__.py#L66), [Pyramid/WSGI ref](https://github.com/reddit/baseplate/blob/master/baseplate/integration/pyramid.py#L117).

Supporting both the old version of headers and B3 shouldn't be too difficult.

cshoe:
Add support for B3 trace headers whil continuing to support the old
trace headers. This change aims to allow both both B3 and the old
headers to be used as long as they contain identical values. For
example:

```
X-Trace: 1234567890
X-B3-TraceId: 1234567890
```

Fixes #187

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -92,6 +92,7 @@ class TraceInfo(_TraceInfo):
     collecting the trace context and passing it along to the server span.
 
     """
+
     @classmethod
     def new(cls):
         """Generate IDs for a new initial server span.
@@ -134,6 +135,62 @@ class TraceInfo(_TraceInfo):
                 raise ValueError("invalid flags value")
 
         return cls(trace_id, parent_id, span_id, sampled, flags)
+
+    @classmethod
+    def extract_upstream_header_values(cls, upstream_header_names, headers):
+        """Extract values from upstream headers.
+
+        This method thinks about upstream headers by a general name as oppposed to the header
+        name, i.e. "trace_id" instead of "X-Trace". These general names are "trace_id",
+        "span_id", "parent_span_id", "sampled" and "flags".
+
+        A dict mapping these general names to corresponding header names is expected.
+
+        For example:
+
+            {
+                "trace_id": ("X-Trace", "X-B3-TraceId"),
+                "span_id": ("X-Span", "X-B3-SpanId"),
+                "parent_span_id": ("X-Parent", "X-B3-ParentSpanId"),
+                "sampled": ("X-Sampled", "X-B3-Sampled"),
+                "flags": ("X-Flags", "X-B3-Flags"),
+            }
+
+        This structure is used to extract relevant values from the request headers resulting
+        in a dict mapping general names to values.
+
+        For example:
+
+            {
+                "trace_id": "2391921232992245445",
+                "span_id": "7638783876913511395",
+                "parent_span_id": "3383915029748331832",
+                "sampled": "1",
+            }
+
+        :param dict upstream_headers_name: Map of general upstream value labels to header names
+        :param dict headers: Headers sent with a request
+        :return: Values found in upstream trace headers
+        :rtype: dict
+
+        :raises: :py:exc:`ValueError` if conflicting values are found for the same header category
+
+        """
+        extracted_values = {}
+        for name, header_names in upstream_header_names.items():
+            values = []
+            for header_name in header_names:
+                if header_name in headers:
+                    values.append(headers[header_name])
+
+            if not values:
+                continue
+            elif not all(value == values[0] for value in values):
+                raise ValueError("Conflicting values found for %s header(s)".format(header_names))
+            else:
+                # All the values are the same
+                extracted_values[name] = values[0]
+        return extracted_values
 
 
 class AuthenticationTokenValidator(object):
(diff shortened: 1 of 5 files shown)


cshoe:
👓 @spladug @pacejackson

baseplate/integration/trace_headers.py
```diff
@@ -0,0 +1,122 @@
...
+LOWER_HEX_16_BYTES = re.compile(b"\A[0-9a-f]{16}\Z")
+LOWER_HEX_32_BYTES = re.compile(b"\A[0-9a-f]{32}\Z")
+
+
+class TraceHTTPHeaders(object):
+    """Helper class for trace headers in an HTTP call.
+
+    """
+    HEADER_NAMES = {
+        "trace_id": ["X-Trace", "X-B3-TraceId"],
```
pacejackson: Should we use tuples rather than lists for these?  Tuples are a little smaller and immutable.

pacejackson: Also, the way these are ordered will favor the old header format over the new, maybe we should put the new one first?

baseplate/integration/trace_headers.py
```diff
@@ -0,0 +1,122 @@
...
+
+LOWER_HEX_16 = re.compile(r"\A[0-9a-f]{16}\Z")
+LOWER_HEX_32 = re.compile(r"\A[0-9a-f]{32}\Z")
+
+LOWER_HEX_16_BYTES = re.compile(b"\A[0-9a-f]{16}\Z")
+LOWER_HEX_32_BYTES = re.compile(b"\A[0-9a-f]{32}\Z")
+
+
+class TraceHTTPHeaders(object):
+    """Helper class for trace headers in an HTTP call.
```
pacejackson: If this is it for the docstring, I think you can fit the closing `"""` on the same line.

cshoe:
💇 

@spladug @pacejackson Thanks for the feedback. My first approach was _way_ too heavy handed and over engineered. I tried to greatly simplify things and focus solely on getting Baseplate to look at B3 headers in addition to those it already supported.

baseplate/integration/pyramid.py
```diff
@@ -116,20 +116,7 @@ def _on_new_request(self, event):
...
-                    flags = int(flags)
-                trace_info = TraceInfo.from_upstream(
-                    trace_id=int(request.headers["X-Trace"]),
-                    parent_id=int(request.headers["X-Parent"]),
-                    span_id=int(request.headers["X-Span"]),
-                    sampled=sampled,
-                    flags=flags,
-                )
-
+                trace_info = TraceInfo.from_http_headers(request.headers)
```
spladug: It's kinda a bummer to make `TraceInfo` have to be aware of the different protocols we're using; if we added more down the line that'd mean more direct modifications in that class rather than just adding the new protocol definitions. The core stuff is meant to be the idealized form of the tracing and then these integrations are where that meets the actual wire. The centralized extraction logic could potentially be exposed as a helper that these integrations use though?

cshoe:
💇

baseplate/core.py
```diff
@@ -135,6 +136,62 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
 
         return cls(trace_id, parent_id, span_id, sampled, flags)
 
+    @classmethod
+    def extract_upstream_header_values(cls, upstream_header_names, headers):
```
spladug: Would it make sense just to merge this into `from_upstream` ? I don't think we'd ever use one without the other?

baseplate/core.py
```diff
@@ -135,6 +136,62 @@ def from_upstream(cls, trace_id, parent_id, span_id, sampled, flags):
...
+        for name, header_names in upstream_header_names.items():
+            values = []
+            for header_name in header_names:
+                if header_name in headers:
+                    values.append(headers[header_name])
+
+            if not values:
+                continue
+            elif not all(value == values[0] for value in values):
+                raise ValueError("Conflicting values found for %s header(s)".format(header_names))
```
spladug: do we want to crash the whole request if the headers are iffy? (i could see either answer being reasonable)

baseplate/core.py
cshoe: Yeah... I wanted to do that but there are subtle differences in how the two protocols handle the header values due to Thrift operating in bytes and HTTP using strings.

https://github.com/reddit/baseplate/pull/188/files#diff-db4cf04c08224f72cb9627e1f9866bbcR121
https://github.com/reddit/baseplate/pull/188/files#diff-d1382533c32ca95376190e31679acc05R162

Some type checking could be done. Or possibly a generic

```
True if extraced_values["sampled"] in ("1", b"1")
```

but this started to feel protocol specific again. If others are added, other protocol specific edge cases could arise making the ``from_upstream`` method kind of gnarly. I'm happy to take a stab at it though, if you'd like.

baseplate/core.py
cshoe: I raised this to follow what is being done in ``new``. Where this is called is [wrapped](https://github.com/reddit/baseplate/pull/188/files#diff-d1382533c32ca95376190e31679acc05R138) so the request doesn't get nuked but rather the trace data is not propagated.

PR #190, opened 2018-07-17
Fix typo where we were calling 'increment' on the 'counter' function, …

pacejackson:
…not a 'Counter' object


pacejackson:
_Probably_...

pacejackson:
💇‍♂️ added a test

PR #191, opened 2018-07-18
Stop error on bad config

Kaitaan:
Make it so that if a bad config is pushed, the parse_experiment call won't throw an exception to the caller.

### DIFF
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -168,7 +168,11 @@ class Experiments(object):
         :return: Variant name if a variant is active, None otherwise.
         """
 
-        experiment = self._get_experiment(name)
+        try:
+            experiment = self._get_experiment(name)
+        except (ValueError, TypeError) as err:
+            logger.error("Invalid configiguration for experiment {}: {}".format(name, err))
+            return None
 
         if experiment is None:
             return None
@@ -228,8 +232,11 @@ class Experiments(object):
         :param kwargs: Additional arguments that will be passed to logger.
 
         """
-
-        experiment = self._get_experiment(experiment_name)
+        try:
+            experiment = self._get_experiment(name)
+        except (ValueError, TypeError) as err:
+            logger.error("Invalid configiguration for experiment {}: {}".format(name, err))
+            return None
 
         if experiment is None:
             return
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -93,7 +93,11 @@ class Experiments(object):
             if not experiment_config:
                 experiment = None
             else:
-                experiment = parse_experiment(experiment_config)
+                try:
+                    experiment = parse_experiment(experiment_config)
+                except (ValueError, TypeError, KeyError) as err:
+                    logger.error("Invalid configuration for experiment {}: {}".format(name, err))
+                    return None
             self._experiment_cache[name] = experiment
         return self._experiment_cache[name]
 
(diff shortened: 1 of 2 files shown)


baseplate/experiments/__init__.py
```diff
@@ -93,7 +93,11 @@ def _get_experiment(self, name):
             if not experiment_config:
                 experiment = None
             else:
-                experiment = parse_experiment(experiment_config)
+                try:
+                    experiment = parse_experiment(experiment_config)
+                except (ValueError, TypeError, KeyError) as err:
```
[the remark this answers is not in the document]
Kaitaan: Yeah, I had considered that, but thought it was "bad form" in python.

@mattknox, DDG is a baseplate service, right? Can it leverage itself to validate configs before shipping them up to ZK?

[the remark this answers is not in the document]
pacejackson: Eh, it's ok when appropriate, what you really don't want to do is just `except:` since that catches things like `ctrl-c`.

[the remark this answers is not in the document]
mattknox: Sure, it will do so (doesn't write to ZK at all yet), but that doesn't guarantee that this will never see bad config.

Kaitaan:
💇‍♂️

tests/unit/experiments/experiment_tests.py
```diff
@@ -529,6 +529,88 @@ def test_that_bucketing_events_not_sent_if_cant_find_experiment(self):
...
+            server_span=self.mock_span,
+            context_name="test",
+            event_logger=self.event_logger,
+        )
+        self.assertEqual(self.event_logger.log.call_count, 0)
+        variant = experiments.variant("test", user=self.user)
+        self.assertEqual(variant, None)
+
+    def test_none_returned_on_variant_call_with_no_experiment(self):
+        self.mock_filewatcher.get_data.return_value = {
```
mattknox: This is intended to test that a call to an unconfigured test doesn't raise, right?

tests/unit/experiments/experiment_tests.py
```diff
@@ -529,6 +529,88 @@ def test_that_bucketing_events_not_sent_if_cant_find_experiment(self):
         experiments.variant("test", user=self.user)
         self.assertEqual(self.event_logger.log.call_count, 0)
 
+    def test_none_returned_on_variant_call_with_bad_id(self):
+        self.mock_filewatcher.get_data.return_value = {
+            "test": {
```
mattknox: what is bad about the id here?  I can't tell.

Kaitaan: it's a string rather than an int

tests/unit/experiments/experiment_tests.py
Kaitaan: yep. If it raised, this would die. Since we're just catching all exceptions now, these tests are a bit redundant, but doesn't hurt to have extra tests.

tests/unit/experiments/experiment_tests.py
mattknox: ah!

Kaitaan:
> This experiment_cache thing gets populated lazily as we call variant() with various experiment names? That is an odd choice.

It may be a bit cleaner to pre-populate it so we can handle all the errors in one shot, but I don't think this is inherently a bad approach. Either way, changing it is outside the scope of current work.

pacejackson:
> It may be a bit cleaner to pre-populate 

Cleaner, maybe, but I don't think it's the right approach, you'll be wasting time on every request parsing dozens of experiment configs that you don't need.  The cache was just added to prevent re-parsing the same config if the same experiment is checked multiple times in the same request.

PR #194, opened 2018-08-02
add targeting for SimpleExperiment

Kaitaan:
Add advanced targeting capabilities to SimpleExperiments.

### DIFF
--- a/baseplate/experiments/providers/simple_experiment.py
+++ b/baseplate/experiments/providers/simple_experiment.py
@@ -14,6 +14,9 @@ from ..variant_sets.single_variant_set import SingleVariantSet
 from ..variant_sets.multi_variant_set import MultiVariantSet
 from ..variant_sets.rollout_variant_set import RolloutVariantSet
 
+from ..targeting.tree_targeting import create_targeting_tree
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -22,6 +25,22 @@ variant_type_map = {
     'multi_variant': MultiVariantSet,
     'feature_rollout': RolloutVariantSet,
 }
+
+
+def _generate_targeting(targeting_config):
+    """Generate the targeting tree for this experiment.
+
+    If no config is provided, then assume we want to target all. If an invalid
+    config is provided, then target none.
+    """
+    if targeting_config is None:
+        return create_targeting_tree({'OVERRIDE': True})
+
+    try:
+        return create_targeting_tree(targeting_config)
+    except Exception:
+        logger.error("Unable to create targeting tree. No targeting applied.")
+        return create_targeting_tree({'OVERRIDE': False})
 
 
 class SimpleExperiment(Experiment):
@@ -34,8 +53,8 @@ class SimpleExperiment(Experiment):
 
     def __init__(self, id, name, owner, start_ts, stop_ts, config,
                  experiment_version, shuffle_version, variant_set,
-                 bucket_seed, bucket_val, enabled=True, log_bucketing=True,
-                 num_buckets=1000):
+                 bucket_seed, bucket_val, targeting, enabled=True,
+                 log_bucketing=True, num_buckets=1000):
         """
         :param int id: The experiment id. This should be unique.
         :param string name: The human-readable name of the experiment.
@@ -87,6 +106,8 @@ class SimpleExperiment(Experiment):
 
         self._log_bucketing = log_bucketing
 
+        self._targeting = targeting
+
         if not self.experiment_version:
             raise ValueError('Experiment version must be provided.')
 
@@ -117,6 +138,9 @@ class SimpleExperiment(Experiment):
 
         bucket_seed = config.get("bucket_seed")
 
+        targeting_config = config.get("targeting")
+        targeting = _generate_targeting(targeting_config)
+
         return cls(
             id=id,
             name=name,
@@ -132,6 +156,7 @@ class SimpleExperiment(Experiment):
             bucket_val=bucket_val,
             num_buckets=num_buckets,
             log_bucketing=log_bucketing,
+            targeting=targeting,
         )
 
     @property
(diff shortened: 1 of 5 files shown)


baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -0,0 +1,192 @@
...
+
+        for node in input_node:
+            self._children.append(create_targeting_tree(node))
+
+    def evaluate(self, **kwargs):
+        for node in self._children:
+            if not node.evaluate(**kwargs):
+                return False
+
+        return True
```
pacejackson: python actually has an `all` method, you can do this with `return all(node.evaluate(**kwargs) for node in self._children)`

baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -0,0 +1,192 @@
...
+
+        for node in input_node:
+            self._children.append(create_targeting_tree(node))
+
+    def evaluate(self, **kwargs):
+        for node in self._children:
+            if node.evaluate(**kwargs):
+                return True
+
+        return False
```
pacejackson: Like `all`, python has an `any` method, you can do this with `return any(node.evaluate(**kwargs) for node in self._children)`

baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -0,0 +1,192 @@
...
+    2. the must not have commented
+    3. they can be logged in or logged out
+    4. they must have cast 5 votes
+
+    """
+    if not len(input_node) == 1:
+        raise ValueError("Call to create_targeting_tree expects a single input key.")
+
+    operator = list(input_node.keys())[0]
+    input_node_value = input_node[operator]
```
pacejackson: You could do this in one line as `operator, input_node_value = list(input_node.values())[0]`.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -24,6 +26,23 @@
...
+
+    If no config is provided, then assume we want to target all. If an invalid
+    config is provided, then target none.
+    """
+    if targeting_config is None:
+        targeting_tree = create_targeting_tree({'OVERRIDE': True})
+    else:
+        try:
+            targeting_tree = create_targeting_tree(targeting_config)
+        except (TypeError, ValueError):
```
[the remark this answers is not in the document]
pacejackson: I.E., `create_targeting_tree` should catch the `TypeError` and raise a custom exception. You also _might_ want to have it raise a custom exception rather than `ValueError` but I think that's less of a concern than the hidden `TypeError`s.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -24,6 +26,23 @@
...
 
 
+def _generate_targeting(targeting_config):
+    """Generate the targeting tree for this experiment.
+
+    If no config is provided, then assume we want to target all. If an invalid
+    config is provided, then target none.
+    """
+    if targeting_config is None:
+        targeting_tree = create_targeting_tree({'OVERRIDE': True})
```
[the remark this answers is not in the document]
Kaitaan: No, if no targeting config is specified, we will Target everything.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -24,6 +26,23 @@
...
+    else:
+        try:
+            targeting_tree = create_targeting_tree(targeting_config)
+        except (TypeError, ValueError):
+            targeting_tree = create_targeting_tree({'OVERRIDE': False})
```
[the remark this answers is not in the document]
Kaitaan: To expand on the above, the logic here is that if no experiment targeting is provided, target everything. If you don't provide one at all, you're not "targeting nothing", you're not targeting at all. An untargeted experiment is one that can hit everything.

The other side of that coin is an invalid targeting config. If you _try_ to provide a config, but there's a problem, you probably don't want to fall back to targeting everything. Instead, target nothing. In theory, whatever is providing the configs would check for and prevent invalid configurations, but they could slip through. In that case, you don't want to risk over-exposing.

baseplate/experiments/targeting/tree_targeting.py
Kaitaan: Doesn't `list(input_node.values())[0]` just return the first _value_? I need both the key and value of the first (only) item in the dict.

pacejackson: Oops, yes that is exactly what I meant.

Kaitaan:
💇‍♂️

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -354,3 +377,45 @@ def test_change_shuffle_version_changes_bucketing(self):
...
+
+        experiment_with_targeting = parse_experiment(cfg)
+
+        experiment_inputs = {
+            'is_mod':True,
+            'is_logged_in':True,
+            'random_numeric':5,
+        }
+
+        self.assertTrue(experiment_with_targeting.is_targeted(**experiment_inputs))
```
pacejackson: 🔕 maybe style nit:  This seems simple enough to just set the kwargs directly rather than using a dict and the `**` syntax.  I feel like that makes it more clear what the call args actually are.  If other's feel differently or don't feel strongly about it, then this is fine too.

baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -0,0 +1,197 @@
+from .base import Targeting
```
spladug: nit: please include the boilerplate `from __future__` things on all new files so we can keep the py2/py3 differences minimal

tests/unit/experiments/providers/simple_experiment_tests.py
spladug: yes please

baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -0,0 +1,197 @@
...
+        {
+            field: <field_name>,
+            value: <accepted_value>
+        }
+        """
+
+        if len(input_node) != 2:
+            raise ValueError("EqualNode expects exactly two fields.")
+
+        if not 'field' in input_node:
```
spladug: nit: it reads a little nicer to say `if 'field' not in input_mode`

PR #195, opened 2018-08-14
Automate PyPI Release

cshoe:
These creds are just for demonstration purposes. If there is an
organizational PyPI account to use or if one of the listed Baseplate
maintainers wants to supply theirs, let's talk!

The hope here is to do smaller incremental releases and publish them to
PyPI. This would allow Reddit projects that are using wheels to make use
of new functionality and bug fixes before doing a fleetwide release.

@spladug, any help you could lend with the credentials would be much
appreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,3 +84,18 @@ script:
     - sphinx-build -M doctest docs/ build/
     - sphinx-build -M spelling docs/ build/
     - make lint
+
+jobs:
+  include:
+    - stage: PyPI Release
+      python: 3.4
+      script: echo "Deploying to PyPI"
+      deploy:
+        provider: pypi
+        install: skip
+        user: reddit
+        password:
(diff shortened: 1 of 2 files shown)


spladug:
Very cool! Although publishing to public PyPI will not enable faster rollouts internally immediately because we don't use public PyPI for production (for security/reproducibility). But this is a great step in reducing the complexity of releases.

Let's get some creds set up just for CI rather than using any individual's account. That's more bus-proof and locked-down. We can then remove all the humans from the baseplate project on PyPI.

Looking at the travis docs, it looks like we might want to do [this step to add bdist into the mix](https://docs.travis-ci.com/user/deployment/pypi/#uploading-different-distributions). If I understand right, this would do the build of the thrift manifests before upload and make it so that installing with pip doesn't require a thrift compiler.

Thank you!

:nail_care:

cshoe:
> make it so that installing with pip doesn't require a thrift compiler.

Ah! This was one of my goals.

cshoe:
💇

spladug:
Cool! Just to confirm: has this been tested to work? It might make sense to fire up a second project on PyPI, take the "on: tags" bit off, and test it there to see that the build works and the resultant wheel looks cromulent.

spladug:
:nail_care:

cshoe:
Quite a bit of flailing here... I was trying to get Travis to work with the test version of PyPI but didn't have any luck. I resulted to temporarily naming the package to ``test-baseplate``.

On a clean VM, I can ``pip install test-baseplate``. It installs just fine but opening an interpreter and attempting to ``import baseplate`` fails with an import error related to Thrift. Installing the ``python-fbthrfit`` system package fixes this. I'm honestly not sure how to handle that dependency since fbthrift doesn't seem to be installable via pip. The standard ``thrift`` package isn't a suitable stand-in as it's layout is different (``fbthrift`` contains ``thrift.util`` which is referenced in ``baseplate.core``).

💇

PR #197, opened 2018-08-16
Catch exceptions that happen outside server span context

spladug:
This can occur when the framework surrounding application code crashes.
For example, when Thrift tries to serialize invalid data like an
1,000,000 in an I16. Without this error reporting it's very annoying to
debug what's going on.

By hooking into the gevent hub, we're tied to using gevent but we
already expect that in Baseplate. This should catch all unhandled
exceptions and give us visibility into them.

cc: @bsimpson63

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -692,7 +692,15 @@ class Baseplate(object):
         :param raven.Client client: A configured raven client.
 
         """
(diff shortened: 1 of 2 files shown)


bsimpson63:
cool

PR #201, opened 2018-10-02
Kaitaan add overrides

Kaitaan:
@pacejackson @mattknox @spladug 

I think this should be the last major experiments addition coming into baseplate for the near future.

### DIFF
--- a/baseplate/experiments/providers/simple_experiment.py
+++ b/baseplate/experiments/providers/simple_experiment.py
@@ -53,8 +53,8 @@ class SimpleExperiment(Experiment):
 
     def __init__(self, id, name, owner, start_ts, stop_ts, config,
                  experiment_version, shuffle_version, variant_set,
-                 bucket_seed, bucket_val, targeting, enabled=True,
-                 log_bucketing=True, num_buckets=1000):
+                 bucket_seed, bucket_val, targeting, overrides, 
+                 enabled=True, log_bucketing=True, num_buckets=1000):
         """
         :param int id: The experiment id. This should be unique.
         :param string name: The human-readable name of the experiment.
@@ -107,6 +107,7 @@ class SimpleExperiment(Experiment):
         self._log_bucketing = log_bucketing
 
         self._targeting = targeting
+        self._overrides = overrides
 
         if not self.experiment_version:
             raise ValueError('Experiment version must be provided.')
@@ -141,6 +142,13 @@ class SimpleExperiment(Experiment):
         targeting_config = config.get("targeting")
         targeting = _generate_targeting(targeting_config)
 
+        overrides = None
+        override_inputs = config.get("overrides")
+        if override_inputs and isinstance(override_inputs, dict):
+            overrides = {}
+            for variant, target in override_inputs:
+                overrides[variant] = _generate_targeting(target)
+
         return cls(
             id=id,
             name=name,
@@ -157,6 +165,7 @@ class SimpleExperiment(Experiment):
             num_buckets=num_buckets,
             log_bucketing=log_bucketing,
             targeting=targeting,
+            overrides=overrides,
         )
 
     @property
@@ -182,6 +191,13 @@ class SimpleExperiment(Experiment):
         """
         return self._targeting.evaluate(**kwargs)
 
+    def is_overridden(self, **kwargs):
+        for variant, targeting in self._overrides:
+            if targeting.evaluate(**kwargs):
+                return variant
+
+        return None
+
     def variant(self, **kwargs):
         if not self._is_enabled():
             return None
@@ -205,6 +221,10 @@ class SimpleExperiment(Experiment):
                 self.name,
             )
             return None
+
+        override = is_overridden(**kwargs):
+        if override:
+            return override
 
         if not self.is_targeted(**kwargs):
             return None
--- a/baseplate/experiments/providers/simple_experiment.py
+++ b/baseplate/experiments/providers/simple_experiment.py
@@ -39,14 +39,8 @@ def _generate_overrides(override_config):
     across targeting trees. Each incoming variant request will check each
     targeting override in sequence until one is matched or none match.
 
-    If the override configuration does not contain a wrapping list (ie, it
-    is just a dictionary), then overrides will work, but override ordering
-    is not guaranteed. This means that if there is overlap between overrides
-    (ie, a particular user is eligible for two overrides), then the results
-    may not be consistent.
-
     In the event of an invalid override:
-    1) if a dict of variant : target is not provided, that list item will be
+    1) if a dict of {variant: target} is not provided, that list item will be
         ignored. All other overrides will still remain active.
     2) if the targeting object is invalid, the override will remain active,
         but will return false for all targeting
@@ -60,9 +54,8 @@ def _generate_overrides(override_config):
         return None
 
     if not isinstance(override_config, list):
-        if override_config:
-            logger.error("Invalid override configuration. Skipping overrides.")
-            return None
+        logger.error("Invalid override configuration. Skipping overrides.")
+        return None
 
     for override_entry in override_config:
         if not isinstance(override_entry, dict) or len(override_entry) > 1:
--- a/baseplate/experiments/providers/simple_experiment.py
+++ b/baseplate/experiments/providers/simple_experiment.py
@@ -27,6 +27,55 @@ variant_type_map = {
 }
 
 
+def _generate_overrides(override_config):
+    """Generate a dictionary of overrides.
+
+    The format of the override config is expected to be a list of dicts,
+    where each dict contains a single key (the treatment to assign) and value
+    (the targeting tree for that treatment).
+
+    This combination is used to ensure that the ordering of the overrides is
+    maintained, allowing consistent overriding for any overlapping area
+    across targeting trees. Each incoming variant request will check each
+    targeting override in sequence until one is matched or none match.
+
+    If the override configuration does not contain a wrapping list (ie, it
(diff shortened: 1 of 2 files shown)


baseplate/experiments/providers/simple_experiment.py
```diff
@@ -27,6 +27,55 @@
...
+    where each dict contains a single key (the treatment to assign) and value
+    (the targeting tree for that treatment).
+
+    This combination is used to ensure that the ordering of the overrides is
+    maintained, allowing consistent overriding for any overlapping area
+    across targeting trees. Each incoming variant request will check each
+    targeting override in sequence until one is matched or none match.
+
+    If the override configuration does not contain a wrapping list (ie, it
+    is just a dictionary), then overrides will work, but override ordering
```
pacejackson: Why even give the option of this?  Couldn't we just say "you have to give a list of dicts"?

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -27,6 +27,55 @@
...
+        but will return false for all targeting
+
+    Multiple targets can be provided for a single variant.
+    """
+
+    override_list = []
+    if override_config and isinstance(override_config, dict):
+        override_config = [override_config]
+
+    if override_config and isinstance(override_config, list):
```
pacejackson: You might want to log a warning if you don't enter this if block since that means someone is giving you an invalid config.

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -27,6 +27,55 @@
...
+        for override_entry in override_config:
+            if not isinstance(override_entry, dict):
+                logger.error("Invalid override configuration. Not applying override."
+                    "Expected dictionary: {}".format(override_entry))
+                continue
+
+            if len(override_entry) > 1:
+                logger.warning("Multiple overrides in a single override entry. "
+                    "Execution order of overrides may not be consistent.")
```
pacejackson: I would say we should just mandate this and skip any that are invalid.

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -63,14 +63,14 @@ def get_simple_config():
...
+                    "size": 0.1,
                 },
                 {
                     "name":"variant_2",
-                    "size":0.1,
+                    "size": 0.1,
                 },
             ],
-            "experiment_version":1,
+            "experiment_version": 1,
```
pacejackson: 🔕 Interesting that this slipped through, I would figure that we had a lint check in CI for this?  Appreciate the clean up!

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -103,6 +103,33 @@ def get_targeting_config():
...
+        }
+    ]
+
+    return override_config
+
+def get_dict_override_config():
+    override_config = {
+        'override_variant_1':{'EQ': {'field':'user_id', 'value':'t2_1'}},
+        'override_variant_2':{'EQ': {'field':'user_id', 'value':'t2_2'}},
+        'override_variant_3':{'EQ': {'field':'user_id', 'values':['t2_2', 't2_3']}},
```
pacejackson: I feel like we shouldn't allow duplicate `key=value` overrides, that sounds like it would get very confusing when someone thinks they added an override but aren't seeing it show up because they have another override earlier in the config that overrides their new one.  I'm also not sure what the benefit of this is?  In this case, `user_id=t2_2` is always going to return `override_variant_2`, so why even let them set it as another override value?

tests/unit/experiments/providers/simple_experiment_tests.py
```diff
@@ -417,3 +444,148 @@ def test_targeting_in_config(self):
...
+            ('override_variant_3', EqualNode),
+            ('override_variant_1', EqualNode),
+        ]
+
+        for i in range(4):
+            self.assertEqual(len(overrides[i]),1)
+            self.assertTrue(isinstance(
+                overrides[i][override_names_and_types[i][0]],
+                override_names_and_types[i][1]
+            ))
```
pacejackson: I think it would be simpler to do this as:

```
for i, override in enumerate(overrides):
    self.assertEqual(len(override), 1)
    variant_name, override_type = override_names_and_types[i]
    self.assertTrue(isinstance(override[variant_name], override_type)
```

or something like that

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: This example is meant to be pretty trivial; think of it instead as the override target contains a set of things based on the input parameters. In this case, the sets are "user_id=t2_2" and "user_id in [t2_2, t2_3]", but the actual sets provided can be largely arbitrary. It could be "user_id is t2_2" and "user is moderator". Determining the overlap there is much harder.
It could even be "user is supertype" and "user is subtype"; without having the type context calculating the overlapping segment is impossible.
I don't think it's worth including a bunch of complicated code to try to calculate overlapping sets here when allowing the "fall through" behaviour is cleaner, simpler, and not application-specific.
_If_ a client wants to disallow overlapping segments, it should be done when creating the configuration with the contextual knowledge of whatever user sets the client has.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: We could; I wanted to give the option of simpler configs if you're not going to have any overriding sets (ie, overriding specific users, for instance). I'm happy to remove it, but I generally like a bit more flexibility when it doesn't greatly increase complexity.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: Good catch. Will do.

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: yeah, that surprised me too. Not sure why this slipped through. Interestingly, I accidentally fixed it because I had the wrong file open, and linting did find that same error in another file.

tests/unit/experiments/providers/simple_experiment_tests.py
Kaitaan: TIL about `enumerate`. Will update.

baseplate/experiments/providers/simple_experiment.py
pacejackson: I imagine most if not all of these configs are going to be generated by code rather than handwritten so the little bit of configuration overhead probably isn't a big deal.  While the code is straight forward, I feel like having forked behavior like this can lead to unexpected results (not bugs, just someone makes a config that doesn't do what they expect) and confusion about what exactly is the "right" thin to do.

tests/unit/experiments/providers/simple_experiment_tests.py
pacejackson: > I don't think it's worth including a bunch of complicated code to try to calculate overlapping sets

That's fair, I didn't consider that

baseplate/experiments/providers/simple_experiment.py
Kaitaan: fair point. Fixed

Kaitaan:
💇‍♂️

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -27,6 +27,54 @@
...
+        ignored. All other overrides will still remain active.
+    2) if the targeting object is invalid, the override will remain active,
+        but will return false for all targeting
+
+    Multiple targets can be provided for a single variant.
+    """
+
+    override_list = []
+
+    if not isinstance(override_config, list):
```
spladug: these ifs are a bit repetitive/commingled. how about:

```python
if not override_config:
    return

if not isinstance(override_config, list):
    logger.error(...)
    return None

for override_entry in override_config:
    ...
```

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -177,11 +230,22 @@ def should_log_bucketing(self):
         return self._log_bucketing
 
     def is_targeted(self, **kwargs):
-        """Determine whether the provided kwargs match targeting parameters
+        """ Determine whether the provided kwargs match targeting parameters
```
spladug: ?

baseplate/experiments/providers/simple_experiment.py
Kaitaan: ah, good catch. Fixed.

Kaitaan:
💇‍♂️

baseplate/experiments/providers/simple_experiment.py
```diff
@@ -27,6 +27,56 @@
...
+    Multiple targets can be provided for a single variant.
+    """
+
+    override_list = []
+
+    if not override_config:
+        return None
+
+    if not isinstance(override_config, list):
+        if override_config:
```
spladug: this if isn't necessary anymore since we're already bailing earlier on now.

baseplate/experiments/providers/simple_experiment.py
pacejackson: I think this docstring still needs to be updated to reflect that change.

baseplate/experiments/providers/simple_experiment.py
Kaitaan: removed

baseplate/experiments/providers/simple_experiment.py
Kaitaan: updated

Kaitaan:
💇‍♂️

PR #205, opened 2018-10-12
Blocking file watcher

### LINKED ISSUE #204: Resolve race conditions caused by use of sidecars
The secret fetcher sidecar is depended on by the live data sidecar which is depended on by the app. If things start up in the wrong order, they can get rather angsty. It's difficult to solve this in an init/scheduling system unless we have it pay attention to the actual generated files (which can be many). Instead, we should get the relevant apps to block or retry intelligently on sidecar-provided resources.

spladug:
This should fix #204

### DIFF
--- a/baseplate/file_watcher.py
+++ b/baseplate/file_watcher.py
@@ -37,6 +37,8 @@ from __future__ import unicode_literals
 import logging
 import os
 
+from baseplate.retry import RetryPolicy
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


baseplate/file_watcher.py
```diff
@@ -61,14 +63,34 @@ class FileWatcher(object):
     :param callable parser: A callable that takes an open file object, parses
         or otherwise interprets the file, and returns whatever data is
         meaningful.
+    :param float timeout: How long, in seconds, to block instantiation waiting
+        for the watched file to become available (defaults to not blocking).
 
     """
-    def __init__(self, path, parser):
+    def __init__(self, path, parser, timeout=None):
```
pacejackson: Should this be a `Timedelta`? That seems like a somewhat common pattern in baseplate.

baseplate/file_watcher.py
```diff
@@ -61,14 +63,34 @@ class FileWatcher(object):
...
+        if timeout is not None:
+            last_error = None
+            for _ in RetryPolicy.new(budget=timeout, backoff=0.01):
+                if self._data is not _NOT_LOADED:
+                    break
+
+                try:
+                    self.get_data()
+                except WatchedFileNotAvailableError as exc:
+                    last_error = exc
```
pacejackson: ~Should there be a short sleep here so it doesn’t just spin on this?~

OHHH, wait, the back off in RetryPolicy does this doesn’t it?

baseplate/file_watcher.py
pacejackson: Also, then it’s easy to just use the timespan config option for this rather than passing it to the `from_config` call

baseplate/file_watcher.py
spladug: yup! :grin:

baseplate/file_watcher.py
spladug: The reason I put it on the `from_config` call instead of in config itself is that I think it's the sort of thing the programmer will know at the call-site rather than the application configuration. You know you want a specific line of code to block so that subsequent lines can safely trust the data's there; a config file doesn't really have the same context.

PR #206, opened 2018-10-19
Move to Apache Thrift

spladug:
This replaces FBThrift with Apache Thrift. This is possible now that
THeaderProtocol is available in Apache Thrift. This is a breaking change
that will require modification to services built on Baseplate, but the
wire protocols should be identical so it should be possible to upgrade
piece by piece.

As an example of what a service would need to do to upgrade, this
is the diff for the [activity service](https://github.com/reddit-archive/reddit-service-activity).

```diff
diff --git a/reddit_service_activity/__init__.py b/reddit_service_activity/__init__.py
index dd8c607..e1acf20 100644
--- a/reddit_service_activity/__init__.py
+++ b/reddit_service_activity/__init__.py
@@ -14,7 +14,7 @@ from baseplate import (
     tracing_client_from_config,
 )
 from baseplate.context.redis import RedisContextFactory
-from baseplate.integration.thrift import BaseplateProcessorEventHandler
+from baseplate.integration.thrift import baseplateify_processor
 
 from .activity_thrift import ActivityService, ttypes
 from .counter import ActivityCounter
@@ -48,7 +48,7 @@ class ActivityInfo(ttypes.ActivityInfo):
         )
 
 
-class Handler(ActivityService.ContextIface):
+class Handler(ActivityService.Iface):
     def __init__(self, counter):
         self.counter = counter
 
@@ -142,8 +142,5 @@ def make_processor(app_config):  # pragma: nocover
 
     counter = ActivityCounter(cfg.activity.window.total_seconds())
     handler = Handler(counter=counter)
-    processor = ActivityService.ContextProcessor(handler)
-    event_handler = BaseplateProcessorEventHandler(logger, baseplate)
-    processor.setEventHandler(event_handler)
-
-    return processor
+    processor = ActivityService.Processor(handler)
+    return baseplateify_processor(processor, logger, baseplate)
```

Applications that use Thrift directly would need a little more work (~some things got renamed; the `Serializer` util isn't provided in Apache thrift, so Baseplate now provides it, and Baseplate's thrift pool got moved to be more consistent~), but over all it should be a pretty mechanical upgrade process.

~For reddit production purposes, we'll need to package Apache Thrift's compiler and Python library off the current upstream HEAD and then use that instead of fbthrift.~ This is done now (see: https://github.com/reddit/thrift)

There are a few things about this PR I don't love. I'll comment on them to call them out and would love suggestions of alternative approaches.

~Tests will be failing in CI right now, unfortunately, since it'll be trying to use FBThrift. I'll try and get some packages ready so we can try it out properly, but until then either trust me that all the checks are passing or pull it down locally and try :)~ All should be passing now.


baseplate/integration/thrift/command.py
```diff
@@ -55,6 +55,20 @@ def run(self):
 
                 self.copy_tree(input_package, output_package)
 
+                # rewrite "from thriftname" to "from package.thriftname_thrift"
```
spladug: i really really don't love this. the generated "remote" script (simple client for local CLI testing) uses absolute imports but because of our shenanigans and the fact that we nest inside the application package, those references are incorrect. this is a horrid hack to try and rewrite the import. :see_no_evil:

Makefile
```diff
@@ -1,7 +1,7 @@
 all: thrift
 
-THRIFT=thrift1
-THRIFT_OPTS=-strict -gen py:utf8strings,slots,new_style
+THRIFT=thrift
+THRIFT_OPTS=-strict -gen py:slots
```
spladug: note: utf8strings and new_style are now the default in more modern thrift compilers and we get a warning about the option going away if we try to set them explicitly

spladug:
:eyeglasses: @dellis23 @bsimpson63 @ckwang8128

baseplate/integration/thrift/__init__.py
```diff
@@ -42,82 +40,103 @@ class RequestContext(object):
...
+    return TraceInfo.from_upstream(
+        int(extracted_values["trace_id"]),
+        int(extracted_values["parent_span_id"]),
+        int(extracted_values["span_id"]),
+        True if extracted_values["sampled"].decode("utf-8") == "1" else False,
+        int(flags) if flags is not None else None,
+    )
+
+
+def baseplateify_processor(processor, logger, baseplate, edge_context_factory=None):
```
spladug: suggestions for a better name gladly accepted

baseplate/context/thrift.py
```diff
@@ -96,22 +96,22 @@ def _call_thrift_method(self, name, *args, **kwargs):
             try:
                 with self.server_span.make_child(trace_name) as span:
                     with self.pool.connection() as prot:
-                        prot.trans.set_header("Trace", str(span.trace_id))
-                        prot.trans.set_header("Parent", str(span.parent_id))
-                        prot.trans.set_header("Span", str(span.id))
+                        prot.trans.set_header(b"Trace", str(span.trace_id).encode())
```
bsimpson63: what's up with these strings?

spladug: like why are they getting bytey? the apache thrift implementation enforces that you're doing raw bytes in the headers. since we're only in the ascii range, it should be no change in wire protocol but we needed to be more clear here for the stricter code.

baseplate/integration/thrift/__init__.py
```diff
@@ -42,82 +40,103 @@ class RequestContext(object):
...
-        )
-
-        try:
-            peer_address, peer_port = server_context.getPeerName()
-        except AttributeError:
-            # the client transport is not a socket
-            pass
-        else:
-            trace.set_tag("peer.ipv4", peer_address)
-            trace.set_tag("peer.port", peer_port)
```
bsimpson63: looks like these tags aren't set anymore

baseplate/integration/thrift/command.py
bsimpson63: is the remote script the thing generated above?

spladug: the above is generating all the thrift code which does include the remote script yeah

baseplate/integration/thrift/__init__.py
spladug: yeah, i should've called that out. apache thrift doesn't have the socket context passed down to the server like fbthrift does, i'll poke around but i don't think it'll be as easy to get these in :(

baseplate/integration/thrift/__init__.py
dellis23: `wrap_processor_for_baseplate`?  `wrap_processor_with_baseplate`? `baseplate_wrap_processor`?

bsimpson63:
🐟 lgtm

baseplate/integration/thrift/__init__.py
```diff
@@ -42,82 +40,103 @@ class RequestContext(object):
...
-            int(flags) if flags is not None else None,
-        )
+    def make_processor_fn(fn_name, processor_fn):
+        def call_processor_with_span_context(self, seqid, iprot, oprot):
+            context = RequestContext()
+
+            headers = iprot.get_headers()
+            try:
+                trace_info = _extract_trace_info(headers)
+                edge_payload = headers.get(b"Edge-Request", None)
```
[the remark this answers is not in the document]
spladug: agreed, that's not how it is in the current code though. OK if i do that in a separate change?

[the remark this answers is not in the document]
pacejackson: sgtm

spladug:
:haircut_man: packages are up in a new PPA (super easy to build thanks to upstream packaging!) and travis is updated. this pointed out that i had an overbroad gitignore and the new baseplate.thrift.utils module (with serializer stuff) wasn't added to git!

baseplate/thrift/utils.py
```diff
@@ -0,0 +1,38 @@
+"""Helpers for working with Thrift."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from thrift.transport.TTransport import TMemoryBuffer
+
+
+def serialize_thrift(protocol_factory, thrift_object):
```
pacejackson: Do we actually need these? Thrift (or at least FBThrift) already has these helper methods.

https://github.snooguts.net/reddit/reddit-service-mailroom/blob/70dc51b9222bc3af80019f1738e8499cae53fd0b/mailroom/rabbitmq/serializer.py#L4

spladug: Apache Thrift does not which is why I'm providing them here.

spladug: OH! i'm an idiot and just found it after grepping again. totally different name though. sweet. will update.

PR #209, opened 2018-11-05
Remove local span on variant call

Kaitaan:
Local spans in variant calls within the experiments are a massive drain on CPU resources on the high volume of variant calls and provide very little value in this context.

This PR removes them from the 0.28 release (similar commits will remove them from all existing releases; the corresponding commit on master will be linked here once it's up).

### DIFF
--- a/baseplate/experiments/__init__.py
+++ b/baseplate/experiments/__init__.py
@@ -182,9 +182,7 @@ class Experiments(object):
         if user:
             inputs.update(user.event_fields())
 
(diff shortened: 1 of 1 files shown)


pacejackson:
Why are we doing this?

Kaitaan:
Note that this PR pulls this change into a new v0.28 branch. This change will not be merged into master directly.

Instead, a new PR will be cut against master with an equivalent change

spladug:
Yes, please include the reasoning in the commit message for posterity.

PR #215, opened 2018-11-21
Add support for more targeting operators

Kaitaan:
Add support for new comparison operator on experiment targeting.
Supports gt, ge, lt, le, and ne operators

### DIFF
--- a/baseplate/experiments/targeting/tree_targeting.py
+++ b/baseplate/experiments/targeting/tree_targeting.py
@@ -1,4 +1,5 @@
 from .base import Targeting
+import operator
 
 
 class TargetingNodeError(Exception):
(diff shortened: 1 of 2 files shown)


baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -183,15 +220,21 @@ def create_targeting_tree(input_node):
...
+    operator_name, input_node_value = list(input_node.items())[0]
+
+    operator_name = operator_name.lower()
 
-    if operator in OPERATOR_NODE_TYPE_MAPPING:
+    if operator_name in OPERATOR_NODE_TYPE_MAPPING:
+        operator_node_type = OPERATOR_NODE_TYPE_MAPPING[operator_name]
         try:
-            subnode = OPERATOR_NODE_TYPE_MAPPING[operator](input_node_value)
+            if operator_node_type == ComparisonNode:
```
spladug: do you want to use `issubclass` to be safe for future extension?

PR #216, opened 2018-11-28
Add BaseplateHeaderTrustHandler to BaseplateConfigurator

dkasper:
In order to allow trace headers to be accepted between services we allow BaseplateConfigurator to be setup with a BaseplateHeaderTrustHandler. The default implementation provided here validates a header X-Trace-Hmac against a secret provided at the path secret/trace/hmac. If this validation passes then the trace headers for the request will be trusted.


dkasper:

baseplate/integration/pyramid.py
```diff
@@ -86,6 +86,45 @@ class ServerSpanInitialized(BaseplateEvent):
...
+    """
+    def __init__(
+        self,
+        trust_trace_headers=False,
+        trust_edge_context_header=False,
+        secret_store=None
+    ):
+        self.trust_trace_headers = trust_trace_headers
+        self.trust_edge_context_header = trust_edge_context_header
+        self.secret_store = secret_store
```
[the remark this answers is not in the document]
dkasper: I can see your point. What I may propose then is that we would could move the BaseplateHeaderTrustHandler to a separate file, and maybe rename to HmacHeaderTrustHandler. That way it can be reused by gateway and other services that need it. We can then create the default version here which simply uses the `trust_trace_headers` flag.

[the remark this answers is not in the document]
pacejackson: They can be in the same file or a separate one, that shouldn't be an issue for re-use.  I think we would want something like:

```
class HeaderTrustHandlerInterface
class StaticTrustHandler(HeaderTrustHandlerInterface)
class HmacSignatureTrustHandler(HeaderTrustHandlerInterface)
```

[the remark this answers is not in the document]
dkasper: Yeah, I meant two unrelated things there. Mainly that we should create the new HmacSignatureTrustHandler class here in baseplate and not in Gateway for reuse, but also moving to a separate spot for code organization purposes.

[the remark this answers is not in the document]
cshoe: I definitely agree with @pacejackson. I think a dependency injection type approach is best -- construct the Trust object that you want and pass it into the configurator.

baseplate/integration/pyramid.py
```diff
@@ -86,6 +86,45 @@ class ServerSpanInitialized(BaseplateEvent):
...
+
+    def should_trust_trace_headers(self, request):
+        if self.secret_store:
```
cshoe: Related to the other comments, I don't think we should have branching logic based on the existence of the secret store. Classes that need the secret store should concretely depend on them, catastrophically failing if it doesn't exist.

baseplate/integration/pyramid.py
```diff
@@ -107,13 +152,21 @@ class BaseplateConfigurator(object):
 
     """
 
-    # TODO: remove the default on trust_trace_headers once apps are updated
     def __init__(self, baseplate, trust_trace_headers=False,
-                 edge_context_factory=None):
+                 edge_context_factory=None, header_trust_handler=None, secret_store=None):
         self.baseplate = baseplate
         self.trust_trace_headers = trust_trace_headers
```
cshoe: If all the trust logic is being moved to a ``header_trust_handler``, this doesn't need to be kept on the instance anymore.

tests/integration/pyramid_tests.py
```diff
@@ -284,3 +288,144 @@ def test_local_trace_in_context(self):
...
+        )
+        configurator.include(self.baseplate_configurator.includeme)
+        self.context_init_event_subscriber = mock.Mock()
+        configurator.add_subscriber(self.context_init_event_subscriber, ServerSpanInitialized)
+        app = configurator.make_wsgi_app()
+        self.test_app = webtest.TestApp(app)
+
+    def test_hmac_valid(self):
+        self.test_app.get("/trace_context", headers={
+            "X-Trace": "1234",
```
cshoe: Let's try to use realistic values here.

> Trace identifiers are 64 or 128-bit, but all span identifiers within a trace are 64-bit. All identifiers are opaque.

[link](https://github.com/openzipkin/b3-propagation#identifiers)

(There's good chance I've violated this in the past. I'm sorry.)

spladug: 0000000000000001234 is a viable 64 bit number :)

baseplate/integration/pyramid.py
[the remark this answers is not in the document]
spladug: Yeah, this would be a nice clean approach and let's us unify the `trust_trace_headers` boolean into the same system. The existing flag can be maintained at the external API, given a deprecation warning, and would just add the equivalent "handler" to the list.

dkasper:
💇‍♂️

baseplate/integration/pyramid.py
```diff
@@ -107,13 +127,23 @@ class BaseplateConfigurator(object):
 
     """
 
-    # TODO: remove the default on trust_trace_headers once apps are updated
     def __init__(self, baseplate, trust_trace_headers=False,
-                 edge_context_factory=None):
+                 edge_context_factory=None, header_trust_handler=None):
         self.baseplate = baseplate
         self.trust_trace_headers = trust_trace_headers
+        if self.trust_trace_headers:
```
pacejackson: You might need to make the default value `None` and have this check be `if trust_trace_headers is not None` and set `self.trust_trace_headers = bool(trust_trace_headers)`

baseplate/integration/pyramid.py
```diff
@@ -107,13 +127,23 @@ class BaseplateConfigurator(object):
...
+                """setting trust_trace_headers is deprecated in favor of using
+                a header trust handler.
+                """
+            )
         self.edge_context_factory = edge_context_factory
 
+        if header_trust_handler:
+            self.header_trust_handler = header_trust_handler
+        else:
+            self.header_trust_handler = StaticTrustHandler(trust_headers=trust_trace_headers)
```
pacejackson: this should probably use `self.trust_trace_headers`

baseplate/integration/pyramid.py
pacejackson: Can we just get rid of `self.trust_trace_headers`?  I'm guessing probably not?

baseplate/integration/pyramid.py
```diff
@@ -123,9 +153,14 @@ def _on_new_request(self, event):
             return
 
         trace_info = None
-        if self.trust_trace_headers:
+        if self.header_trust_handler.should_trust_trace_headers(request):
```
pacejackson: probably just call this once and assign it to a variable

baseplate/integration/pyramid.py
```diff
@@ -127,11 +127,11 @@ class BaseplateConfigurator(object):
...
     """
 
-    def __init__(self, baseplate, trust_trace_headers=False,
+    def __init__(self, baseplate, trust_trace_headers=None,
                  edge_context_factory=None, header_trust_handler=None):
         self.baseplate = baseplate
-        self.trust_trace_headers = trust_trace_headers
-        if self.trust_trace_headers:
+        self.trust_trace_headers = bool(trust_trace_headers)
+        if self.trust_trace_headers is not None:
```
pacejackson: this needs to be `if trust_trace_headers is not None`, `self.trust_trace_headers` is always a boolean.

baseplate/integration/pyramid.py
dkasper: Would be nice, we could add a metric to see if anyone is using it

baseplate/integration/pyramid.py
dkasper: good catch

baseplate/integration/pyramid.py
```diff
@@ -86,18 +86,38 @@ class ServerSpanInitialized(BaseplateEvent):
     pass
 
 
+class HeaderTrustHandlerInterface(object):
```
spladug: just `HeaderTrustHandler` would be fine and would be more consistent with other interface base classes in baseplate.

baseplate/integration/pyramid.py
```diff
@@ -86,18 +86,38 @@ class ServerSpanInitialized(BaseplateEvent):
...
+
+
 # pylint: disable=abstract-class-not-used
 class BaseplateConfigurator(object):
     """Config extension to integrate Baseplate into Pyramid.
 
     :param baseplate.core.Baseplate baseplate: The Baseplate instance for your
         application.
-    :param bool trust_trace_headers: Should this application trust trace
+    :param bool trust_trace_headers: (DEPRECATED) Should this application trust trace
```
spladug: let's just delete the docs for this param

baseplate/integration/pyramid.py
```diff
@@ -86,18 +86,38 @@ class ServerSpanInitialized(BaseplateEvent):
...
         headers from the client? If ``True``, trace headers in inbound requests
         will be used for the server span. If ``False``, new random trace IDs
         will be generated for each request.
     :param baseplate.core.EdgeRequestContextFactory edge_context_factory: A
         configured factory for handling edge request context.
+    :param baseplate.integration.pyramid.HeaderTrustHandlerInterface header_trust_handler:
+        An object which will be used to verify whether the headers for a request
```
spladug: this sorta sounds like any headers would be involved in this. how about:

"to verify whether baseplate should parse request context headers like trace IDs"

baseplate/integration/pyramid.py
```diff
@@ -86,18 +86,38 @@ class ServerSpanInitialized(BaseplateEvent):
...
 
+class HeaderTrustHandlerInterface(object):
+    """Class used by BaseplateConfigurator to validate headers.
+    """
+
+    def should_trust_trace_headers(self, request):
+        raise NotImplementedError
+
+
+class StaticTrustHandler(HeaderTrustHandlerInterface):
```
spladug: this should probably have a docstring and both this and the base class should probably be added to the `.rst` documentation so it shows up in the generated docs

tests/integration/pyramid_tests.py
```diff
@@ -265,12 +265,14 @@ def test_exception_in_exception_view_caught(self):
...
     def test_distrust_headers(self, getrandbits):
         getrandbits.return_value = 1234
-        self.baseplate_configurator.trust_trace_headers = False
+        self.baseplate_configurator.header_trust_handler.trust_headers = False
 
         self.test_app.get("/example", headers={
             "X-Trace": "1234",
             "X-Parent": "2345",
             "X-Span": "3456",
+            "X-Sampled": "1",
```
spladug: i think these can be dropped now right? (with the patches on master to fix the issue that made 'em required)

baseplate/integration/pyramid.py
```diff
@@ -107,13 +127,25 @@ class BaseplateConfigurator(object):
...
-    def __init__(self, baseplate, trust_trace_headers=False,
-                 edge_context_factory=None):
+    def __init__(self, baseplate, trust_trace_headers=None,
+                 edge_context_factory=None, header_trust_handler=None):
         self.baseplate = baseplate
-        self.trust_trace_headers = trust_trace_headers
+        self.trust_trace_headers = bool(trust_trace_headers)
+        if trust_trace_headers is not None:
+            warn_deprecated(
+                """setting trust_trace_headers is deprecated in favor of using
```
spladug: does this end up with a bunch of leading whitespace in the middle of the sentence? you could instead do 

```python
warn_deprecated(
    "blah blah blah "
    "blah blah blah"
)
```

and python will auto-concat those two strings

baseplate/integration/pyramid.py
```diff
@@ -107,13 +127,25 @@ class BaseplateConfigurator(object):
...
     """
 
-    # TODO: remove the default on trust_trace_headers once apps are updated
-    def __init__(self, baseplate, trust_trace_headers=False,
-                 edge_context_factory=None):
+    def __init__(self, baseplate, trust_trace_headers=None,
+                 edge_context_factory=None, header_trust_handler=None):
         self.baseplate = baseplate
-        self.trust_trace_headers = trust_trace_headers
+        self.trust_trace_headers = bool(trust_trace_headers)
```
pacejackson: Maybe we should change this to `self._trust_trace_headers` and add:

```
@property
def trust_trace_headers(self):
    warn_deprecated('don't do this')
    return self._trust_trace_headers
```

And then at some point we could either remove it or have it raise a special exception.

baseplate/integration/pyramid.py
pacejackson: Re-upping this

tests/integration/pyramid_tests.py
dkasper: Yep, I'll remove it for an extra check on that logic.

baseplate/integration/pyramid.py
dkasper: That won't work for hmac validation since the hmac will be different for each request

dkasper:
💇‍♂️ Cleaned up the docs

baseplate/integration/pyramid.py
pacejackson: Not a `self.` variable, just something in here, it'll be different for each request.

```
trust_headers = self.header_trust_handler.should_trust_trace_headers(request)
```

dkasper: oh, got it. Not sure I see the point since it is only used once now that there is should_trust_edge_context_payload

baseplate/integration/pyramid.py
```diff
@@ -123,9 +177,14 @@ def _on_new_request(self, event):
...
 
         trace_info = None
-        if self.trust_trace_headers:
+        if self.header_trust_handler.should_trust_trace_headers(request):
             try:
                 trace_info = self._get_trace_info(request.headers)
+            except (KeyError, ValueError):
+                pass
+
+        if self.header_trust_handler.should_trust_trace_headers(request):
```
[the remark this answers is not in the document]
dkasper: yeah, updated

baseplate/integration/pyramid.py
```diff
@@ -86,34 +86,88 @@ class ServerSpanInitialized(BaseplateEvent):
...
+        """ Whether baseplate should parse the trace headers from the inbound request
+
+        :param request pyramid.util.Request: The request
+
+        :returns bool: Whether baseplate should parse the trace headers from the inbound request.
+        """
+        raise NotImplementedError
+
+    def should_trust_edge_context_payload(self, request):
+        """ Whether baseplate should trust the edge context headers from the inbound request.
```
pacejackson: This should probably say `edge context payload` or `edge context header` rather than `headers`.

baseplate/integration/pyramid.py
```diff
@@ -86,34 +86,88 @@ class ServerSpanInitialized(BaseplateEvent):
...
-    def __init__(self, baseplate, trust_trace_headers=False,
-                 edge_context_factory=None):
+    def __init__(self, baseplate, trust_trace_headers=None,
+                 edge_context_factory=None, header_trust_handler=None):
         self.baseplate = baseplate
-        self.trust_trace_headers = trust_trace_headers
+        self.trust_trace_headers = bool(trust_trace_headers)
+        if trust_trace_headers is not None:
+            warn_deprecated(
+                "setting trust_trace_headers is deprecated in favor of using"
```
spladug: note: you'll wanna have a space at the end of this string or beginning of next since the strings get concatenated. this'll be `in favor of usinga header` right now

PR #218, opened 2018-11-30
Don't report expected Thrift exceptions as errors to span observers

spladug:
Thrift exceptions defined in the IDL are effectively another form of
return value and should not be considered an RPC failure. If
applications wish to monitor rates of specific (expected) error types
they can do so themselves.

This should allow easier monitoring of error rates.

I wanted to define some proper tests for this when getting it in place and found that the test suite was really not in a good place to do that. So I took the opportunity to rewrite it to allow full end-to-end integration testing of our Thrift integrations. Coverage on existing code is increased.

The first commit was done because 1) it was asked of me in the review for the Apache Thrift code and I hadn't done it yet, and 2) it allowed decoupling some of the tests a bit better.

The final commit (the functional change) is best viewed with whitespace diffing off (?w=1) since most of the "changes" are dedenting some code.

### DIFF
--- a/tests/integration/thrift_tests.py
+++ b/tests/integration/thrift_tests.py
@@ -4,31 +4,35 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import contextlib
+import jwt
 import logging
 import unittest
-import jwt
 
-try:
-    from io import BytesIO as StringIO
-except ImportError:
-    from cStringIO import StringIO
+import gevent
+import gevent.monkey
 
+from thrift.protocol.THeaderProtocol import THeaderProtocolFactory
+from thrift.Thrift import TApplicationException
+from thrift.transport.TTransport import TMemoryBuffer, TTransportException
+
+from baseplate import config
 from baseplate.core import (
     Baseplate,
     BaseplateObserver,
     EdgeRequestContext,
     EdgeRequestContextFactory,
     NoAuthenticationError,
+    ServerSpan,
     ServerSpanObserver,
+    SpanObserver,
 )
 from baseplate.context.thrift import ThriftContextFactory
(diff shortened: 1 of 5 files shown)


tests/integration/thrift_tests.py
```diff
@@ -38,267 +42,330 @@
...
+        with serve_thrift(handler, server_span_observer) as server:
+            with raw_thrift_client(server.endpoint) as client:
+                with self.assertRaises(TApplicationException):
+                    client.example()
 
-        self.assertEqual(self.server_observer.on_start.call_count, 1)
-        self.assertEqual(self.server_observer.on_finish.call_count, 1)
-        _, captured_exc, _ = self.server_observer.on_finish.call_args[0][0]
+        server_span_observer.on_start.assert_called_once_with()
+        server_span_observer.on_finish.assert_called_once()
```
pacejackson: This probably doesn't actually work, this was added to `mock` in python 3.6 https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

tests/integration/thrift_tests.py
```diff
@@ -38,267 +42,386 @@
...
+        "max_concurrency": "100",
+    }, listener, processor)
+
+    # figure out what port the server ended up on
+    server_address = listener.getsockname()
+    server_endpoint = config.EndpointConfiguration(
+        family=server_bind_endpoint.family, address=server_address)
+    server.endpoint = server_endpoint
+
+    # run the server until the user our caller with it
```
bsimpson63: typo?

spladug: whoops yeah, also made me realize there's a missing try/finally around the `yield`. thanks!

PR #219, opened 2018-12-01
Use SO_REUSEPORT when binding our own sockets

spladug:
We are seeing a fair amount of imbalance in how many connections each
baseplate-serve process gets while sharing a socket. Using SO_REUSEPORT
should improve the balance and allow us to more consistently use each
server's CPUs.

See https://lwn.net/Articles/542629/ for more information.

If this looks good, I'll try this out on a few servers and gather data to determine if
it's actually effective before merging onto master.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -104,19 +104,30 @@ def configure_logging(config, debug):
 
 
 def make_listener(endpoint):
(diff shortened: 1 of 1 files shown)


spladug:
:eyeglasses: @bsimpson63

PR #220, opened 2018-12-03
Add exponential backoff for event publishers

hhlin:
Currently publishers retry indefinitely every two seconds. In the event that collectors become degraded we prefer that clients back off rather than repeatedly retry.

👓 @spladug

### DIFF
--- a/baseplate/events/publisher.py
+++ b/baseplate/events/publisher.py
@@ -4,18 +4,18 @@ import gzip
 import hashlib
 import hmac
 import logging
-import time
 
 import requests
 
 from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from .. import config, metrics_client_from_config
 from .. _compat import configparser, BytesIO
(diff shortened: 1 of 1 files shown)


baseplate/events/publisher.py
```diff
@@ -4,34 +4,40 @@
...
 
 
 # seconds to wait for the event collector
 POST_TIMEOUT = 3
-# seconds to wait between retries when the collector fails
-RETRY_DELAY = 2
+# Base rate for expontential retry delay
+RETRY_BACKOFF = 2
+# Maximum time spent wait for a message to be sent
+MAX_RETRY_TIME = 15 * 60
```
spladug: that's quite a while! do we want to bail out and let init restart us sooner or maybe lessen the base rate of backoff?

after only a few tries we're waiting minutes between tries which seems pretty intense for recovery.

hhlin:
💇

PR #224, opened 2018-12-04
Vault 0.9.0 compatibility

rram:
A [change in vault 0.9.0 removed the nonce from responses where it was not needed](https://github.com/hashicorp/vault/commit/30aab2aa2feff6dd81ca379f2489132ff5bc284a) in order to avoid logging it. I've added nonce generation to the secret fetcher so that vault doesn't make one and log it for us. I also added a bit of documentation to aid in debugging.

### DIFF
--- a/baseplate/secrets/fetcher.py
+++ b/baseplate/secrets/fetcher.py
@@ -79,6 +79,12 @@ logger = logging.getLogger(__name__)
 K8S_SERVICE_ACCOUNT_TOKEN_FILE = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 NONCE_FILENAME = "/var/local/vault.nonce"
 VAULT_TOKEN_PREFETCH_TIME = datetime.timedelta(seconds=60)
+REAUTHENTICATION_ERROR_MESSAGE = """
+Authenication failed! If this instance previously authenticated with a
+different nonce, a vault operator may need to remove the instance ID from the
+identity whitelist. See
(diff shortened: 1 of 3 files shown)


baseplate/secrets/fetcher.py
```diff
@@ -80,6 +81,7 @@
 VAULT_TOKEN_PREFETCH_TIME = datetime.timedelta(seconds=60)
 
 
+
```
spladug: nit: only two blanks between top level stuff please

baseplate/secrets/fetcher.py
```diff
@@ -92,6 +94,12 @@ def fetch_instance_identity():
     return resp.text
 
 
+def generate_nonce():
+    """Returns a string value suitable for use as a nonce."""
```
spladug: nit: pep257 says this should be "Return a string..." (command form)

baseplate/secrets/fetcher.py
```diff
@@ -205,23 +213,27 @@ def _vault_aws_auth(self):
         """
         identity_document = fetch_instance_identity()
         nonce = load_nonce()
+        if nonce is None:
```
spladug: `if not nonce:` unless you think nonce could legitimately be falsey but non-None

spladug:
oh also it says 0.9 in the commit / PR but we're installing 1.0 in the test suite.

rram:
You have encouraged me to rebase

PR #225, opened 2018-12-05
context.thrift: Include service and method in retry exception

bsimpson63:
r2 app logs frequently have errors like:

 ```Dec  5 13:51:12 r2-app-08fcbf7e67f4981e6 api E: TTransportException: retry policy exhausted while attempting operation, last error was: timed out interacting with socket U: 145735584198 FP: /api/categories_and_posts.json?always_show_media=1&api_type=json&background_fetch=0&expand_srs=1&feature=link_preview&from_detail=1&limit=10&max_recs=1&num_posts=10&num_recs=1&num_subreddits=0&obey_over18=1&raw_json=1&request_timestamp=441625267.3291667&rtj=debug&sr_detail=1```

but it's tough to tell from that line which service or method had the failure.

👓 @spladug


bsimpson63:
💇 `namespace` (e.g. `'thing_service'`) is better than `client_cls.__name__` (`'Client'`)

PR #227, opened 2018-12-10
Make ThriftContextFactory more efficient

spladug:
The previous implementation would re-enumerate all the methods and
re-create all the proxy methods every time it was instantiated,
including on server spans, local spans, and even the retrying()
decorator. This method does the enumeration at factory instantiation and
uses the cached type.

Some unit tests were removed as they were overly knowledgeable about
internals and the new integration tests cover that area.

This is a port to the Apache Thrift version of Baseplate based on
@jesseyli's work in a fork.

:eyeglasses: @jesseyli @bsimpson63 @pacejackson


spladug:
With [this benchmark script](https://gist.github.com/spladug/523922979c9cf96696589bab0c68f5a4) I do see a significant performance improvement doing this:

```
# original baseplate ThriftContextFactory implementation
$ python2.7 benchmark.py 
Time elapsed to create 100000 spans = 1.8929 seconds (18.93μs / per)
$ python3.4 benchmark.py 
Time elapsed to create 100000 spans = 3.4420 seconds (34.42μs / per)

# new implementation with proxy type caching
$ python2.7 benchmark.py 
Time elapsed to create 100000 spans = 0.1106 seconds (1.11μs / per)
$ python3.4 benchmark.py 
Time elapsed to create 100000 spans = 0.1197 seconds (1.20μs / per)

```

Particularly noticeable in py3!

PR #229, opened 2018-12-11
add RangeVariantSet

Kaitaan:
add new range variant set which allows specifying bucketing ranges (rather
than just bucketing range sizes) for each variant.

cc: @matt-knox

### DIFF
--- a/baseplate/experiments/variant_sets/range_variant_set.py
+++ b/baseplate/experiments/variant_sets/range_variant_set.py
@@ -1,0 +1,69 @@
+from .base import VariantSet
+
+
+class RangeVariantSet(VariantSet):
+    """ Variant Set designed to take fixed bucket ranges.
+
+    This VariantSet allows manually setting bucketing ranges.
+    It takes in a variant name, then the range of buckets in
+    that should be assigned to that variant. This enables user-defined
+    bucketing algorithms, as well as simplifies the ability to adjust
+    range sizes in special circumstances.
+    """
+
+    def __init__(self, variants, num_buckets=1000):
+        """ :param list variants: array of dicts, each containing the keys 'name',
+            'range'. Name is the variant name, and range is a list containing the lower
+            (inclusive) and upper (exclusive) ends of the bucketing range.
+            Lower and upper values are expressed as floating point values between 0 and 1,
+            and will be truncated to match the granularity of the number of buckets. This
+            function does not validate that ranges do not overlap.
+        :param int num_buckets: the number of potential buckets that can be
+            passed in for a variant call. Defaults to 1000, which means maximum
+            granularity of 0.1% for bucketing
+        """
+
+        self.variants = variants
+        self.num_buckets = num_buckets
+
+        self._validate_variants()
+
+    def __contains__(self, item):
+        if (self.variants[0].get('name') == item
+                or self.variants[1].get('name') == item):
+            return True
+
+        return False
+
+    def _validate_variants(self):
+
+        if self.variants is None or len(self.variants) == 0:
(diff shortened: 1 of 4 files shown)


baseplate/experiments/variant_sets/range_variant_set.py
```diff
@@ -0,0 +1,69 @@
...
+    def __contains__(self, item):
+        if (self.variants[0].get('name') == item
+                or self.variants[1].get('name') == item):
+            return True
+
+        return False
+
+    def _validate_variants(self):
+
+        if self.variants is None or len(self.variants) == 0:
```
spladug: nit: `if not self.variants:` would be more pythonic + pithy

baseplate/experiments/variant_sets/range_variant_set.py
```diff
@@ -0,0 +1,69 @@
...
+            raise ValueError('No variants provided')
+
+        total_size = 0
+        for variant in self.variants:
+            if variant.get('range') is None or len(variant.get('range')) != 2:
```
spladug: we're doing `.get("range")` a lot in this function, might be good to factor out into a variable

baseplate/experiments/variant_sets/range_variant_set.py
```diff
@@ -0,0 +1,69 @@
...
+        """Deterministically choose a variant. Every call with the same bucket
+        on one instance will result in the same answer
+
+        :param int bucket: an integer bucket representation
+        :return string: the variant name, or None if bucket doesn't fall into
+                          any of the variants
+        """
+
+        for variant in self.variants:
+            if bucket >= variant.get('range')[1] and bucket < variant.get('range')[0]:
```
spladug: based on L48-49, does this say "if it's higher than the upper and lower than the lower return the variant" ?

baseplate/experiments/variant_sets/range_variant_set.py
mattknox:
```suggestion
        if not self.variants or len(self.variants) == 0:
```

mattknox: +1, would pith again

baseplate/experiments/variant_sets/range_variant_set.py
mattknox:
```suggestion
            variant_range = variant.get('range')
            if variant_range is None or len(variant_range) != 2:
```

baseplate/experiments/variant_sets/range_variant_set.py
mattknox:
```suggestion
            if bucket >= variant_range[0] and bucket < variant_range[1]:
```
we definitely want the bottom to be the closed part of the interval (so 0.0, 0.1 means " >= 0.0 and < 0.1")

Kaitaan:
digging into that last comment. Something weird there...

baseplate/experiments/variant_sets/range_variant_set.py
Kaitaan: yeah, this is odd. I'll have some updates here shortly.

Kaitaan:
Closing this. Will re-up with some changes. Looks like the tests for this code weren't actually running (because the unit test filename was missing the suffix '_tests')

baseplate/experiments/variant_sets/range_variant_set.py
pacejackson: if you do `if not self.variants` I think that also checks `len(self.variants) == 0` since an empty container is falsy

PR #230, opened 2018-12-11
add RangeVariantSet

Kaitaan:
add new range variant set which allows specifying bucketing ranges (rather
than just bucketing range sizes) for each variant.

@matt-knox

### DIFF
--- a/baseplate/experiments/variant_sets/range_variant_set.py
+++ b/baseplate/experiments/variant_sets/range_variant_set.py
@@ -1,0 +1,70 @@
+from .base import VariantSet
+
+
+class RangeVariantSet(VariantSet):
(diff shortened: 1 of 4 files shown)


baseplate/experiments/variant_sets/range_variant_set.py
```diff
@@ -0,0 +1,70 @@
...
+    def __contains__(self, item):
+        for variant in self.variants:
+            if variant.get('name') == item:
+                return True
+
+        return False
+
+    def _validate_variants(self):
+
+        if not self.variants or len(self.variants) == 0:
```
spladug: still should be `if not self.variants`

PR #232, opened 2018-12-12
fix: use range instead of xrange

KattMingMing:
This PR fixes a small inconsistency with the mostly python 3 style code by using `range` instead of `xrange`.

Quickly verified with a references check on `range`: https://sourcegraph.com/github.com/reddit/baseplate@master/-/blob/tests/unit/experiments/providers/variant_sets/rollout_variant_set_tests.py#L77:23&tab=references

### DIFF
--- a/baseplate/crypto.py
+++ b/baseplate/crypto.py
@@ -63,7 +63,7 @@ else:
         expected_len = len(expected)
         result = actual_len ^ expected_len
         if expected_len > 0:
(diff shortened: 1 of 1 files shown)


spladug:
Thanks for the patch! That code is actually exclusively for Python 2.7 since the standard library in 3.3+ has the function it's meant to imitate and we only support 3.4+.

PR #237, opened 2019-01-29
thrift: Filter out magic methods from proxy methods

fishy:
In some python implementations, e.g. pypy6-3.5, calling
_enumerate_service_methods on thrift classes will also return magic
methods, e.g. "__new__". If we don't filter them out, they will be
replaced by our pooled thrift proxy method, which relies on the
constructor to be called first, but python calls __new__ before the
constructor, so that will cause us troubles.

This should be a no-op for python implementations that does not return
magic methods in the first place.


fishy:
👓 @pacejackson @spladug

PR #238, opened 2019-02-06
Add logging to retry policies

cshoe:
Adding logging as low as possible in the retry logic. Small step towards
gettting some visibility into what is sending retries.

### DIFF
--- a/baseplate/retry.py
+++ b/baseplate/retry.py
@@ -4,7 +4,11 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import logging
 import time
+
+
+logger = logging.getLogger(__name__)
 
 
 class RetryPolicy(object):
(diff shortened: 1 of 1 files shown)


cshoe:
👓 @spladug @alienth

cshoe:
I'd like to get more context but there doesn't appear to be an easy way unless we added access to a request object.

Yes, it will be spammy but we'll be able to answer the question "Are retries happening?" as opposed to some of the current guess work.

spladug:
Baseplate uses retry policies for quite a few things that aren't service-to-service calls so contextless logging is gonna be really hard to track to something useful. If you care about service-to-service hops then how about logging here instead? https://github.com/reddit/baseplate/blob/master/baseplate/context/thrift.py#L104

spladug:
Also, if it's not intended to stick around forever doing it in the library which releases on a slower cadence than applications might not be the best fit. Could always patch it in on the application side if there's a single place you want to dig into.

PR #242, opened 2019-03-06
Add concurrency monitoring gauges sent periodically

spladug:
Baseplate servers will now report concurrent in-flight requests as a
per-hostname-per-PID gauge every 10 seconds if metrics are configured.

The metrics system should handle aggregation appropriately.

:eyeglasses: @ckwang8128 @bsimpson63

### DIFF
--- a/baseplate/integration/pyramid/__init__.py
+++ b/baseplate/integration/pyramid/__init__.py
@@ -168,6 +168,10 @@ class BaseplateConfigurator(object):
                 trust_headers=self.trust_trace_headers,
             )
 
+    def _on_application_created(self, event):
+        # attach the baseplate object to the application the server gets
+        event.app.baseplate = self.baseplate
+
     def _on_new_request(self, event):
         request = event.request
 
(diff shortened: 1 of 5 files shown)


baseplate/integration/pyramid/__init__.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, baseplate, trust_trace_headers=None,
                 trust_headers=self.trust_trace_headers,
             )
 
+    def _on_application_created(self, event):
+        # attach the baseplate object to the application the server gets
+        event.app.baseplate = self.baseplate
```
bsimpson63: 🔕 what's going on here?

spladug: making the `Baseplate` object accessible to the server in a standard location across thrift/wsgi server types.

bsimpson63: ah, so we can do https://github.com/reddit/baseplate/pull/242/files#diff-ff6fa7da3dc7ab489d21610e7e8b6527R37 ?

spladug: yup!

PR #244, opened 2019-03-15
Add ratelimiting

bradengroom:
This is a second pass at https://github.com/reddit/baseplate/pull/162.
This implementation tries to hide the backend a bit more. Example usage:
```
redis_pool = pool_from_config(app_config)
backend_factory = RedisRateLimitBackendContextFactory(redis_pool)
ratelimiter_factory = RateLimiterContextFactory(backend_factory, allowance, interval, key_prefix='read:')
baseplate.add_to_context('ratelimiter', ratelimiter_factory)

try:
    context.ratelimiter.consume(context.request_context.user.id)
    print('Ratelimit passed')
except RateLimitExceededException:
    print('Too many requests')
```

A call to the ratelimiter can also consume more from the allowance if needed:
```
context.ratelimiter.consume(context.request_context.user.id, amount=3)
```

This only adds support for a Redis cache, but makes it possible to create other types of backends.

Any object can be passed to the ratelimiter as a cache as long as it implements this method:
```
def consume(self, key, amount, max, bucket_size):
    pass
```


bradengroom:
Needs documentation

bradengroom:
This is the most basic implementation that I could think of, but I think there are extensions to this that could make it even more useful:

- Make the ratelimiter usable as a function decorator
- Make the thrown exception customizable. It might be nice if you could have the ratelimiter just raise pyramids `HTTPTooManyRequests` for you
- Allow passing this to `BaseplateProcessorEventHandler` for ratelimiting across all endpoints. I think this would only work if an edge_context were provided and we could assume `context.request_context.user.id` as the ratelimit key
- Make this usable as a context manager on thrift clients to ratelimit downstream calls: `context.downstream_thrift.ratelimited(allowance=1000, interval=60)`. Bonus points if it can be stacked with `retrying`: `context.downstream_thrift.retrying(attempts=3).ratelimited(allowance=1000, interval=60)`

bradengroom:
Tagging folks from the original PR
👓 @bsimpson63 @pacejackson @spladug

pacejackson:
Can we also include memcached in this since that is officially supported by baseplate?

baseplate/ratelimit.py
```diff
@@ -0,0 +1,99 @@
...
+    :param int interval: The interval (in seconds) to reset allowances.
+    :param str key_prefix: A prefix to add to keys during rate limiting.
+        This is useful if you will have two different rate limiters that will
+        receive the same keys.
+
+    """
+
+    def __init__(self, cache, allowance=None, interval=None, key_prefix=''):
+        assert allowance >= 1, 'minimum allowance is 1'
+        assert interval >= 1, 'minimum interval is 1'
```
pacejackson: These should raise `ValueError` rather than being asserts.

bradengroom:
Yeah, I can work on adding that

baseplate/ratelimit.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.cache = cache
```
pacejackson: Should we also check `isinstance(cache, RateLimitCache)`?

bradengroom:
Btw, I would have preferred a leaky bucket type implementation, but I wasn't sure how to implement that on top of Redis without race conditions. I'm interested if anyone has thoughts on this

baseplate/ratelimit.py
```diff
@@ -0,0 +1,99 @@
...
+        :param str key: The name of the rate limit bucket to consume from.
+        :param int amount: The amount to consume from the rate limit bucket.
+
+        """
+        key = self.key_prefix + key
+        if not self.cache.consume(key, amount, self.allowance, self.interval):
+            raise RateLimitExceededException('Rate limit exceeded.')
+
+
+class RateLimitCache:
```
pacejackson: Since baseplate also supports python2, you should include the `(object)` here (you don't need it in python3 but it changes some behavior when you don't have it in python2).

baseplate/ratelimit.py
```diff
@@ -0,0 +1,99 @@
...
+import time
+
+
+class RateLimitExceededException(Exception):
+    """This exception gets raised whenever a rate limit is exceeded.
+    """
+    pass
+
+
+class RateLimiter:
```
pacejackson: Since baseplate also supports python2, you should include the (object) here (you don't need it in python3 but it changes some behavior when you don't have it in python2).

pacejackson:
> Make the ratelimiter usable as a function decorator

_Maybe_ I don't feel super strongly about this, although it sounds cool.

> Make the thrown exception customizable. It might be nice if you could have the ratelimiter just raise pyramids HTTPTooManyRequests for you

I'm not sure we want to get this fancy?  I kind of prefer that the interface for this is "it raises this exception and you should catch it and do with it what you will".

> Allow passing this to BaseplateProcessorEventHandler for ratelimiting across all endpoints. I think this would only work if an edge_context were provided and we could assume context.request_context.user.id as the ratelimit key

This might be doable, but I bet needs another, higher level object that handles configuring the rate limits and calling the `RateLimiter`.

> Make this usable as a context manager on thrift clients to ratelimit downstream calls: context.downstream_thrift.ratelimited(allowance=1000, interval=60). Bonus points if it can be stacked with retrying: context.downstream_thrift.retrying(attempts=3).ratelimited(allowance=1000, interval=60)

👍!  I think you could actually chain the `retrying` the same way that the thrift client does.

bsimpson63:
> I would have preferred a leaky bucket type implementation, but I wasn't sure how to implement that on top of Redis without race conditions

What are the race conditions?

Since this is ratelimiting with a cache there are chances that keys could get evicted and we'd be loose with the ratelimiting. If we could somehow communicate that this was just "best effort" ratelimiting maybe any issues with races would also be forgivable. I'm not sure we want to say this though--making people think about how good/bad the ratelimit is could lead to trouble when they just want a drop in solution.

bsimpson63:
💅 looks great, no comments beyond what Andrew brought up

pacejackson:
About the race conditions, could you address that by doing it in Redis itself via lua?

bradengroom:
> About the race conditions, could you address that by doing it in Redis itself via lua?

I think so. I have seen people doing this else where. I'm not sure if there are any scaling concerns with doing fancy stuff like that. I'm also not sure if memcache has any equivalent, but I guess these two backends don't necessarily have to use the same algorithm.

baseplate/ratelimit.py
```diff
@@ -0,0 +1,134 @@
...
+        :param int amount: The amount to consume from the rate limit bucket.
+        :param int allowance: The maximum allowance for the rate limit bucket.
+        :param int interval: The interval to reset the allowance.
+
+        """
+        current_bucket = _get_current_bucket(interval)
+        key = key + current_bucket
+        ttl = interval * 2
+        self.memcache_client.add(key, 0, expire=ttl)
+        count = self.memcache_client.incr(key, 1)
```
bradengroom: This will fail if there's a GC pause or something between these two memcache calls that lasts the duration of `ttl`

bradengroom: It looks like keys must be set prior to using `incr`, and memcached doesn't have any sort of transactions. I'm not sure that this can be avoided, but maybe someone with more memcache experience knows of something.

bradengroom: Looks like r2 actually handles this case in its ratelimit implementation. Maybe we should do something similar here.

bradengroom: https://github.com/reddit-archive/reddit/commit/df3e8532149ac6d948a18c588a0e0aa52845e679

baseplate/ratelimit.py
```diff
@@ -0,0 +1,169 @@
...
+        counters.
+    :param int allowance: The maximum allowance allowed per key.
+    :param int interval: The interval (in seconds) to reset allowances.
+    :param str key_prefix: A prefix to add to keys during rate limiting.
+        This is useful if you will have two different rate limiters that will
+        receive the same keys.
+
+    """
+
+    def __init__(self, cache, allowance=None, interval=None,
```
pacejackson: These shouldn't have default values if the default values are just going to raise exceptions.

baseplate/ratelimit.py
```diff
@@ -0,0 +1,169 @@
...
+    :param int interval: The interval (in seconds) to reset allowances.
+    :param str key_prefix: A prefix to add to keys during rate limiting.
+        This is useful if you will have two different rate limiters that will
+        receive the same keys.
+
+    """
+
+    # TODO: This feels like a weird way of doing this
+    def __init__(self, cache_context_factory, ratelimit_cache_class,
+                 allowance=None, interval=None, key_prefix=''):
```
pacejackson: I don't think that `allowance` or `interval` should have default values.

baseplate/ratelimit.py
```diff
@@ -0,0 +1,169 @@
...
+class RateLimitExceededException(Exception):
+    """This exception gets raised whenever a rate limit is exceeded.
+    """
+    pass
+
+
+class RateLimiterContextFactory(ContextFactory):
+    """RateLimiter context factory
+
+    :param cache_context_factory: An instance of
```
spladug: are there any other constraints on it that the user should know about?

spladug: what about having the ratelimit factory wrap a cache context factory instead? then this factory only needs to take one parameter and the ratelimit factories can abstract away what info is needed under their hoods.

baseplate/ratelimit.py
spladug: yeah, it's best to catch it so the user doesn't get a crash. i think we can treat the ratelimits as best-effort though so being slightly off because of this sorta thing isn't the end of the world.

baseplate/ratelimit.py
bradengroom: I think I like this direction, but does this mean we need a class per backend?:
`class RedisRateLimiterContextFactory(RedisContextFactory)` and `class MemcacheRateLimiterContextFactory(MemcacheContextFactory)`?

baseplate/ratelimit.py
bradengroom: Hah. Yeah. Thanks for catching this

baseplate/ratelimit.py
spladug:
```python
class RateLimiterContextFactory(ContextFactory):
    def __init__(self, backend, allowance, interval):
        ...

class Backend(object):
    def consume(...):
        ...

class RedisBackend(Backend):
    def __init__(self, redis_context_factory, prefix):
        ...

class MemcachedBackend(Backend):
    def __init__(self, memcache_context_factory, prefix):
        ...
```

then the main ratelimiter is completely agnostic to anything about the backends (including key prefixes) and the backend just handles the protocol of talking to the cache/redis/whatever

bradengroom: I see. Thanks!

bradengroom: Taking a step back, it seems like this is the API we would want:

```
redis_pool = pool_from_config(app_config)
backend = RedisRateLimitBackend(redis_pool)
ratelimiter_factory = RateLimiterContextFactory(backend, allowance, interval)
baseplate.add_to_context('ratelimiter', ratelimiter_factory)
```

So I guess the `Backend` constructors should actually take a cache pool rather than a cache context factory.

spladug: we do want spans for the cache traffic generated by the backend though, right?

bradengroom: Yeah, I think that makes sense. I guess I thought this `pool_from_config` would return a pool of `MonitoredRedisConnection` which would create spans. Maybe I'm missing something though?

bradengroom: Oh yeah `MonitoredRedisConnection` gets created by `RedisContextFactory`

spladug: the base pool is just straight up from the redis library (https://github.com/reddit/baseplate/blob/master/baseplate/context/redis.py#L54) but the context factory adds the baseplatey span stuff. https://github.com/reddit/baseplate/blob/master/baseplate/context/redis.py#L76

spladug: i'd be ok with the ratelimitbackend doing the context factory wrapping automatically rather than requiring that one be passed in. so it'd look like your snippet above but inside we're adding the contextfactory wrapper.

bradengroom: Yeah. That takes us back to your suggested API:
```
redis_pool = pool_from_config(app_config)
redis_factory = RedisContextFactory(redis_pool)
backend = RedisRateLimitBackend(redis_factory)
ratelimiter_factory = RateLimiterContextFactory(backend, allowance, interval)
baseplate.add_to_context('ratelimiter', ratelimiter_factory)
```

Thanks again

bradengroom: > i'd be ok with the ratelimitbackend doing the context factory wrapping automatically rather than requiring that one be passed in. so it'd look like your snippet above but inside we're adding the contextfactory wrapper.

I think this might make for a nicer API since it removes a layer of classes

baseplate/ratelimit.py
bradengroom: Actually, the `pymemcache` docs say this case would return a `None` instead or raising an exception.
https://github.com/pinterest/pymemcache/blob/65560b738a01bfd420cdbeaae3b7bd78503c2855/pymemcache/client/base.py#L571

I think we should just treat it as zero in this case:

```
self.memcache.add(key, 0, expire=ttl)
count = self.memcache.incr(key, 1) or 0
return count <= allowance
```

baseplate/ratelimit.py
```diff
@@ -0,0 +1,203 @@
...
+
+try:
+    from .context.memcache import MonitoredMemcacheConnection
+except ImportError as exc:
+    pass
+
+try:
+    from .context.redis import MonitoredRedisConnection
+except ImportError:
+    pass
```
bradengroom: Is there anything more I should be doing for these import errors?

baseplate/ratelimit.py
bradengroom: Actually, I guess `or amount` is more appropriate.

baseplate/ratelimit.py
pacejackson: You might want to make this a module and put the memcache and redis stuff in it's own files, i.e.:

```
ratelimit/
    __init__.py
    backends/
        __init__.py
        redis.py
        memcache.py
```

pacejackson: Then the import errors only happen when you try to use redis or memcache directly, not just from importing the ratelimiter

bradengroom: 👍

baseplate/ratelimit/backends/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .base import RateLimitBackend
+
+
+__all__ = [
+    'RateLimitBackend',
+]
```
bradengroom: redis & memcache backends not exported here since that would trigger an import of `redis` and `pymemcache`

bradengroom:
💇 I want to poke around with this a bit more in one of my services, but this should be ready for another pass.

baseplate/ratelimit/backends/base.py
```diff
@@ -0,0 +1,21 @@
...
+    :param int amount: The amount to consume from the rate limit bucket.
+    :param int allowance: The maximum allowance for the rate limit bucket.
+    :param int interval: The interval to reset the allowance.
+
+    """
+    def consume(self, key, amount, max, bucket_size):
+        raise NotImplementedError
+
+
+def _get_current_bucket(bucket_size):
```
pacejackson: It might make sense to to make this a `@staticmethod` of `RateLimitBackend` so subclasses don't have to import this separately.

baseplate/ratelimit/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .ratelimit import RateLimiter
+from .ratelimit import RateLimitExceededException
+from .ratelimit import RateLimiterContextFactory
+
+
+__all__ = [
+    'RateLimiter',
+    'RateLimitExceededException',
+    'RateLimiterContextFactory',
+]
```
pacejackson: 🔕 I'm not sure we really care about exporting things like this?  We might just want to leave these `__init__.py`-s empty.

I don't feel strongly either way.

baseplate/ratelimit/backends/memcache.py
```diff
@@ -0,0 +1,51 @@
...
+        :param int amount: The amount to consume from the rate limit bucket.
+        :param int allowance: The maximum allowance for the rate limit bucket.
+        :param int interval: The interval to reset the allowance.
+
+        """
+        current_bucket = _get_current_bucket(interval)
+        key = key + current_bucket
+        ttl = interval * 2
+        self.memcache.add(key, 0, expire=ttl)
+        count = self.memcache.incr(key, amount) or amount
```
pacejackson: Can you add a comment about why you set this to `incr or amount`?

baseplate/ratelimit/ratelimit.py
```diff
@@ -0,0 +1,76 @@
...
+
+    """
+
+    def __init__(self, backend, allowance, interval, key_prefix='rl:'):
+        if allowance < 1:
+            raise ValueError('minimum allowance is 1')
+        if interval < 1:
+            raise ValueError('minimum interval is 1')
+        if not isinstance(backend, RateLimitBackend):
+            raise TypeError('backend must be an instance of RateLimitBackend')
```
pacejackson: 🔕 we might want to move these checks up to the context factory so you get the error when you try to start your service as opposed to when you serve your first request.

bradengroom: Good call

baseplate/ratelimit/__init__.py
bradengroom: I was trying to make the import nicer here:
`from baseplate.ratelimit import RateLimiterContextFactory `
instead of:
`from baseplate.ratelimit.ratelimit import RateLimiterContextFactory `

I don't mind removing them if it's weird though. Do you have a preference @spladug ?

baseplate/ratelimit/ratelimit.py
bradengroom: I put the checks in both places since you can create a rate limiter without going through context factory stuff.

bradengroom:
💇 I added some integration tests too

baseplate/ratelimit/__init__.py
spladug: it's nice to hide the internal details of the ratelimit subpackage from the outside world. import passthrough :+1:

baseplate/ratelimit/__init__.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
```
spladug: does this work on py2? https://github.com/reddit/baseplate/blob/master/baseplate/__init__.py#L4

baseplate/ratelimit/backends/base.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import time
+
+
+class RateLimitBackend(object):
```
spladug: i'd usually just put this in the `__init__`. whatcha think?

baseplate/ratelimit/backends/base.py
```diff
@@ -0,0 +1,24 @@
...
+    :param int amount: The amount to consume from the rate limit bucket.
+    :param int allowance: The maximum allowance for the rate limit bucket.
+    :param int interval: The interval to reset the allowance.
+
+    """
+    def consume(self, key, amount, max, bucket_size):
+        raise NotImplementedError
+
+    @staticmethod
+    def _get_current_bucket(bucket_size):
```
spladug: can this just be a top-level function? even though it's not being used that way, seeing it on the class made me worry about implementation inheritance

baseplate/ratelimit/ratelimit.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class RateLimiter(object):
+    """A class for rate limiting actions.
+
+    :param `RateLimitBackend` backend: The backend to use for storing rate
+        limit counters.
+    :param int allowance: The maximum allowance allowed per key.
+    :param int interval: The interval (in seconds) to reset allowances.
+    :param str key_prefix: A prefix to add to keys during rate limiting.
```
spladug: i feel like calling this a key prefix here puts the parameter in a weird place. i can understand having it on this level of class, but think it might be better to call it "group" or "category" or something similarly vague of how it's used in implementation. or if it goes onto the backend classes, then it could be called prefix and its direct usage be made very clear. 

the default value being a very mcrouter-style prefix route also confuses things a bit. if it's gonna be on this class i think the separator should be backend-specified.

baseplate/ratelimit/ratelimit.py
```diff
@@ -0,0 +1,88 @@
...
+
+        self.backend = backend
+        self.key_prefix = key_prefix
+        self.allowance = allowance
+        self.interval = interval
+
+    def consume(self, key, amount=1):
+        """Consume the given `amount` from the allowance for the given `key`.
+
+        This will rate
```
spladug: s/rate/raise

baseplate/ratelimit/backends/memcache.py
```diff
@@ -0,0 +1,58 @@
...
+    :param memcache_pool: An instance of
+        :py:class:`~pymemcache.client.base.PooledClient`
+
+    """
+
+    def __init__(self, memcache_pool):
+        self.memcache_pool = memcache_pool
+
+    def make_object_for_context(self, name, server_span):
+        memcache = MonitoredMemcacheConnection(name, server_span, self.memcache_pool)
```
spladug: i just wanted to raise these classes peeking into the guts of the memcached/redis context factories. it's probably ok since this is all in baseplate, but it does feel a bit coupled. would it be insane to construct a context factory in `__init__` and call its `make_object_for_context` here to get a monitored connection? too weird with a fake context variable?

baseplate/ratelimit/backends/base.py
pacejackson: I can't find my feedback about this, I made this suggestion so each subclass didn't have to import both this class and the top level function.

pacejackson: If you prefer it to be `get_current_bucket` that's fine too, I don't feel strongly about it :)

baseplate/ratelimit/__init__.py
bradengroom: I don't understand why this wouldn't this work. The tests pass in travis. Is there something more specific you'd like me to look into?

baseplate/ratelimit/backends/__init__.py
```diff
@@ -3,9 +3,22 @@
...
+    :param int amount: The amount to consume from the rate limit bucket.
+    :param int allowance: The maximum allowance for the rate limit bucket.
+    :param int interval: The interval to reset the allowance.
+
+    """
+    def consume(self, key, amount, max, bucket_size):
+        raise NotImplementedError
+
+
+def _get_current_bucket(bucket_size):
```
pacejackson: This should probably not have the `_` since it is meant to be imported.

baseplate/ratelimit/backends/memcache.py
pacejackson: I think this sounds reasonable?

bradengroom: The approach that you describe does seem like better design, but the fake context is kinda weird. I'm okay with this as it is, but I'll implement your advice if you feel strongly about it. I could go either way.

bradengroom: Didn't see Andrew's comment before making mine. I'll change it to use `make_object_for_context ` since I'm the odd one out. 👍

baseplate/ratelimit/backends/__init__.py
bradengroom: I don't think this is meant to be imported from outside of the library. It just gets imported by the two backends because it's a bit of common logic that didn't make it into the interface. I don't have a strong opinion on this though. I can drop the `_` if you want me to.

pacejackson: I figured this might be used by folks who want to write their own, custom back-ends?  Maybe/maybe not?  I don't feel super strongly either way.

bradengroom:
💇 I _think_ the only thing remaining is `_get_current_bucket ` vs `get_current_bucket `. I'd rather not make that function public and document it, but I could be persuaded that we should.

spladug:
This branch is off master right? If we wheel off that, any service using it will need to upgrade to apache thrift too. Might want to rebase onto the 0.28 series if you don't want that hassle. 

Or you could try on activity service which is on apache thrift now, but won't be terribly interesting otherwise.

baseplate/ratelimit/__init__.py
spladug: hm, i can't repro it but it used to fail. maybe something changed in the minor versions of 2.7. :thinking:

baseplate/ratelimit/backends/__init__.py
spladug: eh, they can re-implement it simply for their backend if desired. i'd say keep it private for now and we can open it later if really needed.

PR #246, opened 2019-03-20
Be case-insensitive with Thrift headers

ckwang8128:
There is currently no official spec for Thrift header
case-sensitivity. This follows Pyramid's approach of storing
HTTP headers in case-insensitive dictionary with Thrift
headers to be lenient in accepted header values while
still propagating headers in the format that they came in.

For reference:

The Pyramid implementation details - https://github.com/Pylons/pyramid/blob/c02f228b554d636d475ad1f34945fd91e4b1bfdf/docs/narr/webob.rst (see `req.headers`) 

The requests CaseInsensitiveDict module - https://github.com/kennethreitz/requests/blob/v1.2.3/requests/structures.py#L37 

The fbthrift header object - https://github.com/apache/thrift/blob/66a44c5d8d57bd0c2e2afd228a29b5bc679a6770/lib/py/src/transport/THeaderTransport.py#L96

### DIFF
--- a/baseplate/integration/thrift/__init__.py
+++ b/baseplate/integration/thrift/__init__.py
@@ -20,6 +20,7 @@ from __future__ import unicode_literals
 
 import sys
 
+from requests.structures import CaseInsensitiveDict
 from thrift.Thrift import TException, TApplicationException
 from thrift.transport.TTransport import TTransportException
 from thrift.protocol.TProtocol import TProtocolException
@@ -105,7 +106,9 @@ def baseplateify_processor(processor, logger, baseplate, edge_context_factory=No
         def call_processor_with_span_context(self, seqid, iprot, oprot):
             context = RequestContext()
 
-            headers = iprot.get_headers()
+            # Allow case-insensitivity for THeader headers
+            headers = CaseInsensitiveDict(data=iprot.get_headers())
+
             try:
                 trace_info = _extract_trace_info(headers)
             except (KeyError, ValueError):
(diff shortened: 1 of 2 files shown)


ckwang8128:
:eyeglasses: @spladug @bsimpson63 @pacejackson

baseplate/integration/thrift/__init__.py
```diff
@@ -20,6 +20,7 @@ def make_processor(app_config):
 
 import sys
 
+from requests.structures import CaseInsensitiveDict
```
ckwang8128: We depend on requests already for installation, so this seemed like a reasonable module to include. We could also store our own implementation if desired.

tests/integration/thrift_tests.py
```diff
@@ -195,7 +195,7 @@ def test_header_propagation(self):
         class Handler(TestService.Iface):
             def __init__(self):
                 self.server_span = None
-
+
```
bsimpson63: what is this!

bsimpson63:
🐟 seems reasonable

baseplate/integration/thrift/__init__.py
spladug: yup, that's fine. this is one of the hard deps in `setup.py`.

tests/integration/thrift_tests.py
spladug: hissssssssssssssssssssssssssssssss

ckwang8128: nooooo

pacejackson: rabble

PR #249, opened 2019-03-26
 cassandra: Don't return from result() until callbacks have completed 

bsimpson63:
This fixes a race condition where the server span can complete before
the callback has closed out the child span.

https://github.com/reddit/baseplate/issues/100

👓 @spladug @bradengroom @pacejackson

### DIFF
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -3,6 +3,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import logging
+from threading import Event
+
 # pylint: disable=no-name-in-module
 from cassandra.cluster import Cluster, _NOT_SET
 # pylint: disable=no-name-in-module
@@ -11,6 +14,9 @@ from cassandra.query import SimpleStatement, PreparedStatement, BoundStatement
 from . import ContextFactory
 from .. import config
 from .._compat import string_types
+
+
+logger = logging.getLogger(__name__)
 
 
 def cluster_from_config(app_config, prefix="cassandra.", **kwargs):
@@ -92,15 +98,56 @@ class CQLMapperContextFactory(CassandraContextFactory):
         return cqlmapper.connection.Connection(session_adapter)
 
 
-def _on_execute_complete(_, span):
+class WaitForCallbackResponseFuture(object):
+    """Wrap the ResponseFuture to ensure callbacks have completed.
+
+    This fixes a race condition where the server span can complete before
+    the callback has closed out the child span.
+
+    """
+
+    def __init__(self, span, future, callback_fn, callback_args, errback_fn, errback_args):
+        self.callback_event = Event()
+
+        future.add_callback(callback_fn, callback_args, self.callback_event)
+        future.add_errback(errback_fn, errback_args, self.callback_event)
+
+        self.future = future
+
+    def result(self):
+        exc = None
+
+        try:
+            result = self.future.result()
+        except Exception as e:
+            exc = e
+
+        # wait for either _on_execute_complete or _on_execute_failed to run
+        wait_result = self.callback_event.wait(timeout=0.01)
+        if not wait_result:
+            logger.warning("Cassandra metrics callback took too long. Some metrics may be lost.")
+
+        if exc:
+            raise exc   # pylint: disable=E0702
+
+        return result
+
+
+def _on_execute_complete(result, span, event):
     # TODO: tag with anything from the result set?
     # TODO: tag with any returned warnings
-    span.finish()
+    try:
+        span.finish()
+    finally:
+        event.set()
 
 
-def _on_execute_failed(exc, span):
-    exc_info = (type(exc), exc, None)
-    span.finish(exc_info=exc_info)
+def _on_execute_failed(exc, span, event):
+    try:
+        exc_info = (type(exc), exc, None)
+        span.finish(exc_info=exc_info)
+    finally:
+        event.set()
 
 
 class CassandraSessionAdapter(object):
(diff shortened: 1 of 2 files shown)


baseplate/context/cassandra.py
```diff
@@ -90,7 +96,50 @@ def make_object_for_context(self, name, server_span):
...
+    def result(self):
+        exc = None
+
+        try:
+            result = self.future.result()
+        except Exception as e:
+            exc = e
+
+        # wait for either _on_execute_complete or _on_execute_failed to run
+        wait_result = self.callback_event.wait(timeout=0.1)
```
spladug: can we use a (much) smaller wait timeout? we can spend a little more cpu checking in the rare case it really does take that much longer rather than blocking the response for 100ms at minimum if the result won the race.

baseplate/context/cassandra.py
```diff
@@ -90,7 +96,50 @@ def make_object_for_context(self, name, server_span):
...
+
+        self.future = future
+
+    def result(self):
+        exc = None
+
+        try:
+            result = self.future.result()
+        except Exception as e:
+            exc = e
```
spladug: This will lose the original traceback. According to [this](http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html) we can do:

```python
try:
    ...
except Exception:
    exc_info = sys.exc_info()

...

if exc_info:
    raise exc_info[0], exc_info[1], exc_info[2]
```
which will preserve that even though we're reraising non-locally

baseplate/context/cassandra.py
```diff
@@ -90,7 +96,50 @@ def make_object_for_context(self, name, server_span):
...
+
+    This fixes a race condition where the server span can complete before
+    the callback has closed out the child span.
+
+    """
+
+    def __init__(self, span, future, callback_fn, callback_args, errback_fn, errback_args):
+        self.callback_event = Event()
+
+        def register_completion(fn):
```
spladug: what do you think of passing the event object into `_on_execute_complete` / `_on_execute_failed` and letting them just directly call it explicitly rather than doing this wrapper stuff?

bsimpson63:
> since there is an event object and some extra waiting now involved, we should benchmark this to make sure it's not gonna impact perf too much. maybe we can try it in thing service?

yeah, sounds good

baseplate/context/cassandra.py
bsimpson63: > blocking the response for 100ms at minimum

Isn't this the maximum wait? If `Event.set()` has already been called this will return immediately. If not it will wait/block up to the timeout value.

baseplate/context/cassandra.py
bsimpson63: hmm, this doesn't seem to work in python3. I'll hunt around for a solution

baseplate/context/cassandra.py
spladug: yeah, you're right. ignore me!

bsimpson63: I'll lower the timeout regardless. I don't think we care to wait around more than ~10ms.

baseplate/context/cassandra.py
bsimpson63: the wrapper is kind of nice because it does the `try/finally`, but I'll try moving those to the callback methods and see how it looks

bsimpson63:
💇

baseplate/_compat.py
```diff
@@ -23,6 +23,9 @@
     def iteritems(d):
         return iter(d.items())
 
+    def reraise_exception(exc, exc_info):
+        raise exc
```
spladug: oh wow, TIL this is all better in py3. that's fantastic. https://legacy.python.org/dev/peps/pep-3134/

given the impending doom of py27, do we want to just drop this compat method and let it be a shitty traceback on 2.7?

baseplate/context/cassandra.py
```diff
@@ -90,15 +97,58 @@ def make_object_for_context(self, name, server_span):
...
+        try:
+            result = self.future.result()
+        except Exception as e:
+            exc = e
+            exc_info = sys.exc_info()
+
+        # wait for either _on_execute_complete or _on_execute_failed to run
+        wait_result = self.callback_event.wait(timeout=0.01)
+        if not wait_result:
+            logger.warning("callback took too long")
```
spladug: I think this might be a big vague/confusing of a message to see in logs. If we want to keep it at warning (which sounds reasonable if its presence means we'd be losing metric data) then maybe:

> Cassandra metrics callback took too long. Some metrics may be lost.

? not sure there's much remediation that can be done in this case.

baseplate/_compat.py
bsimpson63: yeah, I think that'd be better than this goofy workaround

baseplate/context/cassandra.py
```diff
@@ -3,12 +3,19 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import logging
+import sys
+from threading import Event
```
spladug: note: this appears to be un-monkeypatched in gevent < 1.3 so we should make sure we're on 1.3+

bsimpson63: Should I import `Event` from `gevent` instead? Or do you mean like do an audit of all our services and ensure they're on 1.3+ or upgrade them before releasing this?

pacejackson: I think we want to avoid importing from gevent in baseplate itself?  I'm not sure if we have a great way to enforce this though other than just doing an audit.

spladug: I'm gonna play around with setup.py a bit, there might be a way to sanely say we need 1.3+ when we're using gevent without pulling it in as a hard dependency.

bsimpson63: Should I merge this now, or hold off until we figure this stuff out?

spladug: let's merge. i've no good answers on versions right now.

Issue #243, opened 2019-03-11
Add `**open_options` to FileWatcher


foklepoint:
Currently, a FileWatcher can't be forced to `open()` a file with a specific encoding type. This is necessary when you don't want to fallback to the platform encoding and instead force a file to be opened as a specific encoding type (i.e: utf-8) 

A proposed solution is to add `**open_options` kwargs to [`FileWatcher.__init__`](https://github.com/reddit/baseplate/blob/8f6e05c88d00ee1e467951a577e22c25aca5ec4a/baseplate/file_watcher.py#L70)

This would also allow the FileWatcher to modify the open operation in the following ways:

```
mode='r'
buffering=-1
encoding=None
errors=None
newline=None
closefd=True
opener=None
```

pacejackson:
Note, this is really only applicable to Python 3.  Python 3 has [several options](https://docs.python.org/3/library/functions.html#open) while Python 2 [does not](https://docs.python.org/2.7/library/functions.html#open)

pacejackson:
Addressed in #250

PR #250, opened 2019-03-26
Support binary mode, encoding, and newline options in FileWatcher

pacejackson:
cc: @foklepoint

This should address https://github.com/reddit/baseplate/issues/243


tests/unit/file_watcher_tests.py
```diff
@@ -115,3 +120,78 @@ def test_timeout(self, mock_retry_factory):
                     parser=json.load,
                     timeout=3,
                 )
+
+    @unittest.skipIf(sys.version_info.major >= 3,
```
pacejackson: I think I might just break these up into separate test classes, one for Py2 and one for Py3

pacejackson:
💇‍♂️

baseplate/file_watcher.py
```diff
@@ -67,11 +91,25 @@ class FileWatcher(object):
...
+        if sys.version_info.major < 3 and open_options:
+            raise TypeError(
+                "'buffering' is the only supported open_options keyword "
+                "argument for FileWatcher() in Python 2"
+            )
+
+        if "file" in open_options:
+            raise TypeError("'file' is an invalid keyword argument for FileWatcher()")
+        if "mode" in open_options:
+            raise TypeError("'mode' is an invalid keyword argument for FileWatcher()")
```
pacejackson: Do we want to support mode?  We could by doing something like:

```
self._open_options.setdefault("mode", "r")
```

and get rid of the `"r"` in the call to open?

pacejackson: I don't think we _can_ support `"closefd"` and `"file"` would be confusing to support.

pacejackson:
👓 @foklepoint @bradengroom @spladug

baseplate/file_watcher.py
pacejackson: Also, if we did support mode, it would have to be restricted to only read types, so we might not want the overhead of that?

pacejackson: Maybe a better way if we want to support binary mode is to have `binary=False` and do `"r+b"` if that is set to `True`

spladug: yeah, agreed. of all the options it looks like `encoding` and `binary` and _maybe_ `newline` are the ones we'd care to let through? probably best to just explicitly add those then? sorry for the bum steer on kwargs.

pacejackson:
💇‍♂️

pacejackson:
It looks like the test failure is unrelated, I think it's just a test that has a small chance of failing since it's testing random distributions:

```
======================================================================
FAIL: test_distribution (tests.unit.random_tests.WeightedChoiceTests)
----------------------------------------------------------------------
  File "/home/travis/build/reddit/baseplate/tests/unit/random_tests.py", line 71, in test_distribution
    self.assertLess(abs(3000 - choices["c"]), 150)
AssertionError: 156 not less than 150
```

baseplate/file_watcher.py
```diff
@@ -63,15 +64,59 @@ class FileWatcher(object):
     :param callable parser: A callable that takes an open file object, parses
         or otherwise interprets the file, and returns whatever data is
         meaningful.
-    :param float timeout: How long, in seconds, to block instantiation waiting
```
spladug: are you hard-wrapping at 80 columns? you can go up to 100!!!

baseplate/file_watcher.py
```diff
@@ -63,15 +64,59 @@ class FileWatcher(object):
...
+            raise TypeError("'encoding' is not supported in binary mode.")
+
+        if binary and newline is not None:
+            raise TypeError("'newline' is not supported in binary mode.")
+
         self._path = path
         self._parser = parser
         self._mtime = 0
         self._data = _NOT_LOADED
+        self._mode = "r+b" if binary else "r"
```
spladug: `r+b` is opening the file for append-writes and reads in binary mode rather than just `rb` which is "read only binary mode"

baseplate/file_watcher.py
pacejackson: In my defense, this docstring was already doing that. 😅

PR #253, opened 2019-03-26
Lazily create context attributes attached via baseplate.add_to_context

bradengroom:
This changes clients added to requests to be created lazily. The idea is that most requests do not access _all_ of the clients added via `baseplate.add_to_context`. We should be able to save work by not creating clients that are never accessed.


bradengroom:
This micro benchmark seems to indicate that there is potential here. We should observe the performance on a real service though.

Benchmark code:
```
# coding=utf8

import time

from baseplate import Baseplate
from baseplate.context.thrift import ThriftContextFactory
from baseplate.thrift_pool import ThriftConnectionPool
from baseplate.integration.thrift import RequestContext
from baseplate.core import ServerSpan


from genpy.test import TestService

ITERATIONS = 100000
CLIENTS = 20

def main():
    baseplate = Baseplate()
    baseplate.configure_logging()
    
    for i in range(CLIENTS):
        pool = ThriftConnectionPool(None)
        context_factory = ThriftContextFactory(pool, TestService.Client)
        baseplate.add_to_context('thrift_client_{}'.format(i), context_factory)

    for clients_to_access in range(CLIENTS):
        finals = []
        for _ in range(3):
            elapsed = 0
            for _ in range(ITERATIONS):
                context = RequestContext()
                start = time.time()
                baseplate.make_server_span(context, "example_service")
                for i in range(clients_to_access):
                    getattr(context, 'thrift_client_{}'.format(i))
                end = time.time()
                elapsed += end-start
            finals.append(elapsed)

        print("Time elapsed to create %d spans (%d clients, %d touched) = %0.4f seconds (%0.2fμs / per)" % (ITERATIONS, CLIENTS, clients_to_access, min(finals), min(finals)/ITERATIONS*1e6))


if __name__ == "__main__":
    main()
```

Current behavior:
```
Time elapsed to create 100000 spans (20 clients, 0 touched) = 6.3250 seconds (63.25μs / per)
Time elapsed to create 100000 spans (20 clients, 1 touched) = 6.3950 seconds (63.95μs / per)
Time elapsed to create 100000 spans (20 clients, 2 touched) = 6.4605 seconds (64.60μs / per)
Time elapsed to create 100000 spans (20 clients, 3 touched) = 6.5612 seconds (65.61μs / per)
Time elapsed to create 100000 spans (20 clients, 4 touched) = 6.7313 seconds (67.31μs / per)
Time elapsed to create 100000 spans (20 clients, 5 touched) = 6.6034 seconds (66.03μs / per)
Time elapsed to create 100000 spans (20 clients, 6 touched) = 6.6752 seconds (66.75μs / per)
Time elapsed to create 100000 spans (20 clients, 7 touched) = 7.1518 seconds (71.52μs / per)
Time elapsed to create 100000 spans (20 clients, 8 touched) = 6.7308 seconds (67.31μs / per)
Time elapsed to create 100000 spans (20 clients, 9 touched) = 6.9472 seconds (69.47μs / per)
Time elapsed to create 100000 spans (20 clients, 10 touched) = 6.9752 seconds (69.75μs / per)
Time elapsed to create 100000 spans (20 clients, 11 touched) = 6.9691 seconds (69.69μs / per)
Time elapsed to create 100000 spans (20 clients, 12 touched) = 6.9107 seconds (69.11μs / per)
Time elapsed to create 100000 spans (20 clients, 13 touched) = 6.9144 seconds (69.14μs / per)
Time elapsed to create 100000 spans (20 clients, 14 touched) = 6.9416 seconds (69.42μs / per)
Time elapsed to create 100000 spans (20 clients, 15 touched) = 7.0143 seconds (70.14μs / per)
Time elapsed to create 100000 spans (20 clients, 16 touched) = 7.0484 seconds (70.48μs / per)
Time elapsed to create 100000 spans (20 clients, 17 touched) = 7.0342 seconds (70.34μs / per)
Time elapsed to create 100000 spans (20 clients, 18 touched) = 7.0700 seconds (70.70μs / per)
Time elapsed to create 100000 spans (20 clients, 19 touched) = 7.1635 seconds (71.64μs / per)
```

New behavior:
```
Time elapsed to create 100000 spans (20 clients, 0 touched) = 3.9924 seconds (39.92μs / per)
Time elapsed to create 100000 spans (20 clients, 1 touched) = 4.4060 seconds (44.06μs / per)
Time elapsed to create 100000 spans (20 clients, 2 touched) = 4.7910 seconds (47.91μs / per)
Time elapsed to create 100000 spans (20 clients, 3 touched) = 5.0533 seconds (50.53μs / per)
Time elapsed to create 100000 spans (20 clients, 4 touched) = 5.3657 seconds (53.66μs / per)
Time elapsed to create 100000 spans (20 clients, 5 touched) = 6.0241 seconds (60.24μs / per)
Time elapsed to create 100000 spans (20 clients, 6 touched) = 6.1715 seconds (61.72μs / per)
Time elapsed to create 100000 spans (20 clients, 7 touched) = 6.5259 seconds (65.26μs / per)
Time elapsed to create 100000 spans (20 clients, 8 touched) = 6.8693 seconds (68.69μs / per)
Time elapsed to create 100000 spans (20 clients, 9 touched) = 7.1974 seconds (71.97μs / per)
Time elapsed to create 100000 spans (20 clients, 10 touched) = 7.5539 seconds (75.54μs / per)
Time elapsed to create 100000 spans (20 clients, 11 touched) = 7.8392 seconds (78.39μs / per)
Time elapsed to create 100000 spans (20 clients, 12 touched) = 8.4427 seconds (84.43μs / per)
Time elapsed to create 100000 spans (20 clients, 13 touched) = 8.6876 seconds (86.88μs / per)
Time elapsed to create 100000 spans (20 clients, 14 touched) = 8.9292 seconds (89.29μs / per)
Time elapsed to create 100000 spans (20 clients, 15 touched) = 9.2837 seconds (92.84μs / per)
Time elapsed to create 100000 spans (20 clients, 16 touched) = 9.5161 seconds (95.16μs / per)
Time elapsed to create 100000 spans (20 clients, 17 touched) = 9.9022 seconds (99.02μs / per)
Time elapsed to create 100000 spans (20 clients, 18 touched) = 10.1899 seconds (101.90μs / per)
Time elapsed to create 100000 spans (20 clients, 19 touched) = 10.5285 seconds (105.29μs / per)
```

baseplate/integration/mixins/lazy_attributes.py
```diff
@@ -0,0 +1,22 @@
+class LazyAttributesMixin(object):
```
bradengroom: I'm looking for some advice on improving this mixin pattern. I'm not a fan of how this relies on the implementing classes to set `self._lazy_attributes = {}`. I _could_ remove that from the implementing classes by modifying `__getattr__` to also:
```
if name == '_lazy_attributes':
  _lazy_attributes = {}
  setattr(self, name, _lazy_attributes)
  return _lazy_attributes
```

This feels gross though.

bradengroom:
👓 @spladug

baseplate/integration/mixins/lazy_attributes.py
```diff
@@ -0,0 +1,22 @@
...
+    """A mixin that allows instances to add a lazily evaluated attribute
+       through `add_lazy_property`.
+    """
+
+    # Called when an attribute is missing
+    def __getattr__(self, name):
+        if name not in self._lazy_attributes:
```
spladug: since we're trying to shave off microseconds here, i wonder if only touching the dictionary once would help?

```python
try:
    lazy_fn = self._lazy_attributes[name]
except KeyError:
    return super(LazyAttributesMixin, self).__getattr__(name)

prop_value = lazy_fn(self)
setattr(self, name, prop_value)
return prop_value
```

baseplate/integration/mixins/lazy_attributes.py
```diff
@@ -0,0 +1,22 @@
...
+            return super(LazyAttributesMixin, self).__getattr__(name)
+        prop_value = self._lazy_attributes[name](self)
+        setattr(self, name, prop_value)
+        return prop_value
+
+    def add_lazy_attribute(self, name, callable):
```
spladug: nit: callable is a builtin. we shouldn't shadow it.

baseplate/integration/mixins/lazy_attributes.py
spladug: ha, that's super meta.

one thing would just be to do the initiation in a constructor and force callers to call the constructor? that at least hides the details of what's going on from consumers?

baseplate/integration/mixins/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .lazy_attributes import LazyAttributesMixin
```
spladug: how about just putting this in `baseplate/integration/__init_.py` and prefixing the class name with `_` to mark it private?

i don't think we really need a whole subpackage for this one thing.

baseplate/context/__init__.py
```diff
@@ -51,8 +51,9 @@ def __init__(self, name, context_factory):
         self.context_factory = context_factory
 
     def on_server_span_created(self, context, server_span):
-        context_attr = self.context_factory.make_object_for_context(self.name, server_span)
-        setattr(context, self.name, context_attr)
+        def context_attr(_req):
```
spladug: `_self` maybe?

pacejackson:
A couple questions:

1. Those benchmarks concern me a bit, once you use 5 or 6 clients, the behavior is about the same and gets much worse the more you access, so I think we would need to test this out in the real world to see it's effect.
2. If we do decide to do this, could we use something like this? https://github.com/reddit/baseplate/blob/0f2cf46345ba37a088078e954f4e5edf5cc37cc9/baseplate/_utils.py#L28

bradengroom:
@pacejackson 
> 1. Those benchmarks concern me a bit, once you use 5 or 6 clients, the behavior is about the same and gets much worse the more you access, so I think we would need to test this out in the real world to see it's effect.

Yeah, it does seem to get worse if you start accessing a lot of things. Neil and I discussed doing exactly what you're suggesting. I think we are gonna try rolling it out to graphql since it is one of the extreme cases with a bunch of clients. We will see how it behaves there and decide if this is worth moving forward with.

> If we do decide to do this, could we use something like this?
So this was actually the first path that I went down. It turns out that you can't just attach a property to an instance. They have to be attached to classes. I dug into how Pyramid's `add_request_method` method works and came to this interesting code:
https://github.com/Pylons/pyramid/blob/80baff6e61d775dc34bb11b7d6c8ef8569da1c40/src/pyramid/util.py#L76

They actually create a class per instance and swap out the `__class__` beneath you. It's pretty interesting, but it ended up being about 4-5x slower than our current code. This was mostly due to the creation of the new class here:
https://github.com/Pylons/pyramid/blob/80baff6e61d775dc34bb11b7d6c8ef8569da1c40/src/pyramid/util.py#L131

Their approach also brings concerns of possible memory leaks if we don't do it right. This led me to try this lighter-weight solution with `__getattr__`

baseplate/integration/pyramid/__init__.py
```diff
@@ -67,6 +69,12 @@ def baseplate_tween(request):
     return baseplate_tween
 
 
+class BaseplatePyramidRequest(Request, _LazyAttributesMixin):
+    def __init__(self, *args, **kwargs):
+        super(BaseplatePyramidRequest, self).__init__(*args, **kwargs)
+        _LazyAttributesMixin.__init__(self)
```
pacejackson: I don't think you need to do this as long as you're calling `super()`

bradengroom: I wasn't sure how exactly multiple inheritance would work here, so I tried it out:

```
>>> class Base(object):
...   def __init__(self):
...     print("Base called")
... 
>>> class Mixin(object):
...   def __init__(self):
...     print("Mixin called")
... 
>>> class Implementer(Base, Mixin):
...   def __init__(self):
...     super(Implementer, self).__init__()
...     print("Implementer called")
... 
>>> Implementer()
Base called
Implementer called
<__main__.Implementer object at 0x10ff78860>
```

The `super()` function is just going to return the first class in `__mro__`, so `Mixin` doesn't get triggered by calling super.

pacejackson: I think if the base classes also call super, then both will get called. 

https://stackoverflow.com/a/16310777

bradengroom: Oh I see. I think this would work for our thrift request object, but not our pyramid request object. It looks like we would make it down to webob's `BaseRequest`  which doesn't appear to call `super`:
https://github.com/Pylons/webob/blob/b5297529ef5a8dd3dd84c6cfa41afd101acfa8a2/src/webob/request.py#L93

pacejackson: Makes sense in it's own weird way, multiple inheritance is kind of a mess 😅

bradengroom:
💇

baseplate/integration/__init__.py
```diff
@@ -5,6 +5,39 @@
...
+    def __init__(self, *args, **kwargs):
+        super(_LazyAttributesMixin, self).__init__(*args, **kwargs)
+        self._lazy_attributes = {}
+
+    # Called when an attribute is missing
+    def __getattr__(self, name):
+        try:
+            lazy_fn = self._lazy_attributes[name]
+        except KeyError:
+            return super(_LazyAttributesMixin, self).__getattr__(name)
```
pacejackson: Out of curiosity, did this change affect the benchmarks much at all?

baseplate/context/__init__.py
```diff
@@ -51,8 +51,9 @@ def __init__(self, name, context_factory):
         self.context_factory = context_factory
 
     def on_server_span_created(self, context, server_span):
-        context_attr = self.context_factory.make_object_for_context(self.name, server_span)
-        setattr(context, self.name, context_attr)
+        def context_attr(_self):
+            return self.context_factory.make_object_for_context(self.name, server_span)
+        context.add_lazy_attribute(self.name, context_attr)
```
pacejackson: Would it make sense to make this optional?  It could default to on but give services the option to turn it off?  I could see it being the case that this helps some services but hurts others.

baseplate/integration/__init__.py
```diff
@@ -5,6 +5,39 @@
...
+
+        prop_value = lazy_fn(self)
+        setattr(self, name, prop_value)
```
pacejackson: Does this actually _do_ anything?  Isn't `__getattr__` always going to check `self._lazy_attributes[name]` first and find the function in there?  I think we would have to `pop` the function out of `self._lazy_attributes` for this to work.

pacejackson: Side note, I think this would make the performance much worse since you're re-building the client object each time you access it.

spladug: `__getattr__` is only called for things not in the instance's `__dict__` (to override _that_, you use `__getattribute__`) so it will stop getting called once the setattr puts the reified value into the instance dict.

pacejackson: ahhhh, that's not confusing at all.......

baseplate/integration/__init__.py
bradengroom: Not any dramatic impact. I think it'd be hard to see something like this come through in my micro benchmark:
```
Time elapsed to create 100000 spans (20 clients, 0 touched) = 3.6920 seconds (36.92μs / per)
Time elapsed to create 100000 spans (20 clients, 1 touched) = 4.1572 seconds (41.57μs / per)
Time elapsed to create 100000 spans (20 clients, 2 touched) = 4.4702 seconds (44.70μs / per)
Time elapsed to create 100000 spans (20 clients, 3 touched) = 4.7819 seconds (47.82μs / per)
Time elapsed to create 100000 spans (20 clients, 4 touched) = 5.0547 seconds (50.55μs / per)
Time elapsed to create 100000 spans (20 clients, 5 touched) = 5.4793 seconds (54.79μs / per)
Time elapsed to create 100000 spans (20 clients, 6 touched) = 5.7219 seconds (57.22μs / per)
Time elapsed to create 100000 spans (20 clients, 7 touched) = 6.0422 seconds (60.42μs / per)
Time elapsed to create 100000 spans (20 clients, 8 touched) = 6.8233 seconds (68.23μs / per)
Time elapsed to create 100000 spans (20 clients, 9 touched) = 6.8550 seconds (68.55μs / per)
Time elapsed to create 100000 spans (20 clients, 10 touched) = 7.6404 seconds (76.40μs / per)
Time elapsed to create 100000 spans (20 clients, 11 touched) = 7.3861 seconds (73.86μs / per)
```

pacejackson: Makes sense, thanks!

bradengroom:
@spladug 
> note: this won't work on r2 but that's probably OK since we're gonna be dropping py2.7 support anyway.

Shouldn't this work as long as r2 uses baseplate's `add_to_context` and `make_server_span` methods? I see lots of calls to `add_to_context` in r2's app_globals file. I'm not sure how span creation happens though.

bradengroom:
I'm gonna see if I can monkey patch this into graphql

spladug:
r2's context object is pylons's `tmpl_context` global. We don't have any code in this PR for mixing-in our base class there so the lazy stuff won't be present.

bradengroom:
> r2's context object is pylons's tmpl_context global. We don't have any code in this PR for mixing-in our base class there so the lazy stuff won't be present.

Oh yeah. You told me this yesterday

pacejackson:
> I'm gonna see if I can monkey patch this into graphql
Since we have wheels, it's probably better to make a wheel of this branch and try using that in graphql?

bradengroom:
> Since we have wheels, it's probably better to make a wheel of this branch and try using that in graphql?

I'd prefer this. I'm just not sure how to make a wheel of this branch. Do I need to reach out to infra?

bradengroom:
I created a ticket to get a wheel of this branch made. I'd just let this sit until then

spladug:
This branch is off master right? If we wheel off that, any service using it will need to upgrade to apache thrift too. Might want to rebase onto the 0.28 series if you don't want that hassle. 

Or you could try on activity service which is on apache thrift now, but won't be terribly interesting otherwise.

bradengroom:
> This branch is off master right? If we wheel off that, any service using it will need to upgrade to apache thrift too. Might want to rebase onto the 0.28 series if you don't want that hassle.

Changing my ticket to point to this branch:
https://github.com/reddit/baseplate/pull/254

baseplate/context/__init__.py
spladug: Maybe a `LazyContextFactory` and `EagerContextFactory` and then alias the old `ContextFactory` to whichever profiles out to be the more common-case friendly one? Then services can decide based on the individual context attributes based on their knowledge of their own usage patterns.

spladug:
I'm going to close this out as f30f1c5f6ca11d5b63aeae4016342217353caf3a..ec802fe16c56e961354742d993947ff8c621a9a5 is merged. Thanks so much for pioneering this and pushing for the improvement.

Issue #248, opened 2019-03-25
metrics batch is too large


bsimpson63:
From Thing service logs:

```
Mar 22 17:08:33 thing-0c5b40c8692e8e5e4 [thing] 15496:17713791717421479062:baseplate.metrics:ERROR:baseplate metrics batch too large: flush more often                 or reduce amount done in one request, length 933145, [Errno 90] Message too long#015
```

spladug:
c6b053936dcfd0a612a052dca9762b63d0c49e9b w/ discussion in #255

PR #255, opened 2019-03-27
metrics: Send metrics in batches to avoid a too long message

bsimpson63:
https://github.com/reddit/baseplate/issues/248

I'm not sure that the max size is set correctly, I got 65507 from https://docstore.mik.ua/manuals/hp-ux/en/B2355-60130/UDP.7P.html

👓 @spladug @pacejackson @bradengroom

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -99,8 +99,13 @@ class BufferedTransport(object):
             self.transport.send(message)
         except socket.error as e:
             logger.error("baseplate metrics batch too large: flush more often \
-                or reduce amount done in one request, length %d, %s",
-                len(message), e)
+                or reduce amount done in one request, length %d, count %d, %s",
+                len(message), len(metrics), e)
+            counters = [
+                metric for metric in metrics if metric.split(b'|')[-1][0] == b'c'
+            ]
+            partial_message = b",".join(counters)
+            logger.exception("counters: %s", partial_message)
 
 
 class BaseClient(object):


baseplate/metrics.py
```diff
@@ -85,22 +85,62 @@ def send(self, serialized_metric):
...
     """A transport which wraps another transport and buffers before sending."""
+    MAX_MESSAGE_SIZE = 65507
+
     def __init__(self, transport):
         self.transport = transport
         self.buffer = []
 
     def send(self, serialized_metric):
-        self.buffer.append(serialized_metric)
+        if len(serialized_metric) >= self.MAX_MESSAGE_SIZE:
```
pacejackson: Should this be `>` like it is in `make_messages`?

baseplate/metrics.py
```diff
@@ -85,22 +85,62 @@ def send(self, serialized_metric):
...
+                proposed_len = len(metric)
+
+            if proposed_len > self.MAX_MESSAGE_SIZE:
+                messages.append(b"\n".join(batch))
+                batch = []
+                batch_len = 0
+
+            batch_len += len(metric)
+            if batch:
+                batch_len += 1  # count the newline
```
pacejackson: It might be good to detail this just a touch more, it took me some thinking to realize that this is counting the newline between the previous value in `batch` and the new one, _not_ the newline that comes after the current metric (I thought you were double counting the newline when checking proposed_len for a min).

tests/unit/metrics_tests.py
```diff
@@ -87,6 +87,39 @@ def test_buffered_exception_is_caught(self, mock_make_socket):
...
+        mocket = mock_make_socket.return_value
+        raw_transport = metrics.RawTransport(EXAMPLE_ENDPOINT)
+        transport = metrics.BufferedTransport(raw_transport)
+        transport.send(b"a" * 40)
+        transport.send(b"b" * 40)
+        transport.send(b"c" * 19)
+        self.assertEqual(mocket.sendall.call_count, 0)
+        transport.flush()
+
+        #self.assertEqual(mocket.sendall.call_count, 2)
```
pacejackson: Why is this commented out?

bsimpson63: oops

pacejackson:
@ckwang8128 thoughts? concerns? jubilations?

bsimpson63:
> I seem to recall there being some discussion around this where folks weren't sure if we wanted to start doing this since this acts as a guard against us flooding in metrics.

I think that we're only seeing thing in thing service because we've started instrumenting more parts of the work done in the request, and not that we're manually creating more timers and counters and stuff. If this is the case it is kind of unfair to the application developer to just drop their metrics and spam up their logs.

pacejackson:
I agree :)

I think graphql has similar issues.

baseplate/metrics.py
pacejackson: The idea being that a single metric can be `MAX_MESSAGE_SIZE` since there won't be a newline character added (I think?).

bsimpson63: yep

bsimpson63:
Thing service's problems with this should be fixed by https://github.snooguts.net/reddit/reddit-service-thing/pull/564. 

Is this PR still something worth doing, or should we switch to more detailed logging or reporting to sentry or something like that?

spladug:
Yeah, the message length clearly isn't enough. Looking at the printing you did, the counters were particularly valuable since they were aggregated in-process. Maybe filter it down to the counters and print those in the log line?

spladug:
Also, if the length of the message is pretty short maybe show a different error message with some diagnostics to try and root out what's going on in graphql?

bsimpson63:
💇 updated to just log the counters when this error occurs, which should help in tracking down the issue. We could also try sending those counters and skip the other metrics 🤷‍♂️

bsimpson63:
I looked at graphql's logs and the messages aren't very large:

```
Mar 25 15:15:02 reddit-service-graphql-6cb9d4697b-nhtn2 reddit-service-graphql-server  1:14121020853690728984:baseplate.metrics:ERROR:baseplate metrics batch too large: flush more often                 or reduce amount done in one request, length 2244, [Errno 90] Message too long
Mar 25 15:15:09 reddit-service-graphql-6cb9d4697b-jvxqh reddit-service-graphql-server  1:6622919049354771326:baseplate.metrics:ERROR:baseplate metrics batch too large: flush more often                 or reduce amount done in one request, length 89, [Errno 90] Message too long
Mar 25 15:15:12 reddit-service-graphql-6c489f6f48-rvlzg reddit-service-graphql-server  1:11593467434533555335:baseplate.metrics:ERROR:baseplate metrics batch too large: flush more often                 or reduce amount done in one request, length 1927, [Errno 90] Message too long
```

Something else must be going on there.

baseplate/metrics.py
```diff
@@ -99,8 +99,13 @@ def flush(self):
             self.transport.send(message)
         except socket.error as e:
             logger.error("baseplate metrics batch too large: flush more often \
```
pacejackson: Would it make sense to change this to `logger.exception` so we can get a little more info about the error?

spladug: that'd just give the stack trace, right? is that useful?

pacejackson: For graphql, it would be 😅

bsimpson63: cool, I can make that change

PR #258, opened 2019-03-29
Remove vestigial logger on context

spladug:
AFAICT this was not meant to be a public interface and was instead for
the WrappedRequestContext to do its own internal logging that no longer
is done.

:eyeglasses: @bradengroom @ckwang8128


bradengroom:
🐟 Thanks!

PR #259, opened 2019-03-29
Fix baseplate-tshell for versions of ipython 5+

bradengroom:
Fixes https://github.com/reddit/baseplate.py/issues/245.

Tested with versions 4.0.0 and 6.5.0.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -257,8 +257,9 @@ def load_and_run_tshell():
     try:
         # try to use IPython if possible
         from IPython.terminal.embed import InteractiveShellEmbed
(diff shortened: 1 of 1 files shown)


bradengroom:
👓 @spladug

spladug:
Looks pretty reasonable to me. Do we use <4.0 anywhere (r2 definitely does, but not tshell-able) and if so does it work with this change?

bradengroom:
Just tested with versions 2.2.0 and 5.8.0 since those are in our index. It seems to work fine there.

spladug:
Great, thank you for confirming! :fish:

PR #260, opened 2019-03-29
Activate readline in baseplate-tshell

### LINKED ISSUE #240: baseplate-tshell doesn't activate readline
This means that the REPL doesn't get all the helpful keyboard shortcuts for moving around in text and is a bit of a pain to use. We should `import readline` (safely) to ensure it's there. We may also want to consider loading `.pythonrc`.

spladug:
This only kicks in when IPython isn't available, but it gives much nicer
keyboard movement when editing text.

Fixes #240

Before, when hitting Ctrl+A to move to the beginning of the line:

```
Baseplate Interactive Shell
Python 3.7.3 (default, Mar 26 2019, 00:55:50) 
[GCC 7.3.0]

Available Objects:

  app          This project's app instance
  context      The context for this shell instance's span
>>> foo^A^A^A
```

afterwards the cursor just moves.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -263,5 +263,14 @@ def load_and_run_tshell():
         import code
         newbanner = "Baseplate Interactive Shell\nPython {}\n\n".format(sys.version)
         banner = newbanner + banner
+
+        # import this just for its side-effects (of enabling nice keyboard
+        # movement while editing text)
+        try:
+            import readline
+            del readline
+        except ImportError:
+            pass
+
         shell = code.InteractiveConsole(locals=env)
         shell.interact(banner)


spladug:
:eyeglasses: @bradengroom

bradengroom:
🐟 This looks good to me and will be very nice. Is catching the `ImportError ` necessary though? It looks like `readline` is in the standard library for [2](https://docs.python.org/2/library/readline.html) and [3](https://docs.python.org/3/library/readline.html).

spladug:
I was cargo culting what Pylons / r2 does, but googling a bit it seems that's a defense against readline (the underlying library) not being available on Windows rather than a more subtle unixey issue. Worth playing it safe or remove it?

bradengroom:
Oh interesting. Sounds like we don't need it, but it's also not hurting anything. Doesn't matter to me

PR #261, opened 2019-03-29
Pass extra baseplate-script args to invoked script

bradengroom:
Fixes part 1 of https://github.com/reddit/baseplate.py/issues/58

This allows users to pass additional args to the invoked script. This should continue to work if no additional args are passed. This allows scripts to parse these args as they usually would with argparse:

```
def run(app_config, *args):
  parser = argparse.ArgumentParser()
  # parser.add_arguments
  args = parser.parse_args(args)
  
```

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -252,16 +252,25 @@ def load_and_run_script():
 
 
 def _check_fn_signature(script_fn, fn_args):
-    argspec = inspect.getfullargspec(script_fn)
-    func_sig_arg_count = len(argspec.args[1:])  # slice off app_config arg
-    cli_arg_count = len(fn_args)
-    if argspec.varargs:
+    signature = inspect.signature(script_fn)
+
+    func_sig_arg_count = 0
+    allows_var_args = False
+    for param in signature.parameters:
+        if param in {param.POSITIONAL_ONLY, param.POSITIONAL_OR_KEYWORD}:
+            func_sig_arg_count += 1
+        elif param == param.VAR_POSITIONAL:
+            func_sig_arg_count += 1
+            allows_var_args = True
+
+    cli_arg_count = len(fn_args) + 1  # add one to account for app_config arg
+    if allows_var_args:
         args_match = cli_arg_count >= func_sig_arg_count
     else:
         args_match = cli_arg_count == func_sig_arg_count
     if not args_match:
         raise ValueError('additional args do not match script function signature: %s' %
-                         inspect.formatargspec(*argspec))
+                         signature)
 
 
 def load_and_run_tshell():
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -246,31 +246,35 @@ def load_and_run_script():
     config = read_config(args.config_file, server_name=None, app_name=args.app_name)
     configure_logging(config, args.debug)
 
-    _check_fn_signature(args.entrypoint, args.args)
+    if _fn_accepts_additional_args(args.entrypoint, args.args):
+        args.entrypoint(config.app, args.args)
+    else:
+        args.entrypoint(config.app)
 
-    args.entrypoint(config.app, args.args)
 
-
-def _check_fn_signature(script_fn, fn_args):
+def _fn_accepts_additional_args(script_fn, fn_args):
+    additional_args_provided = len(fn_args) > 0
     signature = inspect.signature(script_fn)
 
-    func_sig_arg_count = 0
+    positional_arg_count = 0
     allows_var_args = False
-    for param in signature.parameters:
-        if param in {param.POSITIONAL_ONLY, param.POSITIONAL_OR_KEYWORD}:
-            func_sig_arg_count += 1
-        elif param == param.VAR_POSITIONAL:
-            func_sig_arg_count += 1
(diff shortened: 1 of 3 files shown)


bradengroom:
👓 @spladug

spladug:
Clever trick using `*args` for backwards compat. First reaction was a bit squicked because it felt like it'd lock us into the args being strings, but seeing your example usage makes it look pretty clean/clear. I think I'm totally on board with this but maybe we should get one or two more sets of eyes on it just to be sure?

:fish:

bradengroom:
👓 @pacejackson @bsimpson63

spladug:
One thing that comes to mind: the error message the user would get if they accidentally put some extra junk on the command line when the script isn't expecting it would be super confusing. Is there anything we could do to detect that case (inspect the entrypoint signature? :grimacing:) and give a more helpful error message?

bradengroom:
That can be done. I suppose it should cover any scenario where the extra arg count doesn't align with the invoked function signature.

pacejackson:
Would that be solved by encouraging writing your entrypoint like this?:

```
def run(app_config, *args):
  parser = argparse.ArgumentParser()
  # parser.add_arguments
  args = parser.parse_args(args)
```

bradengroom:
Yeah that definitely works moving forward. I think @spladug is concerned about existing scripts that haven't been updated yet

pacejackson:
Ah yes, other people have in fact written code before now 😅 
That's a good point!

baseplate/server/__init__.py
```diff
@@ -20,6 +20,14 @@
 import traceback
 import warnings
 
+try:
+    # Python 3
+    from inspect import getfullargspec as getargspec
+except ImportError:
+    # Python 2
+    from inspect import getargspec
```
bradengroom: `getfullargspec ` is pretty much a drop in replacement for `getargspec ` that was introduced in py 3. `getfullargspec` works with the following syntax while `getargspec` does not:
```
# the keyword argument `kwarg='bar'` used between `*args` and `**kwargs` is not valid in python 2.
def foo(app_config, arg, *args, kwarg='bar', **kwargs):
            pass
```

bradengroom:
💇

baseplate/server/__init__.py
spladug: can this go in the `_compat` module please?

baseplate/server/__init__.py
```diff
@@ -208,11 +216,29 @@ def load_and_run_script():
...
+
+    _check_fn_signature(args.entrypoint, args.args)
+
+    args.entrypoint(config.app, args.args)
+
+
+def _check_fn_signature(script_fn, fn_args):
+    argspec = getargspec(script_fn)
+    func_sig_arg_count = len(argspec.args[1:])  # slice off app_config arg
+    cli_arg_count = len(fn_args)
```
spladug: if we're already looking at function signatures, would it make sense to just do two arguments rather than the `*args` thing? then we only need to handle 1 or 2 args on the function signature

baseplate/server/__init__.py
bradengroom: Rebased a removed py 2 support

bradengroom:
Looks like I should be using the `signature()` api instead of `getfullargspec()` if we are only supporting 3.6+

baseplate/server/__init__.py
bradengroom: I think you're saying instead of defining the script function as:
```
def run(app_config: AppConfig, *args: str):
    pass
```

You'd prefer this?
```
def run(app_config: AppConfig, args: List[str]):
    pass
```

spladug: yessir

spladug: (since that leaves us better open to future arguments should we need them, and is a bit easier to work with as an application)

bradengroom: I'm working on this, but I'm starting to think that this might actually end up being more complicated code. We have to have a check to determine if the script function will accept the second arg and conditionally pass it in if we want to be backwards compatible.

bradengroom:
💇

docs/cli/script.rst
```diff
@@ -38,15 +38,19 @@ and a small script, ``printer.py``:
...
 .. code-block:: python
 
-   def run(app_config):
-       print(app_config["message"])
+   def run(app_config, args):
+       parser = argparse.ArgumentParser()
+       parser.add_argument("name")
+       args = parser.parse_args(args)
+
+       print('%s %s' % (app_config["message"], args.name))
```
spladug: % formatting in our beautiful python 3 future!?!?!? :P

baseplate/server/__init__.py
```diff
@@ -237,11 +238,45 @@ def load_and_run_script():
     parser.add_argument(
         "entrypoint", type=_load_factory, help="function to call, e.g. module.path:fn_name"
     )
+    parser.add_argument(
```
spladug: i wonder if it'd be more error-message-friendly to `parse_known_args` then determine if the entrypoint accepts more args and re-parsing accordingly. :no_bell:

spladug:
Merged as d79ddda5a566019055dd537e80822c932c911d9f. Thanks!

PR #262, opened 2019-03-31
Lots of lint: pydocstyle, pylint, and pycodestyle upgrade

spladug:


spladug:

baseplate/context/redis.py
```diff
@@ -72,8 +73,8 @@ class RedisContextFactory(ContextFactory):
     def __init__(self, connection_pool):
         self.connection_pool = connection_pool
 
-    def make_object_for_context(self, name, server_span):
-        return MonitoredRedisConnection(name, server_span, self.connection_pool)
+    def make_object_for_context(self, name, span):
```
pacejackson: I like this but it _is_ a breaking change if anyone called this with kwargs so we'll have to watch out for that.

baseplate/core.py
```diff
@@ -39,15 +39,12 @@ class SpanObserver(object):
 
     def on_start(self):
         """Do something when the observed span is started."""
-        pass
```
pacejackson: Why was this `pass` removed?  That seems _less_ intuitive.

baseplate/diagnostics/tracing/publisher.py
```diff
@@ -3,12 +3,12 @@
...
 
-from . import MAX_SPAN_SIZE, MAX_QUEUE_SIZE
 from baseplate import config, metrics_client_from_config
 from baseplate._compat import configparser
+from baseplate._compat import urlparse
+from baseplate.diagnostics.tracing import MAX_SPAN_SIZE, MAX_QUEUE_SIZE
 from baseplate.message_queue import MessageQueue, TimedOutError
 from baseplate.retry import RetryPolicy
-from baseplate._compat import urlparse
 from baseplate._utils import BatchFull, RawJSONBatch, TimeLimitedBatch
```
pacejackson: why not use the same style of imports in this file as the rest of baseplate?

baseplate/events/queue.py
```diff
@@ -26,12 +26,15 @@
 from thrift import TSerialization
 from thrift.protocol.TJSONProtocol import TJSONProtocolFactory
 
-from . import MAX_EVENT_SIZE, MAX_QUEUE_SIZE
 from baseplate.context import ContextFactory
 from baseplate.message_queue import MessageQueue, TimedOutError
 from baseplate._utils import warn_deprecated
```
pacejackson: Same deal, maybe we should take this opportunity to have these files import the same way as other files.

baseplate/experiments/providers/base.py
```diff
@@ -7,7 +7,7 @@
 class Experiment(object):
     """Base interface for experiment objects."""
 
-    def get_unique_id(self, **kwargs):
+    def get_unique_id(self, **_kwargs):
```
pacejackson: I don't understand this change, there are other methods that _don't_ have this change.

baseplate/experiments/providers/base.py
```diff
@@ -7,7 +7,7 @@
...
         """Generate a unique ID for this experiment with the given inputs.
 
         Used to determine if a bucketing ivent has alread been fired for a
```
pacejackson: looks like there's a type we can fix in here.

baseplate/experiments/providers/r2.py
```diff
@@ -57,6 +57,7 @@ def max_variant_size(variant_size, num_variants):
           created or when an LoID cookie was generated.
     """
 
+    # pylint: disable=redefined-builtin
```
pacejackson: Can we just ignore this one globally?  I'm not sure how useful it is.

baseplate/metrics.py
```diff
@@ -335,15 +335,14 @@ class BatchCounter(Counter):
...
+        counter.flush()
 
     The above example results in one packet indicating an increase of two
     should be applied to "counter_name".
     """
 
     def __init__(self, transport, name):
-        self.transport = transport
-        self.name = name
+        super(BatchCounter, self).__init__(transport, name)
```
pacejackson: 👍

baseplate/core.py
spladug: pylint has a lint for "unnecessary pass" when something like a docstring has already made it a syntactically valid indenty block.

baseplate/diagnostics/tracing/publisher.py
spladug: I'd like to move to absolute imports only. Do you prefer the relative ones?

baseplate/experiments/providers/base.py
spladug: the arg is unused, _ prefix makes that explicit for the linter and humans

baseplate/diagnostics/tracing/publisher.py
pacejackson: I do not!  I prefer absolute, but I also like consistency 😅 

If we want to move to absolute, this might be a good time to do that but it's also fine to kick that can.

baseplate/experiments/providers/base.py
pacejackson: It's not un-used by subclasses though

pacejackson: This is just an base/interface, not something that is used directly.

spladug: yeah, fair enough. it's probably because there's actually some code in there (it's not _just_ an abstract method. it's only abstract when a condition passes). i'll just pragma it away.

baseplate/diagnostics/tracing/publisher.py
spladug: I solemnly swear to not kick that can too far down the road but I don't want this PR to balloon too much or the merge conflicts will get insane the longer it rots.

baseplate/experiments/providers/r2.py
spladug: I'm sorta ambivalent, but the example given here does seem somewhat like it'd be pretty annoying to run into: https://docs.quantifiedcode.com/python-anti-patterns/correctness/assigning_to_builtin.html

(I just wish they hadn't claimed the name `id` for themselves. Grr.)

baseplate/context/redis.py
spladug: gp. I've done a quick pass on code search in our org and don't see any uses with kwargs, but will note this as a breaking change in the changelog.

PR #263, opened 2019-04-03
A few misc. touchups

### LINKED ISSUE #239: python -m baseplate.message_queue --read prints out b"" strings on Python 3
We should decode and print that out so we don't end up with repr'd b-strings. Very confusing and hard to integrate with pipelines!

spladug:
The first fixes #239 and the latter two are just tidiness.


spladug:

PR #264, opened 2019-04-04
Add country code to edge request context

jesseyli:
👓 @spladug @pacejackson 

Adds country code as a field on the edge request context that gets forwarded on thrift calls. This way upstream services will be able to access the values we get from the CDN. I've thought about adding a place to put generic data into the edge request context, but so far it doesn't seem like we have a need for anything beyond region, country code, and ip.


baseplate/core.py
```diff
@@ -480,6 +480,8 @@ def new(self,
         :param int loid_created_ms: (Optional) Epoch milliseconds when the
             current LoID cookie was created.
         :param str session_id: (Optional) ID for the current session cookie.
+        :param str country_code: (Optional) A two-character ISO 3166-1
+            country code.
```
pacejackson: Can/should we do some validation on this to make sure it's valid?

baseplate/thrift/baseplate.thrift
```diff
@@ -71,4 +71,5 @@ struct Request {
     1: Loid loid;
     2: Session session;
     3: AuthenticationToken authentication_token;
+    4: string country_code;
```
pacejackson: It might make sense to wrap this in a struct for grouping this with other similar data?

jesseyli: Sounds reasonable, how does this sound?
```
struct Client {
    1: string country_code;
}
```

spladug: Isn't the entire edge context about the client?

jesseyli: That's true, based on that I actually don't think we need a struct for this

pacejackson: Sgtm

baseplate/core.py
```diff
@@ -581,6 +584,10 @@ def service(self):
         """:py:class:`~baseplate.core.Service` object for the current context."""
         return Service(self.authentication_token)
 
+    @cached_property
```
spladug: is there any value to this being a `cached_property`? the underlying `_t_request` already is and then from there it's just an attribute access.

baseplate/thrift/baseplate.thrift
spladug: It could be a GeoLocation struct or something like that so that if we changed from ISO country code to some other format down the line we'd have a place to put the multiple fields.

spladug: A custom type alias might be nice too. Just to make the value of the field super obvious.

spladug:
Note: I've put out a request to our privacy team to see what constraints / documentation needs to be added to this field.

jesseyli:
One year later 😅...

baseplate/thrift/baseplate.thrift
```diff
@@ -22,7 +22,7 @@ service BaseplateService {
 }
 
 
-/** The components of the Reddit LoID cookie that we want to propogate between
+/** The components of the Reddit LoID cookie that we want to propagate between
```
spladug, line 25: oof. thanks!

baseplate/thrift/baseplate.thrift
```diff
@@ -87,11 +87,30 @@ struct OriginService {
...
+
 /**A raw authentication token as returned by the authentication service.
 
 */
 typedef string AuthenticationToken
 
+/** A two-character ISO 3166-1 country code
+
+*/
+typedef string CountryCode
```
pacejackson, line 112: Should the typedef come before `Geolocation`?

jesseyli: How about we move all typedefs to after the namespace declarations?

jesseyli:
Weird, I got an intermittent failure here: https://cloud.drone.io/reddit/baseplate.py/831

PR #266, opened 2019-04-09
Add engine_from_config to sqlalchemy integration

pacejackson:
This allows us to specify the different connection params and fetch the password from a SecretsStore.

👓 @spladug @bsimpson63

### DIFF
--- a/baseplate/context/sqlalchemy.py
+++ b/baseplate/context/sqlalchemy.py
@@ -61,7 +61,8 @@ def engine_from_config(app_config, secrets=None, prefix="database.", **kwargs):
 
     if options.url:
         return create_engine(options.url)
-    elif not options.drivername:
+
+    if not options.drivername:
         raise config.ConfigurationError(
             "drivername",
             AttributeError("'drivername' is required if 'url' is not set"),
--- a/baseplate/context/sqlalchemy.py
+++ b/baseplate/context/sqlalchemy.py
@@ -56,21 +56,20 @@ def engine_from_config(app_config, secrets, prefix="database.", **kwargs):
             "query": config.DictOf(config.String),  # optional
         }})
     options = getattr(cfg, config_prefix)
-    url_params = dict(drivername=options.drivername)
+    kwargs.setdefault("drivername", options.drivername)
     if options.username:
-        url_params["username"] = options.username
+        kwargs.setdefault("username", options.username)
     if options.password_secret:
-        url_params["password"] = secrets.get_simple(options.password_secret).decode("utf-8")
+        kwargs.setdefault("password", secrets.get_simple(options.password_secret).decode("utf-8"))
     if options.host:
-        url_params["host"] = options.host
+        kwargs.setdefault("host", options.host)
     if options.port:
-        url_params["port"] = options.port
+        kwargs.setdefault("port", options.port)
     if options.database:
-        url_params["database"] = options.database
+        kwargs.setdefault("database", options.database)
     if options.query:
-        url_params["query"] = options.query
-    url_params.update(kwargs)
-    return create_engine(URL(**url_params))
+        kwargs.setdefault("query", options.query)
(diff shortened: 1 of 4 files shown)


baseplate/context/sqlalchemy.py
```diff
@@ -3,14 +3,70 @@
...
 from sqlalchemy.orm import Session
 
-from ..context import ContextFactory
-from ..core import (
-    ServerSpan,
-    SpanObserver,
-)
+from baseplate import config
+from baseplate.context import ContextFactory
+from baseplate.core import ServerSpan, SpanObserver
```
pacejackson: @spladug how do you feel about me cleaning up these imports while I'm already in here?

spladug: i feel good. very good.

baseplate/context/sqlalchemy.py
```diff
@@ -3,14 +3,70 @@
...
+    * ``database`` (optional): the database name.
+    * ``query`` (optional): a dictionary of options to be passed to the dialect
+        and/or the DBAPI upon connect.
+
+    """
+    assert prefix.endswith(".")
+    config_prefix = prefix[:-1]
+    cfg = config.parse_config(app_config, {
+        config_prefix: {
+            "drivername": config.String,
```
spladug: why not allow the url format?

pacejackson: I kinda feel like that might get confusing about what you are supposed to set?  You can still use `engine_from_config` straight from sqlalchemy if you want which just uses the URL but then you have to embed the password in the URL.

pacejackson:
💇‍♂️ 
I changed it so you can either use `database.url` or `database.drivername`.  If you set the url, it just uses that, it does no go so far to see if you set anything else and raise an error if you do.

As a result, I also moved to make `secrets` an optional argument unless you actually set `password_secret`.

LMK if you think this makes sense or if it's too many conditional requirements (which was my initial concern).

spladug:
What about parsing the URL first then allowing any other config fields to override then kwargs, in that order. That seems like it'd make the most sense (since the common case would be url + password from secret).

baseplate/context/sqlalchemy.py
```diff
@@ -3,14 +3,88 @@
...
+
+    """
+    assert prefix.endswith(".")
+    config_prefix = prefix[:-1]
+    cfg = config.parse_config(app_config, {
+        config_prefix: {
+            "url": config.Optional(config.String),
+            "drivername": config.Optional(config.String),
+            "username": config.Optional(config.String),
+            "password_secret": config.Optional(config.String),
```
spladug: I think we probably want to have username be in here too. If the credentials are being rotated, that'd generally be done by changing user to a new one with a new password, rolling everyone off the old one, then killing the old user. Right?

spladug: (do we want to add a "credentials" secret type?)

pacejackson: 🤔

pacejackson: https://github.com/reddit/baseplate.py/pull/267

pacejackson:
> What about parsing the URL first then allowing any other config fields to override then kwargs, in that order. That seems like it'd make the most sense (since the common case would be url + password from secret).

That seems like it would get really confusing really quickly other than the case of overwriting the username/password?

pacejackson:
Maybe the thing to do is to _only_ support url + optional credentials? @spladug

pacejackson:
💇‍♂️ 
This simplifies it to only support `url` and `credentials` (note the tests will fail because I'm starting to use code from #267 ).

bsimpson63:
🐟 this looks great

pacejackson:
💇‍♂️ 

Screen Shot 2019-04-10 at 5 14 49 PM

PR #267, opened 2019-04-10
Add CredentialSecrets

pacejackson:
Add a new secrets type for storing "credentials" (that is a username/password pair).

### DIFF
--- a/baseplate/secrets/__init__.py
+++ b/baseplate/secrets/__init__.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 
 from .store import (
     CorruptSecretError,
+    CredentialSecret,
     SecretNotFoundError,
     SecretsNotAvailableError,
     SecretsStore,
@@ -17,6 +18,7 @@ from .store import (
 
 __all__ = [
     "CorruptSecretError",
+    "CredentialSecret",
     "SecretNotFoundError",
     "SecretsNotAvailableError",
     "SecretsStore",
--- a/baseplate/secrets/__init__.py
+++ b/baseplate/secrets/__init__.py
@@ -7,7 +7,6 @@ from __future__ import unicode_literals
 
 from .store import (
     CorruptSecretError,
-    CredentialSecret,
     SecretNotFoundError,
     SecretsNotAvailableError,
     SecretsStore,
@@ -18,7 +17,6 @@ from .store import (
 
 __all__ = [
     "CorruptSecretError",
-    "CredentialSecret",
     "SecretNotFoundError",
     "SecretsNotAvailableError",
     "SecretsStore",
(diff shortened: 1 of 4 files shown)


baseplate/secrets/store.py
```diff
@@ -142,6 +160,53 @@ def get_raw(self, path):
...
+        encoding = secret_attributes.get("encoding", "identity")
+
+        if encoding != "identity":
+            raise CorruptSecretError(path, "secret has encoding=%s rather than encoding=identity" % encoding)
+
+        missing_keys = []
+        values = {}
+        for key in ("username", "password"):
+            try:
+                values[key] = str(secret_attributes[key])
```
pacejackson: I wasn't 100% sure what the best way to handle this was.  I don't feel like it makes sense to turn these values into bytes like we do for other secrets since these are text by their nature.  It seemed wrong to put the `str`-ing into `_decode_secret` since that is expected to return `bytes` so it seemed like the thing to do is just `str` the value directly here.

pacejackson:
👓 @spladug @bradengroom

bradengroom:
🐟 This seems reasonable to me

baseplate/secrets/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 from .store import (
     CorruptSecretError,
+    CredentialSecret,
```
spladug: we probably don't need to explicitly export this. versioned secret is in here because of the `VersionedSecret.from_simple_secret` API but otherwise it won't be common for users to manually instantiate a `CredentialSecret` vs. getting one from `get_credentials()`

baseplate/secrets/store.py
spladug: Isn't it already a (unicode) string? This value is coming from `json.load` right?

pacejackson: I imagine it could _technically_ be an int/float/boolean as well?  Although maybe not from vault.

spladug: That sounds like a `CorruptSecretError` to me!

baseplate/secrets/store.py
```diff
@@ -142,6 +161,58 @@ def get_raw(self, path):
...
+                values[key] = val
+            except KeyError:
+                missing_keys.append(key)
+
+        if missing_keys:
+            if len(missing_keys) > 1:
+                message = "keys %s" % (tuple(missing_keys),)
+            else:
+                message = "key '%s'" % missing_keys[0]
+            raise CorruptSecretError(path, "secret does not have %s" % message)
```
pacejackson: Maybe I shouldn't get fancy with this and just `raise` at the first `KeyError`?

pacejackson:
💇‍♂️

baseplate/secrets/store.py
spladug: oh wow, even with proper pluralization. yeah that's probably a bit overkill for an exception message hah. up to you.

PR #270, opened 2019-04-12
Monkeypatch ResponseFuture.result() to wait for callbacks

bsimpson63:
Switching from composition to monkeypatching. `execute_concurrent_with_args` uses the `ResponseFuture` returned by our `session` object so that object needs to behave like a regular `ResponseFuture` object.

I added a test that actually executes `execute_concurrent_with_args` and have tested out the code locally in thing service (by copying baseplate/context/cassandra.py into thing and importing it).

👓 @spladug @pacejackson

### DIFF
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -105,7 +105,9 @@ def wrap_future(response_future, callback_fn, callback_args, errback_fn, errback
     complete before a result will be returned from ResponseFuture.result().
     They are not given precedence over other callbacks or errbacks, so if
     another callback triggers the response from the service (and the server
-    span is closed) the special callback might not complete.
+    span is closed) the special callback might not complete. The special
+    callback is added first and  callbacks are executed in order, so
+    generally the special callback should finish before any other callbacks.
 
     This fixes a race condition where the server span can complete before
     the callback has closed out the child span.
--- a/baseplate/context/cassandra.py
+++ b/baseplate/context/cassandra.py
@@ -7,7 +7,7 @@ import logging
 from threading import Event
 
 # pylint: disable=no-name-in-module
-from cassandra.cluster import Cluster, _NOT_SET
+from cassandra.cluster import Cluster, _NOT_SET, ResponseFuture
 # pylint: disable=no-name-in-module
 from cassandra.query import SimpleStatement, PreparedStatement, BoundStatement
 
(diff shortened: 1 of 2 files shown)


baseplate/context/cassandra.py
```diff
@@ -98,39 +98,34 @@ def make_object_for_context(self, name, span):
...
 
-    The callback_fn and errback_fn passed in the constructor are given
-    special treatment: they must be complete before a result will be
-    returned from result(). They are not given precedence over other
-    callbacks or errbacks, so if another callback triggers the response
-    from the service (and the server span is closed) the special callback might
-    not complete.
+    The callback_fn and errback_fn are given special treatment: they must be
+    complete before a result will be returned from ResponseFuture.result().
+    They are not given precedence over other callbacks or errbacks, so if
```
spladug: but we can generally rely on getting the first callback in and so finishing first, right?

bsimpson63: yeah, I'll update the comment to state that

bsimpson63:
> i got scared when i saw "monkeypatch" but just doing the future instance like this seems pretty reasonable. scary, but reasonable! :)

Yeah 🤷‍♂️. The other options I thought about were:
1. do some nasty stuff with `__getattr__` and `__setattr__` to pass through to the wrapped future
2. use inheritance and then copy the `ResponseFuture` into a new instance of my new class
3. use inheritance and then modify the `__class__` of the `ResponseFuture` object

but all those seemed worse and more brittle than this patching approach.

spladug:
agreed this is the least-worst of those options. could you expand a bit on why composition failed?

bsimpson63:
> could you expand a bit on why composition failed

Our `CassandraSessionAdapter` is used to get futures from `execute_async` https://github.com/datastax/python-driver/blob/master/cassandra/concurrent.py#L135. Then some callbacks are added to handle the result https://github.com/datastax/python-driver/blob/master/cassandra/concurrent.py#L137. This is what I initially attempted to fix by adding passthroughs to the real future object on these callback methods.

The bigger problem arose because the future is then passed into `ResultSet` https://github.com/datastax/python-driver/blob/master/cassandra/concurrent.py#L156, and looking at that class https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L4342 you can see that it accesses lots of attributes and methods of `ResponseFuture`. Trying to plumb all that through to the wrapped object seemed futile and error prone.

PR #271, opened 2019-04-16
Add secret name to error log for baseplate

manishapme:
Open JIRA ticket requested more info for this type of error. 👓 @rram  @spladug


manishapme:
👓 @bsimpson63

baseplate/secrets/fetcher.py
```diff
@@ -282,13 +282,19 @@ def is_about_to_expire(self):
...
+        try:
+            response = self.session.get(
+                urljoin(self.base_url, posixpath.join("v1", secret_name)),
+                headers={
+                    "X-Vault-Token": self.token,
+                },
+            )
+            response.raise_for_status()
+        except requests.HTTPError:
+            logger.exception("Vault client failed to get secret: %s",
```
bsimpson63: can you switch this to `logger.error` so you're not duplicate logging the stack trace (it'll happen here and also when the exception is raised)

bsimpson63:
💅

baseplate/secrets/fetcher.py
spladug: it might also be nice to include the status code or error message from the response if one exists.

baseplate/secrets/fetcher.py
```diff
@@ -282,13 +282,19 @@ def is_about_to_expire(self):
...
-        response.raise_for_status()
+        try:
+            response = self.session.get(
+                urljoin(self.base_url, posixpath.join("v1", secret_name)),
+                headers={
+                    "X-Vault-Token": self.token,
+                },
+            )
+            response.raise_for_status()
+        except requests.HTTPError:
```
manishapme: I do have a question here. In the documentation it looks like the exception would be `requests.exceptions.HTTPError`, but on this [page](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status)) it says it calls the HTTPError, and links to `https://2.python-requests.org/en/master/api/#requests.HTTPError`. So I just want to make sure this is the correct syntax, as there is no easy way to test.

manishapme:
💇

baseplate/secrets/fetcher.py
spladug: The exception is defined in `requests.exceptions` so that's its canonical name, but it is also [imported into `requests`](https://github.com/kennethreitz/requests/blob/master/requests/__init__.py#L120) so available as `requests.HTTPError` as well. It's the same class object either way so it doesn't matter which name you use.

You should be able to test this in the OneVM by writing a short config file for a secret fetcher that points at the test Vault available there. See the integration tests here in baseplate for example usage.

baseplate/secrets/fetcher.py
```diff
@@ -282,13 +282,19 @@ def is_about_to_expire(self):
...
+        try:
+            response = self.session.get(
+                urljoin(self.base_url, posixpath.join("v1", secret_name)),
+                headers={
+                    "X-Vault-Token": self.token,
+                },
+            )
+            response.raise_for_status()
+        except requests.HTTPError as e:
+            logger.error("Vault client failed to get secret: %s, %s",
```
spladug: nit: I think a : between the format strings would be easier to read:

    Vault client failed to get secret: secrets/foo/bar, Not found

vs.

    Vault client failed to get secret: secret/foo/bar: Not found

manishapme:
💇  per convo with @spladug, testing this with onevm would require writing a new authtype for `fetch_instance_identity`. not in scope for this change.

did make fix to readability.

also, i don't have permissions to restart drone, appears to be failing on a random number test.

PR #272, opened 2019-04-16
Update secrets.store.secrets_store_from_config

pacejackson:
* Allow the caller to specify a `prefix`
 * Allow the file timeout to be set in the config file as well as a keyword argument

👓 @spladug @bradengroom


bradengroom:
🐟

pacejackson:
All I had in mind was letting you update this without updating code 😅 .

spladug:
My concern there is that means that _all_ callsites in your application will get updated by that when you may instead only want it in e.g. a cron job but not an in-request thingy. Maybe that's overthinking though since we should generally be doing this in app startup.

pacejackson:
💇‍♂

PR #273, opened 2019-04-16
catch TypeError in experiment targeting and treat as non-match

Kaitaan:
when targeting on strings and attempting to match a long integer,
a TypeError can be thrown. In those cases, catch it and treat as
a non-match.

### DIFF
--- a/baseplate/experiments/targeting/tree_targeting.py
+++ b/baseplate/experiments/targeting/tree_targeting.py
@@ -51,9 +51,12 @@ class EqualNode(Targeting):
 
     def evaluate(self, **kwargs):
         candidate_value = kwargs.get(self._accepted_key)
-        if candidate_value in self._accepted_values:
-            return True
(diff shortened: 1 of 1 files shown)


baseplate/experiments/targeting/tree_targeting.py
```diff
@@ -51,9 +51,12 @@ def __init__(self, input_node):
...
     def evaluate(self, **kwargs):
         candidate_value = kwargs.get(self._accepted_key)
-        if candidate_value in self._accepted_values:
-            return True
-
+        try:
+            if candidate_value in self._accepted_values:
+                return True
+        except TypeError:
+            pass
```
pacejackson: Should this at least have a `logger.warn` or something similar so people don't wonder why their experiments are silently failing?

Kaitaan:
💇‍♂️

PR #277, opened 2019-05-07
Add java namespace to baseplate.thrift

fishy:
There's no way to specify java namespace to thrift compiler command,
line, and without proper java namespace the generated java code will be
kinda unusable, especially when this file is imported by other thrift
files.

The added java namespace should be a no-op for non-java thrift compiler
actions.


fishy:
👓 @spladug

PR #280, opened 2019-05-11
Check trace header sampled value in both byte and str

suzie-su:
Regarding to the change before https://github.com/reddit/baseplate.py/pull/223/files#diff-d1382533c32ca95376190e31679acc05R157 made the str compare with byte, but which is incompatible with python3. so we need to compare with byte and also compare with python3 str

### DIFF
--- a/baseplate/integration/pyramid/__init__.py
+++ b/baseplate/integration/pyramid/__init__.py
@@ -218,7 +218,7 @@ class BaseplateConfigurator(object):
 
     def _get_trace_info(self, headers):
         extracted_values = TraceInfo.extract_upstream_header_values(TRACE_HEADER_NAMES, headers)
-        sampled = bool(extracted_values.get("sampled") == b"1")
+        sampled = bool(extracted_values.get("sampled") in {b"1", "1"})
         flags = extracted_values.get("flags", None)
         return TraceInfo.from_upstream(
             int(extracted_values["trace_id"]),


baseplate/integration/pyramid/__init__.py
```diff
@@ -218,7 +218,7 @@ def _start_server_span(self, request, name, trace_info=None):
 
     def _get_trace_info(self, headers):
         extracted_values = TraceInfo.extract_upstream_header_values(TRACE_HEADER_NAMES, headers)
```
pacejackson: I think the fix is to change `extract_upstream_header_values` to always return the `bytes` form rather than checking if it's `bytes` or `str`

spladug: Or always unicode. Either way. Encoding issues like this are always a bit dodgy. We should be sure we understand what cases causes it to be bytes sometimes and unicode other.

suzie-su: @pacejackson @spladug  or can I just change it to `int(extracted_values.get("sampled")) == 1` 
If I change some `extract_upstream_header_values ` function maybe some services are using this function and use the result to compare to str/bytes and break something?

I do not have any idea what cause it to be bytes.

pacejackson: I _think_ it might just be a mistake?  Thrift returns `bytes` and when this was implemented, they may have assumed that was the case for HTTP as well, I think the core issue is that `extract_upstream_header_values` is returning inconsistent types and that is the issue to address.

pacejackson: The reason this is fine in python2 is that it probably returns `str` in python 2 where `"1" == b"1"`

spladug:
Thanks for your patch. I've merged d75c146804b04eb2907b0d67d8f6b20f233f5740 which should fix this by just dropping my `b"1"` idiocy.

PR #281, opened 2019-05-13
Expect 'sampled' header to be unicode already

### LINKED ISSUE #278: Upstream request.headers trace_info `sampled` set as str not able to send trace in python3
When upstream request.headers trace_info `sampled` set as str,
in python3 apps, when execute `1` == b`1` [here](https://github.com/reddit/baseplate.py/blob/07adf4598972808a171064d2d6d4f8501ca8caa4/baseplate/integration/pyramid/__init__.py#L221) it will return False and set the `server_span.context.trace.sampled` to `False`, so not able to send to sentry. https://github.com/reddit/baseplate.py/blob/07adf4598972808a171064d2d6d4f8501ca8caa4/baseplate/integration/pyramid/__init__.py#L221

spladug:
I added a print statement to L221 of baseplate/integration/pyramid/__init__.py and confirmed that the header is always a unicode object by the time it gets there in both Py2.7 and Py3.7. I don't know why I did this with a byte string, other than potentially stupidly mimicing the `.decode("utf8")` that was there before my change. That apparently was an unnecessary no-op.

This should fix #278 by properly parsing the header.

cc: @suzie-su

### DIFF
--- a/baseplate/integration/pyramid/__init__.py
+++ b/baseplate/integration/pyramid/__init__.py
@@ -218,7 +218,7 @@ class BaseplateConfigurator(object):
 
     def _get_trace_info(self, headers):
         extracted_values = TraceInfo.extract_upstream_header_values(TRACE_HEADER_NAMES, headers)
-        sampled = bool(extracted_values.get("sampled") == b"1")
+        sampled = bool(extracted_values.get("sampled") == "1")
         flags = extracted_values.get("flags", None)
         return TraceInfo.from_upstream(
             int(extracted_values["trace_id"]),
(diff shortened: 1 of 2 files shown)


pacejackson:
I think the question I have is should `TraceInfo.extract_upstream_header_values` always return the same type?  I think the reason this started checking against `bytes` is because that's what it returns for Thrift headers while it's unicode for Pyramid headers.  Do we want to normalize that in `extract_upstream_header_values` so it always returns either `bytes` or `unicode`?  Or do we want that to be a part of the thrift vs pyramid integration (what this PR is doing).

spladug:
Hmm, I hear ya. I worry that if we put some requirement that the wire contents are unicode compatible it'll limit us, but that's probably YAGNI. Honestly, I don't like that function at all and wish we were still completely separate on header parsing for the various integrations.

PR #283, opened 2019-05-14
Do a first pass of removing old Python compatibility shims

spladug:


baseplate/crypto.py
```diff
@@ -46,30 +42,7 @@
...
-                "parameters are of the same type.")
-
-        actual_len = len(actual)
-        expected_len = len(expected)
-        result = actual_len ^ expected_len
-        if expected_len > 0:
-            for i in xrange(actual_len):  # noqa: F821 pylint: disable=undefined-variable
-                result |= ord(actual[i]) ^ ord(expected[i % expected_len])
-        return result == 0
+constant_time_compare = hmac.compare_digest
```
pacejackson: Do we even need to keep this around?

baseplate/file_watcher.py
```diff
@@ -101,10 +88,6 @@ def __init__(self, path, parser, timeout=None, binary=False, encoding=None,
         self._mtime = 0
         self._data = _NOT_LOADED
         self._mode = "rb" if binary else "r"
-        # Since Python 2 does not support these kwargs, we store them as a dict
-        # that we `**` in the call to `open` in `get_data` so we do not have to
-        # call `open` in different ways depending on the Python version.  This
-        # can change if/when Python 2 support is dropped.
         self._open_options = {}
```
pacejackson: Note, we can keep doing this or just store them as separate fields and pass them to open directly, i.e.:

```python
with open(self._path, self._mode, encoding=self._encoding, newline=self._newline) as f:
```

One thing that's kind of nice about this though is we don't have to care about what the default values are in python, so maybe it's worth maintaining this?

spladug: Yeah I kinda like this, it also keeps all the open-related options in one place rather than cluttering up the instance.

baseplate/crypto.py
spladug: Probably not, but I wasn't breaking external compatibility yet.

baseplate/file_watcher.py
pacejackson: SGTM

baseplate/context/memcache/lib.py
```diff
@@ -114,15 +102,12 @@ def dump_and_compress(key, value):  # pylint: disable=unused-argument
         :raises ValueError: if `value` is not JSON serializable
 
         """
-        if isinstance(value, string_types):
+        if isinstance(value, str):
```
bsimpson63: does this mean that bytestrings will get serialized as JSON?

baseplate/context/memcache/lib.py
```diff
@@ -229,15 +211,12 @@ def pickle_and_compress(key, value):  # pylint: disable=unused-argument
         :returns: value serialized as str, flags int.
 
         """
-        if isinstance(value, string_types):
+        if isinstance(value, str):
```
bsimpson63: bytestrings will get serialized as pickles?

baseplate/context/memcache/lib.py
spladug: this shouldn't be any change in behavior since `string_types` was defined as `(str,)` on py3 already https://github.com/reddit/baseplate.py/pull/283/files#diff-b1da7a5976a089e9c05aae386c2586e3L19

bsimpson63: ahh ok cool

tests/integration/pyramid_tests.py
```diff
@@ -1,9 +1,5 @@
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
 
 # webtest doesn't play well with unicode literals for headers on py2 :(
```
bsimpson63: remove?

bsimpson63:
🐟 looks great

PR #288, opened 2019-05-30
Add mypy infrastructure and type up the metrics module

spladug:
This adds the basic requirements and makefile infrastructure for running mypy on baseplate.py. The metrics module is the first ~victim~ place to get annotated up. Sphinx will now pull in type annotations instead of having to do it manually in the docstring. I've also made mypy spit out an HTML report that looks like this:

Screenshot from 2019-05-30 14-41-14

which should help us track our progress on typing everything up.

Notes:

* We're targeting Python 3.6 so we can't use `from __future__ import annotations` so in places that need forward declarations of types, we just use strings. This feels double plus ungood.
* I've turned off all failed import warnings. Maybe this is too broad and I should explicitly turn off the ones we don't have? flake8/pylint will warn us of true typos though.
* None of the strictness settings are on yet as we don't have anywhere near full coverage. Hopefully that can go crazy later on.
* The `LoggingLogAdapter` change is because mypy doesn't like the import in try/except thing. :man_shrugging:   

:eyeglasses: @pacejackson @bradengroom @bsimpson63

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -43,116 +43,123 @@ ends.
 
 """
 
+import collections
 import logging
 import socket
 import time
 
-from collections import defaultdict
+from types import TracebackType
+from typing import List, Dict, DefaultDict, Optional, Type
+
+from baseplate.config import EndpointConfiguration
 
 
 logger = logging.getLogger(__name__)
 
 
-def _metric_join(*nodes):
+def _metric_join(*nodes: bytes) -> bytes:
     return b".".join(node.strip(b".") for node in nodes)
 
 
 class Transport:
-    def send(self, serialized_metric):
+    def send(self, serialized_metric: bytes) -> None:
+        raise NotImplementedError
+
+    def flush(self) -> None:
         raise NotImplementedError
 
 
 class NullTransport(Transport):
     """A transport which doesn't send messages at all."""
 
-    def send(self, serialized_metric):
+    def send(self, serialized_metric: bytes) -> None:
         for metric_line in serialized_metric.splitlines():
             logger.debug("Would send metric %r", metric_line)
+
+    def flush(self) -> None:
+        pass
 
 
 class RawTransport(Transport):
     """A transport which sends messages on a socket."""
 
-    def __init__(self, endpoint):
+    def __init__(self, endpoint: EndpointConfiguration):
         self.socket = socket.socket(endpoint.family, socket.SOCK_DGRAM)
         self.socket.connect(endpoint.address)
 
-    def send(self, serialized_metric):
+    def send(self, serialized_metric: bytes) -> None:
         self.socket.sendall(serialized_metric)
 
+    def flush(self) -> None:
+        pass
 
-class BufferedTransport:
+
+class BufferedTransport(Transport):
     """A transport which wraps another transport and buffers before sending."""
 
-    def __init__(self, transport):
+    def __init__(self, transport: Transport):
         self.transport = transport
-        self.buffer = []
+        self.buffer: List[bytes] = []
 
-    def send(self, serialized_metric):
+    def send(self, serialized_metric: bytes) -> None:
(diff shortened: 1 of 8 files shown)


baseplate/metrics.py
```diff
@@ -260,60 +269,63 @@ def stop(self):
...
 class Counter:
     """A counter for counting events over time."""
 
-    def __init__(self, transport, name):
+    def __init__(self, transport: Transport, name: bytes):
         self.transport = transport
         self.name = name
 
-    def increment(self, delta=1, sample_rate=1.0):
+    def increment(self, delta: float = 1, sample_rate: float = 1.0) -> None:
```
pacejackson: if delta is a float, should the default be `1.0`?

pacejackson: It looks like this happens in several places.

baseplate/metrics.py
```diff
@@ -342,28 +354,28 @@ class BatchCounter(Counter):
     should be applied to "counter_name".
     """
 
-    def __init__(self, transport, name):
+    def __init__(self, transport: Transport, name: bytes):
         super(BatchCounter, self).__init__(transport, name)
-        self.packets = defaultdict(int)
+        self.packets: DefaultDict[float, float] = collections.defaultdict(int)
```
pacejackson: These types don't actually match.

baseplate/metrics.py
pacejackson: Or should `delta` be `int`?

baseplate/metrics.py
```diff
@@ -417,27 +429,27 @@ def replace(self, new_value):
...
+        :param new_value: The new value to store in the gauge.
 
         """
         assert new_value >= 0, "gauges cannot be replaced with negative numbers"
         serialized = self.name + (":{:g}|g".format(new_value).encode())
         self.transport.send(serialized)
 
 
-def make_client(namespace, endpoint):
+def make_client(namespace: str, endpoint: EndpointConfiguration) -> Client:
```
pacejackson: Why isn't `Client` in quotes here like some of the other functions?  Typo or quirk of python type hints?

spladug: because client is defined by this point so we can use the actual type instead of the quotes for a forward declaration

spladug: (on py3.8 or 3.7 w/ `from __future__ import annotations` annotations are done in a second pass and don't have this restriction)

baseplate/metrics.py
spladug: delta should be float, but for typing purposes ints count as floats. i don't know that it matters, but i'll  go be more precise about this.

baseplate/metrics.py
pacejackson: Ahh, thanks for the info :)

baseplate/server/wsgi.py
```diff
@@ -11,10 +11,10 @@
 
 try:
     # pylint: disable=no-name-in-module
-    from gevent.pywsgi import LoggingLogAdapter
+    from gevent.pywsgi import LoggingLogAdapter  # type: ignore
 except ImportError:
     # LoggingLogAdapter is from gevent 1.1+
-    class LoggingLogAdapter:
+    class LoggingLogAdapter:  # type: ignore
```
bradengroom: Why are the type ignore flags needed here?

bradengroom:
This is exciting. I'm not seeing anything that @pacejackson hasn't pointed out. This looks good to me if his comments are addressed.

baseplate/server/wsgi.py
bradengroom: Ah never mind. I see your comment above

spladug: here's the upstream bug for more context https://github.com/python/mypy/issues/1153

spladug:
:haircut_man: floats are fixed up, thanks for the comments!

bradengroom:
🐟

PR #289, opened 2019-06-03
Style Style Style

spladug:
This applies [Black] to the whole codebase. It also enables flake8 on the test suite which found some incomplete tests and generally caused a bunch of minor cleanup. The style changes all look pretty mundane and good with the exception of config parsers now being a bit ugly. Oh well. Consistency wins.

Commit-wise "Format all code using Black" and "Do a flake8-guided cleanup pass" are big ugly diffs of style fixups. The other two are the actual config changes.

[Black]: https://github.com/python/black/


bsimpson63:
very exciting!

PR #290, opened 2019-06-03
Absolute imports absolutely all the time

spladug:
Cleaned up and now enforced. Yay.


spladug:
:haircut_man: Forced the PYTHONPATH to make pylint work in Drone.

PR #291, opened 2019-06-04
Trace publisher heartbeat

cshoe:
Currently, the logs for the trace publisher are silent if nothing is going wrong. All emitted metrics are also tied to the ZipkinPublisher class. This means that if no traces are being published, no metrics are emitted.

These changes will make the logs slightly noisier by bumping to default level to ``INFO``. A metric and log line will also be emitted every 15 seconds so allow the monitoring and alerting to be established around the main loop and not the publisher.

### DIFF
--- a/baseplate/diagnostics/tracing/publisher.py
+++ b/baseplate/diagnostics/tracing/publisher.py
@@ -156,9 +156,12 @@ def publish_traces():
     while True:
         current_ts = int(round(time.time() * 1000))
         if (current_ts - last_heartbeat_ts) >= HEARTBEAT_INTERVAL:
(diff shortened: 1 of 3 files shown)


spladug:
Where does POSIX queue length fit into the monitoring/alerting plan?

cshoe:
My original goal was to get something in place that could answer the question "Is the trace publisher still running?". I added a gauge for the queue size as I think that will be very useful as well.

spladug:
Ah sorry, I meant the metrics we're already pulling via Diamond from the mqueues. They're monitored at the infrastructural level.

PR #293, opened 2019-06-05
Add a bunch of type annotations to a bunch of modules

spladug:
This brings us from 32% "imprecise" to 28%. Progress.

I'm not super happy with all of this and it definitely shows places where it's really hard to type stuff because it's nice and dynamic. We'll basically never get to have strict types for the request/context object. Bummer.

:eyeglasses: @pacejackson @bradengroom

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -1,13 +1,19 @@
 import os
 import sys
 
+from typing import TYPE_CHECKING
+
 from baseplate import config, metrics
 from baseplate.core import Baseplate
 from baseplate.diagnostics import tracing
 from baseplate._utils import warn_deprecated
 
 
-def metrics_client_from_config(raw_config):
+if TYPE_CHECKING:
+    import raven
+
+
+def metrics_client_from_config(raw_config: config.RawConfig) -> metrics.Client:
     """Configure and return a metrics client.
 
     This expects two configuration options:
@@ -33,7 +39,7 @@ def metrics_client_from_config(raw_config):
     return metrics.make_client(cfg.metrics.namespace, cfg.metrics.endpoint)
 
 
-def make_metrics_client(raw_config):
+def make_metrics_client(raw_config: config.RawConfig) -> metrics.Client:
     warn_deprecated(
         "make_metrics_client is deprecated in favor of the more "
         "consistently named metrics_client_from_config"
@@ -41,7 +47,9 @@ def make_metrics_client(raw_config):
     return metrics_client_from_config(raw_config)
 
 
-def tracing_client_from_config(raw_config, log_if_unconfigured=True):
+def tracing_client_from_config(
+    raw_config: config.RawConfig, log_if_unconfigured: bool = True
+) -> tracing.TracingClient:
     """Configure and return a tracing client.
 
     This expects one configuration option and can take many optional ones:
@@ -105,7 +113,9 @@ def tracing_client_from_config(raw_config, log_if_unconfigured=True):
     )
 
 
-def make_tracing_client(raw_config, log_if_unconfigured=True):
+def make_tracing_client(
+    raw_config: config.RawConfig, log_if_unconfigured: bool = True
+) -> tracing.TracingClient:
     warn_deprecated(
         "make_tracing_client is deprecated in favor of the more "
         "consistently named tracing_client_from_config"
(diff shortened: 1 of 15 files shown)


baseplate/__init__.py
```diff
@@ -1,13 +1,19 @@
...
 from baseplate.diagnostics import tracing
 from baseplate._utils import warn_deprecated
 
 
-def metrics_client_from_config(raw_config):
+if TYPE_CHECKING:
+    import raven
+
+
+def metrics_client_from_config(raw_config: Dict[str, str]) -> metrics.Client:
```
bradengroom: The funding instrument service saw this `Dict[str, str]` show up in a lot of places, so we decided to define a variable for it to make it more obvious to the reader. I'm not sure if I like this pattern or not, but I'm interested to get your thoughts. I think it's all the same to mypy.

https://github.snooguts.net/reddit/reddit-service-funding-instruments/blob/1d490bbd9c79d0329f34884dcc353ab8d413a805/funding_instruments/common_types.py#L4
https://github.snooguts.net/reddit/reddit-service-funding-instruments/blob/1d490bbd9c79d0329f34884dcc353ab8d413a805/funding_instruments/models/sql.py#L7

spladug: Yeah! Then you can just import those type defs too. That should be really nice. I'll get some aliases going.

baseplate/file_watcher.py
```diff
@@ -32,49 +32,69 @@
...
 
 from baseplate.retry import RetryPolicy
 
 
 logger = logging.getLogger(__name__)
 
 
-_NOT_LOADED = object()
+class _NOT_LOADED:
+    pass
```
bradengroom: I went down a rabbit hole looking at this change. It looks like they recommend using an enum for these singleton instances since they can't be subclassed:
https://github.com/python/typing/pull/240/files

I like the way you've done it more tbh, but wanted to point out that the pep recommends something else. I'm totally okay with what you've got, but wanted to share

spladug: Ah yeah, I saw the enum thing. But [this comment](https://github.com/python/typing/issues/236#issuecomment-232554039) linked to an implementation ticket in mypy that is still open: https://github.com/python/mypy/issues/1803

so i think it doesn't work yet? didn't actually try because I'm a jerk. Would you prefer I give the proper way a shot?

bradengroom: Yours feels cleaner and I prefer it. I don't think anyone is going to subclass this `_NOT_LOADED` class and get confused.

baseplate/secrets/store.py
```diff
@@ -306,18 +293,20 @@ def make_object_for_context(self, name, span):
...
+    def __init__(self, filewatcher: FileWatcher):  # pylint: disable=super-init-not-called
         self._filewatcher = filewatcher
 
     @cached_property
     def _data(self):
-        return super(_CachingSecretsStore, self)._get_data()
+        return super()._get_data()
 
-    def _get_data(self):
+    def _get_data(self) -> Dict:
```
bradengroom: I'm guessing we don't know anything about what's in this dictionary?

baseplate/__init__.py
pacejackson: +1

baseplate/config.py
```diff
@@ -318,62 +338,73 @@ def one_of(text):
...
 
 
 class ConfigNamespace(dict):
     def __init__(self):
-        super(ConfigNamespace, self).__init__()
+        super().__init__()
         self.__dict__ = self
 
+    def __getattr__(self, name: str) -> Any:
+        ...
```
pacejackson: What's this?

baseplate/crypto.py
```diff
@@ -124,7 +121,7 @@ def make_signature(secret, message, max_age):
     return base64.urlsafe_b64encode(header + digest)
 
 
-def validate_signature(secret, message, signature):
+def validate_signature(secret: "VersionedSecret", message: str, signature: bytes) -> SignatureInfo:
```
pacejackson: I don't think this should need the quotes right?

baseplate/config.py
spladug: It's a type stub for `__getattr__` so that type checking knows we can return anything we want via attribute references. Prior art is here:

* https://github.com/python/typeshed/blob/master/stdlib/2and3/argparse.pyi#L268
* https://github.com/python/typeshed/issues/25

spladug: tl;dr "good luck!!!!!! i'll do whatever i want!!!!"

baseplate/crypto.py
spladug: gp :hocho:

baseplate/secrets/store.py
spladug: yeah, we could get a bit more precise in that the top level will be str->dict but then that underlying dict will be a lot more undefined. not sure how precise we need to be.

spladug:
:haircut_man:

PR #298, opened 2019-06-07
Metrics error handling and bugfix

spladug:
The first commit tries to clarify error messaging around metrics batches. I was watching the thing service logs yesterday and noticed a fair number of the "batch too big" warnings were actually for other stuff like "connection refused". I think this also makes the error handling a bit cleaner. The new messaging looks like:

```
Metrics batch of 370109 bytes is too large to send, flush more often or reduce amount done in this request. Top counters: example_service.busy_counter=2345, example_service.normal_counter=37
```

or 

```
Failed to send metrics batch: [Errno 111] Connection refused
```

The second commit fixes an error encountered by the data team in the event collectors.

### DIFF
--- a/baseplate/metrics.py
+++ b/baseplate/metrics.py
@@ -44,6 +44,7 @@ ends.
 """
 
 import collections
+import errno
 import logging
 import socket
 import time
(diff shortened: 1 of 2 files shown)


spladug:
:eyeglasses: @bsimpson63 @pacejackson

PR #308, opened 2019-06-26
Add 'component' annotation to traces

ckwang8128:
The 'component' annotation is a standard across
tracing implementations that designates the software
package, framework, or library that generates a given span.
Adding this annotation to Baseplate will make it easier
to see Baseplate-generated spans in a trace.


spladug:
LGTM modulo the style nit :nail_care:

ckwang8128:
fixed formatting! :haircut:

PR #312, opened 2019-06-28
Add a one-stop shop method for configuring diagnostics observers

### LINKED ISSUE #311: Rework wire-up of the core diagnostics observers
This is pretty manual right now and we can simplify it as well. That potentially means easier upgrades for services down the line too.

```python
    tracing_client = tracing_client_from_config(app_config)

    baseplate = Baseplate()
    baseplate.configure_logging()
    baseplate.configure_tracing(tracing_client)
```

could, e.g., become

```python
    baseplate = Baseplate()
    baseplate.configure_observers(app_config)
```

where we automatically wire up the observers found in the config file.

The manual way should continue to exist for advanced use cases. This is just sugar.

spladug:
This is just a convenience layer but should help simplify app wire-up
and make it easier to upgrade to future versions with different
observers by just changing config files.

This fixes #311

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -738,6 +738,56 @@ class Baseplate:
 
         self.register(SentryBaseplateObserver(client))
 
+    def configure_observers(self, app_config: config.RawConfig, module_name: str) -> None:
+        """Configure diagnostics observers based on application config file.
+
+        This installs all the currently supported observers that have settings
+        in the configuration file.
+
+        For the individual configurables, see the documentation for:
+
+        * :py:func:`~baseplate.error_reporter_from_config`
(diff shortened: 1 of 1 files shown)


baseplate/core.py
```diff
@@ -767,6 +767,55 @@ def configure_error_reporting(self, client):
 
         self.register(SentryBaseplateObserver(client))
 
+    def configure_observers(self, app_config: config.RawConfig, module_name: str) -> None:
```
pacejackson: Actually, one tiny comment, should this be `configure_monitoring_observers`?

spladug:
rebased to get the flaky test fix and added a pylint directive for the cyclic import

PR #313, opened 2019-07-01
Support thrift include statements, document integration

pwildani:
The approach for handling thrift include statements is basically tell the thrift compiler to put the built code in the right place up front, rather than trying to fix up its output. 

So, the new build_thrift pass first patches all of the thrift files to have the correct namespace for including them python side, then builds them all, and finally copies the result out of the working directories back into the rest of the source.

This should make fully handling https://reddit.atlassian.net/browse/IO-2804 simpler.

### DIFF
--- a/baseplate/integration/thrift/README.rst
+++ b/baseplate/integration/thrift/README.rst
@@ -11,7 +11,6 @@ So a file named ``foo/bar/baz.thrift``` is accessed as a python module named
 If the baz.thrift file already has a ``namespace py quux`` declaration. It will be
 accessed in python as ``foo.bar.quux``, not the thrift default of ``quux``.
 
-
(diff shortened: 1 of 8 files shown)


bradengroom:
There's a previous PR here with some relevant feedback about maintaining backwards compatibility:
https://github.com/reddit/baseplate.py/pull/183

I wonder if @spladug would be okay with this being a breaking change in the 1.0 release though.

spladug:
I'm going to close this out as it's been idle for a while and doesn't look like it's ready for merge. I think we're on the cusp of some pretty different Thrift stuff soon with Artifactory coming online internally, so hopefully we won't need to use the thrift build steps in here anymore at all. Please let me know if you'd still like to work on this stuff.

PR #314, opened 2019-07-09
Simplify the request context laziness setup

spladug:
The stuff I merged a couple of weeks ago (#304) has some problems with Pyramid integration. It looks like using `add_request_method(..., reify=True)` leads to a situation where the new request method gets a copy of the underlying Pyramid request object rather than the baseplate-provided wrapper. This means that those methods can't use any context attributes like `secrets` or whatever. This broke things when trying to upgrade GraphQL for canarying.

I poked around a bit and ended up realizing that what I built then was too complicated and instead of trying to wrap the object and do all the `__getattribute__` shenanigans we should go with the mixin approach @bradengroom pioneered in #253. The big difference from that PR remains that we're passing around the `context_config` dictionary rather than building a span observer for every attribute. This means we only copy the dictionary ref around on span creation rather than doing all the setup of lazy attributes so we keep the big perf win for span setup:

Time per span by % of attributes accessed (25 attributes)

But it also means that we're no longer doing any proxying of attributes except in local span situations. This means that accessing the ContextFactory-generated attributes after initial setup are very fast again:

Time to read ContextFactory-generated attribute(1)

and that accessing non-Baseplate stuff like `request.content_type` isn't penalized anymore:

Time to read non-baseplate attribute (e g  pyramid request properties)

### DIFF
--- a/baseplate/core.py
+++ b/baseplate/core.py
@@ -2,7 +2,7 @@ import collections
 import logging
 import random
 
-from types import TracebackType, SimpleNamespace
+from types import TracebackType
 from typing import Tuple, Optional, Type, NamedTuple, Any, Dict
 
 import jwt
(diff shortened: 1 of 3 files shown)


spladug:

bradengroom:
🐟

spladug:
Time per request by how many clients accessed

PR #315, opened 2019-07-11
Reference Cycles

### LINKED ISSUE #310: Audit for reference cycles
These cause GC pauses so let's get rid of the ones we're making in the framework.

spladug:
Python objects track how many references are held to them and automatically cease to exist when the count drops to zero. In the case of a reference cycle, this system doesn't work and the `gc` module's garbage collector occasionally kicks in to find unreferenced cycles and clean them up. This can be bad for latency as the garbage collector stops request processing at non-deterministic intervals.

This pull request adds some tooling for finding reference cycles in baseplate.py apps and then uses that tooling to remove a bunch of cycles found in the framework itself. It looks like at least some of the libraries we use (sqlalchemy/psycopg2, kombu, and cassandra-driver) have their own reference cycles which we can't immediately fix here but we may try to work with upstream to fix those.

The new tooling hangs off of the runtime monitor system and should not be used constantly in production as it will almost certainly slow things down. When configured, it tells the gc to save all collected objects to `gc.garbage` then uses [`objgraph`](https://mg.pov.lt/objgraph/) to make pretty pictures of the lingering objects:

image

The important parts are the arrows pointing back up as those are the cycles keeping the whole visible graph alive. Once those were eliminated from the framework code, a basic request that generated that graph no longer made any cycles at all!

This fixes #310


spladug:
:eyeglasses: @pacejackson @bsimpson63 @bradengroom

baseplate/server/__init__.py
```diff
@@ -208,6 +209,9 @@ def load_app_and_run_server():
     if args.reload:
         reloader.start_reload_watcher(extra_files=[args.config_file.name])
 
+    # clean up leftovers from initialization before we get into requests
+    gc.collect()
+
```
bradengroom: I'm a little confused about what we are cleaning up here if we're are saying that we no longer have cycles and the ref counter can handle things.

bradengroom:
🐟 looks good. Just have one question above

baseplate/server/__init__.py
spladug: This is just a one time thing during startup. There are definitely ref cycles in the libraries we use that show up in this stage. For example, the CQL driver and CQLMapper have all sorts during init. By cleaning up here, we make sure they don't bother real requests or show up in our monitoring of steady-state garbage collection.

bradengroom: I see. I forgot about other libraries. Sounds good

PR #320, opened 2019-07-19
tracing: Turn all exception logging in tracing module to warning

ckwang8128:
Tracing instrumentation is not application-critical. This changes
exception and error logging in the module to 'warning' so non-critical
instrumentation issues do not spam application logs.


ckwang8128:
:eyeglasses: @pacejackson @spladug

PR #321, opened 2019-07-22
Sort imports

spladug:
Not a big fan of the growing duplication between the makefile and Drone config. Could install make in drone or bite the bullet and integrate pre-commit.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,6 +10,7 @@ steps:
               from_secret: pip_options
       commands:
           - pip install $PIP_OPTIONS -r requirements-lint.txt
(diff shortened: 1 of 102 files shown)


pacejackson:
IMO, I would look into installing `make` in drone rather than going with pre-commit.  I've found pre-commit to be kind of a headache personally. :/

spladug:
Yeah, `black` and `isort` will happily do the `()` from imports, but `reorder-python-imports` opinionedly chooses to split them all out line-by-line to improve mergeability.  I don't have a strong preference and `reorder-python-imports` was _much_ easier to use out of the box because of its better detection of what's first and third party.

spladug:
:haircut_man:

PR #322, opened 2019-07-25
A few housecleaning tasks

spladug:
I've got a lot more type annotations coming, but wanted to clear these out before that so the review doesn't get too large.


baseplate/lib/config.py
```diff
@@ -228,8 +228,8 @@ def File(mode: str = "r") -> Callable[[str], IO]:  # noqa: D401
     def open_file(text: str) -> IO:
         try:
             return open(text, mode=mode)
-        except IOError:
-            raise ValueError("could not open file: %s" % text)
+        except OSError:
```
bradengroom: I didn't know the exception thrown by `open` changed. Nice catch

baseplate/lib/events.py
```diff
@@ -32,105 +23,6 @@
 MAX_QUEUE_SIZE = 10000
 
 
-# pylint: disable=pointless-string-statement,no-init
-class FieldKind(Enum):
```
bradengroom: I'm not super familiar with this part of baseplate. I'm assuming this is going away because v2 events are a thing now?

baseplate/lib/config.py
spladug: looks like IOError and a bunch of others (socket.error! yay!) got moved under OSError. pyupgrade does this for us. v nice.

baseplate/lib/events.py
spladug: Yeah, the v1 events are on their way out. I confirmed with data that we're good to delete this from 1.0.

bradengroom:
🐟

baseplate/server/thrift.py
```diff
@@ -25,12 +23,7 @@ def __init__(self, processor, *args, **kwargs):
                 THeaderClientType.UNFRAMED_BINARY,
             ]
         )
-        super(GeventServer, self).__init__(*args, **kwargs)
-
-    def serve_forever(self, stop_timeout=None):
-        signal.signal(signal.SIGINT, lambda sig, frame: self.stop())
-        signal.signal(signal.SIGTERM, lambda sig, frame: self.stop())
-        super(GeventServer, self).serve_forever(stop_timeout=stop_timeout)
```
pacejackson: What happened here?

spladug: The base server stuff stopped using `serve_forever` in a4144dfb8b772b3d477da614eeb4145e1ee81273

PR #323, opened 2019-07-25
baseplate: Add tracing debug tag populating for spans

ckwang8128:
"Sampled" spans are ones that match a sampled request, which could be at the whim of any number of upstream sampling lotteries. We need a way to differentiate "debug" spans - spans that represent part or all of a request that we have designated specifically as "debug"-required.

Add differentiating between a sampled span and a debug span, and
update TraceObserver to determine whether a span is a debug span
using its flags.

Additionally, add a 'debug' binary annotation to debug spans
for trace storage systems to utilizing in filtering mechanisms.

### DIFF
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -36,6 +36,8 @@ ANNOTATIONS = {
     "SERVER_RECEIVE": "sr",
     "LOCAL_COMPONENT": "lc",
     "COMPONENT": "component",
+    "DEBUG": "debug",
+    "ERROR": "error",
 }
 
 # Feature flags
@@ -187,7 +189,13 @@ class TraceSpanObserver(SpanObserver):
 
     def on_finish(self, exc_info):
         if exc_info:
-            self.on_set_tag("error", True)
+            self.on_set_tag(ANNOTATIONS["ERROR"], True)
+
+        # Set a debug annotation for downstream processing to
+        # utilize in span filtering.
+        if self.span.flags and (self.span.flags & FLAGS["DEBUG"]):
+            self.on_set_tag(ANNOTATIONS["DEBUG"], True)
+
         self.end = current_epoch_microseconds()
         self.elapsed = self.end - self.start
         self.record()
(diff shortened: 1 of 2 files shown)


ckwang8128:
Some general open questions/design decisions:

- I opted not to make child spans automatically inherit the debug property of their parent and rather leave it up to observers (e.g. TracingObserver) to determine at creation time. In Tracing context, having child spans always be `debug` makes sense, but there might be other sets of observers that turn debug on or off for certain spans within a service. Maybe I'm overthinking this.

ckwang8128:
:eyeglasses: @spladug @pacejackson @cshoe

ckwang8128:
After some discussion (huge thanks @spladug!) this opts for a simpler model without a full change to the Span API to add a "debug" mechanism. Developers can still leverage the tracing `debug` functionality via upstream client specifications as needed. There hasn't been other direct request for this `debug` setting, so let's simplify this for now and should the need arise for another observer use case we can reevaluate.

baseplate/observers/tracing.py
```diff
@@ -179,6 +180,12 @@ def __init__(self, service_name, hostname, span, recorder):
         self.elapsed = None
         self.binary_annotations = []
         self.on_set_tag(ANNOTATIONS["COMPONENT"], "baseplate")
+
+        # Set a debug annotation for downstream processing to
+        # utilize in span filtering.
+        if self.span.flags and (self.span.flags & FLAGS["DEBUG"]):
```
spladug: should we do this in `finish()` so the user can do `request.trace.flags = DEBUG` mid-request to turn it on?

baseplate/observers/tracing.py
```diff
@@ -179,6 +181,7 @@ def __init__(self, service_name, hostname, span, recorder):
         self.elapsed = None
         self.binary_annotations = []
         self.on_set_tag(ANNOTATIONS["COMPONENT"], "baseplate")
+
```
pacejackson: style nit, we probably don't want to leave this extra blank line.

ckwang8128: thanks for catching!

ckwang8128:
:haircut: @spladug 

* moved to `on_finish`.
* lint cleanup.

PR #324, opened 2019-07-29
Make mypy stricter and type annotate the rest of the library

spladug:
There were a few small dormant bugs found via this process (yay!) but it
should mostly not have any runtime effect.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -3,14 +3,23 @@ import random
 
 from types import TracebackType
 from typing import Any
+from typing import Callable
 from typing import Dict
+from typing import List
 from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
 from typing import Type
+from typing import TYPE_CHECKING
 
 from baseplate.lib import config
-from baseplate.lib import warn_deprecated
+from baseplate.lib import metrics
+
+
+if TYPE_CHECKING:
+    import baseplate.clients
+    import baseplate.observers.tracing
+    import raven
 
 
 logger = logging.getLogger(__name__)
@@ -19,14 +28,13 @@ logger = logging.getLogger(__name__)
 class BaseplateObserver:
     """Interface for an observer that watches Baseplate."""
 
-    def on_server_span_created(self, context: Any, server_span: "Span") -> None:
+    def on_server_span_created(self, context: "RequestContext", server_span: "ServerSpan") -> None:
         """Do something when a server span is created.
 
         :py:class:`Baseplate` calls this when a new request begins.
 
         :param context: The :term:`context object` for this request.
-        :param baseplate.ServerSpan server_span: The span representing
-            this request.
+        :param server_span: The span representing this request.
 
         """
         raise NotImplementedError
@@ -41,13 +49,13 @@ class SpanObserver:
     def on_start(self) -> None:
         """Do something when the observed span is started."""
 
-    def on_set_tag(self, key: str, value: str) -> None:
+    def on_set_tag(self, key: str, value: Any) -> None:
         """Do something when a tag is set on the observed span."""
 
-    def on_log(self, name: str, payload: str) -> None:
+    def on_log(self, name: str, payload: Any) -> None:
         """Do something when a log entry is added to the span."""
 
-    def on_finish(self, exc_info: _ExcInfo) -> None:
+    def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
         """Do something when the observed span is finished.
 
         :param exc_info: If the span ended because of an exception, the
@@ -61,7 +69,7 @@ class SpanObserver:
         :py:class:`SpanObserver` objects call this when a new child span is
         created.
 
-        :param baseplate.Span span: The new child span.
+        :param span: The new child span.
 
         """
 
@@ -81,7 +89,7 @@ class TraceInfo(NamedTuple):
 
     trace_id: int
     parent_id: Optional[int]
-    span_id: Optional[int]
+    span_id: int
     sampled: Optional[bool]
     flags: Optional[int]
 
@@ -136,11 +144,24 @@ class TraceInfo(NamedTuple):
 
 
 class RequestContext:
-    def __init__(self, context_config, prefix=None, span=None, wrapped=None):
+    def __init__(
+        self,
+        context_config: Dict[str, Any],
+        prefix: Optional[str] = None,
+        span: Optional["Span"] = None,
+        wrapped: Optional["RequestContext"] = None,
+    ):
         self.__context_config = context_config
         self.__prefix = prefix
         self.__wrapped = wrapped
-        self.trace: ServerSpan = span
+
+        # the context and span reference eachother (unfortunately) so we can't
+        # construct 'em both with references from the start. however, we can
+        # guarantee that during the valid life of a span, there will be a
+        # reference. so we fake it here and say "trust us".
+        #
+        # this would be much cleaner with a different API but this is where we are.
+        self.trace: "Span" = span  # type: ignore
 
     def __getattr__(self, name: str) -> Any:
         try:
@@ -168,6 +189,11 @@ class RequestContext:
         setattr(self, name, obj)
         return obj
 
+    # this is just here for type checking
+    # pylint: disable=useless-super-delegation
+    def __setattr__(self, name: str, value: Any) -> None:
+        super().__setattr__(name, value)
+
     def clone(self) -> "RequestContext":
         return RequestContext(
             context_config=self.__context_config,
@@ -186,27 +212,27 @@ class Baseplate:
 
     """
 
-    def __init__(self):
-        self.observers = []
-        self._metrics_client = None
-        self._context_config = {}
+    def __init__(self) -> None:
+        self.observers: List[BaseplateObserver] = []
+        self._metrics_client: Optional[metrics.Client] = None
+        self._context_config: Dict[str, Any] = {}
 
-    def register(self, observer):
+    def register(self, observer: BaseplateObserver) -> None:
         """Register an observer.
 
-        :param baseplate.BaseplateObserver observer: An observer.
+        :param observer: An observer.
 
         """
         self.observers.append(observer)
 
-    def configure_logging(self):
+    def configure_logging(self) -> None:
         """Add request context to the logging system."""
         # pylint: disable=cyclic-import
         from baseplate.observers.logging import LoggingBaseplateObserver
 
         self.register(LoggingBaseplateObserver())
 
-    def configure_metrics(self, metrics_client):
+    def configure_metrics(self, metrics_client: metrics.Client) -> None:
         """Send timing metrics to the given client.
 
         This also adds a :py:class:`baseplate.lib.metrics.Batch` object to the
@@ -214,8 +240,7 @@ class Baseplate:
         your own application-specific metrics. The batch is automatically
         flushed at the end of the request.
 
-        :param baseplate.lib.metrics.Client metrics_client: Metrics client to send
-            request metrics to.
+        :param metrics_client: Metrics client to send request metrics to.
 
         """
         # pylint: disable=cyclic-import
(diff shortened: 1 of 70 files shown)


baseplate/__init__.py
```diff
@@ -224,41 +249,30 @@ def configure_metrics(self, metrics_client):
...
 
-        :param baseplate.observers.tracing.TracingClient tracing_client: Tracing
-            client to send request traces to.
+        :param tracing_client: Tracing client to send request traces to.
 
         """
         # pylint: disable=cyclic-import
-        from baseplate.observers.tracing import make_client, TraceBaseplateObserver, TracingClient
-
-        # the first parameter was service_name before, so if it's not a client
```
spladug: Dropped support for the deprecated way of setting up trace observers to make typing simpler here. This whole method shouldn't really get seen much now with `configure_observers()`.

baseplate/clients/memcache/__init__.py
```diff
@@ -125,135 +148,148 @@ class MonitoredMemcacheConnection:
...
             return self.pooled_client.get_many(keys)
 
-    def gets(self, key, **kwargs):
+    def gets(
+        self, key: Key, default: Optional[Any] = None, cas_default: Optional[Any] = None
+    ) -> Tuple[Any, Any]:
         with self._make_span("gets") as span:
             span.set_tag("key", key)
-            return self.pooled_client.gets(key, **kwargs)
+            return self.pooled_client.gets(key, default=default, cas_default=cas_default)
```
spladug: I might need to change this to look like `get()`'s internals on L215? I think this was done because those params don't exist in some versions of redis-py?

baseplate/clients/memcache/lib.py
```diff
@@ -111,11 +121,13 @@ def dump_and_compress(key, value):  # pylint: disable=unused-argument
             serialized = json.dumps(value)
             flags = Flags.JSON
 
+        serialized_bytes = serialized.encode("utf8")
```
spladug: bug caught by typing!

`zlib.compress` on py3 takes a bytes object, not a str. we need to encode first. this shows up a few other times in this file too.

baseplate/frameworks/pyramid/csrf.py
```diff
@@ -108,26 +116,30 @@ def get_csrf_token(self, request):
...
+        if token_version != self.VERSION:
+            return False
+
         _, payload = _make_csrf_token_payload(
             version=token_version, account_id=request.authenticated_userid
         )
 
         try:
-            validate_signature(self._get_secret(), payload, signature)
+            validate_signature(self._get_secret(), payload, signature.encode())
```
spladug: similar need to clean up byte encoding

baseplate/frameworks/pyramid/csrf.py
```diff
@@ -108,26 +116,30 @@ def get_csrf_token(self, request):
...
+    def check_csrf_token(self, request: Any, supplied_token: str) -> bool:
         """Return True if the supplied_token is valid.
 
         This is called automatically by Pyramid if you have configured it
         to require CSRF.
         """
         try:
-            token_version, sep, signature = supplied_token.partition(".")
+            version_str, sep, signature = supplied_token.partition(".")
+            token_version = int(version_str)
```
spladug: explicitly int-ified first so the type signature of `_make_csrf_token_payload` could always take an int

baseplate/frameworks/queue_consumer.py
```diff
@@ -145,7 +176,7 @@ def get_batch(self, max_items, timeout):
         return batch
 
 
-class KombuConsumer(BaseKombuConsumer):
+class KombuConsumer:
```
spladug: Broke up inheritance here and opted for composition instead. The signatures of the `get_` methods are different so claiming it's a subclass is a lie.

baseplate/lib/experiments/providers/simple_experiment.py
```diff
@@ -168,14 +178,25 @@ def __init__(
 
         self.variant_set = variant_set
 
-        self._seed = bucket_seed
-        if self._seed is None:
+        if bucket_seed:
```
spladug: An example of reordering `if`s a bit to convince the type checker we're handling `None`s properly

baseplate/lib/experiments/providers/simple_experiment.py
```diff
@@ -168,14 +178,25 @@ def __init__(
...
+        owner: str,
+        start_ts: int,
+        stop_ts: int,
+        config: Dict[str, Any],
+        variant_type: str,
+        enabled: bool = True,
+    ) -> "SimpleExperiment":
         bucket_val = config.get("bucket_val", "user_id")
-        version = config.get("experiment_version")
+        version = config["experiment_version"]
```
spladug: You'll also see this a few times in experiment code where we get rid of an `Optional` on the type by using the subscript form for a dictionary. These cases should all be pulling out data that's required so it should be the same effect.

baseplate/lib/message_queue.py
```diff
@@ -172,7 +174,7 @@ def queue_tool():
             print(item.decode())
     elif args.mode == "write":
         for line in sys.stdin:
-            queue.put(line.rstrip("\n"))
+            queue.put(line.rstrip("\n").encode())
```
spladug: encoding bug!

baseplate/observers/tracing.py
```diff
@@ -330,19 +372,32 @@ def on_child_span_created(self, span):
...
+        )
 
         return self._to_span_obj(annotations, self.binary_annotations)
 
 
-class BaseBatchRecorder:
-    def __init__(self, max_queue_size, num_workers, max_span_batch, batch_wait_interval):
-        self.span_queue = queue.Queue(maxsize=max_queue_size)
+class Recorder:
+    def send(self, span: TraceSpanObserver) -> None:
```
spladug: this was a bit of a surprising type signature to me. i hope that having explicit annotations required in the future will make this sort of thing more obvious during code review.

baseplate/observers/tracing.py
```diff
@@ -430,12 +493,12 @@ def __init__(
...
         self.endpoint = f"http://{endpoint}/api/v1/spans"
 
-    def flush_func(self, spans):
+    def flush_func(self, spans: List[Dict[str, Any]]) -> None:
         """Send a set of spans to remote collector."""
         try:
             self.session.post(
                 self.endpoint,
-                data=json.dumps(spans),
+                data=json.dumps(spans).encode("utf8"),
```
spladug: encoding bug!

spladug:

baseplate/__init__.py
```diff
@@ -3,14 +3,23 @@
...
 
 from baseplate.lib import config
-from baseplate.lib import warn_deprecated
+from baseplate.lib import metrics
+
+
+if TYPE_CHECKING:
+    import baseplate.clients
+    import baseplate.observers.tracing
+    import raven
```
bradengroom: I'm pretty sure it doesn't, but does this have any impact on the "required-ness" of the raven library? This just means that you will need to have raven installed in order to perform type checking right? Nothing changes for actual users of the library?

spladug: correct, `TYPE_CHECKING` will always be `False` at runtime so this code is only ever seen by static analysis tooling.

baseplate/__init__.py
```diff
@@ -465,35 +491,35 @@ def start(self):
...
-    def set_tag(self, key, value):
+    def set_tag(self, key: str, value: Any) -> None:
         """Set a tag on the span.
 
         Tags are arbitrary key/value pairs that add context and meaning to the
         span, such as a hostname or query string. Observers may interpret or
         ignore tags as they desire.
 
-        :param str key: The name of the tag.
-        :param value: The value of the tag, must be a string/boolean/number.
```
pacejackson: Is this restriction no longer true? If it is, I think you could use a Union to enforce it if it is rather than just using `Any`.

spladug: The current tracing implementation accepts bools, ints, and anything with `__str__`. We're placing some exception instances as tags right now. Could clean that up and tighten the param type, or leave as-is since at worst we'll just get a stringy version of whatever's passed in (I should definitely update the docs if we go the latter, good catch).

baseplate/lib/edge_context.py
```diff
@@ -19,15 +28,14 @@ class NoAuthenticationError(Exception):
 class AuthenticationTokenValidator:
     """Factory that knows how to validate raw authentication tokens."""
 
-    def __init__(self, secrets):
+    def __init__(self, secrets: SecretsStore):
         self.secrets = secrets
 
-    def validate(self, token):
+    def validate(self, token: bytes) -> "AuthenticationToken":
```
pacejackson: Not sure how much we care about this but would it make sense to just move the class definition for this up so we don't have to use a string?

baseplate/lib/edge_context.py
```diff
@@ -353,11 +365,13 @@ class EdgeRequestContext:
...
 
-    def __init__(self, authn_token_validator, header):
+    def __init__(
+        self, authn_token_validator: AuthenticationTokenValidator, header: Optional[bytes]
+    ):
         self._authn_token_validator = authn_token_validator
         self._header = header
 
-    def attach_context(self, context):
+    def attach_context(self, context: Any) -> None:
```
pacejackson: Should we just go ahead and type this to `RequestContext`?

baseplate/lib/experiments/variant_sets/rollout_variant_set.py
```diff
@@ -12,28 +17,29 @@ class RolloutVariantSet(VariantSet):
...
         as a floating point value between 0 and 1.
-    :param int num_buckets: the number of potential buckets that can be
+    :param num_buckets: the number of potential buckets that can be
         passed in for a variant call. Defaults to 1000, which means maximum
         granularity of 0.1% for bucketing
 
     """
 
-    def __init__(self, variants, num_buckets=1000):
+    # pylint: disable=super-init-not-called
```
pacejackson: should we just go ahead and call `super` or is there a reason we didn't?

baseplate/observers/tracing.py
pacejackson: 🔕 Out of curiosity, why was it surprising?

baseplate/server/wsgi.py
```diff
@@ -15,18 +20,18 @@
 except ImportError:
     # LoggingLogAdapter is from gevent 1.1+
```
pacejackson: should we just "enforce" that we be on a newer version of baseplate by not catching this anymore?

baseplate/lib/experiments/variant_sets/rollout_variant_set.py
spladug: super raises `NotImplementedError`!

spladug: (i'm not fond of this, but it was the simplest way I could find to type the dictionary mapping their names to types -- open to suggestions)

baseplate/server/wsgi.py
spladug: yeah, I've got a branch coming with some version restriction updates

baseplate/observers/tracing.py
spladug: the recorder takes the observer?! the variable name is `span`. that seems _very_ weird.

pacejackson: 😅

baseplate/lib/edge_context.py
spladug: I think it's probably fine to leave like this. In 3.8 (or 3.7 w/ `from __future__ import annotations`) _all_ annotations become strings without having to quote them so ordering doesn't matter.

baseplate/lib/edge_context.py
spladug: done, thanks

baseplate/server/reloader.py
```diff
@@ -25,15 +30,15 @@ def _get_loaded_modules():
...
 
-def _get_watched_files(extra_files):
+def _get_watched_files(extra_files: Sequence[str]) -> Iterator[str]:
     """Yield filenames for all files to be watched for modification."""
     yield from _get_loaded_modules()
     yield from extra_files
 
 
-def _reload_when_files_change(extra_files):
+def _reload_when_files_change(extra_files: Sequence[str]) -> NoReturn:
```
bradengroom: 🔕 I wasn't aware that this return type was an option. Good to know.

PR #331, opened 2019-08-01
Add 'fmt' command to Makefile

pacejackson:
This command will run 'black' and 'reorder-python-imports' for you

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,10 @@ docs:
 test:
 	tox
 
+fmt:
+	reorder-python-imports --py3-plus --separate-from-import --separate-relative $(shell find baseplate/ tests/ -name '*.py')
+	black .
+
 lint:
 	reorder-python-imports --diff-only --py3-plus --separate-from-import --separate-relative $(shell find baseplate/ tests/ -name '*.py')
 	black --diff --check .
@@ -46,4 +50,4 @@ clean:
 realclean: clean
 	-rm -rf baseplate.egg-info/
 
-.PHONY: docs spelling clean realclean tests lint checks
+.PHONY: docs spelling clean realclean tests fmt lint checks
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,6 +30,10 @@ docs:
 test:
 	tox
 
+fmt:
+	reorder-python-imports --py3-plus --separate-from-import --separate-relative $(shell find baseplate/ tests/ -name '*.py')
```
spladug: should the `--py3-plus --separate-from-import --separate-relative` parts be split out into a variable so we're not duplicating them between targets?

also, i think we need the flag for exiting with 0 even on changes on this one?

pacejackson: > should the --py3-plus --separate-from-import --separate-relative parts be split out into a variable so we're not duplicating them between targets?

Makes sense.

> also, i think we need the flag for exiting with 0 even on changes on this one?

I don't think we need it, would we want it though?

spladug: We'd want it (or starting the line with `-` to ignore the return code) so that if `reorder-python-imports` and `black` both have jobs to do when we run `make fmt` we don't just stop immediately after RPI did its thing because it returned an "error" code.

spladug:
I just added a `make fmt` to the upgrader tool's makefile after your PR here, here's what that looks like in case it's helpful:

```make
REORDER_PYTHON_IMPORTS := reorder-python-imports --py37-plus --separate-from-import --separate-relative
PYTHON_FILES = $(shell find baseplate_py_upgrader/ tests/ -name '*.py')


all:


lint:
	$(REORDER_PYTHON_IMPORTS) --diff-only $(PYTHON_FILES)
	black --diff baseplate_py_upgrader/
	flake8 baseplate_py_upgrader/
	mypy baseplate_py_upgrader/


fmt:
	$(REORDER_PYTHON_IMPORTS) --exit-zero-even-if-changed $(PYTHON_FILES)
	black baseplate_py_upgrader/


test:
	pytest


.PHONY: lint fmt test
```

Makefile
pacejackson: Oops, I think my brain was a little fried, I thought that flag was already in there (I added it in another repo I was working on) and read your comment as the opposite of what you actually said.

Makefile
```diff
@@ -1,3 +1,6 @@
+REORDER_PYTHON_IMPORTS := reorder-python-imports --py37-plus --separate-from-import --separate-relative
```
spladug: we're not 37 yet on baseplate, still supporting 36. this has to be `--py3-plus` sorry for confusion from the copy paste i put into the comments

Issue #301, opened 2019-06-19
Add an overview document on baseplate's metrics


spladug:
Aggregate it all across the various places we generate metrics and explain what they all mean.

spladug:
0f4f9c2bc4ab4e503c70477a71bd5da9f8bf1659

Issue #309, opened 2019-06-27
Build out narrative docs for initial learning


spladug:
The docs are too referency and not good at explaining what's going on or how to use the system.

spladug:
0f4f9c2bc4ab4e503c70477a71bd5da9f8bf1659

PR #336, opened 2019-08-11
Overhaul the documentation

spladug:
This is a biiiig overhaul of the documentation. You can see it [all rendered in my S3 bucket](https://spladug.s3.amazonaws.com/baseplate/index.html) or pull this branch down and run `make docs` yourself.

* Add new tutorial. More chapters to come later.
* Drop a bunch of internal stuff that didn't need to be in the public docs.
* Get doctests and spellchecker going again. Fix up issues this exposed.
* Verify links with linkcheck; fix a bunch of cross-refs and update intersphinx.
* Add a bunch of examples.
* Fully document observers.
* And a ton of other editing, polish, and tightening.
* Updates to theme, TOC, and navigation to make easier / cleaner discoverability.

# To Do for this PR

* [x] add introduction to each of clients/frameworks/observers describing the purpose of that tool for people who are new to it.
* [x] figure out a way to link timespan better from various config docs

# Out of scope for this PR

(or yell at me that it should be in scope :)

* Thrift chapters in tutorial


spladug:
Fixes #301 and #309

baseplate/clients/redis.py
```diff
@@ -26,10 +26,10 @@ def pool_from_config(
 
     * ``url`` (required): a URL like ``redis://localhost/0``.
     * ``max_connections``: an integer maximum number of connections in the pool
-    * ``socket_connect_timeout``: a timespan of how long to wait for sockets
-        to connect. e.g. ``200 milliseconds``.
-    * ``socket_timeout``: a timespan of how long to wait for socket operations,
-        e.g. ``200 milliseconds``.
+    * ``socket_connect_timeout``: how long to wait for sockets to connect. e.g.
```
gtaylor: This may be answered in one of the other files touched in this PR, but I'm not sure what valid input looks like for these timeout kwargs. The example is `200 milliseconds`, so it seems to be a string with the unit name as the second token. Would it stand to reason that I could also do `0.2 seconds`? If I just pass an int, does that work? If so, what does that mean?

This might be a more general question for any other kwarg/param that accepts relative times, but I am not yet sure of whether we have others that fit this description.

docs/api/baseplate/clients/cassandra.rst
```diff
@@ -0,0 +1,62 @@
...
+.. code-block:: ini
+
+   [app:main]
+
+   ...
+
+   # required: a comma-delimited list of hosts to contact to find the ring
+   foo.contact_points = cassandra-01.local, cassandra-02.local
+
+   # optional: the port to connect to on each cassandra server
```
gtaylor: 🔕 worth specifying the default if no value is provided? Or is that best done/already done elsewhere? I always have to dig up these Cassandra ports after longer periods of not having touched Cass or an app that uses Cass.

docs/api/baseplate/clients/cassandra.rst
```diff
@@ -0,0 +1,62 @@
+``baseplate.clients.cassandra``
+===============================
+
+.. automodule:: baseplate.clients.cassandra
+
+This integration supports both the base Python Cassandra driver and the
```
gtaylor: This may be a 🔕 , but would it be worth a sentence or two that says something like "Cassandra is a X. This class is a client that does Y", with a link to Apache Cassandra. Rationale being that some engineers may not have seen Cassandra prior to their stay at reddit.

docs/api/baseplate/clients/hvac.rst
```diff
@@ -0,0 +1,47 @@
+``baseplate.clients.hvac``
+==========================
+
+.. automodule:: baseplate.clients.hvac
+
+To integrate HVAC with your application, add the appropriate client declaration
```
gtaylor: In the same line as my Cass feedback, would it be worth intro'ing Vault and/or HVAC? When I read this first sentence from the perspective of someone who doesn't know Vault or its ecosystem, I am left to dig elsewhere in the docs to figure out what HVAC is/means.

docs/api/baseplate/clients/kombu.rst
```diff
@@ -0,0 +1,63 @@
+``baseplate.clients.kombu``
+===========================
+
+This integration adds support for sending messages to queue brokers via Kombu.
```
gtaylor: Worth linking to Kombu?

docs/api/baseplate/clients/sqlalchemy.rst
```diff
@@ -3,6 +3,41 @@
 
 .. automodule:: baseplate.clients.sqlalchemy
 
+To integrate SQLAlchemy with your application, add the appropriate client
```
gtaylor: Worth linking to SQLAlchemy? It's widely used but couldn't hurt!

docs/api/baseplate/observers/tracing.rst
```diff
@@ -0,0 +1,36 @@
+Tracing
+=======
+
+The tracing observer reports span information to a Zipkin-compatible
```
gtaylor: Worth adding a link to Zipkin?

docs/cli/index.rst
```diff
@@ -0,0 +1,14 @@
+CLI Tools
+=========
+
+.. toctree::
+   :titlesonly:
+
+   baseplate-healthcheck: Is your service alive? <healthcheck>
```
gtaylor: Worth making this a declarative sentence instead of interrogative? The other CLI tools describe what each thing does. This one poses a question that seems to infer that this tool checks for health. We could state as much and not leave that up to interpretation.

docs/tutorial/chapter1/index.rst
```diff
@@ -0,0 +1,151 @@
...
+   Also creating executable in /home/user/tutorial/venv/bin/python
+   Installing setuptools, pkg_resources, pip, wheel...done.
+   $ source venv/bin/activate
+
+.. _`virtual environment`: https://virtualenv.pypa.io/en/stable/
+
+Build a simple Pyramid service
+------------------------------
+
+Pyramid is a mature web framework for Python that we build HTTP services with.
```
gtaylor: Can we link Pyramid here?

docs/index.rst
```diff
@@ -1,78 +1,52 @@
...
+Baseplate.py
+============
 
-Baseplate is a framework to build services on and a library of common and
-well-tested code to share. Its goal is to provide all the common things a
-service needs with as few surprises as possible. It can be divided up into
-three distinct categories of components, listed below.
+It's much easier to manage a bunch of services when they all have the same
+shape: the way they're developed, the way they interact with the infrastructure
+they run on, and the way they interact with each other.  Baseplate is reddit's
```
gtaylor: I feel so petty in raising this but this period is double spaced whereas the rest of the docs seem to mostly be single-spaced.

docs/tutorial/chapter3/index.rst
```diff
@@ -0,0 +1,138 @@
+Observers --- Looking under the hood
+====================================
+
+In the previous chapter, we took our basic Pyramid service and made it
```
gtaylor: Can probably remove this comma for better flow.

README.rst
```diff
@@ -4,10 +4,10 @@ baseplate.py
 |Build Status|
 
 It's much easier to manage a bunch of services when they all have the same
-shape: how they're developed, how they interact with the infrastructure, and
-how they interact with eachother. Baseplate is reddit's specification for the
-common shape of our services, and this library, Baseplate.py, is the Python
-implementation of that specification.
+shape: the way they're developed, the way they interact with the infrastructure
+they run on, and the way they interact with each other.  Baseplate is reddit's
+specification for the common shape of our services, and this library,
```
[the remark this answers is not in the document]
bradengroom: Shouldn't Reddit be capitalized?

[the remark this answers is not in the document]
gtaylor: :popcorn:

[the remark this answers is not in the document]
bradengroom: Or maybe not? I guess it's not in the logo. I'm not so sure now. It is a proper noun though. 🤷‍♂️
I only called it out because my last company was very strict about always capitalizing the brand because it might get confused with a fruit otherwise 😛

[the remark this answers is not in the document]
spladug: Fixed. re: capitalization... :hear_no_evil: :innocent: :)

baseplate/clients/redis.py
spladug: It's one of these: https://spladug.s3.amazonaws.com/baseplate/api/baseplate/lib/config.html#baseplate.lib.config.Timespan

What would be a good way to link back to that?

docs/api/baseplate/clients/cassandra.rst
spladug: Updated. It's 9042. I can't keep track of 'em either.

docs/cli/index.rst
spladug: SGTM. Fixed.

docs/index.rst
spladug: ha! weird. My personal preference would be for double spaces in the source (since it's monospace) and let smartypants/sphinx correct the typography for the actual renderings. That said, it looks like I've been horribly inconsistent about this but leaning towards single space after period.

```console

$ git grep '[^.]\. \S' | wc -l
547

$ git grep '\.  ' | wc -l
120
```

Fixed this one.

README.rst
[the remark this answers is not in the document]
spladug: (brand guidelines aside, we are actually incorporated as `reddit, Inc.` :)

spladug:
> Q: How much do we assume that our audience knows about the Python ecosystem (common libraries, in particular)?

> For the various clients...

Good questions. For the clients, I was initially coming at them from the angle of "you already know you want this library, here's how to use it with baseplate" but I think you have a good point that these docs are potentially the first time someone has heard about a particular library or the database it integrates with. I'll look at adding some front-matter to each of the docs for context and maybe add a "what we expect you to know already" to the tutorial.

> Could we stash a really basic description of Thrift somewhere? What is it? In what cases do we use it instead of HTTP? Maybe a link to an example IDL. Perhaps this could be covered in a tutorial.

Strongly agreed. I'd like to add at least 3 more chapters to the tutorial, with Thrift being a big part of it. The reason I stopped here was that adding thrift gets a lot more complex infra-structurally (you now have to get the compiler installed, set up the build process, write an IDL, etc.) so the "just edit helloworld.py" nature gets lost. What I'm thinking is I'll make a tutorial repo and have those chapters ask the user to clone it down (and maybe run it in vagrant or something?). I just didn't want to block this whole process on getting that stuff written up as it sounds a lot more involved.

> If there was somewhere that we could define the expected format for duration config values/params/kwargs, that might be worth having and linking to judiciously.

https://spladug.s3.amazonaws.com/baseplate/api/baseplate/lib/config.html#baseplate.lib.config.Timespan I'll see if I can link to that better.

docs/api/baseplate/observers/tracing.rst
spladug: we don't actually use zipkin internally anymore, right? just use something that's compatible with its span ingest? I wasn't sure how much to call out the zipkin angle and confuse people.

gtaylor:
> What I'm thinking is I'll make a tutorial repo and have those chapters ask the user to clone it down (and maybe run it in vagrant or something?)

That sounds great! I think you are right about not holding things up for that, FWIW. No need to block this PR.

> https://spladug.s3.amazonaws.com/baseplate/api/baseplate/lib/config.html#baseplate.lib.config.Timespan I'll see if I can link to that better.

Ahh, that's perfect. I had missed that.

docs/api/baseplate/observers/tracing.rst
gtaylor: Correct, no more Zipkin. It sounds like the wider ecosystem might be heading in the direction of OpenTracing, but I think we are sending Zipkin-formatted traces to Wavefront (which accepts Zipkin + OpenTracing traces).

spladug:
Added a bunch of introductory bits to each of the clients/frameworks/observers. How's that?

spladug:
Also have a sorta-crappy linking to timespans going on now. Not in the examples (can't link there) but in the docs for the `_from_config` functions. How's that?

baseplate/clients/redis.py
spladug: Took a stab at just linking this from the various docs for `_from_config` functions

docs/api/baseplate/clients/hvac.rst
```diff
@@ -0,0 +1,70 @@
+``baseplate.clients.hvac``
+==========================
+
+`Vault`_ is a high-security store for secret tokens, credentials, and other
+sensitive information. `HVAC`_ is a Python client library for Vault.
+
+.. note:: The :py:class:`~baseplate.lib.secrets.SecretsStore` handles the most
```
gtaylor: :kisses-fingers:

PR #344, opened 2019-08-19
Add incr_tag interface to spans for observer-generic counters

### LINKED ISSUE #305: Make metrics observer send counters based on span tags
We already map spans to timers, instead of having apps directly send counters let's add tags to spans and send from there. Then it's one big unified API.

spladug:
This is a generic interface on spans that allows you to "increment" a tag. The tracing observer just coalesces these into a final tag, while the metrics observer can now listen into these and send counters as well. The idea is that applications can use this interface and not tie themselves to a specific metrics implementation.

This makes it so that an application doing

```python
context.trace.incr_tag("foo")
```

is equivalent to

```python
context.trace.set_tag("counter.foo", "1")
context.metrics.counter("foo").increment()
```

but without the application ever having to know which observers are configured which gives us more flexibility to change observers in the future and keeps application code simpler.

Fixes #305

### DIFF
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -221,7 +221,7 @@ class TraceSpanObserver(SpanObserver):
         self.elapsed = self.end - typing.cast(int, self.start)
 
         for key, value in self.counters.items():
-            self.binary_annotations.append(self._create_binary_annotation(key, value))
+            self.binary_annotations.append(self._create_binary_annotation(f"counter.{key}", value))
 
         self.recorder.send(self)
 
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -1,4 +1,5 @@
 """Components for processing Baseplate spans for service request tracing."""
+import collections
 import json
 import logging
 import queue
@@ -10,6 +11,7 @@ import typing
 
 from datetime import datetime
 from typing import Any
+from typing import DefaultDict
 from typing import Dict
 from typing import List
 from typing import NamedTuple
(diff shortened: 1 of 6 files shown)


spladug:
:eyeglasses: @pacejackson @ckwang8128

baseplate/observers/tracing.py
```diff
@@ -216,7 +219,11 @@ def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
 
         self.end = current_epoch_microseconds()
         self.elapsed = self.end - typing.cast(int, self.start)
-        self.record()
+
+        for key, value in self.counters.items():
+            self.binary_annotations.append(self._create_binary_annotation(key, value))
```
pacejackson: Any concerns about cardinality with this?  Also, should the `key` include something like `counter-` in it so we don't have to worry about some random metric colliding with something else?

baseplate/observers/tracing.py
```diff
@@ -282,10 +292,6 @@ def _serialize(self) -> Dict[str, Any]:
 
         return self._to_span_obj(annotations, self.binary_annotations)
 
-    def record(self) -> None:
```
pacejackson: 🔕 I'm assuming this got removed from the interface and we just missed this one?

spladug: nah, just removing a pretty unnecessary function. it was always like this :man_shrugging:

baseplate/observers/tracing.py
spladug: good points, will do

spladug: Added a namespace. I don't think cardinality is an issue for tag values on a tracing system, and is no worse than `client.counter("...").increment()` for metrics. Do you think this API is too abusable?

spladug:
:haircut_man:

baseplate/observers/tracing.py
pacejackson: I don’t think so, just didn’t know if we needed to worry about tracing tag cardinality. 🙂

PR #355, opened 2019-09-18
Add QueueConsumerServer

pacejackson:
This PR adds a new way to run queue consumers as a "server".  This means that you will run your queue consumers using `baseplate-serve` rather than `baseplate-script`.  The server allows you to process multiple messages concurrently (useful when you don't care about ordering and are IO bound rather than CPU bound) and also will attempt to shut down your consumers gracefully.

Note that this approach largely sidesteps and deprecates the existing queue consumer flow rather than repurposing it.

This PR also adds a framework for using custom serialization options with Kombu (such as Thrift which is also implemented here).


pacejackson:
@bsimpson63 @spladug If either of ya'll get a chance, I would appreciate some feedback on this and the approach I'm taking before diving too deep into writing a bunch of docs and tests.

pacejackson:
👓 @spladug @bsimpson63

pacejackson:
I think this is _mostly_ ready for review, I still need to update some documentation, do some manual testing, and figure out how we can help developers move off of `consume` and onto the `server` model.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+    """An error that signals that the queue process should exit.
+
+    Raising an Exception that is a subclass of AbortError will cause the
+    KombuMessageHandler to re-raise the exception rather than swallowing it
+    which will cause the handler thread/process to stop.  This, in turn, will
+    gracefully shut down the QueueConsumerServer currently running.
+
+    Exceptions of this nature should be reserved for errors that are due to
+    problems with the environment rather than the message itself.  For example,
+    a node that cannot get it's AWS credentials.
```
bsimpson63: nit: its

baseplate/clients/kombu.py
```diff
@@ -60,6 +66,53 @@ def exchange_from_config(app_config: config.RawConfig, prefix: str, **kwargs: An
...
+class KombuSerializer(abc.ABC):
+    """Interface for wrapping non-built-in serializers for Kombu."""
+
+    @property
+    @abc.abstractmethod
+    def name(self) -> str:
+        pass
+
+    @abc.abstractmethod
+    def serialize(self, obj: Any) -> bytes:
```
pacejackson: Note, I want to take a look at generics and see if that is appropriate in some way.

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+        """Run the worker.
+
+        This is the method called by the the PumpWorker Thread within the
+        QueueConsumerServer and is ultimately responsible for getting messages
+        off of the message queue and putting them on the work queue.
+        """
+
+    @abc.abstractmethod
+    def stop(self) -> None:
+        """Signal the PumpWorker that it should stop receiving new messages from it's message queue."""
```
bsimpson63: its

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+    @abc.abstractmethod
+    def handle(self, message: Any) -> None:
+        """Handle a message supplied by a PumpWorker.
+
+        This method is what the server calls when it reads a new message off of
+        the internal work queue that is fed by the PumpWorker.  This is both the
+        method where you should handle processing the message and setting up
+        the RequestContext and ServerSpan for baseplate.
+
+        While not required, it is reccomended that allow the caller to pass in the
```
bsimpson63: recommended to allow ?

bsimpson63: Looking at `KombuMessageHandler` it appears that the caller passes the function to process the message in the constructor, is that right? That all makes sense to me, but the description here makes it seem like that function would be an argument to `handle()` (rather than writing implementations/subclasses of `MessageHandler.handle()` that do the message processing)

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+                # new message.  If we did not do this, we would not be able to
+                # wait for all of our workers to finish before stopping the
+                # server.
+                message = self.work_queue.get(timeout=self._queue_timeout)
+            except queue.Empty:
+                pass
+            else:
+                # Ensure that if self.message_handler.handle throws a `queue.Empty`
+                # error, that bubbles up and is not treated as though `self.work_queue`
+                # is empty
```
bsimpson63: is this because we don't know anything about how `message_handler.handle` works so it could be running its own queue and hit `queue.Empty`?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+        work_queue: WorkQueue,
+        serializer: Optional[KombuSerializer] = None,
+    ):
+        self.connection = connection
+        self.queues = queues
+        self.work_queue = work_queue
+        self.serializer = serializer
+
+    def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
+        args = dict(queues=self.queues, on_message=self.work_queue.put)
```
bsimpson63: what happens when `work_queue` is at its max size and `kombu.Consumer` tries to put a new message on?

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+        assert self.started
+        if not self.stopped:
+            os.kill(os.getpid(), signal.SIGTERM)
+
+    def _watch(self) -> None:
+        """Watches the threads to see if they have stopped prematurely.
+
+        Loops through the pump thread and handler threads periodically to check
+        if they are still alive and terminates the server if one is no longer
+        alive.
```
bsimpson63: sweeeeet

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        assert self.started
+        watched = [self.pump_thread] + self.threads
+        while not self.stopped:
+            for thread in watched:
+                if not thread.is_alive():
+                    logger.debug("Dead thread found, terminating server.")
+                    self._terminate()
```
bsimpson63: will we get any metrics around this? will we be able to see if the servers are constantly crashing (other than reading the logs)?

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,268 @@
...
+        self.stopped = True
+        logger.debug("Stopping pump thread.")
+        self.pump.stop()
+        logger.debug("Stopping message handler threads.")
+        for handler in self.handlers:
+            handler.stop()
+        logger.debug("Waiting for message handler threads to drain.")
+        for thread in self.threads:
+            thread.join()
+        logger.debug("Stopping healthcheck server.")
```
bsimpson63: is there any ordering requirement here? is there any for the startup?

bsimpson63:
💅 this looks awesome. I just had a couple questions and comments

baseplate/server/queue_consumer.py
pacejackson: > it appears that the caller passes the function to process the message in the constructor, is that right

Correct, I'll touch up the language here.  All it's trying to express is that you should do what `KombuMessageHandler` does, not implement the logic to actually process the message directly in the `handle` method of the `MessageHandler` subclass you implement.

baseplate/server/queue_consumer.py
pacejackson: Correct.

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: It blocks until something is removed.  When running under gevent, this is monkey-patched so you yield to the event loop.

The reason for doing this is if you don't bind the queue so the consumer has to wait on it, the consumer will inhale messages and cause the process to OOM if you have enough messages.

baseplate/server/queue_consumer.py
pacejackson: FYI, I think there's probably a better way to do this that I'm gonna put up soon 😅 

```python
def watcher(fn):
    def _run_and_terminate(*a, **kw):
        try:
            return fn(*a, **kw)
        except Exception:
            self._terminate()

    return _run_and_terminate

self.pump_thread = Thread(target=watcher(self.pump.run), daemon=True)
self.threads = [Thread(target=watcher(handler.run)) for handler in self.handlers]
```

baseplate/server/queue_consumer.py
pacejackson: AFAIK, we don't really have a good way to report metrics outside of the context of a request since the actual sending of metrics is tied to that lifecyle, so not specifically here.

1. We should have metrics on errors in the handler threads which would cause this
2. I would _like_ to support metrics like this if there's a way to @spladug ?

baseplate/server/queue_consumer.py
pacejackson: As this is currently written, yes for startup (you want to start the watcher thread after all of the other threads it is meant to watch).

For shutdown, the ordering of `handler.stop` before `thread.join` is important (otherwise you'll be waiting forever on threads that haven't been stopped) but the rest is _less_ important. Stopping the pump first is nice so you stop consuming messages from your external message queue and stopping the healthchecker last is probably a good idea in case you have a system that aggressively kills processes that fail their health checks.

pacejackson:
💇‍♂ Addressed the feedback, I'm going to try and find some time this week to do some testing and updating the documentation.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+            queues=self.queues,
+            work_queue=work_queue,
+            serializer=self.serializer,
+        )
+
+    def build_message_handler(self) -> KombuMessageHandler:
+        return KombuMessageHandler(self.baseplate, self.name, self.handler_fn)
+
+    def build_health_checker(self, listener: socket.socket) -> StreamServer:
+        baseplate_configurator = BaseplateConfigurator(self.baseplate)
```
spladug: do we actually need this whole baseplatey pyramid stack for this? could just make a trivial wsgi app directly?

```python
def healthcheck(environ, start_response):
    start_response("200 OK", [("Content-Type", "text/plain")])
    return ["all good, boss"]

return WSGIServer(listener=listener, application=healthcheck)
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+                delivery_info = message.delivery_info
+                span.set_tag("kind", "consumer")
+                span.set_tag("routing_key", delivery_info.get("routing_key", ""))
+                span.set_tag("consumer_tag", delivery_info.get("consumer_tag", ""))
+                span.set_tag("delivery_tag", delivery_info.get("delivery_tag", ""))
+                span.set_tag("exchange", delivery_info.get("exchange", ""))
+                self.handler_fn(context, message)
+        except Exception as exc:
+            logger.exception(
+                "Unhandled error while trying to process a message, re-queueing the message."
```
spladug: is there a clearer terminology for requeue that captures the possibility of deadletter queues? (i see the kombu method is called that but this is our messaging to the end user that i want to be clear in)

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+    def handle(self, message: kombu.Message) -> None:
+        context = self.baseplate.make_context_object()
+        try:
+            # We place the call to ``baseplate.make_server_span`` inside the
+            # try/except block because we still want Baseplate to see and
+            # handle the error (publish it to error reporting)
+            with self.baseplate.make_server_span(context, self.name) as span:
+                delivery_info = message.delivery_info
+                span.set_tag("kind", "consumer")
+                span.set_tag("routing_key", delivery_info.get("routing_key", ""))
```
spladug: maybe prefix these tags with `amqp.` like is done for HTTP-specific tags? https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+    which will cause the handler thread/process to stop.  This, in turn, will
+    gracefully shut down the QueueConsumerServer currently running.
+
+    Exceptions of this nature should be reserved for errors that are due to
+    problems with the environment rather than the message itself.  For example,
+    a node that cannot get its AWS credentials.
+    """
+
+
+class KombuConsumerWorker(ConsumerMixin, PumpWorker):  # pylint: disable=abstract-method
```
spladug: why disable abstract-method?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+Handler = Callable[[RequestContext, kombu.Message], None]
+HealthCheck = Callable[[RequestContext], bool]
+
+
+class AbortError(Exception):
```
spladug: is this just meant for users to use? how about something like `ConsumerFatalError`? worried it's a bit generic of a name.

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,269 @@
...
+
+from gevent.server import StreamServer
+
+import baseplate.lib.config
+
+logger = logging.getLogger(__name__)
+
+
+class PumpWorker(abc.ABC):
+    """Reads messages off of a message queue and puts them into a queue.Queue for handling by a  MessageHandler.
```
spladug: nit: mysterious double space before MessageHandler

baseplate/server/queue_consumer.py
```diff
@@ -0,0 +1,269 @@
...
+                # server.
+                message = self.work_queue.get(timeout=self._queue_timeout)
+            except queue.Empty:
+                pass
+            else:
+                # Ensure that if self.message_handler.handle throws a `queue.Empty`
+                # error, that bubbles up and is not treated as though `self.work_queue`
+                # is empty
+                self.message_handler.handle(message)
+        logger.debug("Consumer <%s> stopping.", self.id)
```
spladug: there's a lot of debug stuff around here that seems useful during dev of this module but might be annoying when doing `baseplate-serve --debug` like in dev. should we define a TRACE level? clean these up? deal with it?

tests/unit/clients/kombu_tests.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+from baseplate.clients.kombu import KombuThriftSerializer
+from baseplate.thrift.ttypes import Loid
+from baseplate.thrift.ttypes import Request
+from baseplate.thrift.ttypes import Session
+
+
+class TestKombuThriftSerializer:
```
spladug: since we're all pytesty now, any reason to use a class at all?

tests/unit/server/queue_consumer_tests.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def test_stop_before_start(self, server):
+        with pytest.raises(AssertionError):
+            server.stop()
+
+    def test_pump_exception_terminates(self, build_server):
+        server = build_server(max_concurrency=1, pump_raises=Exception())
+        server._terminate = mock.Mock()
+        server.start()
+        time.sleep(2)
```
spladug: oof, this'll add a chunk o' time to the tests. any way to do this with shorter sleeps?

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: `run` is defined by `ConsumerMixin` but the linter doesn't seem to be aware of that, I'll try running it again though just to double check.

tests/unit/clients/kombu_tests.py
pacejackson: If `kombu.py` was _only_ the serializer stuff, I wouldn't use a class.  In this case the tests file is only testing one thing but the file it's testing has several distinct things that could be tested, in which case I think it's nice to use classes just to organize the tests.  Given that, I thought it made sense to use a class here so if/when someone decides to add tests for other parts of the file, they don't have to try and figure out a way to organize the existing tests.

baseplate/frameworks/queue_consumer/kombu.py
spladug: if you swap the order of the base classes does that change it?

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: Hmm, maybe say something like `"the message has been re-delivered to the message queue"?  (Redelivered is the term RabbitMQ appears to use).

spladug: or like "returned to the queue broker" or something?

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: The order actually _is_ already swapped from when I first ran this, I actually update the order because it wasn't working.  So pylint caught my problem and I told it to go away. 🤣 

I confirmed that I don't need this when the code actually works.

baseplate/server/queue_consumer.py
pacejackson: I'm not sure that they would be that annoying?  The messages only happen around server startup and shutdown, not while the server itself is running.

pacejackson: I also don't mind removing them if that is preferable.

pacejackson:
💇‍♂ modulo updating the documentation (I'll try to have that ready for review before the end of the week).

baseplate/server/queue_consumer.py
spladug: yeah, that's fair. i was overthinking. :+1:

baseplate/clients/kombu.py
```diff
@@ -73,14 +143,17 @@ class KombuProducer(config.Parser):
 
     """
 
-    def __init__(self, max_connections: Optional[int] = None):
+    def __init__(
+        self, max_connections: Optional[int] = None, serializer: Optional[KombuSerializer] = None
+    ):
         self.max_connections = max_connections
+        self.serializater = serializer
```
spladug: serializater???

baseplate/frameworks/queue_consumer/__init__.py
```diff
@@ -0,0 +1,22 @@
+from baseplate.frameworks.queue_consumer._depr import BaseKombuConsumer
+from baseplate.frameworks.queue_consumer._depr import consume
+from baseplate.frameworks.queue_consumer._depr import Handler
+from baseplate.frameworks.queue_consumer._depr import KombuConsumer
+from baseplate.frameworks.queue_consumer._depr import WorkQueue
+from baseplate.frameworks.queue_consumer.kombu import KombuConsumerWorker as _ConsumerWorker
+from baseplate.lib import warn_deprecated
+
+warn_deprecated(
```
spladug: does this not get called when importing the non-depr version?

pacejackson: Gah, I was hoping it would only happen when you directly imported from `queue_consumer` but that doesn't appear to be the case.  😭

baseplate/clients/kombu.py
pacejackson: 🤦‍♂ 

That one is just impressive

tests/unit/clients/kombu_tests.py
```diff
@@ -1,10 +1,96 @@
...
+            does_not_raise(),
+            {
+                "hostname": "rabbit.local",
+                "userid": "spez",
+                "password": "hunter2",
+                "virtual_host": "/",
+            },
+        ),
+    ],
+)
```
pacejackson: How do you feel about this super pytest-y parameterize test style @spladug ?  I think it's neat but maybe isn't the best use case for a test like this?  Happy write it either way, more was just interested in writing one and seeing how folks feel about it.

pacejackson:
💇‍♂ 

I think this is ready for another round of review

1. Squashed the previous changes down into 2 commits
2. I fixed a bug that came up when manually testing this
3. I took a stab at updating the documentation
4. Added support for "credentials" to get the username and password when creating a kombu connection
5. Added "stop_timeout" to the server to keep it in line with the other baseplate servers.

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,53 @@
+``baseplate.frameworks.queue_consumer.kombu``
+=============================================
+
+:doc:`Kombu <kombu:index>` is a library for interacting with queue brokers.
+
+This module provides a :py:class:`~baseplate.server.queue_consumer.QueueConsumerFactory`
+that allows you to run a :py:class:`~baseplate.server.queue_consumer.QueueConsumerServer`
+using Kombu that also integrates Baseplate's facilities into Kombu's message lifecycle.
```
spladug: "to run a QueueConsumerServer that integrates with Kombu" maybe?

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,53 @@
...
+
+An abbreviated example of it in use::
+
+    def process_links(context, body, message):
+        print(f"processing {body}")
+
+    def make_consumer_factory(app_config):
+        baseplate = Baseplate()
+        exchange=Exchange("reddit_exchange", "direct"),
```
spladug: these bits look like they're written as params to a function but they're top-level statements.

maybe turn this into a doctest so at least we know it's valid syntax even if we don't go all the way with it?

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,53 @@
...
+
+:doc:`Kombu <kombu:index>` is a library for interacting with queue brokers.
+
+This module provides a :py:class:`~baseplate.server.queue_consumer.QueueConsumerFactory`
+that allows you to run a :py:class:`~baseplate.server.queue_consumer.QueueConsumerServer`
+using Kombu that also integrates Baseplate's facilities into Kombu's message lifecycle.
+
+An abbreviated example of it in use::
+
+    def process_links(context, body, message):
```
spladug: worth adding type annotations to this example so people know what they're getting?

docs/api/baseplate/frameworks/queue_consumer/deprecated.rst
```diff
@@ -0,0 +1,144 @@
...
+sections to what you would have for an HTTP or Thrift service.
+
+.. code-block:: diff
+
+    [DEFAULT]
+    rabbitmq.hostname = amqp://rabbit.local:5672
+    rabbitmq.exchange_name = my-exchange
+    rabbitmq.exchange_type = direct
+
+    + [app:main]
```
spladug: maybe a name other than `main`?

baseplate/clients/kombu.py
```diff
@@ -60,6 +87,70 @@ def exchange_from_config(app_config: config.RawConfig, prefix: str, **kwargs: An
...
+        """Deserialize the message bytes into an object for consuming."""
+
+
+class KombuThriftSerializer(KombuSerializer[T]):  # pylint: disable=unsubscriptable-object
+    """Thrift object serializer for Kombu."""
+
+    def __init__(
+        self,
+        thrift_class: Type[T],
+        factory_class: TProtocolFactory = TBinaryProtocolAcceleratedFactory,
```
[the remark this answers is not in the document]
pacejackson: > why not allow the instantiation to happen outside?

I dunno... it seems kinda silly, I'll update it.

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
pacejackson: > these bits look like they're written as params to a function but they're top-level statements.

Sorry about that, I fixed that in a later commit (copy/paste oversight).

docs/api/baseplate/frameworks/queue_consumer/deprecated.rst
pacejackson: I wasn't sure about that?  I didn't want people to think that they _had_ to set an app/server name when doing this and I thought that if I put that in the example, it might give that impression.

spladug: fair enough. the whole thing's a bit confusing anyway :(

pacejackson: Ya, I can see it either way :/

baseplate/frameworks/queue_consumer/deprecated.py
```diff
@@ -72,6 +69,11 @@ def consume(
     :param handler: The handler method.
 
     """
+    warn_deprecated(
+        "baseplate.frameworks.queue_consumer is deprecated and will be removed "
+        "in the next major release.  You should migrate your consumers to use "
+        "baseplate.server.queue_consumer."
```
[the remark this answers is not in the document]
pacejackson: > shortlink if possible

Is that a sphinx thing?  Since this is outside of that realm I'm guessing I can't do that?

pacejackson:
💇‍♂

baseplate/server/queue_consumer.py
spladug: oh, I missed this one. We've got the runtime monitor stuff now! You can report server-level metrics there. 

That said, if we just want to see if it's in a crash loop, the infra can tell us how long the process has been alive.

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,61 @@
...
+    ):
+        print(f"processing {body}")
+
+    def make_consumer_factory(app_config):
+        baseplate = Baseplate()
+        exchange = Exchange("reddit_exchange", "direct"),
+        connection = Connection(
+          hostname="amqp://guest:guest@reddit.local:5672",
+          virtual_host="/",
+        ),
```
spladug: still extraneous comma

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,61 @@
...
+        queue_name = "process_links_q",
+        routing_keys = ["link_created"],
```
spladug: and here

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,61 @@
...
+        print(f"processing {body}")
+
+    def make_consumer_factory(app_config):
+        baseplate = Baseplate()
+        exchange = Exchange("reddit_exchange", "direct"),
+        connection = Connection(
+          hostname="amqp://guest:guest@reddit.local:5672",
+          virtual_host="/",
+        ),
+        queue_name = "process_links_q",
```
spladug: and here

docs/api/baseplate/frameworks/queue_consumer/kombu.rst
```diff
@@ -0,0 +1,61 @@
...
+    def process_links(
+        context: RequestContext,
+        body: Any,
+        message: kombu.Message,
+    ):
+        print(f"processing {body}")
+
+    def make_consumer_factory(app_config):
+        baseplate = Baseplate()
+        exchange = Exchange("reddit_exchange", "direct"),
```
spladug: and here

PR #356, opened 2019-09-19
Pylint

jennLam:
add pylint plugin to check for string formatting in cassandra queries

👓 @spladug


baseplate/lint/cql_string_format_plugin.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class NoCQLStringFormatChecker(BaseChecker):
+    __implements__ = IAstroidChecker
+
+    name = "no-cql-string-format"
+    priority = -1
+    msgs = {
+        "W0001": (
+            "Python string formatting found in Cassandra database query",
```
spladug: Would this also fire if someone does this with a SQLAlchemy query? I think that's a good thing but we may want to update the messages to not be Cassandra-specific in that context.

jennLam:
Changes made since last review:

- changed linter name from NoCQLStringFormatChecker to NoDbQueryStringFormatChecker
- add example checker with comments
- add example checker test

- add linters doc

baseplate/lint/db_query_string_format_plugin.py
```diff
@@ -0,0 +1,89 @@
...
+            variable_is_query = True
+        elif (
+            isinstance(node.value, nodes.Call)
+            and isinstance(node.targets[0], nodes.Name)
+            and isinstance(node.value.func, nodes.Attribute)
+            and node.value.func.attrname == "format"
+            and self.check_string_is_query(node.value.func.expr.value)
+        ):
+            variable_is_query = True
+        if variable_is_query and node.targets[0].name not in self.string_sub_queries:
```
spladug: readability thing: could you add a blank line before this `if` block? on first read, i thought these were more `elif` branches because they're all jammed together so i missed that `variable_is_query` gets used in these conditionals.

baseplate/lint/example_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+            # message-help shown to user when calling pylint --help-msg
+            "Ensure variables are not reassigned.",
+        )
+    }
+
+    def __init__(self, linter: PyLinter = None):
+        super().__init__(linter)
+        self.variables: set = set()
+
+    # The following two methods are how we visit and leave desired nodes from the astroid library
```
spladug: might be good to mention that these get called for us by pylint/astroid as it walks the tree

docs/lint/index.rst
```diff
@@ -0,0 +1,63 @@
...
+.. _`Install Pylint`: https://pylint.readthedocs.io/en/latest/user_guide/installation.html
+
+Follow the `Pylint user guide`_ for instructions to generate a default pylintrc configuration file and run Pylint.
+
+.. _`Pylint user guide`: https://pylint.readthedocs.io/en/latest/user_guide/run.html
+
+Adding Custom Checkers
+^^^^^^^^^^^^^^^^^^^^^^
+
+In your pylintrc file, add baseplate.lint to the [MASTER] load-plugins configuration.
```
spladug:
```suggestion
In your ``pylintrc`` file, add ``baseplate.lint`` to the ``[MASTER]`` load-plugins configuration.
```

tests/unit/lint/db_query_string_format_plugin_tests.py
```diff
@@ -0,0 +1,162 @@
...
+from baseplate.lint import db_query_string_format_plugin
+
+
+class TestNoCQLStringFormatChecker(pylint.testutils.CheckerTestCase):
+    CHECKER_CLASS = db_query_string_format_plugin.NoDbQueryStringFormatChecker
+
+    def test_finds_variable_binop_string_format_query(self):
+        assign_node_a, call_node_b = astroid.extract_node(
+            """
+        query = "select * from %s" % "users" #@
```
pacejackson: I don't believe the Cassandra provide any mechanism for inserting the table name into a query other than string substitution like this so I'm not sure we _want_ to catch these case?

baseplate/lint/db_query_string_format_plugin.py
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    def __init__(self, linter: PyLinter = None):
+        super().__init__(linter)
+        self.string_sub_queries: set = set()
+        self.query_verbs = {"select", "update", "delete", "insertinto", "truncate"}
+
+    def check_string_is_query(self, string: str) -> bool:
+        query = string.split(" ")
+        if query[0].lower() in self.query_verbs and ("{}" or "%s" in query):
```
pacejackson: Note, I don't think this will catch f-strings or the case where you name the format values, both of which are `"{some_name_or_code}"`.

tests/unit/lint/db_query_string_format_plugin_tests.py
spladug: Do we actually want to encourage that sort of query building though? I think the test string could be written to put the %s in the WHERE to be absolutely clear about the intent, but having done this sort of table-name-insertion before I think it's generally a bad idea and if it's _really really_ necessary then the dev can use the `# pylint: disable` syntax.

pacejackson: I think it's unavoidable, this comes up in several places:

1. You are trying to write a generic function to load multiple models concurrently: 
https://github.snooguts.net/reddit/baseplate-cookiecutter/blob/3d04301ef573e87ec30c98af3d48f2255d29c29f/baseplate_cookiecutter/make_model/templates/model_init.py#L8
2. Adding a non-hardcoded `LIMIT` to your query.

The Cassandra drivers don't have a proper query building interface the same way sqlalchemy does and it can only do substitution on arguments, not components of a query.

spladug: OK, so you're saying this will need to fully parse and understand the query and only apply the rule in the dangerous positions?

pacejackson: Probably although doing the `# pylint: disable` in those sort of cases probably isn't the end of the world (although that can also build up de-sensitivity to the lint rule itself so devs might just ignore it or always silence it even when not appropriate), but I don't think string substitution is something we can fully avoid.

spladug: How about we require that the % or {} is preceded by one of the operators like `=` so we know it's in value position of some expression? Or maybe if it's surrounded by quotes? Some initial heuristic to catch a good number of the options.

@largenocream maybe you have some thoughts on this one too?

spladug: Talked some more with Andrew and I think we're going to go forward with the checks as-is and move the current main contender to a well-tested library. https://github.com/reddit/cqlmapper/pull/20/files

JordanMilne: I think there are some cases of interpolation we're always fine with, like the `d` format specifier case (necessary for dynamic `LIMIT`s and `TTL`s like @pacejackson mentioned.) `r` is generally bad in addition to the `s` we're checking for.

Only looking behind in the format string to see if we think we're in a value position and only disallowing those is problematic because string interpolation outside of the value position.

Consider something like `"SELECT %s from foo" % ','.join(wanted_fields)` where `wanted_fields` is user-supplied. That kind of dynamism is probably uncommon enough that we should have to specifically silence the check when it's necessary and have a reviewer verify that there's not a better way to do it / the data isn't user-supplied.

To that end, is there any way to have PyLint mention that there are new silenced errors in the test / summon security if they're security-relevant?

spladug: Cool, we can perhaps make sure that `%d` is allowed and then continue ahead with the current level of disallowing all other formatting, regardless of operators etc. Since the main other violator right now is the function Andrew's moving to CQLMapper we should be OK. (and getting that out of copy-paste mode is a win anyway imo).

> To that end, is there any way to have PyLint mention that there are new silenced errors in the test / summon security if they're security-relevant?

Not pylint, but we could definitely figure out something to hook up in CI for that.

PR #358, opened 2019-10-01
Default to 10 second wait for graceful server shutdown

spladug:
The default was previously 0. I thought that meant wait forever, but
instead it meant die immediately. After fixing graceful shutdown, this
meant we'd still be hard-killing in flight requests but now we'd also be
getting the GreenletExit errors from it. This default of 10 seconds is
meant to give some time for long-running requests while not holding up
shutdown too long.

### DIFF
--- a/baseplate/server/thrift.py
+++ b/baseplate/server/thrift.py
@@ -1,3 +1,4 @@
+import datetime
 import socket
 
 from typing import Any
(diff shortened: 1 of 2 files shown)


baseplate/server/thrift.py
```diff
@@ -59,7 +59,7 @@ def make_server(server_config: Dict[str, str], listener: socket.socket, app: Any
         server_config,
         {
             "max_concurrency": config.Integer,
-            "stop_timeout": config.Optional(config.Integer, default=0),
+            "stop_timeout": config.Optional(config.Integer, default=10),
```
m104: These should probably be `config.Timespan` as a config value of `15 seconds` doesn't work with `config.Int`.

baseplate/server/thrift.py
```diff
@@ -59,13 +60,15 @@ def make_server(server_config: Dict[str, str], listener: socket.socket, app: Any
         server_config,
         {
             "max_concurrency": config.Integer,
-            "stop_timeout": config.Optional(config.Integer, default=0),
+            "stop_timeout": config.Optional(
+                config.TimespanWithLegacyFallback, default=datetime.timedelta(seconds=10)
```
pacejackson: TIL

baseplate/server/thrift.py
spladug: thanks! definitely want to be consistent on using timespans. updated.

m104: If we use a value like `15` right now, that will be backwards compatible with the update, correct?

spladug: correct

m104: Perfecto, thanks!

PR #360, opened 2019-10-08
Allow backoff value to be set for filewatchers

diffyqgirl:
Reduce spam in graphql logs caused by excessive retries from secrets and experiments filewatchers.

### DIFF
--- a/baseplate/lib/experiments/__init__.py
+++ b/baseplate/lib/experiments/__init__.py
@@ -15,7 +15,6 @@ from baseplate.lib.events import DebugLogger
 from baseplate.lib.events import EventLogger
 from baseplate.lib.experiments.providers import parse_experiment
 from baseplate.lib.experiments.providers.base import Experiment
-from baseplate.lib.file_watcher import DEFAULT_FILEWATCHER_BACKOFF
 from baseplate.lib.file_watcher import FileWatcher
 from baseplate.lib.file_watcher import WatchedFileNotAvailableError
 
@@ -49,7 +48,7 @@ class ExperimentsContextFactory(ContextFactory):
         path: str,
         event_logger: Optional[EventLogger] = None,
         timeout: Optional[float] = None,
-        backoff: Optional[float] = DEFAULT_FILEWATCHER_BACKOFF,
+        backoff: Optional[float] = None,
     ):
         self._filewatcher = FileWatcher(path, json.load, timeout=timeout, backoff=backoff)
         self._event_logger = event_logger
@@ -310,6 +309,6 @@ def experiments_client_from_config(
     if options.backoff:
         backoff = options.backoff.total_seconds()
     else:
-        backoff = DEFAULT_FILEWATCHER_BACKOFF
+        backoff = None
 
     return ExperimentsContextFactory(options.path, event_logger, timeout=timeout, backoff=backoff)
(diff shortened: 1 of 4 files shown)


diffyqgirl:
👓 @m104

diffyqgirl:
👓 @spladug

diffyqgirl:
Mystifying weirdness: Originally I had several commits that passed tests, then I squashed and force-pushed them to clean up the commit log. However, the resulting commit failed one asynchronity test, only in python 3.7 (not 3.6), seen here:
https://cloud.drone.io/reddit/baseplate.py/495/2/8

The test seemed totally unrelated to what I'm changing here so I committed an empty commit to get drone to rerun and everything passed. Is this something I need to worry about?

baseplate/lib/experiments/__init__.py
```diff
@@ -15,6 +15,7 @@
 from baseplate.lib.events import EventLogger
 from baseplate.lib.experiments.providers import parse_experiment
 from baseplate.lib.experiments.providers.base import Experiment
+from baseplate.lib.file_watcher import DEFAULT_FILEWATCHER_BACKOFF
```
pacejackson: Importing this default all over and setting it as the default seems weird to me, I think a cleaner way to do it would be to always default to `None` and in the actual file_watcher code have something like:

```python
backoff = backoff or 0.01
```

spladug:
> The failing test seemed totally unrelated to what I'm changing here so I committed an empty commit to get drone to rerun and everything passed. Is this something I need to worry about?

Sounds like a flakey test. Don't worry about it for your PR. I've opened #361 to track fixing it.

baseplate/lib/experiments/__init__.py
```diff
@@ -44,9 +44,13 @@ class ExperimentsContextFactory(ContextFactory):
     """
 
     def __init__(
-        self, path: str, event_logger: Optional[EventLogger] = None, timeout: Optional[float] = None
+        self,
+        path: str,
+        event_logger: Optional[EventLogger] = None,
+        timeout: Optional[float] = None,
+        backoff: Optional[float] = None,
```
spladug: can you add documentation of this new parameter to the docstring above please?

baseplate/lib/file_watcher.py
```diff
@@ -100,6 +102,7 @@ def __init__(
         binary: bool = False,
         encoding: Optional[str] = None,
         newline: Optional[str] = None,
+        backoff: Optional[float] = None,
```
spladug: same request for documentation please

diffyqgirl:
> Just some docs please. Looks good otherwise. Alternatively, is there any reason not to just hard code a backoff and skip threading it through all the APIs?

Previously it was hardcoded to 0.01 seconds, you mean hardcode it to a higher value? We could do that, I didn't because 0.01 was working fine for some of graphql's use cases, just not all of them.

baseplate/lib/experiments/__init__.py
diffyqgirl: Done.

baseplate/lib/secrets.py
```diff
@@ -335,11 +335,17 @@ def secrets_store_from_config(
         app_config,
         {
             config_prefix: {
-                "path": config.Optional(config.String, default="/var/local/secrets.json")
+                "path": config.Optional(config.String, default="/var/local/secrets.json"),
+                "backoff": config.Optional(config.Timespan),
```
pacejackson: This needs a doc update as well.

diffyqgirl: Done.

baseplate/lib/file_watcher.py
diffyqgirl: Done.

baseplate/lib/experiments/__init__.py
diffyqgirl: Done.

spladug:
Thank you!

PR #365, opened 2019-10-31
support pool_recycle in engine_from_config

prthiaga:
As part of thing->k8s migration, we noticed that baseplate does not support pool_recycle while creating the db engine.

Added the support for pool_recycle in https://github.snooguts.net/reddit/reddit-service-thing/commit/cc0bbdea642023f5dfc3c6798974596ab5f72f88

We wanted to push the changes to baseplate as well so other services that use pool_recycle do not run into a similar issue when moving to k8s.

👓 @bsimpson63 @spladug

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -45,11 +45,17 @@ def engine_from_config(
         credentials from ``secrets`` as a :py:class:`~baseplate.lib.secrets.CredentialSecret`.
         If this is supplied, any credentials given in ``url`` we be replaced by
         these.
+    * ``pool_recycle`` (optional): this setting causes the pool to recycle connections after
+        the given number of seconds has passed. It defaults to -1, or no timeout.
 
     """
     assert prefix.endswith(".")
     parser = config.SpecParser(
-        {"url": config.String, "credentials_secret": config.Optional(config.String)}
+        {
+            "url": config.String,
+            "credentials_secret": config.Optional(config.String),
+            "pool_recycle": config.Optional(config.String),
+        }
     )
     options = parser.parse(prefix[:-1], app_config)
     url = make_url(options.url)
@@ -61,7 +67,10 @@ def engine_from_config(
         url.username = credentials.username
         url.password = credentials.password
 
-    return create_engine(url, **kwargs)
+    if options.pool_recycle:
+        return create_engine(url, pool_recycle=int(options.pool_recycle), **kwargs)
+    else:
+        return create_engine(url, **kwargs)
 
 
 class SQLAlchemySession(config.Parser):
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -60,7 +60,7 @@ def engine_from_config(
     options = parser.parse(prefix[:-1], app_config)
     url = make_url(options.url)
 
-    if options.pool_recycle:
+    if options.pool_recycle is not None:
         kwargs.setdefault("pool_recycle", options.pool_recycle)
 
     if options.credentials_secret:
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -54,7 +54,7 @@ def engine_from_config(
         {
             "url": config.String,
             "credentials_secret": config.Optional(config.String),
-            "pool_recycle": config.Optional(config.String),
+            "pool_recycle": config.Optional(config.Integer),
         }
     )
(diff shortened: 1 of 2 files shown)


baseplate/clients/sqlalchemy.py
```diff
@@ -61,7 +67,10 @@ def engine_from_config(
         url.username = credentials.username
         url.password = credentials.password
 
-    return create_engine(url, **kwargs)
+    if options.pool_recycle:
```
spladug: Instead of having two whole calls to `create_engine`, it'd be cleaner to follow the pattern used in other clients where we just `setdefault()` on `kwargs`.

https://github.com/reddit/baseplate.py/blob/master/baseplate/clients/cassandra.py#L77-L78

baseplate/clients/sqlalchemy.py
```diff
@@ -45,11 +45,17 @@ def engine_from_config(
...
+        the given number of seconds has passed. It defaults to -1, or no timeout.
 
     """
     assert prefix.endswith(".")
     parser = config.SpecParser(
-        {"url": config.String, "credentials_secret": config.Optional(config.String)}
+        {
+            "url": config.String,
+            "credentials_secret": config.Optional(config.String),
+            "pool_recycle": config.Optional(config.String),
```
spladug: since this is an integer, you should use `config.Optional(config.Integer)`. then the config parser will `int()` it for you and yell during config parsing (rather than later on) if the user has an invalid value in config.

tests/unit/clients/sqlalchemy_tests.py
```diff
@@ -39,6 +39,7 @@ def test_credentials(self, create_engine_mock):
             {
                 "database.url": "postgresql://fizz:buzz@localhost:9000/db",
                 "database.credentials_secret": "secret/sql/account",
+                "database.pool_recycle": 60,
```
spladug: the value on the right should be a string to match how a config dictionary will work.

baseplate/clients/sqlalchemy.py
prthiaga: yes. I tried it but i was getting an error that said something like 
```
int() can't convert non-string with explicit base
```

Then I noticed that growth ran into something similar as well as switched to using String
https://reddit.slack.com/archives/CBDQ2E41X/p1540326246000100

baseplate/clients/sqlalchemy.py
prthiaga: way cleaner!

tests/unit/clients/sqlalchemy_tests.py
prthiaga: I think this is how it will be. 

https://github.snooguts.net/reddit/puppet/blob/919fead76757608bb5e26848a8a8616892a75cec/modules/reddit_thing_server/templates/thing.ini.erb#L31

prthiaga:
The reason this was found during the k8s move was because Thing Service always called `engine_from_config` method in `sqlalchemy` and not baseplate. 

While moving to baseplate, we were forced to use secrets to pass around the database credentials, that meant we had to call baseplate's version of `engine_from_config` that understood `credentials_secret`. But baseplate did not know about `pool_recycle`.

tests/unit/clients/sqlalchemy_tests.py
spladug: No, it'll be a string because `configparser` only deals in strings. The `config.RawConfig` type is `Dict[str, str]`.

baseplate/clients/sqlalchemy.py
spladug: That's because you've got an integer in your test. If you make that a string it'll be like reading from a real config and all will go fine.

baseplate/clients/sqlalchemy.py
```diff
@@ -45,15 +45,24 @@ def engine_from_config(
...
+        {
+            "url": config.String,
+            "credentials_secret": config.Optional(config.String),
+            "pool_recycle": config.Optional(config.Integer),
+        }
     )
     options = parser.parse(prefix[:-1], app_config)
     url = make_url(options.url)
 
+    if options.pool_recycle:
```
spladug: would the user ever want to set this to `0` legitimately? (in which case we should check `is not None` to disambiguate) or is that silly?

prthiaga: probably good to be defensive against bad values. I will update it

prthiaga: I have changed it to check `is not None` just so we do what the user says to do

PR #371, opened 2019-11-22
Add loid and scopes into AuthenticationToken property

suzie-su:
## Description
In authentication service 
2 fields was added into authentication token payload 
1: Scopes(deployed) 
2: Loid(field exists): we are logging more information in this field (PR)

So in order to access those fields easier by other services, add those fields into `property` instead of hiding in the `payload`

@pacejackson @bsimpson63

### DIFF
--- a/baseplate/lib/edge_context.py
+++ b/baseplate/lib/edge_context.py
@@ -86,7 +86,7 @@ class AuthenticationToken:
         raise NotImplementedError
 
     @property
-    def loid_id(self) -> Optional[str]:
+    def loid(self) -> Optional[str]:
         raise NotImplementedError
 
     @property
@@ -119,7 +119,7 @@ class ValidatedAuthenticationToken(AuthenticationToken):
         return set(self.payload.get("scopes") or [])
 
     @property
-    def loid_id(self) -> Optional[str]:
+    def loid(self) -> Optional[str]:
         return (self.payload.get("loid") or {}).get("id")
 
     @property
@@ -149,7 +149,7 @@ class InvalidAuthenticationToken(AuthenticationToken):
         raise NoAuthenticationError
 
     @property
-    def loid_id(self) -> Optional[str]:
+    def loid(self) -> Optional[str]:
         raise NoAuthenticationError
 
     @property
--- a/baseplate/lib/edge_context.py
+++ b/baseplate/lib/edge_context.py
@@ -81,9 +81,21 @@ class AuthenticationToken:
     def oauth_client_type(self) -> Optional[str]:
         raise NotImplementedError
 
+    @property
+    def scopes(self) -> Set[str]:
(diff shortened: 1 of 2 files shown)


baseplate/lib/edge_context.py
```diff
@@ -81,9 +81,21 @@ def oauth_client_id(self) -> Optional[str]:
     def oauth_client_type(self) -> Optional[str]:
         raise NotImplementedError
 
+    @property
+    def scopes(self) -> Set[str]:
+        raise NotImplementedError
+
+    @property
+    def loid_id(self) -> Optional[str]:
```
bsimpson63: why the extra id? doesn't loid mean logged out id?

suzie-su: the loid naming is very confusing sometimes means id sometimes means dict of loid and create_ms. i am open to any suggestino

bsimpson63: how about just `loid`

suzie-su:
Changed the naming

suzie-su:
@bsimpson63  resolve the review and conflict 
💇 
'not able to click the refresh button in reviewer part'

baseplate/lib/edge_context.py
```diff
@@ -92,7 +104,7 @@ def subject(self) -> Optional[str]:
 
     @cached_property
     def user_roles(self) -> Set[str]:  # type: ignore # pylint: disable=W0236
-        return set(self.payload.get("roles", []))
+        return set(self.payload.get("roles") or [])
```
bsimpson63: why is this changing?

baseplate/lib/edge_context.py
```diff
@@ -81,9 +81,21 @@ def oauth_client_id(self) -> Optional[str]:
...
+        raise NotImplementedError
+
+    @property
+    def loid_created_ms(self) -> Optional[int]:
+        raise NotImplementedError
+
 
 class ValidatedAuthenticationToken(AuthenticationToken):
-    def __init__(self, payload: Dict[str, str]):
+    def __init__(self, payload: Dict[str, Any]):
```
bsimpson63: was this wrong before? it looks like `payload[roles]` is a `List[str]`, not `str`

suzie-su: yea i dont know why before set as str

baseplate/lib/edge_context.py
suzie-su: what if pass the role as None then it will get the errors if we set(None)

suzie-su:
💇

baseplate/lib/edge_context.py
bsimpson63: but is that changing, or was that always possible (and not causing problems)?

suzie-su:
💇

PR #375, opened 2019-11-26
Request Timeouts

### LINKED ISSUE #337: Server-side timeout
We should be able to set a maximum length of request. If the greenlet takes longer, it gets killed. This means we don't let greenlets choke up the server for too long.

spladug:
The bulk of this is in the final commit. The rest is sorta cleanup/setup. Rather than writing a bunch here, I'd like to see if the new docs do a decent job of explaining. So without further ado, check this out: https://github.com/spladug/baseplate.py/blob/1e45989d8b28a06a60eb460177194fb66c5f226b/docs/api/baseplate/observers/timeout.rst

Fixes #337

:eyeglasses: @pacejackson @eaceaser

### DIFF
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -45,8 +45,8 @@ class _ContextAwareHandler:
                 # this is an expected exception, as defined in the IDL
                 span.finish()
                 raise
-            except Exception:
-                # the handler crashed!
+            except:  # noqa: E722
+                # the handler crashed (or timed out)!
                 span.finish(exc_info=sys.exc_info())
                 raise
             else:
--- a/baseplate/observers/sentry.py
+++ b/baseplate/observers/sentry.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import sys
 
@@ -15,6 +16,7 @@ from baseplate import RequestContext
 from baseplate import ServerSpanObserver
 from baseplate import Span
 from baseplate.lib import config
+from baseplate.observers.timeout import ServerTimeout
 
 
 if TYPE_CHECKING:
@@ -88,7 +90,7 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
             break
 
     # pylint: disable=maybe-no-member
-    return raven.Client(
+    client = raven.Client(
         dsn=cfg.sentry.dsn,
         site=cfg.sentry.site,
         release=release,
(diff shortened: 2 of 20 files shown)


baseplate/frameworks/thrift/__init__.py
```diff
@@ -45,7 +45,7 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 # this is an expected exception, as defined in the IDL
                 span.finish()
                 raise
-            except Exception:
+            except:  # noqa: E722
                 # the handler crashed!
```
pacejackson:
```suggestion
                # the handler crashed (or timed out)!
```

baseplate/observers/sentry.py
```diff
@@ -158,10 +163,19 @@ def __init__(self, hub: "Hub", client: raven.Client):
...
+        context: Any,
         exc_type: Optional[Type[BaseException]],
         value: Optional[BaseException],
         tb: Optional[TracebackType],
     ) -> None:
         self.raven.captureException((exc_type, value, tb))
-        self.original_print_exception(context, exc_type, value, tb)
+
+        if value and isinstance(value, ServerTimeout):
+            logging.warning(
```
pacejackson: Should this be doing the `logger = logging.getLogger(__name__)` song and dance and using `logger`?

docs/api/baseplate/observers/timeout.rst
```diff
@@ -0,0 +1,62 @@
...
+
+   # optional: defaults to false. if enabled, tracebacks will be
+   # printed to the logs when timeouts occur.
+   server_timeout.debug = true
+
+   # optional: timeout values for specific endpoints. the name
+   # used must match the name of the server span generated.
+   # this overrides the default timeout.
+   # - thrift services: the name of the thrift RPC method
+   # - pyramid services: the name of the route
```
pacejackson: 🔕  It might be useful to call attention to where the name of the route is defined to make it easier for devs to know where to look for it.

https://docs.pylonsproject.org/projects/pyramid/en/1.10-branch/api/config.html#pyramid.config.Configurator.add_route

spladug:
:haircut_man: post-holiday since usernames don't map in harold currently

baseplate/observers/sentry.py
```diff
@@ -155,13 +160,23 @@ class SentryUnhandledErrorReporter:
     def __init__(self, hub: "Hub", client: raven.Client):
         self.original_print_exception = getattr(hub, "print_exception")
         self.raven = client
+        self.logger = logging.getLogger(__name__)
```
pacejackson: 🔕 OOC, why set `self.logger` here rather than doing a global `logger` at the top of the file?

spladug: The hub stuff has to be super defensive about interpreter shutdown where globals can suddenly be `None`d out on you. I'm being a little cargo-culty here but didn't want to take the risk of a crash in weird situations and since this is only instantiated once per app there shouldn't be any cost to doing it this way.

PR #376, opened 2019-11-27
Rework runtime metrics reporting of concurrency

### LINKED ISSUE #353: Rethink max_concurrency and active_requests
These values are based on the number of live _connections_ rather than actual active requests. In many cases, those connections are idle. We need to be limiting and tracking concurrency by actual requests instead.

spladug:
The previous number incorrectly conflated the number of open connections
with the number of in-flight requests. An idle connection has very
little impact on the server, but in-flight requests contend for CPU
time.

Example output:

```
example_server.runtime.reddit.PID2505.open_connections:5|g
example_server.runtime.reddit.PID2505.active_requests:2|g
```

This is a redefinition of the existing `active_requests` gauge. I didn't think it made sense to keep the old one just for backwards compatibility since it was a confusing name. Would it be better to come up with a third name and leave the old one blank instead?

Fixes #353 

### DIFF
--- a/baseplate/server/runtime_monitor.py
+++ b/baseplate/server/runtime_monitor.py
@@ -11,12 +11,18 @@ from typing import Dict
 from typing import List
 from typing import NoReturn
 from typing import Optional
+from typing import Set
 
 import gevent.events
 
 from gevent.pool import Pool
 
+from baseplate import _ExcInfo
 from baseplate import Baseplate
+from baseplate import BaseplateObserver
+from baseplate import RequestContext
+from baseplate import ServerSpan
+from baseplate import ServerSpanObserver
 from baseplate.lib import config
 from baseplate.lib import metrics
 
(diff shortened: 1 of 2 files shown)


baseplate/server/runtime_monitor.py
```diff
@@ -32,12 +37,36 @@ def report(self, batch: metrics.Batch) -> None:
...
+        batch.gauge("active_requests").replace(len(self.live_requests))
+
+
+class _ActiveRequestsServerSpanObserver(ServerSpanObserver):
+    def __init__(self, reporter: _ActiveRequestsObserver, trace_id: int):
+        self.reporter = reporter
+        self.trace_id = trace_id
+
+    def on_start(self) -> None:
+        self.reporter.live_requests.add(self.trace_id)
```
spladug: There's a non-zero (but hopefully very close to zero) chance of this leaving requests counting even though they are long gone. I was considering storing a timestamp and periodically sweeping out very old ones or something similar, but felt that's probably overcomplicated since processes restart occasionally anyway. Thoughts?

spladug:
:haircut_man: post-holiday since usernames don't map in harold currently

baseplate/server/runtime_monitor.py
pacejackson: > processes restart occasionally anyway

That's true in the legacy deployments, but not in k8s.  In k8s, they only restart if they crash or when there's a deploy (although deploy's aren't technically a restart in k8s like they are in legacy deployments).

spladug: indeed, I still think that should change.

pacejackson: > I was considering storing a timestamp and periodically sweeping out very old ones or something similar

How about `self.live_requests: Dict[int, Set[int]] = defaultdict(set)` and key-ing that using `int(time.time()`?  Then you can just `pop` any keys less than or equal to `int(time.time()) - 15` (or however many seconds you want to keep around) at the top of `report` and use something like `len(list(itertools.chain(self.live_requests.values()))` to get the count?

pacejackson:
```python
class _ActiveRequestsObserver(BaseplateObserver, _Reporter):
    def __init__(self) -> None:
        self.live_requests: Dict[int, Set[int]] = defaultdict(set)

    def add(self, trace_id: int) -> int:
        key = int(time.time())
        self.live_requests[key].add(trace_id)
        return key

    def remove(self, timestamp: int, trace_id: int) -> None:
        if timestamp in self.live_requests:
            self.live_requests[timestamp].remove(trace_id)

    def on_server_span_created(self, context: RequestContext, server_span: ServerSpan) -> None:
        observer = _ActiveRequestsServerSpanObserver(self, server_span.trace_id)
        server_span.register(observer)

    def report(self, batch: metrics.Batch) -> None:
        cutoff = int(time.time()) - 15
        for expired in filter(lambda k: k <= cutoff, self.live_requests.keys()):
            self.live_requests.pop(expired, None)
        batch.gauge("active_requests").replace(len(list(itertools.chain(self.live_requests.values()))))
```

spladug: Yeah I was just doing `self.live_requests[trace_id] = time.time()` and then scanning them away but `time.time()` isn't free so I felt iffy on the value.

spladug: That `15` is also a little sketch to me especially with request timeouts coming in at the same time. Ideally we'd want to cull things only after their request timeout, but that gets real complex.

spladug:
:haircut_man: haha sorry for the timing, I was just giving it a shot with a 60 second timeout. seems like a reasonable tradeoff to avoid bad metrics. hopefully the `time.time()` isn't too costly.

spladug:
that look OK?

PR #378, opened 2019-12-03
Fix ServerTimeout exemption on raven<6.1

spladug:
Support for non-string values was only added in 6.1. Some of our
services are still using older versions.

https://github.com/getsentry/raven-python/blob/master/CHANGELOG.md#610-2017-05-25


spladug:
:eyeglasses: @rram @pacejackson

PR #379, opened 2019-12-03
Fix redis client parameter munging and make test stricter

spladug:
This regressed in 1aa93077e24041478b7c38d41c4c65a734bc8e76.


spladug:
:eyeglasses: @pacejackson @rram

PR #380, opened 2019-12-04
Allow custom callback from the secrets fetcher

bradengroom:
Allows applications to defined a callback to run after the secret fetcher lays down its file.

test_config.ini
```
[secret-fetcher]
vault.url = http://127.0.0.1:8200
vault.role = my-server-role

output.path = secrets.json
output.owner = vagrant
output.group = vagrant
output.mode = 0400


secrets =
	secret/data/dbuser,
	secret/data/dbpass

callback = scripts/my-transformer
```

scripts/my-transformer
```
#!/usr/bin/python

import json
import sys

secrets_file = sys.argv[1]

with open(secrets_file) as f:
    secrets = json.load(f)

with open('secrets.properties', 'w') as f:
    f.write("DB_USER=" + secrets['secrets']['secret/data/dbuser']['data']['value'] + "\n")
    f.write("DB_PASS=" + secrets['secrets']['secret/data/dbpass']['data']['value'] + "\n")
```

Running:
```
$ python -m baseplate.sidecars.secrets_fetcher test_config.ini --once
2019-12-05 17:18:09,737:INFO:Running secret fetcher once
2019-12-05 17:18:09,738:INFO:Fetching secrets.
2019-12-05 17:18:09,744:INFO:Secrets fetched.
```

secrets.properties
```
DB_USER=reddit
DB_PASS=password
```

### DIFF
--- a/baseplate/sidecars/secrets_fetcher.py
+++ b/baseplate/sidecars/secrets_fetcher.py
@@ -344,7 +344,7 @@ def trigger_callback(
     callback: Optional[str], secrets_file: str, last_proc: Optional[subprocess.Popen] = None
 ) -> Optional[subprocess.Popen]:
     if callback:
-        if last_proc is not None and last_proc.poll() is None:
+        if last_proc and last_proc.poll() is None:
             logger.info("Previous callback process is still running. Skipping")
         else:
             return subprocess.Popen([callback, secrets_file])
--- a/baseplate/sidecars/secrets_fetcher.py
+++ b/baseplate/sidecars/secrets_fetcher.py
@@ -19,6 +19,8 @@ The file should contain a section looking like:
         secret/two,
         secret/three,
 
+    callback = scripts/my-transformer  # optional
+
 where each secret is a path to look up in Vault. The daemon authenticates with
 Vault as a role using a token obtained from an auth backend designated by `auth_type`.
 
@@ -52,6 +54,11 @@ be refetched.
 The `store` module in this package contains utilities for interacting with this
 file from a running service.
 
+Some applications require a specific format for their secrets. The `callback` option may
+be provided if a different format is needed. This script will be invoked as a subprocess
+with the JSON file as its first and only argument. This allows you to read in the secrets,
+write to a new file in whatever format needed, and restart other services if necessary.
+
 """
 import argparse
 import configparser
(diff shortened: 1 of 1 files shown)


bradengroom:
Untested at the moment. I'll be trying this out with my service tomorrow.

bradengroom:
👓 @spladug

baseplate/sidecars/secrets_fetcher.py
```diff
@@ -332,6 +340,11 @@ def fetch_secrets(
     return soonest_expiration
 
 
+def trigger_callback(callback: Optional[str], secrets_file: str) -> None:
+    if callback:
+        subprocess.Popen([callback, secrets_file])
```
spladug: I've got a few concerns about the external script in failure scenarios, not sure how many of these matter:

* if the script takes a while to run, its output will be interleaved with the fetcher daemon's. I do think we want its logs to go to the same place for debugging, but I'm not really sure of a cleaner way here. Probably just a note and not actionable.
* if the script takes a while to run and secrets are fetched again, we might end up with multiple of the external script piled up. should we only fire the callback if the callback isn't already running? or say that's up to the end user to care about?

bradengroom: Yeah #2 could be bad. `subprocess.Popen` gives us the pid, so we can check if it's still running on the next run and log some message if we skip it. How frequently does the fetcher typically run?

bradengroom:
> if the script takes a while to run and secrets are fetched again, we might end up with multiple of the external script piled up. should we only fire the callback if the callback isn't already running? or say that's up to the end user to care about?

I added some logic to address this

baseplate/sidecars/secrets_fetcher.py
```diff
@@ -332,6 +340,17 @@ def fetch_secrets(
     return soonest_expiration
 
 
+def trigger_callback(
+    callback: Optional[str], secrets_file: str, last_proc: Optional[subprocess.Popen] = None
+) -> Optional[subprocess.Popen]:
+    if callback:
+        if last_proc is not None and last_proc.poll() is None:
```
spladug:
```suggestion
        if last_proc and last_proc.poll() is None:
```

PR #382, opened 2019-12-06
Deprecate the max_concurrency setting for Thrift/WSGI servers

spladug:
:eyeglasses: @pacejackson

### DIFF
--- a/baseplate/server/thrift.py
+++ b/baseplate/server/thrift.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import socket
 
 from typing import Any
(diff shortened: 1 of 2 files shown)


baseplate/server/thrift.py
```diff
@@ -59,13 +63,18 @@ def make_server(server_config: Dict[str, str], listener: socket.socket, app: Any
...
-            "max_concurrency": config.Integer,
+            "max_concurrency": config.Optional(config.Integer),
             "stop_timeout": config.Optional(
                 config.TimespanWithLegacyFallback, default=datetime.timedelta(seconds=10)
             ),
         },
     )
 
+    if cfg.max_concurrency is not None:
+        logger.warning(
```
pacejackson: Should this be using `warn_deprecated`?

spladug: I decided to use logging instead because giving a traceback showing the baseplate code wouldn't be terribly useful.

PR #383, opened 2019-12-09
Circuitbreaking thrift

fizx:
Though circuit-breaking is often referenced as a thing clients can do (particularly if clients are responsible for service discovery), it also makes sense on the server.   

In this case, the simplest form of circuit-breaking that achieves robustness is limiting concurrency.  In the past this was done by limiting the number of connections.  Unfortunately, I see at least two problems with this.  

1. It appears that we are still accepting connections above the connection limit, just not processing work on them.  This will cause stalls, and is the likely reason that people are reluctant to use max_concurrency.
2. The expensive part of the lifecycle isn't handling the connection, it's doing the request work.  Therefore, the limitation is better applied on a per request basis.
3. When the limit is hit, the correct behavior is to queue a small number of requests, then drop requests that will never be serviced inside the SLA of the service.  In practice, you don't need an explicit queue.  Assuming a typical db-bound web service, if max_concurrency is 100, and the db has a connection pool of size 10, then that's a 90 item implicit queue.  Therefore, the correct behavior is to reject requests that exceed max_concurrency.

This will have the effect of replacing stalls with failures.  Ultimately, this is healthier behavior, because it is more explicit over implicit, it allows the upstream service to react faster, and it is quicker and less wasteful of server resources.


baseplate/server/thrift.py
```diff
@@ -7,10 +7,15 @@
 from typing import Tuple
 from typing import Union
 
+import wrapt
```
fizx: Note the new dependency.  This replaces 200 lines of boilerplate.  Our choice whether we carry this to production, but having an ObjectProxy around really helps a lot of framework use cases.  For example, I'm not sure how we currently send statsd for every thrift method, but an ObjectProxy would do it nicely.

tests/integration/test.thrift
```diff
@@ -4,4 +4,5 @@ exception ExpectedException {
 
 service TestService {
     bool example() throws (1: ExpectedException exc),
+    bool sleep() throws (1: ExpectedException exc),
```
fizx: Always useful to have a slow version of the rpc when you're testing concurrency.

tests/integration/thrift_tests.py
```diff
@@ -81,7 +83,7 @@ def on_server_span_created(self, context, server_span):
     # bind a server socket on an available port
     server_bind_endpoint = config.Endpoint("127.0.0.1:0")
     listener = make_listener(server_bind_endpoint)
-    server = make_server({"max_concurrency": "100"}, listener, processor)
+    server = make_server({"max_concurrency": "5"}, listener, processor)
```
fizx: It's a lot of work to have to drive 100 concurrent requests, when a few will do.

requirements-test.txt
```diff
@@ -50,7 +50,7 @@ repoze.lru==0.7
 requests==2.21.0
 six==1.12.0
 SQLAlchemy==1.3.0
-thrift==0.12.1
+thrift>=0.12.1
```
zmsaul: 🔕 I'm a little nervous about dependencies that can upgrade themselves. I generally prefer pinning to the exact version.

fizx: Yeah, it's a bit unclear what versions of thrift we support at the moment.  Mostly because they seem to be different on mac and linux :|

spladug:
Thanks for doing this! I'm going to close this PR out since we've got the design doc and the WIP PR for implementation. As for error messaging, @bradengroom is about to drop a design doc for some standard error structs that I think would solve that problem.

spladug:
(Let's catch up on the current status of all this soon)

PR #387, opened 2019-12-30
Add go namespace to baseplate.thrift

fishy:


fishy:
👓 @spladug @pacejackson

PR #392, opened 2020-01-08
Performance fixes

fishy:
Return error early at 2 places to improve performances:

1. In edge context, when the error is jwt.ExpiredSignatureError there's
   no point of trying the next version of public key.

2. In message signature verification, if the expiration timestamp is
   already in the past there's no point of calculating the digest and do
   the comparison.

### DIFF
--- a/baseplate/lib/crypto.py
+++ b/baseplate/lib/crypto.py
@@ -155,15 +155,15 @@ def validate_signature(secret: VersionedSecret, message: str, signature: bytes) 
     except (struct.error, KeyError, binascii.Error, TypeError, ValueError):
         raise UnreadableSignatureError
 
+    if time.time() > expiration:
+        raise ExpiredSignatureError(expiration)
+
     for secret_value in secret.all_versions:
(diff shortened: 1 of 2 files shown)


fishy:
👓 @spladug @pacejackson

fishy:
the `pass` part is in a for loop. In best/worst case scenario changing that to return saves you 2 `jwt.decode` calls. more if we could have more than 3 versions in versioned secret.

spladug:
Ok, yeah, that's fair. I still think it's valuable to include proof of actual performance benefit when claiming one.

fishy:
I just noticed them as low hanging fruits while implementing the go version. I see no scenario these changes will make things slower (or incorrect, except in signing part the caller can no longer assume `ExpiredSignatureError` means the signature was correct otherwise), as `_compute_digest` and `jwt.decode` are both cpu heavy calls.

spladug:
Alright, sounds good. Thanks for the patch.

PR #393, opened 2020-01-16
Make the redis client compatible with versions 3.0 of the redis library

bradengroom:


spladug:
Can you update the version pin in https://github.com/reddit/baseplate.py/blob/master/setup.py#L29 please?

baseplate/clients/redis.py
```diff
@@ -3,7 +3,13 @@
...
 from typing import Optional
 
-import redis.client
+import redis
+
+# redis.client.StrictPipeline was renamed to redis.client.Pipeline in version 3.0
+try:
+    from redis.client import StrictPipeline as Pipeline
+except ImportError:
+    from redis.client import Pipeline  # type: ignore
```
bradengroom: I feel bad adding this `type: ignore`, but I was unable to find a better solution. Open issue in mypy:
https://github.com/python/mypy/issues/1153

spladug: seems reasonable, can't please the linter all the time.

spladug:
For the record, I looked at [the breaking changes in 3.0](https://github.com/andymccurdy/redis-py/blob/master/CHANGES#L167-L199) and other than the pipeline rename, it looks like the rest are pretty application-specific so should be fine from a baseplate point of view. Applications will know when they're upgrading the library, since it's in their control, and can pay attention to the other breaking changes for their own purposes.

PR #397, opened 2020-01-28
Add avro producer and consumer to baseplate

manishapme:
This moves the avro consumer and producer code into baseplate service for access to all services. Matches code live in Thing Service now.

### DIFF
--- a/baseplate/clients/kafka.py
+++ b/baseplate/clients/kafka.py
@@ -1,18 +1,21 @@
 import logging
 import time
 
+from concurrent.futures import wait
 from typing import Any
 from typing import Dict
+from typing import List
+from typing import Optional
 
 import confluent_kafka
 
 from confluent_kafka.avro import AvroProducer
 from confluent_kafka.avro import Producer
+from gevent.threadpool import ThreadPoolExecutor
 
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
-
 
 logger = logging.getLogger(__name__)
 
@@ -21,7 +24,14 @@ CHECK_INTERVAL_SECONDS = 0.001
 
 
 class AvroProducerContextFactory(ContextFactory):
-    def __init__(self, bootstrap_servers: str, schema_registry: str, acks: int):
+    def __init__(
+        self,
+        bootstrap_servers: str,
+        schema_registry: str,
+        acks: int,
+        thread_pool_max_size: int = 10,
+        prefetch_schema_ids: Optional[List[int]] = None,
+    ):
         self.avro_producer = AvroProducer(
             {
                 "bootstrap.servers": bootstrap_servers,
@@ -30,10 +40,55 @@ class AvroProducerContextFactory(ContextFactory):
                 # force sending 1 message at a time--don't wait for larger batches
                 "queue.buffering.max.ms": 0,
                 "max.in.flight": 1,
+                # hardcode the defaults here because we reference them with our manual refreshing
+                "topic.metadata.refresh.interval.ms": 300000,
+                "metadata.max.age.ms": 900000,
             }
         )
+        self.threadpool = ThreadPoolExecutor(thread_pool_max_size)
+        self.initialize(self.threadpool, prefetch_schema_ids or [])
+        self.last_initialized_at = time.time()
+
+    def initialize(
+        self, threadpool: ThreadPoolExecutor, prefetch_schema_ids: List[int], block: bool = True
+    ) -> None:
+        """Do some setup tasks to make sure the producer is ready to use."""
+        futures = []
+
+        # make a metadata request, otherwise we'll have to do this on our
+        # first produce which can add up to 1s of extra latency.
+        # list_topics() is a blocking call, so we have to use a separate
+        # thread to avoid blocking the gevent event loop. We can't easily
+        # make confluent_kafka gevent compatible here because it's wrapping
+        # the C library librdkafka.
+        futures.append(threadpool.submit(self.avro_producer.list_topics))
+
+        # fetch and cache all the schemas we plan to use
+        if prefetch_schema_ids:
+            for schema_id in prefetch_schema_ids:
+                futures.append(
+                    threadpool.submit(
+                        self.avro_producer._serializer.registry_client.get_by_id, schema_id
+                    )
+                )
+
+        if block:
+            _, not_done = wait(futures, timeout=1)
+            if not_done:
+                logger.info("AvroProducer timeout while initializing")
 
     def make_object_for_context(self, name: str, span: Span) -> "KafkaAvroProducer":
+        # When initializing the producer we request the metadata by calling list_topics()
+        # Metadata is automatically refreshed every 5 minutes (topic.metadata.refresh.interval.ms)
+        # and expires from cache after 15 minutes (metadata.max.age.ms). We should attempt to
+        # refresh manually before it expires (after 10 minutes).
+        METADATA_REFRESH_INTERVAL = 600
+
+        now = time.time()
+        if now - self.last_initialized_at > METADATA_REFRESH_INTERVAL:
+            self.last_initialized_at = now
+            self.initialize(threadpool=self.threadpool, prefetch_schema_ids=[], block=False)
+
         return KafkaAvroProducer(name, span, self.avro_producer)
 
 
@@ -42,7 +97,6 @@ class KafkaAvroProducer:
         self.name = name
         self.span = span
         self.avro_producer = avro_producer
-        self.initialized = False
 
     def produce(self, topic: str, schema_id: int, value: Dict[str, Any]) -> None:
         """Encode `value` using the `schema_id` and produce the message to Kafka.
@@ -58,33 +112,10 @@ class KafkaAvroProducer:
         if not topic.endswith(TOPIC_SUFFIX):
             raise ValueError("Avro Producers must publish to topics ending with '_avro'")
 
-        if not self.initialized:
-            # make a metadata request, otherwise we'll have to do this on our
-            # first produce which can add up to 1s of extra latency
-            with self.span.make_child(f"{self.name}.avro_producer.list_topics"):
-                for _ in range(3):
-                    try:
-                        # list_topics() is a blocking call, so we have to use
-                        # a very short timeout and retry after a short sleep.
-                        self.avro_producer.list_topics(timeout=0.001)
-                        break
-                    except confluent_kafka.KafkaException:
-                        time.sleep(0.001)
-                else:
-                    logger.info("list_topics() failed after 3 attempts")
-
-            self.initialized = True
-
         serializer = self.avro_producer._serializer
 
-        encode_trace_name = f"{self.name}.serializer.encode_record_with_schema_id"
-        encode_span = self.span.make_child(encode_trace_name)
-
-        with encode_span:
+        with self.span.make_child(f"{self.name}.serializer.encode_record_with_schema_id"):
             encoded = serializer.encode_record_with_schema_id(schema_id, value)
-
-        producer_trace_name = f"{self.name}.avro_producer.produce"
-        producer_span = self.span.make_child(producer_trace_name)
 
         # Producer.produce() is asynchronous so we have to jump through some
         # hoops to get the result here at the call site.
@@ -95,7 +126,7 @@ class KafkaAvroProducer:
             delivery_result["message"] = msg
             delivery_result["complete"] = True
 
-        with producer_span:
+        with self.span.make_child(f"{self.name}.avro_producer.produce"):
             # call the base class Producer.produce() method so we can do our own
             # interactions with schema registry to avoid a bug where the schema
             # can be re-registered.
@@ -141,6 +172,19 @@ def avro_producer_from_config(
     :param app_config: The raw configuration information.
     :param prefix: The name of the kafka cluster to publish messages to.
 
+    The keys useful to :py:func:`avro_producer_from_config` should be prefixed, e.g.
+    ``kafka_cluster_name.bootstrap_servers`` etc. The ``prefix`` argument specifies the
+    prefix used to filter keys. It should map to the name of the kafka cluster
+    that will be published too.
+
+    Supported keys:
+
+    * ``bootstrap_servers`` (required): comma delimited list of kafka brokers to
+      try connecting to, including the port, kafka-01.data.net:9092.
+    * ``schema_registry``: url to connect to the schema-registry on, http://cp-schema-registry.data.net:80.
+    * ``threadpool_size`` (optional): metadata is requested from kafka in it's own thread, option to specify the size of that threadpool.
+    * ``prefetch_schema_ids`` (optional): to optimize initialization pass in list of int ids used by producers.
+
     """
     assert prefix.endswith(".")
 
@@ -149,10 +193,16 @@ def avro_producer_from_config(
             "bootstrap_servers": config.String,
             "schema_registry": config.String,
             "acks": config.Integer,
+            "threadpool_size": config.Optional(config.Integer, default=10),
+            "prefetch_schema_ids": config.Optional(config.TupleOf(config.Integer), default=()),
         }
     )
     options = parser.parse(prefix[:-1], app_config)
 
     return AvroProducerContextFactory(
-        options.bootstrap_servers, options.schema_registry, options.acks
+        options.bootstrap_servers,
+        options.schema_registry,
+        options.acks,
+        options.threadpool_size,
+        options.prefetch_schema_ids,
     )
--- a/baseplate/clients/kafka.py
+++ b/baseplate/clients/kafka.py
@@ -1,47 +1,48 @@
 import logging
+import time
 
 from typing import Any
 from typing import Dict
-from typing import Optional
+
+import confluent_kafka
 
 from confluent_kafka.avro import AvroProducer
+from confluent_kafka.avro import Producer
 
-from baseplate import _ExcInfo
 from baseplate import Span
-from baseplate import SpanObserver
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class AvroProducerSpanObserver(SpanObserver):
+    """Automatically flush to kafka at the end of each request."""
+
+    def __init__(self, kafka_avro_producer: KafkaAvroProducer):
+        self.avro_producer = kafka_avro_producer.avro_producer
+
+    def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
+        self.avro_producer.flush()
```
pacejackson: Why are we flushing the producer at the end of the request rather than at the call time?  That seems like something we wouldn't want to do since you wouldn't be able to rely on on knowing that your message was actually published or not when making the function call.

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+        encode_trace_name = "{}.{}".format(self.name, "serializer.encode_record_with_schema_id")
+        encode_span = self.span.make_child(encode_trace_name)
+
+        with encode_span:
+            buffer = serializer.encode_record_with_schema_id(schema_id, value)
+
+        producer_trace_name = "{}.{}".format(self.name, "avro_producer.produce")
+        producer_span = self.span.make_child(producer_trace_name)
+
+        producer = super(AvroProducer, self.avro_producer)  # pylint: disable=bad-super-call;
```
pacejackson: Why do we have this `super` call here, some quirk of the kafka library?  Could you add a comment about it?

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -235,17 +298,22 @@ def new(
...
         """
 
-        service_name, _, group_name = group_id.partition(".")
-        assert service_name and group_name, "group_id must start with 'SERVICENAME.'"
-        assert name == f"kafka_consumer.{group_name}"
+        validate_group_id(name, group_id)
+        consumer = make_kafka_consumer(bootstrap_servers, group_id, topics, schema_registry)
 
-        consumer = cls.make_kafka_consumer(bootstrap_servers, group_id, topics)
+        if message_unpack_fn is None:
```
pacejackson: Do you want to raise an exception if both `message_unpack_fn` and `schema_registry` are supplied?

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -167,6 +168,69 @@ def handle(self, message: confluent_kafka.Message) -> None:
...
+    assert service_name and group_name, "group_id must start with 'SERVICENAME.'"
+    assert name == f"kafka_consumer.{group_name}"
+
+
+def make_kafka_consumer(
+    bootstrap_servers: str,
+    group_id: str,
+    topics: Sequence[str],
+    schema_registry: Optional[str] = None,
+    consumer_config: Dict[str, Any] = {},
```
bsimpson63: probably don't want a mutable default argument https://docs.python-guide.org/writing/gotchas/

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -235,17 +300,22 @@ def new(
         :param kafka_consume_batch_size: The number of messages the `KafkaConsumerWorker`
             reads from Kafka in each batch. Defaults to 1.
         :param message_unpack_fn: A function that takes one argument, the `bytes` message body
-            and returns the message in the format the handler expects. Defaults to `json.loads`.
+            and returns the message in the format the handler expects. If no input is provided,
+            a function is automatically picked one based on whether a schema_registry was provided or not.
```
bsimpson63: If no value is provided, we will use avro decode_message when using schema registry, and json.loads otherwise

?

baseplate/frameworks/queue_consumer/kafka.py
bsimpson63: this confused me too, but we could have a case where we actually intended to specify both, like if we're using avro and set `message_unpack_fn=lambda m: consumer._serializer.decode_message(m).bar`

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -167,6 +169,69 @@ def handle(self, message: confluent_kafka.Message) -> None:
             raise
 
 
+def validate_group_id(name: str, group_id: str) -> None:
```
bsimpson63: can you inline this? it's only called one place and it'd make it easier to read

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
+import logging
```
bsimpson63: it'd mirror the existing structure if this were in baseplate/clients/kafka.py (see https://github.com/reddit/baseplate.py/blob/master/baseplate/clients/kombu.py)

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+        :param topic: topic name
+        :param schema_id: schema id associated with the schema that `value` conforms to. It will
+                          be used to look up the schema from Schema Registry. This prevents callers
+                          from using unregistered schemas
+       :param value: object to serialize
+
+        """
+        serializer = self.avro_producer._serializer
+
+        encode_trace_name = "{}.{}".format(self.name, "serializer.encode_record_with_schema_id")
```
bsimpson63: fstring?

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+        encode_span = self.span.make_child(encode_trace_name)
+
+        with encode_span:
```
bsimpson63: why not make the span all on this line?

```
with self.span.make_child(f"{self.name}.serializer.encode_record_with_schema_id"):
```

baseplate/frameworks/queue_producer/avro_kafka.py
bsimpson63: can we replace it with `super(AvroProducer, self.avro_producer).produce(topic=topic, value=buffer)` and comment that we're doing this because we're manually encoding avro because the library might try to re-register the schema for us

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+            buffer = serializer.encode_record_with_schema_id(schema_id, value)
+
+        producer_trace_name = "{}.{}".format(self.name, "avro_producer.produce")
+        producer_span = self.span.make_child(producer_trace_name)
+
+        producer = super(AvroProducer, self.avro_producer)  # pylint: disable=bad-super-call;
+        with producer_span:
```
bsimpson63: can we make the span here inline?

baseplate/frameworks/queue_producer/avro_kafka.py
bsimpson63: I think the thinking is that we can let the background publishing do its thing and not block, but we only want to be sure the message was produced by the time we are exiting

pacejackson: Right, but why do that?  We don't do that for any other IO clients like this and it means that the caller now no longer knows for sure that its message was actually published to Kafka.  If the app crashes or there's a network failure before this happens, we might have told the caller that their request succeeded but their message did not actually make it from Kafka, so they don't even get the chance to retry and that message has been lost.

baseplate/frameworks/queue_consumer/kafka.py
manishapme: oh, yeah, i've read that and forgot. i was matching changes in thing, but I can have them be much more discrete. will move this back into the class and then don't need to pass in.

baseplate/frameworks/queue_producer/avro_kafka.py
bsimpson63: So do you think we should depend on the client calling `flush()`? or build that into our wrapper around the real avro client?

baseplate/frameworks/queue_producer/avro_kafka.py
manishapme: it seems there was conversation in the Thing PR about having to do it this way to avoid linting errors. https://github.snooguts.net/reddit/reddit-service-thing/pull/796/files#r247408

but I will try it. 
The difference is in being able to call the `encode_record_with_schema_id` instead of `encode_record_with_schema` which causes schemas to be registered.

baseplate/frameworks/queue_producer/avro_kafka.py
```diff
@@ -0,0 +1,97 @@
...
+class KafkaAvroProducer:
+    def __init__(self, name: str, span: Span, app_config: Dict[str, Any]):
+        self.name = name
+        self.span = span
+        self.avro_producer = AvroProducer(app_config)
+
+    def produce(self, topic: str, schema_id: int, value: Dict[str, Any]) -> None:
+        """Encode `value` using the `schema_id` and produce the message to Kafka.
+
+        :param topic: topic name
```
bsimpson63: can we add an assert to check that the topic ends with `"_avro"`

baseplate/frameworks/queue_producer/avro_kafka.py
prthiaga: The reason we call the parent class' `produce` is because `AvroProducer`s `produce` method tries to register the schema under `<subject_name>-value` automatically. I think a fix was just pushed in to not have it auto register schemas.

So, we decided to encode the message to Avro ourselves by calling https://github.com/confluentinc/confluent-kafka-python/blob/e575850e4e5d21544d590b76afc7d4f9876a04c9/confluent_kafka/avro/serializer/message_serializer.py#L120 instead that does not auto-register schemas.

The reason it was split into two lines was because I ran into a error cycle between pylint and black that both wanted the code to be a certain way.

baseplate/frameworks/queue_producer/avro_kafka.py
pacejackson: Even if we go with this offline publishing approach, I don't think we can use `flush` since it'll block gevent:

https://github.com/confluentinc/confluent-kafka-python/issues/612

pacejackson: I think we should build calling `poll(timeout=0)` into the client.  You can pass in a callback to the produce function that'll get called after the message is delivered, one potential option would be something like this:

```
def produce(message):
    q = queue.Queue(maxsize=1)
    func on_delivery(message, err):
        q.add((message, err))
    kafka.publish(message, on_delivery=on_delivery)
    message, err = flush(q)
    return message, err

def _flush(q):
    kafka.poll(timeout=0)
    while q.empty():
        kafka.poll(timeout=0)
    return q.get(block=False)
```

bsimpson63: why use a queue instead of calling `poll` or `flush` in a loop with 0 timeout and small sleep and then waiting until the number of pending messages is 0 https://docs.confluent.io/current/clients/confluent-kafka-python/index.html#confluent_kafka.Producer.flush?


something like:

```
def produce(...):
    kafka.produce(topic, value)

def on_finish(...):
    pending_messages = self.avro_producer.flush(timeout=0)
    while pending_messages:
            # we can't use timeouts on the kafka consumer methods
            # because they call out to C code that ends up blocking
            # the event loop. gevent will successfully yield control
            # on this sleep call though.
            # see https://tech.wayfair.com/2018/07/blocking-io-in-gunicorn-gevent-workers/
            logger.debug("waiting 0.001s for pending messages to flush")
            time.sleep(0.001)
            pending_messages = self.avro_producer.flush(timeout=0)
```

pacejackson: I think there's 2 things, one is that we probably don't want the flush logic to live in `on_finish`, I still think we want to flush within the `produce` call so the client using the `produce` method can know that their message has been delivered.

Now, about how we accomplish that, the queue _might_ be overkill but it does let us know that our message was delivered, regardless if any other pending messages were delivered.  If we want to just spin until `pending` is 0, we could be waiting an indeterminate amount of time for other messages to publish.

pacejackson: Actually, I would think we would still want the queue (or something like it), since that seems to be the only way to actually get the message from the `on_delivery` callback into the main call stack to return up to the caller of `publish` (or raise `err` possibly?).

bsimpson63: Ok how about having the flush in `on_finish` but also providing a "safer" producer message that ensures the message is actually sent?

bsimpson63: OK how about something like https://gist.github.snooguts.net/bsimpson/7a63c0da25075781d873a6843460a853 using thread locals to get the result back to the caller?

manishapme: why is it important to be in the `on_finish` call instead of at the time `produce` is called? if you keep the `flush` in `on_finish` aren't you enabling developers to write unexpectedly slow/blocking producers?

pacejackson: Thread local approach looks good!

baseplate/frameworks/queue_producer/avro_kafka.py
bsimpson63: ok I think the way python wants us to do this is

```
Producer.produce(self.avro_producer, topic=topic, value=value)
```

baseplate/frameworks/queue_producer/avro_kafka.py
bsimpson63: @manishapme yeah I think for now we can drop the `on_finish` and wait for the response in our `produce()` method

manishapme:
💇 used the new threading.local approach that Brian wrote into the thing PR. addressed other comments around tests and formatting. Verified producer code in onevm.

:shakes_fist linting errors.

baseplate/clients/kafka.py
```diff
@@ -0,0 +1,138 @@
...
+        with encode_span:
+            encoded = serializer.encode_record_with_schema_id(schema_id, value)
+
+        producer_trace_name = "{}.{}".format(self.name, "avro_producer.produce")
+        producer_span = self.span.make_child(producer_trace_name)
+
+        # use a thread local to get delivery result from the callback.
+        # Producer.produce() is asynchronous so we have to jump through some
+        # hoops to get the result here at the call site.
+        delivery_result = threading.local()
```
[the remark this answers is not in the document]
manishapme: updated to include latest changes deployed to Thing Service.

manishapme:
💇 Addressed comments, added tests and matched current state of Thing Service code plus this PR https://github.snooguts.net/reddit/reddit-service-thing/pull/811.

manishapme:
💇 there are outstanding issues to prevent merging but this PR is ready for technical review, and matches code running in Thing. Modifications improved one time latency on startup. Other comments were addressed.

baseplate/clients/kafka.py
```diff
@@ -0,0 +1,208 @@
...
+                "queue.buffering.max.ms": 0,
+                "max.in.flight": 1,
+                # hardcode the defaults here because we reference them with our manual refreshing
+                "topic.metadata.refresh.interval.ms": 300000,
+                "metadata.max.age.ms": 900000,
+            }
+        )
+        self.threadpool = ThreadPoolExecutor(thread_pool_max_size)
+        self.initialize(self.threadpool, prefetch_schema_ids or [])
+        self.last_initialized_at = time.time()
```
pacejackson: Can you set this in `initialize` instead?

baseplate/clients/kafka.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def initialize(
+        self, threadpool: ThreadPoolExecutor, prefetch_schema_ids: List[int], block: bool = True
```
pacejackson: I don't think we need to pass in the threadpool since it's already set on `self`, you can just access it there.

baseplate/clients/kafka.py
```diff
@@ -0,0 +1,208 @@
...
+            _, not_done = wait(futures, timeout=1)
+            if not_done:
+                logger.info("AvroProducer timeout while initializing")
+
+    def make_object_for_context(self, name: str, span: Span) -> "KafkaAvroProducer":
+        # When initializing the producer we request the metadata by calling list_topics()
+        # Metadata is automatically refreshed every 5 minutes (topic.metadata.refresh.interval.ms)
+        # and expires from cache after 15 minutes (metadata.max.age.ms). We should attempt to
+        # refresh manually before it expires (after 10 minutes).
+        METADATA_REFRESH_INTERVAL = 600
```
pacejackson: Can you make this either an object or module constant and also use:

```python
timedelta(minutes=10).total_seconds()
```

to initialize it so it's clear what the interval is and what units it is in.

baseplate/clients/kafka.py
```diff
@@ -1,27 +1,37 @@
...
 
 class AvroProducerContextFactory(ContextFactory):
-    def __init__(self, bootstrap_servers: str, schema_registry: str, acks: int):
+    def __init__(
+        self,
+        bootstrap_servers: str,
+        schema_registry: str,
+        acks: int,
+        thread_pool_max_size: int = 10,
+        prefetch_schema_ids: Optional[List[int]] = None,
```
pacejackson: Nit: maybe type this as `Sequence` rather than `List`?  Then you can pass in a list or a tuple, just anything that you can iterate over.

spladug:
Hi, my understanding is this project will not be continuing forward right now, so I'm going to close this PR. Feel free to reopen should this come back up later.

spladug:
Since we're not moving forward with this project right now, I'm going to close this out.

PR #400, opened 2020-02-11
add timer sampling option to client observers

diffyqgirl:
Default option will be sampling rate of 1, so that the default behaviour doesn't change.

### DIFF
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -200,9 +200,7 @@ class Batch(BaseClient):
     """
 
     # pylint: disable=super-init-not-called
-    def __init__(
-        self, transport: Transport, namespace: bytes, timer_sampling_rate: float = 1.0
-    ):
+    def __init__(self, transport: Transport, namespace: bytes, timer_sampling_rate: float = 1.0):
         self.transport = BufferedTransport(transport)
         self.namespace = namespace
         self.counters: Dict[bytes, BatchCounter] = {}
@@ -467,7 +465,9 @@ class Gauge:
         self.transport.send(serialized)
 
 
-def make_client(namespace: str, endpoint: config.EndpointConfiguration, timer_sampling_rate: float = 1.0) -> Client:
+def make_client(
+    namespace: str, endpoint: config.EndpointConfiguration, timer_sampling_rate: float = 1.0
+) -> Client:
     """Return a configured client.
 
     :param namespace: The root key to prefix all metrics with.
@@ -506,13 +506,12 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
     cfg = config.parse_config(
         raw_config,
         {
-            "metrics": {
-                "namespace": config.String,
-                "endpoint": config.Optional(config.Endpoint),
-            },
+            "metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)},
             "metrics_observer": {"timer_sampling_rate": config.Optional(config.Float)},
         },
     )
 
     # pylint: disable=maybe-no-member
-    return make_client(cfg.metrics.namespace, cfg.metrics.endpoint, cfg.metrics_observer.timer_sampling_rate)
+    return make_client(
+        cfg.metrics.namespace, cfg.metrics.endpoint, cfg.metrics_observer.timer_sampling_rate
+    )
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -501,12 +501,8 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
     """
     cfg = config.parse_config(
         raw_config,
-        {
-            "metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)},
-        }
+        {"metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)}},
     )
 
     # pylint: disable=maybe-no-member
-    return make_client(
-        cfg.metrics.namespace, cfg.metrics.endpoint
-    )
+    return make_client(cfg.metrics.namespace, cfg.metrics.endpoint)
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -507,7 +507,7 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
         raw_config,
         {
             "metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)},
-            "metrics_observer": {"timer_sampling_rate": config.Optional(config.Float)},
+            "metrics_observer": {"timer_sampling_rate": config.Optional(config.Float, default=1.0)},
         },
     )
 
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -130,10 +130,9 @@ class BufferedTransport(Transport):
 
 
 class BaseClient:
-    def __init__(self, transport: Transport, namespace: str, timer_sampling_rate: float = 1.0):
+    def __init__(self, transport: Transport, namespace: str):
         self.transport = transport
         self.namespace = namespace.encode("ascii")
-        self.timer_sampling_rate = timer_sampling_rate
 
     def timer(self, name: str) -> "Timer":
         """Return a Timer with the given name.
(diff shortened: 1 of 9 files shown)


baseplate/lib/metrics.py
```diff
@@ -501,7 +503,7 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
     """
     cfg = config.parse_config(
         raw_config,
-        {"metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)}},
+        {"metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint), "histogram_sampling_rate": config.Optional(config.Float)}},
```
diffyqgirl: Is it better to get the `histogram_sampling_rate` config option here in the same place where we get the other config options, then pass it through several constructors, or to pull it out of the raw_config later on when it's actually needed?

baseplate/lib/metrics.py
```diff
@@ -501,7 +505,13 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
     """
     cfg = config.parse_config(
         raw_config,
-        {"metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)}},
+        {
+            "metrics": {
+                "namespace": config.String,
+                "endpoint": config.Optional(config.Endpoint),
+                "histogram_sampling_rate": config.Optional(config.Float),
```
diffyqgirl: Is it better to pull the `histogram_sampling_rate` out of the config here in the same place where we get a bunch of other config data, or later on where we wouldn't have to pass it through so many constructors?

baseplate/observers/metrics.py
```diff
@@ -93,15 +94,18 @@ def __init__(self, batch: metrics.Batch, span: Span):
         self.batch = batch
         self.base_name = f"clients.{span.name}"
         self.timer = batch.timer(self.base_name)
+        self.sample_histogram = random() < batch.histogram_sampling_rate
```
m104: nit: only run the `random()` check if `histogram_sampling_rate` is less than 1

fishy:
>Why are we only sampling the timers and not the counters?

We rarely need to sample counters. Counters are always aggregated locally first, and only send the aggregated result to statsd collector (e.g. telegraf). Timers/Histograms cannot be aggregated and need to be send to statsd collector one-by-one (it could and should still be batched, but the packet size of the batch would be linear to the number of histograms/timers reported in the code). This is why sampling histograms/timers is much more useful than sampling counters, and the configs should be separated.

Also in statsd the only difference between histograms and timers is the "unit" (`|h` vs. `|ms`), so they should be treated equally/interchangeably in most cases.

spladug:
> Also in statsd the only difference between histograms and timers is the "unit" (|h vs. |ms), so they should be treated equally/interchangeably in most cases.

My point is that the terminology used in the code is confusing. It should be consistent.

diffyqgirl:
> > Also in statsd the only difference between histograms and timers is the "unit" (|h vs. |ms), so they should be treated equally/interchangeably in most cases.
> 
> My point is that the terminology used in the code is confusing. It should be consistent.

Agreed, I will change the name

diffyqgirl:
> * It does feel rather weird to be putting the config for this in the metrics client and then actually using it in the metrics observer. Maybe we should reorganize how the config parsing works there so that the metrics observer can get its own config?

Hmm so I tried to just have the observer configure itself, but the problem is that the observer doesn't know about `raw_config` and therefore it doesn't have access to the configuration data, except through the metrics client or through manually registering an observer. Ideally I'd like this to be compatible with `register_observers`. I think what makes the most sense, unless you know a good way of getting the `raw_config` in the observer class, is to load the configurations in the metrics client but rename them to make it clearer where they are used.

> I'd really like to try any changes like this out in a service before we upstream it here into the library please.

I agree, I opened the PR because I wanted feedback. Once it looks good what is the correct way to try it without merging it? Do I merge it into my own fork and make a wheel?

baseplate/lib/metrics.py
```diff
@@ -501,8 +505,13 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
     """
     cfg = config.parse_config(
         raw_config,
-        {"metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)}},
+        {
+            "metrics": {"namespace": config.String, "endpoint": config.Optional(config.Endpoint)},
```
spladug: Rather than forcing the metrics client to know about span sampling, which is only a concern of the observer, I think it'd make more sense to have the observer do its own config parsing. That could look like adding a new classmethod to `MetricsBaseplateObserver` like so:

```python
class MetricsBaseplateObserver(BaseplateObserver):
    @classmethod
    def from_config_and_client(cls, raw_config: config.RawConfig, client: metrics.Client) -> "MetricsBaseplateObserver":
        cfg = config.parse_config(raw_config, {
            "metrics_observer": {
               "sample_rate": config.Optional(config.Float, default=1.0),
            }
        })

       return cls(client, sample_rate=cfg.metrics_observer.sample_rate)

    def __init__(self, client: metrics.Client, sample_rate: float):
        self.client = client
        self.sample_rate = sample_rate
```

Then you can update `Baseplate.configure_metrics` to use that factory function rather than directly calling the constructor:

```python
def configure_metrics(self, metrics_client: metrics.Client) -> None:
    from baseplate.observers.metrics import MetricsBaseplateObserver

    self._metrics_client = metrics_client
    observer = MetricsBaseplateObserver.from_client_and_config(metrics_client, self._app_config or {})
    self.register(observer)
```

Since that relies on `self._app_config` being present, this setting will only be available to clients using the modern `Baseplate(app_config)` setup, which should probably be documented.

baseplate/observers/metrics.py
```diff
@@ -44,15 +45,18 @@ def __init__(self, batch: metrics.Batch, server_span: Span):
         self.batch = batch
         self.base_name = "server." + server_span.name
         self.timer = batch.timer(self.base_name)
+        self.sample_timer = batch.timer_sampling_rate == 1.0 or random() < batch.timer_sampling_rate
```
spladug: I'm still a bit concerned about doing the randomness multiple times during the same request. This could lead to some weird counting vs. if we sample at the whole request level.

Since we can scale the counters with the `sample_rate` parameter, what do you think of the proposal to do this sampling in the root `BaseplateObserver` and not even registering the sub-observers when sampling's off? Then we get consistent whole-request level sampling and do less work in the case of unsampled requests.

m104: Agreed on this. Let's sample the whole request.

baseplate/observers/metrics.py
```diff
@@ -29,82 +31,105 @@ class MetricsBaseplateObserver(BaseplateObserver):
...
+
+    @classmethod
+    def from_config_and_client(
+        cls, raw_config: Optional[config.RawConfig], client: metrics.Client
+    ) -> "MetricsBaseplateObserver":
+        sample_rate = 1.0
+        if raw_config:
+            cfg = config.parse_config(
+                raw_config,
+                {"metrics_observer": {"sample_rate": config.Optional(config.Float, default=1.0)}},
```
spladug: it'd be clearer for this to be `config.Percent` so the config looks like `metrics_observer.sample_rate = 37%` which is more self-describing than wondering what the range of the value is. the output will still be exactly a float between 0 and 1.0 like you've got here.

spladug:
oh! sorry, one more request: it'd be good to document the new configurable here https://github.com/reddit/baseplate.py/blob/master/docs/api/baseplate/observers/statsd.rst

diffyqgirl:
💇 @spladug

baseplate/observers/metrics.py
spladug: looks like this still needs to be done?

docs/api/baseplate/observers/statsd.rst
```diff
@@ -30,6 +30,10 @@ the StatsD metrics observer.
    # if not specified, metrics will only be emitted to debug logs.
    metrics.endpoint = statsd.local:8125
 
+   # optional: the fraction of statsd metrics to sample
+   # if not specified, it will default to 1.0 (all metrics sent)
```
spladug: would be good to add a note that they must be passing config to the `Baseplate` constructor for this to work

PR #402, opened 2020-02-12
Convert boolean span tags to strings

pacejackson:
Our current system for storing trace data only supports string values for binary annotations so we want to also encode boolean values as lower case strings.

### DIFF
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -258,7 +258,7 @@ class TraceSpanObserver(SpanObserver):
         endpoint_info = self._endpoint_info()
 
         # Annotation values must be str type.
-        if isinstance(bool):
+        if isinstance(annotation_value, bool):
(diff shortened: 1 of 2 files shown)


baseplate/observers/tracing.py
```diff
@@ -257,9 +257,9 @@ def _create_binary_annotation(
         """
         endpoint_info = self._endpoint_info()
 
-        # Annotation values must be bool or str type.
-        if not isinstance(annotation_value, (bool, str)):
-            annotation_value = str(annotation_value)
+        # Annotation values must be str type.
+        if not isinstance(annotation_value, str):
+            annotation_value = str(annotation_value).lower()
```
spladug: do we want things that weren't originally bools to get lowercased too?

pacejackson: I'm not sure it matters too much but better safe than sorry.

PR #405, opened 2020-02-21
Add file-age-watcher that monitors the age of files

raffy4284:
This expects a file with configs of:
``file.<name>.dest = <file_location>``

So ideally, the files that you're watching from live-data and secrets fetcher are put in here as entries;

This script will run as a sidecar with all N pods from reddit-service-listing ( or any other service ) and based on the config, it will get the age of the file and send to metrics backend as a histogram.

From here, we can figure out if at least one of the N pods contains a stale file. From here, alert that live-data-fetcher may be faulty.

### DIFF
--- a/baseplate/sidecars/file_age_watcher.py
+++ b/baseplate/sidecars/file_age_watcher.py
@@ -34,7 +34,7 @@ def main() -> NoReturn:
     parser = configparser.RawConfigParser()
     parser.read(args.config_file.name)
     watcher_config = dict(parser.items("file-age-watcher"))
-    cfg = config.parse_config(watcher_config, {"file": config.DictOf({"dest": config.String})})
+    cfg = config.parse_config(watcher_config, {"file": config.DictOf({"path": config.String})})
 
     metrics_client = metrics_client_from_config(watcher_config)
     while True:
@@ -42,7 +42,7 @@ def main() -> NoReturn:
         now = time.time()
         for name, file in cfg.file.items():
             try:
-                mtime = os.path.getmtime(file.dest)
+                mtime = os.path.getmtime(file.path)
             except OSError:
                 mtime = 0
 
--- a/baseplate/sidecars/file_age_watcher.py
+++ b/baseplate/sidecars/file_age_watcher.py
@@ -1,0 +1,54 @@
+"""Calculate the age of files sync'd to disk by the baseplate live-data watcher.
+
+The watcher periodically updates the mtime even if data hasn't changed, so we
+can alert on excessive age as an indication that data sync isn't working.
+"""
+import argparse
+import configparser
+import os
+import sys
+import time
+
+from typing import NoReturn
+
+from baseplate import config
+from baseplate.lib.metrics import metrics_client_from_config
+
+
+HEARTBEAT_INTERVAL = 10
+
+
+def main() -> NoReturn:
+    """
+    Sidecar that tracks the age of the file monitored by live-data watcher and secrets fetcher.
+
+    Use this with live_data_watcher and secrets_fetcher to monitor their files and ensure that
+    these sidecars are not failing silently.
+    """
+    arg_parser = argparse.ArgumentParser(description=sys.modules[__name__].__doc__)
+    arg_parser.add_argument(
+        "config_file", type=argparse.FileType("r"), help="path to a configuration file"
+    )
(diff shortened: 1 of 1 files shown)


raffy4284:
to @spladug:
Do you know what this error means?
``Implicit return in function which does not return``

spladug:
> Do you know what this error means?
> Implicit return in function which does not return

You marked the `main()` function as `-> NoReturn` which is used for functions that won't ever finish executing normally (they throw an exception, exit the program, etc.). But the function actually does end when the for loop finishes and all the gauges are replaced.

You probably want to `-> None` rather than `-> NoReturn`.

baseplate/sidecars/file_age_watcher.py
```diff
@@ -0,0 +1,49 @@
+"""Calculate the age of files sync'd to disk by the baseplate live-data watcher.
+
+The watcher periodically updates the mtime even if data hasn't changed, so we
```
spladug: this says "periodically" but it only runs once. do you intend this to be a one-shot sidecar that gets spun up periodically by the hosting environment or for the sidecar to run indefinitely and manage its own cycle of checking file health?

baseplate/sidecars/file_age_watcher.py
```diff
@@ -0,0 +1,49 @@
...
+    metrics_client = metrics_client_from_config(watcher_config)
+    now = time.time()
+    for name, file in cfg.file.items():
+        try:
+            mtime = os.path.getmtime(file.dest)
+        except OSError:
+            mtime = 0
+
+        age = now - mtime
+        metrics_client.gauge("file-age-watcher.{}.age".format(name)).replace(age)
```
spladug: how will this work when there are multiple pods in prod and each one has one of these sidecars? they'll all be overwriting the same gauge?

spladug: (nit: please use `f"file-age-watcher.{name}.age"` rather than `.format()`)

baseplate/sidecars/file_age_watcher.py
```diff
@@ -0,0 +1,56 @@
...
+        time.sleep(HEARTBEAT_INTERVAL)
+        now = time.time()
+        for name, file in cfg.file.items():
+            try:
+                mtime = os.path.getmtime(file.dest)
+            except OSError:
+                mtime = 0
+
+            age = now - mtime
+            metrics_client.gauge(f"file-age-watcher.{hostname}.{name}.age").replace(age)
```
raffy4284: I don't need to do some sort of metrics_client.flush right?

raffy4284:
Tested this within onevm within graphql's env ( since theirs is up-to-date )....
Will change the base image of where I'm using this to be the general baseplate upgraded version image

baseplate/sidecars/file_age_watcher.py
spladug: no, that's only for batches (though you might want to make a batch for these metrics :man_shrugging: )

baseplate/sidecars/file_age_watcher.py
```diff
@@ -0,0 +1,57 @@
...
+        time.sleep(HEARTBEAT_INTERVAL)
+        now = time.time()
+        for name, file in cfg.file.items():
+            try:
+                mtime = os.path.getmtime(file.dest)
+            except OSError:
+                mtime = 0
+
+            age = now - mtime
+            metrics_client.histogram(f"file-age-watcher.{name}.age").add_sample(age)
```
raffy4284: Is this ok?
Out of N hosts...
Let each host emit this file age and put them into some sort of histogram.
From the histogram, get the ``max`` value which indicates the oldest file in the buckets.
If ``max`` hits a certain threshold, emit an alert, indicating that at least 1 of the N hosts has had their file stale

raffy4284:
@spladug I *think* this will work? since all we care about are just the max age number ( which indicates the oldest file )

baseplate/sidecars/file_age_watcher.py
spladug: Since this is a new sidecar and opt-in, let's give it a shot.

raffy4284:
💇‍♂ 
also spoke with whitney a while back. He said this should be fine as long as we have some tooling on how to find out which host is bad.

Sorry for the lag! I've got a lot of stuff sliding on me

PR #406, opened 2020-02-25
A couple of doc example fixes

spladug:
:eyeglasses: @kemitche


spladug:
oops! good catch, thank you.

PR #408, opened 2020-03-04
Add options to let redis-py decode strings

markis:
`redis-py` has an option `decode_responses`, which will automatically decode all string values.  Having redis-py do all of the conversions makes the calling code cleaner and possibly slightly more performant.

When `decode_responses` is True, `redis-py` will take the values returned from the server and call [`bytes.decode`](https://docs.python.org/3/library/stdtypes.html#bytes.decode).  `encoding` and `encoding_errors` are passed into the decode function.


markis:
Nevermind, I am doing it all wrong.

PR #411, opened 2020-03-11
create context.metrics even when request is not sampled

diffyqgirl:
Fixup to this pr:
https://github.com/reddit/baseplate.py/pull/400

Many applications access context.metrics directly, which currently throws an error because `context.metrics` is not assigned if the request is not sampled.

### DIFF
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -124,9 +124,10 @@ class BufferedTransport(Transport):
         self.buffer.append(serialized_metric)
 
     def flush(self) -> None:
-        metrics, self.buffer = self.buffer, []
-        message = b"\n".join(metrics)
-        self.transport.send(message)
+        if self.buffer:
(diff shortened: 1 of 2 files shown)


diffyqgirl:
👓 @spladug

baseplate/lib/metrics.py
```diff
@@ -347,6 +347,8 @@ def send(self, delta: float, sample_rate: float) -> None:
         :param sample_rate: What rate this counter is sampled at. [0-1].
 
         """
+        if delta == 0:
```
spladug: while i don't think this is really a bad change, what's the context for it happening here?

diffyqgirl: The context is that I misunderstood your suggestion to not be sending packets if there's no data. I'm happy to take or leave it.

diffyqgirl:
💇

PR #413, opened 2020-03-24
Kombu: Add ability to pass in error_handler_fn

punitrathore:
Current behavior when an error occurs is to requeue the message. This
allows the ability to override that behavior by passing in a function

### DIFF
--- a/baseplate/frameworks/queue_consumer/kombu.py
+++ b/baseplate/frameworks/queue_consumer/kombu.py
@@ -109,6 +109,7 @@ class KombuMessageHandler(MessageHandler):
                 span.set_tag("amqp.delivery_tag", delivery_info.get("delivery_tag", ""))
                 span.set_tag("amqp.exchange", delivery_info.get("exchange", ""))
                 self.handler_fn(context, message_body, message)
+                message.ack()
         except Exception as exc:
             logger.exception(
                 "Unhandled error while trying to process a message.  The message "
@@ -118,12 +119,10 @@ class KombuMessageHandler(MessageHandler):
                 self.error_handler_fn(context, message_body, message)
             else:
                 message.requeue()
+
             if isinstance(exc, FatalMessageHandlerError):
                 logger.info("Recieved a fatal error, terminating the server.")
                 raise
-        else:
-            if not self.error_handler_fn:
-                message.ack()
 
 
 class KombuQueueConsumerFactory(QueueConsumerFactory):
@@ -157,8 +156,6 @@ class KombuQueueConsumerFactory(QueueConsumerFactory):
         :param routing_keys: List of routing keys that you will create :py:class:`~kombu.Queue` s
             to consume from.
         :param handler_fn: A function that will process an individual message from a queue.
-            If `error_handler_fn` is passed in, this function is responsible for calling
-            `message.ack` or `message.requeue` when a message is successfully processed.
         :param error_handler_fn: A function that will be called when an error is thrown
             while executing the `handler_fn`. This function will be responsible for calling
             `message.ack` or `message.requeue` as it will not be automatically called by
@@ -204,8 +201,6 @@ class KombuQueueConsumerFactory(QueueConsumerFactory):
         :param routing_keys: List of routing keys that you will create
             :py:class:`~kombu.Queue` s to consume from.
         :param handler_fn: A function that will process an individual message from a queue.
-            If `error_handler_fn` is passed in, this function is responsible for calling
-            `message.ack` or `message.requeue` when a message is successfully processed.
         :param error_handler_fn: A function that will be called when an error is thrown
             while executing the `handler_fn`. This function will be responsible for calling
             `message.ack` or `message.requeue` as it will not be automatically called by
--- a/baseplate/frameworks/queue_consumer/kombu.py
+++ b/baseplate/frameworks/queue_consumer/kombu.py
@@ -81,10 +81,17 @@ class KombuConsumerWorker(ConsumerMixin, PumpWorker):
 
 
 class KombuMessageHandler(MessageHandler):
-    def __init__(self, baseplate: Baseplate, name: str, handler_fn: Handler):
+    def __init__(
+        self,
+        baseplate: Baseplate,
+        name: str,
+        handler_fn: Handler,
+        error_handler_fn: Optional[Handler] = None,
+    ):
         self.baseplate = baseplate
         self.name = name
         self.handler_fn = handler_fn
+        self.error_handler_fn = error_handler_fn
 
     def handle(self, message: kombu.Message) -> None:
         context = self.baseplate.make_context_object()
@@ -105,7 +112,10 @@ class KombuMessageHandler(MessageHandler):
                 "Unhandled error while trying to process a message.  The message "
                 "has been returned to the queue broker."
             )
-            message.requeue()
+            if self.error_handler_fn:
+                self.error_handler_fn(context, message.decode(), message)
+            else:
+                message.requeue()
             if isinstance(exc, FatalMessageHandlerError):
                 logger.info("Recieved a fatal error, terminating the server.")
                 raise
@@ -130,6 +140,7 @@ class KombuQueueConsumerFactory(QueueConsumerFactory):
         connection: kombu.Connection,
         queues: Sequence[kombu.Queue],
         handler_fn: Handler,
+        error_handler_fn: Optional[Handler] = None,
         health_check_fn: Optional[HealthcheckCallback] = None,
         serializer: Optional[KombuSerializer] = None,
     ):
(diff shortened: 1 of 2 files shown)


punitrathore:
👓 @spladug @pacejackson

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -182,8 +195,11 @@ def new(
         :param queue_name: Name for your queue.
         :param routing_keys: List of routing keys that you will create
             :py:class:`~kombu.Queue` s to consume from.
-        :param handler_fn: A `baseplate.frameworks.queue_consumer.komub.Handler`
+        :param handler_fn: A `baseplate.frameworks.queue_consumer.kombu.Handler`
```
spladug: :grimacing: thanks

spladug: actually, in both of these: is it actually necessary to spell out the type like this? it should be auto-linked by the type hint in the sphinx docs.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -105,7 +112,10 @@ def handle(self, message: kombu.Message) -> None:
                 "Unhandled error while trying to process a message.  The message "
                 "has been returned to the queue broker."
             )
-            message.requeue()
+            if self.error_handler_fn:
+                self.error_handler_fn(context, message.decode(), message)
```
spladug: what happens if decoding was the problem in the first place? would it make sense to re-use the decoded value from the happy path here and default it to something like `None` should the failure have happened before decoding was complete?

punitrathore:
💇 also added additional changes to only `ack` the message if there is no `error_handler_fn`. Reasoning is that if the `ack` happens then the `error_handler_fn` cannot do things like `message.reject` or `message.nack` which is required if we want to dead letter the message.

punitrathore:
@pacejackson see the comment above as to why we need the `error_handler_fn`. Currently the `KombuMessageHandler`'s `handle` function always calls `message.ack` which limits the things you can do by overriding the `handler_fn`.

baseplate/frameworks/queue_consumer/kombu.py
punitrathore: Good point, I did not know how these comments are parsed out. I've updated it as suggested

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -142,8 +156,9 @@ def __init__(
         :param queue_name: Name for your queue.
         :param routing_keys: List of routing keys that you will create :py:class:`~kombu.Queue` s
             to consume from.
-        :param handler_fn: A `baseplate.frameworks.queue_consumer.komub.Handler`
-            function that will process an individual message from a queue.
+        :param handler_fn: A function that will process an individual message from a queue.
+        :param error_handler_fn: A function that will be called when an error is thrown
+            while executing the `handler_fn`.
```
pacejackson: Good point about why we need this, can you update this comment to reflect that `message.ack` and/or `message.requeue` will _not_ be called automatically if this is set when an error is raised and you need to ensure that this function handles that?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -182,8 +199,9 @@ def new(
         :param queue_name: Name for your queue.
         :param routing_keys: List of routing keys that you will create
             :py:class:`~kombu.Queue` s to consume from.
-        :param handler_fn: A `baseplate.frameworks.queue_consumer.komub.Handler`
-            function that will process an individual message from a queue.
+        :param handler_fn: A function that will process an individual message from a queue.
+        :param error_handler_fn: A function that will be called when an error is thrown
```
pacejackson: Same question about updating the comments.

punitrathore:
💇‍♂ updated the docstring

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -94,23 +101,29 @@ def handle(self, message: kombu.Message) -> None:
...
+                self.error_handler_fn(context, message_body, message)
+            else:
+                message.requeue()
             if isinstance(exc, FatalMessageHandlerError):
                 logger.info("Recieved a fatal error, terminating the server.")
                 raise
         else:
-            message.ack()
+            if not self.error_handler_fn:
+                message.ack()
```
pacejackson: I don't think we want to do this, we should just `ack` if there's no error, I don't think that adding an `error_handler_fn` should affect `handler_fn`.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -142,8 +156,13 @@ def __init__(
         :param queue_name: Name for your queue.
         :param routing_keys: List of routing keys that you will create :py:class:`~kombu.Queue` s
             to consume from.
-        :param handler_fn: A `baseplate.frameworks.queue_consumer.komub.Handler`
-            function that will process an individual message from a queue.
+        :param handler_fn: A function that will process an individual message from a queue.
+            If `error_handler_fn` is passed in, this function is responsible for calling
+            `message.ack` or `message.requeue` when a message is successfully processed.
```
pacejackson: I don't think this is accurate, (or what we want). The only difference should be when an error is raised, `error_handler_fn` should finalize the message (`ack`, `requeue`, etc), not the case where there is no error.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -94,23 +101,29 @@ def handle(self, message: kombu.Message) -> None:
...
         except Exception as exc:
             logger.exception(
                 "Unhandled error while trying to process a message.  The message "
                 "has been returned to the queue broker."
             )
-            message.requeue()
+            if self.error_handler_fn:
+                self.error_handler_fn(context, message_body, message)
+            else:
+                message.requeue()
```
spladug: nit: please add a blank after this so it doesn't read at first glance like these two ifs are related

baseplate/frameworks/queue_consumer/kombu.py
punitrathore: The problem with `always` calling `ack` in a finally clause is that if we call `reject` in the `error_handler_fn`, it doesn't allow you to call `ack` on the message again.

punitrathore: I guess what you're saying is to remove the `ack` from the `else` clause and move it in the `try` clause

punitrathore:
💇‍♂ 🤞 hoping this is the last change!

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -94,23 +101,28 @@ def handle(self, message: kombu.Message) -> None:
...
+
             if isinstance(exc, FatalMessageHandlerError):
                 logger.info("Recieved a fatal error, terminating the server.")
                 raise
-        else:
-            message.ack()
```
pacejackson: I don't think this needs to be moved, it's in an `else` not a `finally` so it runs if an exception does not trigger.  Also, we don't want to catch errors that this raises and trigger the error handler/requeue.

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: We wouldn't call `ack` in that case because this code only runs if an exception isn't raised (it's an `else`, not a `finally`), we want this outside of the try block since we don't want to catch `ack` errors, the try block is just for errors during message handling.

baseplate/frameworks/queue_consumer/kombu.py
punitrathore: Ahhh, TIL.

punitrathore:
💇‍♂ @pacejackson added the else clause

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -94,18 +101,24 @@ def handle(self, message: kombu.Message) -> None:
...
-                self.handler_fn(context, message.decode(), message)
+                self.handler_fn(context, message_body, message)
         except Exception as exc:
             logger.exception(
                 "Unhandled error while trying to process a message.  The message "
                 "has been returned to the queue broker."
             )
-            message.requeue()
+            if self.error_handler_fn:
+                self.error_handler_fn(context, message_body, message)
```
pacejackson: Last one, I promise! Sorry I missed this before but we should also pass `exc` to the error handler.

punitrathore: Good catch. that's pretty important

punitrathore:
@spladug got some weird transient failures re:cassandra tests, which got fixed by retriggering the build. lemme know if this is good to go and whats the next step are to getting this available as a dev dependency?

PR #414, opened 2020-03-26
Add sampling to metrics created by timers

diffyqgirl:
The previous sampling PR failed to account for the fact that some counters are created implicitly by timers. These counters need to know about sampling or telegraf won't be able to compensate for it correctly on the backend.

This should also make sampling more backwards compatible with the old context.metrics syntax by removing the error where context.metrics doesn't exist.

### DIFF
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -135,14 +135,14 @@ class BaseClient:
         self.transport = transport
         self.namespace = namespace.encode("ascii")
 
-    def timer(self, name: str) -> "Timer":
+    def timer(self, name: str, sample_rate: float = 1.0) -> "Timer":
         """Return a Timer with the given name.
 
         :param name: The name the timer should have.
 
         """
         timer_name = _metric_join(self.namespace, name.encode("ascii"))
-        return Timer(self.transport, timer_name)
+        return Timer(self.transport, timer_name, sample_rate)
 
     def counter(self, name: str) -> "Counter":
         """Return a Counter with the given name.
@@ -268,12 +268,13 @@ class Timer:
 
     """
 
-    def __init__(self, transport: Transport, name: bytes):
+    def __init__(self, transport: Transport, name: bytes, sample_rate: float = 1.0):
         self.transport = transport
         self.name = name
 
         self.start_time: Optional[float] = None
         self.stopped: bool = False
+        self.sample_rate = sample_rate
 
     def start(self) -> None:
         """Record the current time as the start of the timer."""
@@ -302,6 +303,9 @@ class Timer:
 
         """
         serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        if self.sample_rate < 1.0:
+            sampling_info = f"@{self.sample_rate:g}".encode()
+            serialized = serialized + b"|" + sampling_info
         self.transport.send(serialized)
 
     def __enter__(self) -> None:
@@ -350,7 +354,7 @@ class Counter:
         """
         parts = [self.name + (f":{delta:g}".encode()), b"c"]
 
-        if sample_rate != 1.0:
+        if sample_rate < 1.0:
             parts.append(f"@{sample_rate:g}".encode())
 
         serialized = b"|".join(parts)
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -274,7 +274,7 @@ class Timer:
 
         self.start_time: Optional[float] = None
         self.stopped: bool = False
-        self.sample_rate = None
+        self.sample_rate = 1.0
 
     def start(self, sample_rate: float = 1.0) -> None:
         """Record the current time as the start of the timer."""
@@ -304,7 +304,7 @@ class Timer:
 
         """
         serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
-        if self.sample_rate and self.sample_rate < 1.0:
+        if self.sample_rate < 1.0:
             sampling_info = f"@{self.sample_rate:g}".encode()
             serialized = serialized + b"|" + sampling_info
         self.transport.send(serialized)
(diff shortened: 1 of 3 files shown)


baseplate/lib/metrics.py
```diff
@@ -302,6 +303,9 @@ def send(self, elapsed: float) -> None:
 
         """
         serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        if self.sample_rate != 1.0:
```
fishy: I don't know how python handle float numbers, but the general rule of thumb is that don't use `==` and `!=` on float numbers. In this case I'd suggest to use `self.sample_rate < 1.0` instead.

diffyqgirl: 👍

diffyqgirl:
👓 @m104

diffyqgirl:
👓 @spladug

baseplate/observers/metrics.py
```diff
@@ -53,15 +53,35 @@ def on_server_span_created(self, context: RequestContext, server_span: Span) ->
...
+    def on_child_span_created(self, span: Span) -> None:
+        pass
 
 
 class MetricsServerSpanObserver(SpanObserver):
     def __init__(self, batch: metrics.Batch, server_span: Span, sample_rate: float = 1.0):
         self.batch = batch
         self.base_name = "server." + server_span.name
-        self.timer: Optional[metrics.Timer] = None
+        self.timer: Optional[metrics.Timer] = batch.timer(self.base_name, sample_rate)
```
spladug: this will get replaced on L88

baseplate/observers/metrics.py
```diff
@@ -101,7 +121,9 @@ def on_child_span_created(self, span: Span) -> None:
 class MetricsLocalSpanObserver(SpanObserver):
     def __init__(self, batch: metrics.Batch, span: Span, sample_rate: float = 1.0):
         self.batch = batch
-        self.timer = batch.timer(typing.cast(str, span.component_name) + "." + span.name)
+        self.timer = batch.timer(
```
spladug: can probably f-string this now: `f"{span.component_name}.{span.name}"

if mypy still catches the potential `None` there the cast can go inside the f-string (:nauseated_face:)

baseplate/lib/metrics.py
```diff
@@ -302,6 +303,9 @@ def send(self, elapsed: float) -> None:
 
         """
         serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        if self.sample_rate < 1.0:
+            sampling_info = f"@{self.sample_rate:g}".encode()
+            serialized = serialized + b"|" + sampling_info
```
spladug: worth benchmarking to be sure, but it might be more performant to do this the same way `Counter.send` works with building up a list of byte strings and then joining them at the end. that's generally one of the faster ways of doing this sort of string concatenation in python.

baseplate/lib/metrics.py
```diff
@@ -135,14 +135,14 @@ def __init__(self, transport: Transport, namespace: str):
         self.transport = transport
         self.namespace = namespace.encode("ascii")
 
-    def timer(self, name: str) -> "Timer":
+    def timer(self, name: str, sample_rate: float = 1.0) -> "Timer":
```
spladug: this makes timers have a pretty different interface from counters. would it make sense to put the sample rate on the `start()` and `send()` methods of a timer? feels kinda weird too.

diffyqgirl: Yeah I thought about this, I'd be happy to move it to `start`. I wasn't sure if the same timer could get restarted or how that might interact.

diffyqgirl:
💇 @spladug

baseplate/lib/metrics.py
```diff
@@ -302,6 +304,9 @@ def send(self, elapsed: float) -> None:
 
         """
         serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
```
spladug: like the counters, all the parts should be in the list building setup, not just the sample rate.

```python
        parts = [f"{self.name}:{(elapsed * 1000.0):g}".encode()), b"ms"]

        if sample_rate < 1.0:
            parts.append(f"@{sample_rate:g}".encode())

        serialized = b"|".join(parts)
```

baseplate/lib/metrics.py
```diff
@@ -302,6 +304,9 @@ def send(self, elapsed: float) -> None:
```
spladug: i think we'd want `send()` to take a sample_rate as well. 

```python
def stop(self):
    ...
    self.send(elapsed, sample_rate=self.sample_rate)
    ...

def send(self, elapsed: float, sample_rate: float) -> None:
    ...
    if sample_rate < 1.0:  # note: _not_ self.sample_rate
```

baseplate/lib/metrics.py
```diff
@@ -274,9 +274,11 @@ def __init__(self, transport: Transport, name: bytes):
 
         self.start_time: Optional[float] = None
         self.stopped: bool = False
+        self.sample_rate = 1.0
 
-    def start(self) -> None:
+    def start(self, sample_rate: float = 1.0) -> None:
         """Record the current time as the start of the timer."""
+        self.sample_rate = sample_rate
```
spladug: nit: can this go down with `self.start_time = ` so it's below the guard statements as well?

diffyqgirl: 👍

baseplate/lib/metrics.py
diffyqgirl: 👍

baseplate/lib/metrics.py
diffyqgirl: This gets complicated because name is bytes but everything else is strings, so name gets double-encoded.

diffyqgirl:
💇 @spladug

PR #415, opened 2020-03-26
Thrift: catch exceptions in span.start() as part of flow

spladug:
This is required to get proper error handling for observers that use
on_start() to throw an exception to kill the request preemptively, such
as the [upcoming concurrency limiter.][limit]

[limit]: https://github.com/reddit/baseplate.py/compare/master...spladug:active-request-limit


spladug:
cc: @fizx

spladug:
:haircut_man: sorry, one more.

PR #416, opened 2020-03-26
Revert "Add file-age-watcher that monitors the age of files"

raffy4284:
Reverts reddit/baseplate.py#405

Going to revert this...trying to add this as a side car adds another failure point for the main service!

Going to hold for a bit and see what katie has to say within the channel!


raffy4284:
👓 @spladug

PR #417, opened 2020-03-28
Add 'request.origin_service' to the edge request context

pacejackson:
This is intended to identify the 'origin' service of this request, that is the service that is handling the request made from the client


pacejackson:
cc: @jesseyli

baseplate/thrift/BaseplateService.py
```diff
@@ -1,5 +1,5 @@
 #
-# Autogenerated by Thrift Compiler (0.12.1)
+# Autogenerated by Thrift Compiler (0.13.0)
```
pacejackson, line 2: @spladug is this ok? 😬

spladug: seems fine. only change is the `_on_message_begin()` thing which looks harmless

PR #418, opened 2020-03-30
Add support for wavefront tagging to baseplate

vanessawuhoo:


vanessawuhoo:
@pacejackson

baseplate/lib/metrics.py
```diff
@@ -49,20 +49,33 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 def _metric_join(*nodes: bytes) -> bytes:
-    return b".".join(node.strip(b".") for node in nodes)
+    return b".".join(node.strip(b".") for node in filter(None, nodes))
+
+
+def _format_tags(tags: Optional[Dict]) -> str:
```
spladug, line 70: Since the metrics code is called a _lot_ in an application and I've seen it show up in performance profiling in the past, I was curious if this was the most efficient way to serialize the tags. Particularly since the standard wisdom is that f-strings and `str.join` are faster than `+` concatenation. The results were kinda surprising! 

https://gist.github.com/spladug/b936c5a9528ced0d1095a01aa1136075

So if we want this to be as quick as possible, something like the following would appear good:

```python
def format_tags_concatenate_fstrings(tags: Optional[Dict]) -> str:
    if not tags:                                                  
        return ""                                                 
                                                                  
    formatted_str = "|#"                                          
    for key, value in tags.items():                               
        formatted_str += f"{key}:{value},"                        
    return formatted_str[:-1]                                     
```

(it's a tiny bit faster without stripping the trailing comma off, but I think wasting bytes in our network packets is probably worse)

Feel free to play around with benchmarking and find a faster one too :)

baseplate/lib/metrics.py
```diff
@@ -186,6 +199,17 @@ def batch(self) -> "Batch":
         """
         return Batch(self.transport, self.namespace)
 
+    def tagged_batch(self) -> "Batch":
```
spladug, line 202: I don't think we need this now, right? Since we're recommending modern services don't put a namespace at all in favor of telegraf doing it for them?

baseplate/lib/metrics.py
```diff
@@ -303,12 +327,26 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
         :param elapsed: The elapsed time in seconds to report.
 
         """
-        serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        # serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        serialized = (
+            self.name + (f":{(elapsed * 1000.0):g}|ms".encode()) + _format_tags(tags).encode()
```
spladug, line 332: maybe write this like `self.name + f":{(elapsed * 1000.0):g}|ms{_format_tags(tags)}".encode()`? (also maybe worth benchmarking)

baseplate/lib/metrics.py
```diff
@@ -303,12 +327,26 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
         :param elapsed: The elapsed time in seconds to report.
 
         """
-        serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        # serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
```
spladug, line 330: nit: these comments can go away

baseplate/lib/metrics.py
```diff
@@ -329,24 +367,28 @@ def __init__(self, transport: Transport, name: bytes):
         self.transport = transport
         self.name = name
 
-    def increment(self, delta: float = 1.0, sample_rate: float = 1.0) -> None:
+    def increment(
```
spladug, line 370: Would it work to put the tags on the construction of the counter/timer/etc.? That feels more conceptually correct since the tag is part of the metric name not value. It also means we don't need `stop_tagged()` on timers since we'll just `.send()` regardless. So that'd look like this instead:

```python
batch.counter("foo", tags={"bar": "baz"}).increment()
```

baseplate/lib/metrics.py
```diff
@@ -512,3 +560,40 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
 
     # pylint: disable=maybe-no-member
     return make_client(cfg.metrics.namespace, cfg.metrics.endpoint)
+
+
+def metrics_observer_from_config(raw_config: config.RawConfig, client: Client) -> Any:
```
spladug, line 565: 1. The metrics library doesn't generally know the observers even exist, so it feels a bit odd to have this in this file.
2. Since we're already doing branching like this in `configure_observers`, I think we can probably skip the indirection here? Then we just ask the right observer directly to configure itself. A common pattern in baseplate for this is a `.from_config()` classmethod:

```python
class TaggedMetricsBaseplateObserver(BaseplateObserver):
    @classmethod
    def from_config_and_client(cls, raw_config, client):
        cfg = config.parse_config(raw_config, {
            "metrics_observer:" {
               ...
            }
        })
        return cls(cfg, client)

    def __init__(self, cfg, client):
        ...
```

baseplate/lib/metrics.py
```diff
@@ -303,12 +327,26 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
...
-        serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        # serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+        serialized = (
+            self.name + (f":{(elapsed * 1000.0):g}|ms".encode()) + _format_tags(tags).encode()
+        )
         if sample_rate < 1.0:
             sampling_info = f"@{sample_rate:g}".encode()
-            serialized = b"|".join([serialized, sampling_info])
+            # serialized = b"|".join([serialized, sampling_info])
+            serialized = b"|".join([serialized, sampling_info]) + _format_tags(tags).encode()
```
spladug, line 337: hmm! if `format_tags` didn't add `|` itself it might be possible to have `.join()` here do that for us? worth benchmarking to see if it's worth it?

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,154 @@
...
+        self.client = client
+        self.whitelist = self.parse_tags(cfg)
+
+    def on_server_span_created(self, context: RequestContext, server_span: Span) -> None:
+        batch = self.client.tagged_batch()
+        context.metrics = batch
+        observer = TaggedMetricsServerSpanObserver(batch, server_span, self.whitelist)
+        server_span.register(observer)
+
+    def parse_tags(self, raw_config: config.RawConfig) -> list:
```
spladug, line 49: this would be a good fit for the `from_config` pattern talked about in the other comment

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,154 @@
...
+        cfg = config.parse_config(
+            raw_config,
+            {
+                "metrics": {
+                    "tagging": config.Optional(config.Boolean, default=False),
+                    "whitelist": config.Optional(config.String, default=""),
+                }
+            },
+        )
+        return [tag.replace(" ", "") for tag in cfg.metrics.whitelist.split(",")]
```
spladug, line 59: https://baseplate.readthedocs.io/en/stable/api/baseplate/lib/config.html#baseplate.lib.config.TupleOf

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,154 @@
...
+        observer = TaggedMetricsServerSpanObserver(batch, server_span, self.whitelist)
+        server_span.register(observer)
+
+    def parse_tags(self, raw_config: config.RawConfig) -> list:
+        cfg = config.parse_config(
+            raw_config,
+            {
+                "metrics": {
+                    "tagging": config.Optional(config.Boolean, default=False),
+                    "whitelist": config.Optional(config.String, default=""),
```
spladug, line 55: does this mean that every application will need to add the built-in tags to the whitelist? maybe we should just hard-code those as acceptable?

tests/integration/test_thrift/test/__init__.py
```diff
@@ -0,0 +1 @@
+__all__ = ["ttypes", "constants", "TestService"]
```
spladug, line 1: what's going on with all these test files?

baseplate/lib/metrics.py
vanessawuhoo: I added this just so we could get a Batch with an empty namespace, since it's something attached automatically in batch() without being passed in

baseplate/observers/metrics_tagged.py
vanessawuhoo: Currently that's how it is set up because I was concerned about the cardinality limits (especially the 20 tag hard cap). Open to hard coding them if that's not a concern though

tests/integration/test_thrift/test/__init__.py
vanessawuhoo: I actually...do not know. they popped up when I ran the linter but I didn't actually touch them when I was coding.

baseplate/lib/metrics.py
vanessawuhoo: Is there another way I could do this? I didn't want to redo the batch() function since that might hurt backwards compatibility

baseplate/lib/metrics.py
vanessawuhoo: I think that makes sense!

baseplate/lib/metrics.py
spladug: I think we could just ignore namespaces altogether, right? and make sure we throw an error if they used tags + a configured namespace?

vanessawuhoo: * set a rule that says namespace must be blank in config

baseplate/lib/metrics.py
vanessawuhoo: configure_observers() will decide which metrics observer to use

vanessawuhoo: will require configure_observers use, logic for which metrics observer to attach will occur there

vanessawuhoo: remove metrics_observer_from_config as parsing of cfg will occur inside TaggedMetricsObserver

baseplate/lib/metrics.py
vanessawuhoo: * benchmark the ways of formatting tags for performance

baseplate/observers/metrics_tagged.py
vanessawuhoo: rename to style of "from style" and also turn config.String -> config.TupleOf

baseplate/observers/metrics_tagged.py
vanessawuhoo: just add stuff in default= 
* endpoint
* success

tests/integration/test_thrift/test/__init__.py
vanessawuhoo: *reset and rm from pr

vanessawuhoo:
💇  @spladug Fixed up and ready for a second pass!

baseplate/lib/metrics.py
```diff
@@ -424,17 +450,28 @@ class Histogram:
     backend must support the :code:`h` key, e.g. :code:`metric_name:320|h`.
     """
 
-    def __init__(self, transport: Transport, name: bytes) -> None:
+    def __init__(
+        self,
+        transport: Transport,
+        name: bytes,
+        tags: Optional[Dict] = None,  # pylint: disable=bad-whitespace
```
spladug, line 457: eeps, what's pylint doing that requires this pragma?

baseplate/lib/metrics.py
```diff
@@ -249,7 +259,7 @@ def counter(self, name: str) -> "Counter":
         counter_name = _metric_join(self.namespace, name.encode("ascii"))
         batch_counter = self.counters.get(counter_name)
```
spladug, line 260: i think we need to figure out how to handle tags here. right now, the first metric of a given name will apply its tags to all subsequent ones. e.g. "foo#a=b" and "foo#x=y" will collide because we're only storing the BatchCounter objects based on metric name not tag

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+
+class Errors(Enum):
+    TIMED_OUT = 1
+    CONCURRENCY_LIMITED = 2
+    EXCEPTION = 3
+
+
+error_messages = {
+    Errors.TIMED_OUT: "timed_out",
+    Errors.CONCURRENCY_LIMITED: "concurrency_limited",
```
spladug, line 31: ah sorry, i only mentioned this as a heads up this would be coming but it's not necessary to include yet since that patch set isn't fully ready to merge at this time

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+    @classmethod
+    def from_config_and_client(
+        cls, raw_config: config.RawConfig, client: metrics.Client
+    ) -> "TaggedMetricsBaseplateObserver":
+        cfg = config.parse_config(
+            raw_config,
+            {
+                "metrics": {
+                    "whitelist": config.Optional(
+                        config.TupleOf(config.String), default="endpoint,success,error"
```
spladug, line 67: should the `default` value be a list of strings rather than the unparsed one?

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+    def on_child_span_created(self, span: Span) -> None:
+        observer: SpanObserver
+        if isinstance(span, LocalSpan):
+            observer = TaggedMetricsLocalSpanObserver(self.batch, span, self.whitelist)
+        else:
+            observer = TaggedMetricsClientSpanObserver(self.batch, span, self.whitelist)
+        span.register(observer)
+
+    def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
+        # elapsed = self.timer.stop_tagged()
```
spladug, line 113:
```suggestion

```

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+        server_span.register(observer)
+
+
+class TaggedMetricsServerSpanObserver(SpanObserver):
+    def __init__(self, batch: metrics.Batch, server_span: Span, whitelist: list):
+        self.batch = batch
+        self.span = server_span
+        self.base_name = "baseplate.server"
+        self.whitelist = whitelist
+        self.tags = {}  # type: Dict[str, str]
```
spladug, line 87:
```suggestion
        self.tags: Dict[str, str] = {}
```

would be a little nicer than using the comments, since it's actually available to the language itself then

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+            observer = TaggedMetricsLocalSpanObserver(self.batch, span, self.whitelist)
+        else:
+            observer = TaggedMetricsClientSpanObserver(self.batch, span, self.whitelist)
+        span.register(observer)
+
+    def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
+        # elapsed = self.timer.stop_tagged()
+
+        if not exc_info:
+            self.span.set_tag("success", "true")
```
spladug, line 116: i don't think we want to be setting tags on the abstract spans from inside the observers. that'll then cause this observer's `on_set_tag()` to get called which is pretty funky and would lead to ordering issues with the other observers (if we get called after them they'll never see this tag).

spladug: maybe just add the error stuff directly to the timer's tags?

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,187 @@
...
+        self.timer = batch.timer(self.base_name)
+
+    def on_start(self) -> None:
+        self.span.set_tag("endpoint", self.span.name)
+        self.timer.start()
+
+    def on_incr_tag(self, key: str, delta: float) -> None:
+        self.batch.counter(key, self.tags).increment(delta, True)
+
+    def on_set_tag(self, key: str, value: Any) -> None:
```
spladug, line 97: would it make sense to call `update_tags()` directly in here to get those tags onto the timer so we don't have to build up a parallel dictionary of stuff in this object?

baseplate/lib/metrics.py
```diff
@@ -62,7 +62,17 @@
 
 
 def _metric_join(*nodes: bytes) -> bytes:
-    return b".".join(node.strip(b".") for node in nodes)
+    return b".".join([node.strip(b".") for node in nodes if node])
```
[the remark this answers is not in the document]
vanessawuhoo: ohhhh cool! I did not know you could do this :til:

baseplate/lib/metrics.py
vanessawuhoo: flatten dictionary and use in key for self.counters

isugimpy:
I don't want to be too much of a pain, but baseplate.go already implements labels on metrics, and is using the influxstatsd style instead of dogstatsd.  Given that there's already prior art for labels in an existing baseplate library, could we standardize on one format or the other?

baseplate/observers/metrics_tagged.py
vanessawuhoo: currently update_tags completely replaces its current tags with a new dict of tags, so callign it would lose any existing tags already on the timer. I'm open to changing it so that update_tags is smarter and adds if not already on the tags and replaces if it is though. Not sure of which is preferable

spladug: ah yeah, if update tags were like:

```python
def update_tags(self, tags):
    self.tags.update(tags)
```

that should do the trick?

baseplate/lib/metrics.py
```diff
@@ -316,17 +316,15 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
...
             parts.append(formatted_tags.encode())
         serialized = b"|".join(parts)
         self.transport.send(serialized)
 
     def update_tags(self, tags: Dict) -> None:
-        if self.stopped:
-            logger.warning("cannot update tags on stopped timer")
-        else:
-            self.tags = tags
+        assert(not self.stopped)
```
spladug, line 326:
```suggestion
        assert not self.stopped
```

vanessawuhoo:
@spladug  ready for pass 3!

vanessawuhoo:
@isugimpy Hi Jayme! sorry for the late response, I didn't have a lot of time last sprint to work on this. influxstatsd makes some of the things we're trying to do with batchcounters cleaner so I pushed the change to switch to it. Thanks for raising this issue!!

vanessawuhoo:
💇‍♀️ @pacejackson  @spladug

baseplate/lib/metrics.py
```diff
@@ -299,16 +311,24 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
...
+        formatted_tags = _format_tags(self.tags)
+        if formatted_tags:
+            parts = [self.name + formatted_tags.encode() + f":{(elapsed * 1000.0):g}|ms".encode()]
+        else:
+            parts = [self.name + f":{(elapsed * 1000.0):g}|ms".encode()]
         if sample_rate < 1.0:
             sampling_info = f"@{sample_rate:g}".encode()
-            serialized = b"|".join([serialized, sampling_info])
+            parts.append(sampling_info)
+        serialized = b"|".join(parts)
```
spladug, line 325: I think this can go back to how it was before since with influx we're always either one or two components, never three.

baseplate/lib/metrics.py
```diff
@@ -62,7 +62,17 @@
...
+
+
+def _format_tags(tags: Optional[Dict]) -> Optional[str]:
+    if not tags:
+        return None
+
+    parts = [""]
+    for key, value in tags.items():
+        parts.append(f"{key}={value}")
+    return ",".join(parts)
```
spladug, line 75: since we have to call `encode()` every time we sue this function, how bout doing the encode here to save time for callers?

baseplate/lib/metrics.py
```diff
@@ -299,16 +311,24 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
...
         :param elapsed: The elapsed time in seconds to report.
-
         """
-        serialized = self.name + (f":{(elapsed * 1000.0):g}|ms".encode())
+
+        formatted_tags = _format_tags(self.tags)
+        if formatted_tags:
+            parts = [self.name + formatted_tags.encode() + f":{(elapsed * 1000.0):g}|ms".encode()]
+        else:
+            parts = [self.name + f":{(elapsed * 1000.0):g}|ms".encode()]
```
spladug, line 321:
```suggestion
        serialized = self.name
        if self.tags:
            serialized += _format_tags(self.tags)
        serialized += f":{(elapsed * 1000.0):g}|ms".encode()
```

would be a bit less repetitive. though this probably bears a bit of benchmarking?

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,183 @@
...
+error_messages = {
+    Errors.TIMED_OUT: "timed_out",
+    Errors.EXCEPTION: "internal_server_error",
+}
+
+
+class TaggedMetricsBaseplateObserver(BaseplateObserver):
+    """Metrics collecting observer.
+
+    This observer reports metrics to statsd in the dogstatsd format. It does three important things:
```
spladug, line 37:
```suggestion
    This observer reports metrics to statsd in the Influx StatsD format. It does three important things:
```

baseplate/observers/metrics_tagged.py
spladug: ah, i'm sorry, i mislead you here. the update_tags as-is works for the timer but not the counters. i think we do have to .update() on the self.tags and like you originally had.  sorry :( (the rest is fine though)

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, batch: metrics.Batch, server_span: Span, whitelist: list):
+        self.batch = batch
+        self.span = server_span
+        self.base_name = "baseplate.server"
+        self.whitelist = whitelist
+        self.tags: Dict[str, str] = {}
+        self.timer = batch.timer(self.base_name)
+
+    def on_start(self) -> None:
+        self.span.set_tag("endpoint", self.span.name)
```
spladug, line 89:
```suggestion
        self.tags["endpoint"] = self.span.name
```

another instance of not wanting to call set_tag() on the abstract span from within an observer

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, batch: metrics.Batch, span: Span, whitelist: list):
+        self.batch = batch
+        self.span = span
+        self.tags = {}  # type: Dict[str, str]
+        self.timer = batch.timer("baseplate.local")
+        self.whitelist = whitelist
+
+    def on_start(self) -> None:
+        self.timer.start()
+        self.span.set_tag("endpoint", self.span.name)
```
spladug, line 134: same set_tag() worry

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,183 @@
...
+        self.batch = batch
+        self.span = span
+        self.base_name = "baseplate.client"
+        self.tags = {}  # type: Dict[str, str]
+        self.timer = batch.timer(self.base_name)
+        self.whitelist = whitelist
+
+    def on_start(self) -> None:
+        self.timer.start()
+        self.span.set_tag("endpoint", self.span.name.split(".")[1])
```
spladug, line 162: and here

vanessawuhoo:
@spladug 💇 and also pulled out logic for the whitelisting check so we can use it outside of on_set_tag

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+        self.tags: Dict[str, str] = {}
+        self.timer = batch.timer(self.base_name)
+
+    def on_start(self) -> None:
+        if _is_whitelisted("endpoint", self.whitelist):
+            self.tags["endpoint"] = self.span.name
+        self.timer.start()
+
+    def on_incr_tag(self, key: str, delta: float) -> None:
+        self.batch.counter(key, self.tags).increment(delta, True)
```
spladug, line 93: what's the True here?

baseplate/lib/metrics.py
```diff
@@ -313,21 +313,19 @@ def send(self, elapsed: float, sample_rate: float = 1.0) -> None:
...
             sampling_info = f"@{sample_rate:g}".encode()
-            parts.append(sampling_info)
-        serialized = b"|".join(parts)
+            serialized = b"|".join([serialized, sampling_info])
         self.transport.send(serialized)
 
     def update_tags(self, tags: Dict) -> None:
         assert not self.stopped
-        self.tags.update(tags)
+        self.tags = tags
```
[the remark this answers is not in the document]
vanessawuhoo: I think I wrote this under the assumption that the tags were being kept in parallel on the span, so any updates could just completely overwrite the existing tags. However, this may not work for users and other use cases so I'll change it back

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+error_messages = {
+    Errors.TIMED_OUT: "timed_out",
+    Errors.EXCEPTION: "internal_server_error",
+}
+
+
+def _is_whitelisted(key: str, whitelist: List[str]) -> bool:
+    return key in whitelist
```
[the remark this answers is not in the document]
vanessawuhoo: I was of two minds about it 😅it just makes it a little more readable imo. I don't feel strongly about keeping or getting rid of it

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, batch: metrics.Batch, server_span: Span, whitelist: list):
+        self.batch = batch
+        self.span = server_span
+        self.base_name = "baseplate.server"
+        self.whitelist = whitelist
+        self.tags: Dict[str, str] = {}
+        self.timer = batch.timer(self.base_name)
+
+    def on_start(self) -> None:
+        if _is_whitelisted("endpoint", self.whitelist):
```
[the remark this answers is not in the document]
vanessawuhoo: that sounds good and was not something I considered! will do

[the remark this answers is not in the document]
vanessawuhoo: The only issue is making sure unwhitelisted tags are not put on the timer- the timer doesn't know about the whitelist.
So we can either pass the whitelist to the entities like the timer, counter, etc and let them handle filtering themselves, or only set tags on the entitites at the very end when we know the span's tags have stopped changing, or something else I haven't thought of yet

baseplate/observers/metrics_tagged.py
vanessawuhoo: this is a remainder from when I was using increment(delta, tagged=true), will remove!

vanessawuhoo:
💇‍♀️ @spladug 
- Rewrote to filter tags at end and send counters, update timer tags in on_finish()
- Added documentation (draft)
- Fixed up tests

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+                    "whitelist": config.Optional(
+                        config.TupleOf(config.String),
+                        default=["client", "endpoint", "success", "error"],
+                    ),
+                }
+            },
+        )
+        return cls(client, whitelist=cfg.metrics.whitelist)
+
+    def on_server_span_created(self, context: RequestContext, server_span: Span) -> None:
```
spladug, line 68: we don't need to block merging this in, but before this gets into a release we'll want to retrofit the sampling changes to the metrics observer e.g. https://github.com/reddit/baseplate.py/blob/master/baseplate/observers/metrics.py#L49

baseplate/lib/metrics.py
```diff
@@ -62,7 +62,17 @@
...
 def _metric_join(*nodes: bytes) -> bytes:
-    return b".".join(node.strip(b".") for node in nodes)
+    return b".".join(node.strip(b".") for node in nodes if node)
+
+
+def _format_tags(tags: Optional[Dict]) -> Optional[bytes]:
+    if not tags:
+        return None
+
+    parts = [""]
```
pacejackson, line 72: 🔕  does this start with an empty string so there's a `","` at the beginning of the string?  It might be clearer just to return `b"," + ",".join(parts).encode()`.  It would be good to have a comment explaining this if this is the approach we take so it's clear what's happening.

baseplate/lib/metrics.py
```diff
@@ -135,16 +145,16 @@ def __init__(self, transport: Transport, namespace: str):
         self.transport = transport
         self.namespace = namespace.encode("ascii")
 
-    def timer(self, name: str) -> "Timer":
+    def timer(self, name: str, tags: Optional[Dict] = None) -> "Timer":
```
pacejackson, line 148: Should the tag dict be `Dict[str, Any]`?  If so' it might be good to use that and define it as `TagDict = Dict[str, Any]`.

Alternatively, would keyword args make sense?

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+
+class Errors(Enum):
+    TIMED_OUT = 1
+    EXCEPTION = 2
+
+
+error_messages = {
+    Errors.TIMED_OUT: "timed_out",
+    Errors.EXCEPTION: "internal_server_error",
+}
```
pacejackson, line 26: Note, you can define enums as:

```python
class Errors(Enum):
    TIMED_OUT = "timed_out"
    EXCEPTION = "internal_server_error"
```

and use `err_enum.value` to get the string value.

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def on_server_span_created(self, context: RequestContext, server_span: Span) -> None:
+        batch = self.client.batch()
+        context.metrics = batch
+        observer = TaggedMetricsServerSpanObserver(batch, server_span, self.whitelist)
+        server_span.register(observer)
+
+
+class TaggedMetricsServerSpanObserver(SpanObserver):
+    def __init__(self, batch: metrics.Batch, server_span: Span, whitelist: list):
```
pacejackson, line 76: `List[str]` (which you import from typing), not `list`.

```suggestion
    def __init__(self, batch: metrics.Batch, server_span: Span, whitelist: List[str]):
```

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+        self.tags = filtered_tags
+        for key, delta in self.counters.items():
+            self.batch.counter(key, self.tags).increment(delta)
+        self.timer.update_tags(self.tags)
+        self.timer.stop()
+        self.batch.flush()
+
+
+class TaggedMetricsLocalSpanObserver(SpanObserver):
+    def __init__(self, batch: metrics.Batch, span: Span, whitelist: list):
```
pacejackson, line 127:
```suggestion
    def __init__(self, batch: metrics.Batch, span: Span, whitelist: List[str]):
```

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+        for key, delta in self.counters.items():
+            self.batch.counter(key, self.tags).increment(delta)
+        self.tags = filtered_tags
+        self.timer.update_tags(self.tags)
+        self.timer.stop()
+        self.batch.flush()
+
+
+class TaggedMetricsClientSpanObserver(SpanObserver):
+    def __init__(self, batch: metrics.Batch, span: Span, whitelist: list):
```
pacejackson, line 160:
```suggestion
    def __init__(self, batch: metrics.Batch, span: Span, whitelist: List[str]):
```

baseplate/observers/metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+    ``metrics`` attribute on the :py:class:`~baseplate.RequestContext`.
+
+    :param client: The client where metrics will be sent.
+    :param cfg: the parsed application config with the tag whitelist
+
+    """
+
+    def __init__(self, client: metrics.Client, whitelist: List[str]):
+        self.client = client
+        self.whitelist = whitelist
```
pacejackson, line 49: Should we convert `whitelist` to a `set` since we're just checking `x in whitelist`?

baseplate/lib/metrics.py
vanessawuhoo: That is indeed why, I wasn't sure which was cleaner

baseplate/lib/metrics.py
vanessawuhoo: I think Dict[str, Any] is fine, but it does have to stay Optional

pacejackson: If you want to default to an empty value, yes, because you want to use `None`.  You don't want to use an empty dict as a default value (or any mutable container).  The reason being that "default" value is a global value so if you mutate it, that mutation applies to the "default" so anytime you use it after that, it has those values set.

vanessawuhoo: Actually, I think it should be Dict[str,str], unless you think we should be able to accept booleans as values?

vanessawuhoo:
@spladug @pacejackson 💇 
- retrofitted observer with sample_rate changes
- changed dict to [str, Any]
- changed whitelist to a Set for faster performance
- doc updates 
- simplified enum definition

baseplate/lib/metrics.py
```diff
@@ -424,17 +448,30 @@ class Histogram:
...
         """
-        serialized = self.name + (f":{value:g}|h".encode())
+
+        formatted_tags = _format_tags(self.tags)
+        if formatted_tags:
+            parts = [self.name + formatted_tags + (f":{value:g}|h".encode())]
+        else:
+            parts = [self.name + (f":{value:g}|h".encode())]
+
+        serialized = b"|".join(parts)
```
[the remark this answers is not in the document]
vanessawuhoo: mm you are correct! this is leftover from when we were doing the dogstatsd format

vanessawuhoo:
💇 @pacejackson @spladug 
- included sample rate in tagged metrics docs
- linked to tagged metrics docs from index.rst
- simplified code where joining and putting in array was redundant

docs/api/baseplate/observers/tagged_statsd.rst
```diff
@@ -32,6 +32,11 @@ the service.
    # defaults to [ endpoint, success, error ] if no value is provided
    metrics.whitelist = endpoint, success, error
 
+    # optional: the percent of statsd metrics to sample
```
pacejackson, line 35: looks like this has an extra space in front of it?

PR #419, opened 2020-03-31
Add convenience function for creating quick server span contexts

spladug:
This pattern comes up pretty frequently:

    context = baseplate.make_context_object()
    with baseplate.make_server_span(context, "foo"):
        context.foo.bar()

and while we do need the context object creation and span creation to be separate for different frameworks to have full control of things we can also have some nice sugar for the common case:

    with baseplate.server_context("foo") as context:
        context.foo.bar()

:eyeglasses: @melissacole @pacejackson

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -2,10 +2,12 @@ import inspect
 import logging
 import random
 
+from contextlib import contextmanager
 from types import TracebackType
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import Iterator
 from typing import List
 from typing import NamedTuple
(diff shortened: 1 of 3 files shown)


spladug:
I don't really see anywhere inside baseplate that this would be that helpful. All of the places I saw are also accessing the span object (setting tags, making child spans, etc.) so it'll just trade the awkward. Applications on the other hand rarely set tags.

MelissaCole:
:fish: beautiful :) thanks!

pacejackson:
Oh, I'm a dork, I get the difference.

PR #421, opened 2020-04-01
Handle ServerTimeouts while connecting to a thrift service

m104:
This update wraps the `ThriftConnectionPool._acquire()` call within a
try/finally block to ensure that the connection slot is returned to the
pool in case a ServerTimeout is raised.


m104:
👓 @spladug @diffyqgirl

PR #422, opened 2020-04-01
Send and receive service names in Thrift

spladug:
This gives services an identity and sends it to downstreams in Thrift
calls. Thrift servers also accept this identity and attach it to the
server span as the peer.service tag.

The idea is to allow servers to get better information about who is
calling them.

### DIFF
--- a/baseplate/lib/__init__.py
+++ b/baseplate/lib/__init__.py
@@ -1,5 +1,6 @@
 """Internal library helpers."""
 import functools
+import inspect
 import warnings
 
 from typing import Callable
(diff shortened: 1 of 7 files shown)


spladug:
:eyeglasses: @bradengroom

baseplate/lib/__init__.py
```diff
@@ -45,3 +46,15 @@ def __get__(self, instance: T, owner: Type[T]) -> R:
...
         return ret
+
+
+class UnknownCallerError(Exception):
+    def __init__(self) -> None:
+        super().__init__("Could not determine calling module's name")
+
+
+def get_calling_module_name() -> str:
+    module = inspect.getmodule(inspect.stack()[2].frame)
```
pacejackson, line 57: 🔕 is there any way this can raise an `IndexError`?

spladug: I don't want to say know cus I can't say for sure... but I hope not?

PR #425, opened 2020-04-24
Make the queue consumer HTTP healthcheck return JSON

spladug:
This brings it in line with the baseplate healthcheck protocol and
allows `baseplate-healthcheck` to work with it.


spladug:
now with a test to enforce the correctness of the JSON response

PR #427, opened 2020-05-01
Add info fields to TTransportException

foreverest:
Something like this would be useful. This is not the full PR but just a proposal to discuss.

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -21,6 +21,14 @@ from baseplate.lib.thrift_pool import thrift_pool_from_config
 from baseplate.lib.thrift_pool import ThriftConnectionPool
 
 
+class ThriftTransportException(TTransportException):
+    def __init__(self, namespace: str, name: str, last_error: Optional[str], **kwargs: Any):
+        super().__init__(**kwargs)
+        self.namespace = namespace
+        self.name = name
+        self.last_error = last_error
+
+
 class ThriftClient(config.Parser):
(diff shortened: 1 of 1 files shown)


spladug:
hiya, sorry for the delay.  I just saw this. (Had some poorly configured email settings for GH.com -- all fixed)

Just for my understanding, what's your intended use case for the extra fields on the exceptions?

foreverest:
In one of the services, we found it useful to send information about these exceptions to Wavefront instead of letting them spam Sentry. There are three "sub-types" of the exception (based on `last_error`): timeout, pool exhausted, and upstream exception. In a Wavefront graph, we see a separate chart for each of the sub-types per service (`namespace`) per method ('name'). Currently, we have to parse the message text to extract that information. I thought it might be useful to preserve the three pieces of information in the exception instance.

spladug:
got it! would it make sense to add individual subtypes for the different error cases? some come from the thrift pool and some from the client itself so there'd have to be two groups but that's not too crazy. makes it really easy to add them in try/except branches then?

foreverest:
Yeah, and then we just re-raise `last_error` [here](https://github.com/reddit/baseplate.py/pull/427/files#diff-0e434287ad163ea2c27ebbc59e86c8d5R211), right? That would work too. But note that we'd still need to add the `namespace` and `name` fields to the exception just before re-raising.
(Sorry for delaying the response. Just got to it.)

foreverest:
There is no great need for this.

PR #429, opened 2020-05-13
Raises warning instead of exception on empty Kafka message

paulkiernan:
Skips attempting to deserialize an empty Kafka message and raises a suppressible warning instead. This is useful for cases such as consuming CDC row updates from kSQL where keyed messages with a null value indicate no change to the row.

### DIFF
--- a/baseplate/frameworks/queue_consumer/kafka.py
+++ b/baseplate/frameworks/queue_consumer/kafka.py
@@ -3,7 +3,6 @@ import logging
 import queue
 import socket
 import time
-import warnings
 
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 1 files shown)


baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -131,6 +132,10 @@ def handle(self, message: confluent_kafka.Message) -> None:
 
                 blob: bytes = message.value()
 
+                if not blob:
+                    warnings.warn("empty message body", BytesWarning)
```
spladug, line 136: `warnings.warn` will generate a warning which is separate from logging and only show up in the logs in debug mode and only the first time this ever happens, not subsequent ones. Is that intentional? Alternatively you could use `logger.warning()` which would generate a message at the WARNING level in the logs on every time this is hit.

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -130,25 +130,32 @@ def handle(self, message: confluent_kafka.Message) -> None:
...
-                    # we can't guarantee that all publishers populate this field
-                    # v2 events publishers (event collectors) do, but future
-                    # kafka publishers may not
-                    message_latency = None
-
-                self.handler_fn(context, data, message)
+                data: Any = None
+                message_latency: Optional[int] = None
+
+                if not blob:
```
spladug, line 136: as a stylistic nit, I'd generally prefer to put the normal case first if we're branching the whole thing. Makes it easier to read through the happy path.

spladug:
hey, I'm going to close this out for now since it sounds like we should let the application do this. Please feel free to re-open if you think we should do it here or something else has changed!

PR #431, opened 2020-05-22
Thrift pool overflow #1

foreverest:
When an exception (other than `queue.Empty`) happens in line 180,
https://github.com/reddit/baseplate.py/blob/887f9be99f367be27d0e58fecdb274cb090a1585/baseplate/lib/thrift_pool.py#L179-L184
it results in a call of
https://github.com/reddit/baseplate.py/blob/887f9be99f367be27d0e58fecdb274cb090a1585/baseplate/lib/thrift_pool.py#L267-L268
which puts an excess item in the queue (above `self.size`).

One of the subsequences of it is:
```
  File "/usr/local/lib/python3.7/dist-packages/baseplate/server/runtime_monitor.py", line 146, in report
    reporter(batch)
  File "/usr/local/lib/python3.7/dist-packages/baseplate/clients/thrift.py", line 86, in report_runtime_metrics
    batch.gauge("pool.in_use").replace(self.pool.checkedout)
  File "/usr/local/lib/python3.7/dist-packages/baseplate/lib/metrics.py", line 467, in replace
    assert new_value >= 0, "gauges cannot be replaced with negative numbers"
```

This PR proposes a fix for the issue. @m104 @spladug please take a look.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -142,9 +142,6 @@ class ThriftConnectionPool:
 
     """
 
-    class _PoolCheckoutException(Exception):
-        pass
-
     # pylint: disable=too-many-arguments
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


baseplate/lib/thrift_pool.py
```diff
@@ -257,6 +266,10 @@ def connection(self) -> Generator[TProtocolBase, None, None]:
             # (expected) errors which should be safe for the connection.
             # don't close the transport here!
             raise
+        except self._PoolCheckoutException:
```
spladug, line 269: Won't this swallow something like a `ServerTimeoutError`? That'd effectively make timeouts that happen during this phase never affect that request? One option would be to unwrap the original exception from the PoolCheckoutException. Another would be to decompose the two things and handle their error cases separately.

baseplate/lib/thrift_pool.py
```diff
@@ -142,6 +142,9 @@ class ThriftConnectionPool:
 
     """
 
+    class _PoolCheckoutException(Exception):
```
spladug, line 145: Can this be at the module level so we don't have to access it with cls/self?

baseplate/lib/thrift_pool.py
foreverest: 👍

baseplate/lib/thrift_pool.py
foreverest: I removed it altogether

spladug:
@m104 would you mind taking another look? this is sufficiently subtle stuff that i'd love another pair of eyes (sorry for the un-harolded ping -- github's not doing anything when I click :recycle:)

PR #432, opened 2020-05-26
Preserve previous behavior of deprecated queue consumer

pacejackson:
We used to catch the queue.Empty error and would just 'break' when that happened in get_batch

### DIFF
--- a/baseplate/frameworks/queue_consumer/deprecated.py
+++ b/baseplate/frameworks/queue_consumer/deprecated.py
@@ -148,17 +148,21 @@ class BaseKombuConsumer:
             block = False
         else:
             block = True
-
         batch = []
         retry_policy = RetryPolicy.new(attempts=max_items, budget=timeout)
         for time_remaining in retry_policy:
-            item = self.work_queue.get(block=block, timeout=time_remaining)
+            item = self._get_next_item(block=block, timeout=time_remaining)
             if item is None:
                 break
-
             batch.append(item)
 
         return batch
+
+    def _get_next_item(self, block: bool, timeout: Optional[float]) -> Optional[Message]:
+        try:
+            return self.work_queue.get(block=block, timeout=timeout)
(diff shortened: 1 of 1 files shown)


pacejackson:
How this used to work, for reference:

https://github.com/reddit/baseplate.py/commit/5b8054ed4e952d173076f2c157e0272eb248ad52#diff-bb850ef72799fd806255bcc3eddb7dc6L91-L107

pacejackson:
cc: @mimichen226

baseplate/frameworks/queue_consumer/deprecated.py
```diff
@@ -148,18 +148,22 @@ def get_batch(self, max_items: int, timeout: Optional[float]) -> Sequence[Messag
...
         else:
             block = True
-
         batch = []
         retry_policy = RetryPolicy.new(attempts=max_items, budget=timeout)
         for time_remaining in retry_policy:
-            item = self.work_queue.get(block=block, timeout=time_remaining)
+            item = self._get_next_item(block=block, timeout=time_remaining)
             if item is None:
                 break
```
spladug, line 156: I'm not really sure why we introduce a new function here when it's pretty clear inlined:

```suggestion
            try:
                item = self.work_queue.get(block=block, timeout=time_remaining)
            except queue.Empty:
                break
```
```

pacejackson: So, that’s what I did originally but the thing about this code it it’s _not_ actually the same behavior as it was originally. 

One of the tests started failing because it was relying on the old behavior, which is checking `item is None`.  It’s probably debatable whether that’s even the behavior we actually want but for something like this, I erred on the side of just preserving the old behavior.

spladug: yuck. ok, thanks for the explanation.

PR #433, opened 2020-05-29
Fix how we return loid of an user

fishy:
We should return LoID in this order (fallback to the next one if it's
empty):

1. The logged in user id from the JWT token
2. The loid from the thrift payload
3. The loid from the JWT token

### DIFF
--- a/baseplate/lib/edge_context.py
+++ b/baseplate/lib/edge_context.py
@@ -167,7 +167,7 @@ class User(NamedTuple):
     """Wrapper for the user values in AuthenticationToken and the LoId cookie."""
 
     authentication_token: AuthenticationToken
-    loid: str
+    loid_: str
     cookie_created_ms: int
 
     @property
@@ -215,16 +215,36 @@ class User(NamedTuple):
 
     def event_fields(self) -> Dict[str, Any]:
         """Return fields to be added to events."""
-        if self.is_logged_in:
-            user_id = self.id
-        else:
-            user_id = self.loid
+        loid: Optional[str] = self.loid
+        if loid == "":
+            loid = None
 
         return {
-            "user_id": user_id,
+            "user_id": loid,
             "logged_in": self.is_logged_in,
             "cookie_created_timestamp": self.cookie_created_ms,
         }
+
+    @property
+    def loid(self) -> str:
+        # First, if it's logged in user, return logged in user id.
+        try:
+            user_id = self.id
+            if user_id is not None:
+                return user_id
+        except NoAuthenticationError:
+            pass
+
+        # Next, return the loid from thrift payload if it's non-empty
+        if self.loid_:
+            return self.loid_
+
+        # Finally, return loid from authentication token
+        loid = self.authentication_token.loid
+        if loid:
+            return loid
+
+        return ""
 
 
 class OAuthClient(NamedTuple):
@@ -433,7 +453,7 @@ class EdgeRequestContext:
         """:py:class:`~baseplate.lib.edge_context.User` object for the current context."""
         return User(
             authentication_token=self.authentication_token,
-            loid=self._t_request.loid.id,
+            loid_=self._t_request.loid.id,
             cookie_created_ms=self._t_request.loid.created_ms,
         )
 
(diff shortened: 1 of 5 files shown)


fishy:
👓 @spladug @pacejackson

baseplate/lib/edge_context.py
```diff
@@ -167,7 +167,7 @@ class User(NamedTuple):
     """Wrapper for the user values in AuthenticationToken and the LoId cookie."""
 
     authentication_token: AuthenticationToken
-    loid: str
+    loid_: str
```
pacejackson, line 170:
```suggestion
    _loid: str
```

baseplate/lib/edge_context.py
```diff
@@ -215,17 +215,37 @@ def has_role(self, role: str) -> bool:
...
+            if user_id is not None:
+                return user_id
+        except NoAuthenticationError:
+            pass
+
+        # Next, return the loid from thrift payload if it's non-empty
+        if self.loid_ != "":
+            return self.loid_
+
+        # Finally, return loid from authenticatino token
```
pacejackson, line 242:
```suggestion
        # Finally, return loid from authentication token
```

baseplate/lib/edge_context.py
fishy: It's not allowed in NamedTuple.

baseplate/lib/edge_context.py
fishy: Done,

baseplate/lib/edge_context.py
pacejackson: :til:

tests/unit/lib/experiments/experiment_tests.py
```diff
@@ -34,7 +34,7 @@ def setUp(self):
         self.mock_authentication_token.user_roles = set()
         self.user = User(
             authentication_token=self.mock_authentication_token,
-            loid="t2_1",
+            loid_="t2_1",
```
spladug, line 37: Depending on how public we feel this API is this is technically a breaking change. That said, it doesn't look like the constructor's parameters are publicly documented and I don't see any usage of this in our services in a quick github search so... :see_no_evil:

baseplate/lib/edge_context.py
```diff
@@ -215,17 +215,37 @@ def has_role(self, role: str) -> bool:
...
+        # First, if it's logged in user, return logged in user id.
+        try:
+            user_id = self.id
+            if user_id is not None:
+                return user_id
+        except NoAuthenticationError:
+            pass
+
+        # Next, return the loid from thrift payload if it's non-empty
+        if self.loid_ != "":
```
spladug, line 239: maybe

```suggestion
        if self.loid_:
```
?

technically None != ""

fishy:
> what's the impact on applications with this change? how should we document this?

For that part, I plan to have a meeting with experiment council next week to understand more.

baseplate/lib/edge_context.py
fishy: it's mypy complaining, actually: https://cloud.drone.io/reddit/baseplate.py/904/1/2

not sure if there's a way to override that behavior in mypy.

spladug: https://docs.python.org/3/library/collections.html#collections.namedtuple

> Any valid Python identifier may be used for a fieldname except for names starting with an underscore.

so mypy's just enforcing the docs for namedtuple

spladug:
just to follow up here: was there ever a decision from experiments council?

fishy:
> just to follow up here: was there ever a decision from experiments council?

experiments council is fine merging this. but doing it now has a potential to break some other thing. we plan to fix the other thing towards the end of Q3, after that's fixed we could merge this PR.

spladug:
ok, thanks for the update! i'll pop this in the milestone for v1.5 (which would be released the first week of december)

spladug:
@fishy should this still be on track for v1.5 (releasing Nov 30) ?

fishy:
> @fishy should this still be on track for v1.5 (releasing Nov 30) ?

Unfortunately no. New ETA is sometime Q1 2021 :(

spladug:
Understood. Thanks for the update. I've bumped it to 2.0's milestone. Let me know if there's anything I can do to help.

fishy:
I think we can close this PR in favor of https://github.com/reddit/edgecontext.py/pull/2 now?

spladug:
Sounds good, thanks!

PR #446, opened 2020-06-04
Allow disabling server timeouts and disable them by default

spladug:
They don't make sense in all situations (e.g. cron jobs) and the ideal
is that we explicitly configure them in the future but in the interest
of compatibility we'll just emit a deprecation warning for now.

### DIFF
--- a/baseplate/observers/timeout.py
+++ b/baseplate/observers/timeout.py
@@ -1,5 +1,3 @@
-import datetime
-
 from typing import Optional
 
 import gevent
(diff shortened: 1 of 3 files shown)


baseplate/observers/timeout.py
```diff
@@ -30,28 +29,37 @@ def from_config(cls, app_config: config.RawConfig) -> "TimeoutBaseplateObserver"
             app_config,
             {
                 "server_timeout": {
-                    "default": config.Optional(
-                        config.Timespan, default=datetime.timedelta(seconds=10)
-                    ),
+                    "default": config.Optional(config.TimespanOrInfinite, default=None),
```
pacejackson, line 32: should this default to `"infinite"`?

spladug: no, so the deprecation warning below can trigger

pacejackson: ahh, makes sense

PR #447, opened 2020-06-06
Thrift pool overflow #2

foreverest:
This PR solves another instance of the issue described in #431 (i.e. excess items get into the pool, `thrift_pool.checkedout` returns a negative value).

This time the issue happens when the retry policy exhausted trying to connect to a service. In this case, `self.pool.put(None)` gets called twice: in both `_acquire` and `_release` methods.

Note: I removed two assertions from the `test_max_retry_on_connect` test as they no longer make sense (and fail).

@spladug @m104 Please take a look.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -205,8 +205,6 @@ class ThriftConnectionPool:
 
             return prot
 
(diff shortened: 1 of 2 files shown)


spladug:
Thanks!

Issue #285, opened 2019-05-20
Wrap HTTP X-Edge-Request payload in base64


spladug:
We haven't ever used this, but if we did it'd probably break fast on null bytes etc.

PR #449, opened 2020-06-09
Decode edge payload before passing to edge context factory

madepsilon:
👓  @pacejackson

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -190,14 +190,14 @@ class BaseplateConfigurator:
         if self.header_trust_handler.should_trust_edge_context_payload(request):
             try:
                 edge_payload = request.headers.get("X-Edge-Request", None)
-                edge_payload_decoded = base64.urlsafe_b64decode(edge_payload)
-                if self.edge_context_factory:
+                if self.edge_context_factory and edge_payload:
+                    edge_payload_decoded = base64.b64decode(edge_payload)
                     edge_context = self.edge_context_factory.from_upstream(edge_payload_decoded)
                     edge_context.attach_context(request)
                 else:
                     # just attach the raw context so it gets passed on
                     # downstream even if we don't know how to handle it.
-                    request.raw_request_context = edge_payload_decoded
+                    request.raw_request_context = edge_payload
             except (KeyError, ValueError):
                 pass
 
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -190,10 +190,9 @@ class BaseplateConfigurator:
         if self.header_trust_handler.should_trust_edge_context_payload(request):
             try:
                 edge_payload = request.headers.get("X-Edge-Request", None)
-                if edge_payload:
-                    edge_payload = base64.b64decode(edge_payload)
-                if self.edge_context_factory:
-                    edge_context = self.edge_context_factory.from_upstream(edge_payload)
+                if self.edge_context_factory and edge_payload:
+                    edge_payload_decoded = base64.b64decode(edge_payload)
+                    edge_context = self.edge_context_factory.from_upstream(edge_payload_decoded)
                     edge_context.attach_context(request)
                 else:
                     # just attach the raw context so it gets passed on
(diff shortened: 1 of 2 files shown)


spladug:
Oh cool! This fixes #285. Thank you!

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -189,13 +190,14 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         if self.header_trust_handler.should_trust_edge_context_payload(request):
             try:
                 edge_payload = request.headers.get("X-Edge-Request", None)
+                edge_payload_decoded = base64.urlsafe_b64decode(edge_payload)
```
pacejackson, line 193: This will throw a `TypeError` if `edge_payload` is `None`.  We might want to change this to:

```python
edge_payload = request.headers.get("X-Edge-Request", None)
if self.edge_context_factory and edge_payload is not None:
    edge_payload_decoded = base64.urlsafe_b64decode(edge_payload)
    edge_context = self.edge_context_factory.from_upstream(edge_payload_decoded)
    edge_context.attach_context(request)
```

spladug:
The baseplate spec says 

> For HTTP, the header is named X-Edge-Request and the payload is sanitized using Base 64 encoding.

does baseplate.go use any particular base64 encoding already? should we update the spec to say URL-safe base64? or change this to use standard base64?

pacejackson:
Ah, oops it's actually standard b64 encoding, _not_ urlsafe:

https://github.com/reddit/baseplate.go/blob/b78f1bd84153a0d0f3a97dfe30fd90c90861e285/httpbp/headers.go#L69-L84

So the spec is right and this should be changed.

baseplate/frameworks/pyramid/__init__.py
pacejackson: Also, as neil noted, this should be `b64decode`, not `urlsafe_`

madepsilon:
💇

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -189,8 +190,9 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         if self.header_trust_handler.should_trust_edge_context_payload(request):
             try:
                 edge_payload = request.headers.get("X-Edge-Request", None)
-                if self.edge_context_factory:
-                    edge_context = self.edge_context_factory.from_upstream(edge_payload)
+                if self.edge_context_factory and edge_payload:
+                    edge_payload_decoded = base64.b64decode(edge_payload)
```
pacejackson, line 194: I think this is going to be a `string` and `b64decode` wants `bytes`, so you'll want to `.encode()` edge payload here.

tests/integration/pyramid_tests.py
```diff
@@ -139,6 +144,7 @@ def test_trace_headers(self):
             "/example",
             headers={
                 "X-Trace": "1234",
+                "X-Edge-Request": base64.b64encode(SERIALIZED_EDGECONTEXT_WITH_NO_AUTH),
```
pacejackson, line 147: This should also be `base64.b64encode(SERIALIZED_EDGECONTEXT_WITH_NO_AUTH).decode()` so it's the type that the code expects.

madepsilon:
💇

tests/integration/pyramid_tests.py
```diff
@@ -20,7 +21,11 @@
 except ImportError:
     raise unittest.SkipTest("pyramid/webtest is not installed")
 
-from .. import AUTH_TOKEN_PUBLIC_KEY, SERIALIZED_EDGECONTEXT_WITH_VALID_AUTH
+from .. import (
+    AUTH_TOKEN_PUBLIC_KEY,
+    SERIALIZED_EDGECONTEXT_WITH_VALID_AUTH,
+    SERIALIZED_EDGECONTEXT_WITH_NO_AUTH,
+)
```
pacejackson, line 28: 🔕  I _thiiiink_ we might want to do this as multiple imports?  i.e.

```python
from .. import AUTH_TOKEN_PUBLIC_KEY
from .. import SERIALIZED_EDGECONTEXT_WITH_VALID_AUTH
from .. import SERIALIZED_EDGECONTEXT_WITH_NO_AUTH
```

madepsilon: just curious, what's the difference? I just do what the linter tells me to  :sob:

spladug: this is why the linter wants separate lines: https://github.com/asottile/reorder_python_imports#why-this-style

pacejackson: In practice, not much, it's just what we do earlier in the file.  I thought `reorder-python-imports` forced this but maybe not for this style of import `from ... import`?

PR #450, opened 2020-06-10
Updates the event publisher to not exit on validation errors

paulkiernan:
This PR updates the event publisher sidecar to print, but not exit on `HTTPUnprocessableEntity` responses from a downstream event collector. This status code is returned when an event cannot be validated by the collector service.

### DIFF
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -128,9 +128,11 @@ class BatchPublisher:
 
                 # we should crash if it's our fault
                 response = getattr(exc, "response", None)
-                if response is not None and response.status_code < 500:
+                if response and response.status_code < 500:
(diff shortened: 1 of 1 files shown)


baseplate/sidecars/event_publisher.py
```diff
@@ -128,10 +128,13 @@ def publish(self, payload: SerializedBatch) -> None:
 
                 # we should crash if it's our fault
                 response = getattr(exc, "response", None)
-                if response is not None and response.status_code < 500:
+                if response and response.status_code < 500:
```
spladug, line 131: `Response` has a `__bool__` definition that "Returns True if :attr:`status_code` is less than 400.". That's why we were checking `is not None` instead of truthiness earlier. I think that still needs to be around so this can ever be true for 4xx errors.

PR #456, opened 2020-06-17
pyramid: encode edge_payload before setting request.raw_edge_payload

madepsilon:
👓  @pacejackson

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -188,21 +188,20 @@ class BaseplateConfigurator:
                 pass
 
         if self.header_trust_handler.should_trust_edge_context_payload(request):
+            edge_payload: Optional[bytes]
             try:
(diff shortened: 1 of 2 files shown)


pacejackson:
@madepsilon I don't think the commit message really encapsulates this change or why we're making it.  

We were already checking if edge_payload was `None` before encoding, the problem was we were setting `request.raw_edge_request` to the value we got from the header without encoding it, which would break when you tried to set it as a thrift header since _that_ expects `bytes`.

Maybe something along the lines of:

`pyramid: encode edge_payload before setting request.raw_edge_payload`?

pacejackson:
_actually_ I think the issue is a little more complex.  `request.raw_request_context = edge_payload` isn't right anymore, we _have_ to try to b64decode _before_ setting that, I think we need to do that sooner and edge_payload needs to fall back to `None` if it can't be decoded?

PR #461, opened 2020-07-01
Add function for converting milliseconds timestamp to datetime

bradengroom:
This adds an `epoch_milliseconds_to_datetime` function that is intended to be used with the thrift TimestampMilliseconds typedef.

### DIFF
--- a/baseplate/lib/datetime.py
+++ b/baseplate/lib/datetime.py
@@ -15,6 +15,13 @@ def datetime_to_epoch_seconds(dt: datetime) -> int:
     return datetime_to_epoch_milliseconds(dt) // 1000
 
 
+def epoch_milliseconds_to_datetime(ms: int) -> datetime:
+    """Convert epoch milliseconds to UTC datetime."""
+    return datetime.utcfromtimestamp(ms // 1000).replace(
+        microsecond=ms % 1000 * 1000, tzinfo=timezone.utc
+    )
+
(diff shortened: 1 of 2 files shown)


baseplate/lib/datetime.py
```diff
@@ -15,6 +15,13 @@ def datetime_to_epoch_seconds(dt: datetime) -> int:
     return datetime_to_epoch_milliseconds(dt) // 1000
 
 
+def epoch_milliseconds_to_datetime(ms: int) -> datetime:
+    """Convert epoch milliseconds to UTC datetime."""
+    return datetime.utcfromtimestamp(ms // 1000).replace(
```
spladug: what's the advantage of truncating to whole seconds then doing the millisecond->microsecond conversion ourselves? it looks like it works if you just do a float division:

```python
>>> from datetime import datetime
>>> ms = 1594057784005
>>> datetime.utcfromtimestamp(ms // 1000).replace(microsecond=ms % 1000 * 1000)
datetime.datetime(2020, 7, 6, 17, 49, 44, 5000)
>>> datetime.utcfromtimestamp(ms / 1000)
datetime.datetime(2020, 7, 6, 17, 49, 44, 5000)
```

bradengroom: There's no advantage. We should do it the way you proposed.

PR #462, opened 2020-07-06
Fix typo in thrift comment

fishy:
🤦

Also slightly tweak the wording as we could use either string or int
values in the http health checks.


fishy:
👓 @spladug

PR #465, opened 2020-07-13
Add requests-based HTTP clients

### LINKED ISSUE #448: Add HTTP client

spladug:
This adds two HTTP clients: 

* The internal client is for use with first-party services and propagates trace and edge context. 
* The external client is for use with third-party services and does not propagate the extra context.

We use @jordanmilne's Advocate library (now updated for modern requests!) to enforce that the internal client only talks to local addresses and the external client only talks to external addresses. The cost of doing this is about 100 microseconds per request.

Fixes #448

### DIFF
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -1,0 +1,322 @@
+import base64
+import ipaddress
+
+from typing import Any
+from typing import Type
+from typing import Union
+
+from advocate import AddrValidator
+from advocate import ValidatingHTTPAdapter
+from requests import PreparedRequest
+from requests import Request
+from requests import Response
+from requests import Session
+from requests.adapters import HTTPAdapter
+
+from baseplate import Span
+from baseplate.clients import ContextFactory
+from baseplate.lib import config
+
+
+def http_adapter_from_config(
+    app_config: config.RawConfig, prefix: str, **kwargs: Any
+) -> HTTPAdapter:
+    """Make an HTTPAdapter from a configuration dictionary.
+
+    The keys useful to :py:func:`http_adapter_from_config` should be prefixed,
+    e.g. ``http.pool_connections``, ``http.max_retries``, etc. The ``prefix``
+    argument specifies the prefix used. Each key is mapped to a corresponding
+    keyword argument on the :py:class:`~requests.adapters.HTTPAdapter`
+    constructor.
+
+    Supported keys:
+
+    * ``pool_connections``: The number of connections to cache (default: 10).
+    * ``pool_maxsize``: The maximum number of connections to keep in the pool
+      (default: 10).
+    * ``max_retries``: How many times to retry DNS lookups or connection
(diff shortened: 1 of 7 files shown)


spladug:
cc (just FYI) @pacejackson

baseplate/clients/requests.py
```diff
@@ -0,0 +1,322 @@
...
+            span.set_tag("http.method", request.method)
+            span.set_tag("http.url", request.url)
+
+            # we cannot re-use the same session every time because sessions re-use the same
+            # CookieJar and so we'd muddle cookies cross-request. if the application wants
+            # to keep track of cookies, it should do so itself.
+            #
+            # note: we're still getting connection pooling because we're re-using the adapter.
+            session = Session()
+            session.mount("http://", self.adapter)
```
MelissaCole, line 204: :no_bell: both of these are needed so it can support sending `http://` requests?

baseplate/clients/requests.py
```diff
@@ -0,0 +1,322 @@
...
+            request.headers["X-Sampled"] = "1"
+        if self.span.flags is not None:
+            request.headers["X-Flags"] = str(self.span.flags)
+
+        try:
+            edge_context = self.span.context.raw_request_context
+        except AttributeError:
+            pass
+        else:
+            request.headers["X-Edge-Context"] = base64.b64encode(edge_context).decode()
```
MelissaCole: Should this be `X-Edge-Request` to be consistent with what's in pyramid.py?

MelissaCole:
:fish: looks great! ty for adding this

baseplate/clients/requests.py
spladug: well, that's super embarrassing. you're absolutely right.

i've updated this and also reworked the test suite a bit to use baseplate-integrated pyramid on the receiving end with higher-level tests of correctness. this made it obvious the headers we were sending were also wrong: they were for the parent span not the child span. so i've fixed that as well.

thanks for calling this out!

baseplate/clients/requests.py
spladug: yup!

docs/api/baseplate/clients/requests.rst
```diff
@@ -0,0 +1,85 @@
...
+   foo.filter.ip_allowlist = 1.2.3.0/24
+
+   ...
+
+
+and then use the attached :py:class:`~requests.Session`-like object in
+request::
+
+   def my_method(request):
+       request.foo.get("http://html5zombo.com")
```
MelissaCole, line 67: Hah TIL they made an html5 version

MelissaCole:
:fish: looks amazing!

docs/api/baseplate/clients/requests.rst
spladug: it would have been so sad if it'd died with flash!

PR #467, opened 2020-07-14
Add explanation of what to do with "metrics batch too large"

### LINKED ISSUE #426: Document what to do when "metrics batch too large" and link it in error message
This comes up semi-frequently and I give the same spiel every time but that's rather unscalable.

The spiel is along these lines:

> the absolute max is slightly south of 65535 bytes (the maximum size of a UDP datagram)
> but this is generally indicative of something going wrong
> the batching is done at the level of a server span. if this is in request, that means there's a lot going on in that request (like in the past that has meant calling a database in a loop e.g.). outside of request it usually means there's a span set up for something really big like the whole lifetime of a cronjob or qproc but it would be better to make spans for each individual record you process instead.

spladug:
I have grand plans for the new "guide" section of the docs.

This fixes #426.

### DIFF
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -242,7 +242,7 @@ class Batch(BaseClient):
             )
             logger.warning(
                 "Metrics batch of %d bytes is too large to send, flush more often or reduce "
-                "amount done in this request. Top counters: %s",
+                "amount done in this request. See https://git.io/JJscR. Top counters: %s",
                 exc.message_size,
                 ", ".join(f"{c.name.decode()}={c.total:.0f}" for c in counters_by_total[:10]),
             )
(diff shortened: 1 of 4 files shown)


baseplate/lib/metrics.py
```diff
@@ -242,7 +242,7 @@ def flush(self) -> None:
             )
             logger.warning(
                 "Metrics batch of %d bytes is too large to send, flush more often or reduce "
-                "amount done in this request. Top counters: %s",
+                "amount done in this request. See https://git.io/JJscR. Top counters: %s",
```
spladug, line 245: note: this link won't work until this is merged to master. it's a fun little hack to get a nice git.io shortlink for something not hosted on github.com (eek?): it points here: https://github.com/reddit/baseplate.py/blob/gh-pages/shortlinks/metrics-batch-too-large.html

MelissaCole:
:fish: looks great!

PR #468, opened 2020-07-17
Baseplate health check v2 take 2

fishy:
Revert is_healthy endpoint defined in BaseplateService, and add health
check v2 into BaseplateServiceV2, and deprecate BaseplateService.

The healthcheck binary is switched to BaseplateServiceV2. It still works
with thrift services using BaseplateService. Added
tests/integration/thrift_healthcheck_tests.py to make sure of that.


fishy:
👓 @spladug @pacejackson

baseplate/thrift/baseplate.thrift
```diff
@@ -48,7 +69,7 @@ Your service should inherit from this one so that common tools can interact
 with any expected interfaces.
 
 */
-service BaseplateService {
+service BaseplateServiceV2 {
```
fishy, line 72: Or maybe just `BaseplateService2`, not sure which one do y'all prefer.

pacejackson (approved):
V2 seems fine to me

fishy:
:haircut: Added an integration test to make sure that v2 health check client works with v1 server.

tests/integration/thrift_healthcheck_tests.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class GeventPatchedTestCase(unittest.TestCase):
+    def setUp(self):
+        gevent.monkey.patch_socket()
+
+    def tearDown(self):
+        import socket
+
+        reload(socket)
```
spladug: this feels like a ton of duplication. could these be parameterized in the original module and either imported here or the test suite below just moved there? (i lean towards the latter as this is just a subset of thrift integration tests)

spladug (changes requested):
looks great. thanks for the tests, that really improves confidence this is non-breaking and continues to be going forward. only thing is making the test suite a little less duplicated there.

spladug (approved):

PR #469, opened 2020-07-20
Add docs page explaining application startup sequence

spladug:
This is the first longer-form doc in the new "guide" section. Its intent is to give an overview of everything that happens during application startup to make it less opaque. I intend a future section about request lifecycle for once the application is started up.

### DIFF
--- a/docs/guide/index.rst
+++ b/docs/guide/index.rst
@@ -4,3 +4,4 @@ User Guide
 .. toctree::
 
    faq
+   startup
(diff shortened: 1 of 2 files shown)


bradengroom (approved):
Looks good to me. The beginning of `The Python Interpreter` section that talks about virtualenvs/Docker felt unnecessary to me since it's not specific to Baseplate.py. I'm fine with it though if you like it.

MelissaCole:
:fish: very awesome! Looking forward to the request lifecycle :)

spladug:
:+1: cut out the bit about virtualenv/docker. i just needed to get that off my chest but you're right it's not really relevant. thanks both of you!

PR #480, opened 2020-08-13
Allow redis locking

spladug:
The upstream implementation calls back through our client so this ends up being instrumented just fine, but rather granularly. Is this OK or should we go to the effort of figuring out a higher level span system for it?

The spans emitted look like: `redis.SET` and `redis.EVALSHA` (the latter running a LUA script server-side)

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -155,10 +155,6 @@ class MonitoredRedisConnection(redis.StrictRedis):
         """Not currently implemented."""
         raise NotImplementedError
 
-    def lock(self, *args: Any, **kwargs: Any) -> Any:
-        """Not currently implemented."""
-        raise NotImplementedError
-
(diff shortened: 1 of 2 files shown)


MelissaCole:
:fish: lgtm after fixing the tests. I didn't find a need for more granular metrics when I used it so it's probably fine for now (?)

spladug:
aghh the tests are the flakey issue. i thought they'd gone away when drone's disks got fixed but they seem to be back. i've reopened the ticket to fix 'em. https://github.com/reddit/baseplate.py/issues/361

spladug:
Ok there was another flakiness in the new test as well. Because we're looking at the super-granular spans level sometimes there was an error span when the Lua script needed to be loaded. This wasn't ever an application-level error but did fail our "error is None" assertion. I've just removed that assertion.

Thinking about this one got me a solid lead for the other issue as well. Patch for that coming soon.

PR #481, opened 2020-08-13
Use random node path in zookeeper writer tests to avoid clobbering

### LINKED ISSUE #361: Clean out or fix flakey tests
https://github.com/reddit/baseplate.py/pull/360#issuecomment-539760238

spladug:
This should hopefully fix #361, my hypothesis for the flakiness is that
the test suites for each Python version are clobbering each other since
they run simultaneously and before this patch they worked on the same
node in zookeeper.

### DIFF
--- a/tests/integration/live_data/writer_tests.py
+++ b/tests/integration/live_data/writer_tests.py
@@ -1,4 +1,5 @@
 import unittest
+import uuid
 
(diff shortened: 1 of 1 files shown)


MelissaCole:
:fish:!

bradengroom (approved):

PR #483, opened 2020-08-25
Batched kombu consumer

gfmio:
This PR adds support for batch processing of messages in kombu consumer servers.

A typical use case for this would be a queue consumer that has better performance characteristics when performing batched processing, for example if performing a batch of database mutations in a single transaction is significantly faster than performing individual writes.

The `KombuQueueConsumerFactory` now accepts additional parameters to configure the batch processing. If the `batch_size` is `None` (default case), messages are consumed individually, so the behaviour is the same as before. If the `batch_size` is set, the messages will be processed in batches. 

The kombu pump worker class `KombuConsumerWorker` accepts the same parameters and performs the actual batching. If `batch_size` is set, it stores new messages in a batch and processes them once the `batch_size` is reached or a timeout occurs.

The PR also adds a `KombuBatchMessageHandler`, which handles message sequences rather than individual messages. The `KombuQueueConsumerFactory` will return the correct type of message handler depending on whether `batch_size` is set.

The timeout in the `KombuBatchQueueConsumer` is implemented using a new `Timer` helper class which internally uses the library `schedule` to schedule the work. To prevent blocking, the default scheduler is extended to run on a separate thread.

The `Timer` and `KombuConsumerWorker` have safeguards in place for requeueing unprocessed messages in the event of a server shutdown or errors during batch processing.


baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +133,148 @@ def handle(self, message: kombu.Message) -> None:
...
+class KombuBatchQueueConsumer(QueueConsumer):
+    """
+    Wrapper around a MessageHandler object that interfaces with the work_queue and starts/stops the handle loop.
+
+    The batched queue consumer consumes messages until the batch_size is reached
+    or a the batch_timeout has occurred. Then, if the batch is not empty, the
+    messages are consumed. The timer is stopped after a batch is processed if
+    there are no new messages and started when there is at least one message in
+    the batch.
+    """
```
pacejackson: I don't think extending the base `QueueConsumer` is really necessary, couldn't the batch gathering be done in a new PumpWorker implementation and instead of putting work directly on the `work_queue` it just does what this is doing (accumulating  work into a batch and then putting the batch on the queue)?

baseplate/server/queue_consumer.py
```diff
@@ -164,23 +139,62 @@ def run(self) -> None:
...
+
+    @abc.abstractmethod
+    def build_health_checker(self, listener: socket.socket) -> StreamServer:
+        """Build an HTTP server to service health checks."""
+
+    def build_queue_consumer(
+        self, work_queue: queue.Queue, message_handler: MessageHandler
+    ) -> QueueConsumer:
+        """Build a queue consumer."""
+        return QueueConsumer(work_queue=work_queue, message_handler=message_handler)
```
pacejackson: To go along with my comment about not needing to extend the base `QueueConsumer` I don't think these changes are really necessary either?

baseplate/lib/timer.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(
+        self,
+        function: Callable[[], None],
+        interval: float,
+        scheduler_interval: Optional[float] = None,
+        is_repeating: bool = False,
+    ) -> None:
+        self.function = function
+        self.interval = interval
+        self.scheduler_interval = scheduler_interval if scheduler_interval is not None else interval
```
pacejackson: having 2 intervals seems confusing, why not just have one?

baseplate/lib/timer.py
```diff
@@ -0,0 +1,56 @@
...
+        self.is_repeating = is_repeating
+        self.state = TimerState.Stopped
+        self.scheduler = ThreadedScheduler()
+
+    def start(self) -> None:
+        if self.state == TimerState.Running:
+            return
+        self.state = TimerState.Running
+        self.scheduler.every(self.interval).seconds.do(self._invoke)
+        self.scheduler.run_continuously(self.scheduler_interval)
```
pacejackson: I don't think there's a method called `run_continuously` did you mean to use `run`?

baseplate/lib/timer.py
```diff
@@ -0,0 +1,56 @@
...
+class TimerState(Enum):
+    Running = "Running"
+    Stopped = "Stopped"
+
+
+class Timer:
+    def __init__(
+        self,
+        function: Callable[[], None],
+        interval: float,
```
pacejackson: I think, in general using `timedelta`s rather than raw floats/ints makes for a cleaner interface?

pacejackson (changes requested):

baseplate/frameworks/queue_consumer/kombu.py
gfmio: Makes sense. I chose to extend the QueueConsumer because I felt it would make more sense to make this the responsibility of the consumer rather than the pump worker.

baseplate/lib/timer.py
gfmio: The `scheduler_interval` is the resolution of the scheduler, i.e. how often it checks for pending work. The `interval` is the actual frequency at which we want to execute our command. But thinking about it, it's probably sufficient to make them the same interval in this instance because there's only ever this one job scheduled.

baseplate/lib/timer.py
gfmio: Yup, mistake during refactoring

baseplate/lib/timer.py
gfmio: I've updated everything to use `timedelta`s

baseplate/server/queue_consumer.py
gfmio: removed

gfmio:
@pacejackson I've made the requested changes. The pump worker is now handling the batching. However, I had to introduce these rather ugly `@overload` declarations to address the type changes. Do you have an idea how to make this nicer?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -57,20 +68,53 @@ class KombuConsumerWorker(ConsumerMixin, PumpWorker):
     PumpWorker because the ConsumerMixin class already defines it.
     """
 
+    @overload
```
pacejackson: I think it's probably better to just have 2 different `PumpWorker`-s rather than having these overloads and trying to re-use this existing `KombuConsumerWorker`.  They really don't share much code and it's probably easier to work with/understand as separate classes.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -137,6 +288,7 @@ class KombuQueueConsumerFactory(QueueConsumerFactory):
     use the constructor directly.
     """
 
+    @overload
```
pacejackson: Similar to the pump worker, I think this is better handled with a separate `BatchKombuQueueConsumerFactory`.  There might be some cases where we want to share code than with the pump worker, so we _might_ want to consider something like having a base class for the `KombuQueueConsumerFactory` where you plug in functions for building the pump worker and message handler?

baseplate/lib/threaded_scheduler.py
```diff
@@ -0,0 +1,43 @@
+import time
+
+from datetime import timedelta
+from threading import Event
+from threading import Thread
+from typing import Optional
+
+from schedule import Scheduler
```
pacejackson: I'm not sure that adding `schedule` as a dependency and using `Scheduler` is really getting us anything? Ultimately, we just want to set up a thread that maybe executes something in the future, which you can just do with a basic thread by putting a `time.sleep` at the beginning of the `target` function.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -80,6 +124,58 @@ def stop(self) -> None:
...
+            self._timer.stop()
+
+        messages = self._read_and_empty_batch()
+
+        if messages:
+            logger.debug("Requeueing %i messages", len(messages))
+            message: kombu.Message
+            for message in messages:
+                if not message.acknowledged:
+                    message.requeue()
```
pacejackson, line 130: I think this should probably be up to the handler rather than putting it in the pump worker?

pacejackson (changes requested):
In addition to the comments about using separate classes, I think we can simplify things a bit further by putting the actual batch management into it's own object rather than having it be in the pump worker itself:

```python
class BatchedQueue:
    def __init__(self, work_queue: Queue, batch_size: int, flush_interval: timedelta):
        assert batch_size >= 1
        assert work_queue is not None
        self._work_queue = work_queue
        self._batch = []
        self._batch_size = batch_size
        self._flush_interval = flush_interval
        self._flusher = None

    def put(self, data):
        self._batch.append(data)
        if len(self._batch) >= self._batch_size:
            self._flush()
        else:
            self._queue_flush()

    def _flush(self):
        self._flusher.stop()
        self._flusher = None
        batch = copy(self._batch)
        self._batch = []
        self._work_queue.put(batch)

    def _queue_flush(self):
        if self._flusher != None:
            return
        self._flusher = DelayedWorker(delay=self._flush_interval, fn=self._flush)


class DelayedWorker:
    def __init__(self, delay: timedelta, fn: Callable):
        self._delay = delay
        self._fn = fn
        self._stopped = False
        self._thread = threading.Thread(target=self._run)
        self._thread.run()

    def stop(self):
        self._stopped = True

    def _run():
        time.sleep(self._delay)
        if self._stopped:
            return
        self._fn()

    def __del__(self):
        self.stop()
```

So then the pump consumer is just `put`-ing the message into a `BatchedQueue` rather than directly into a `queue.Queue`, you might even be able to make it easy to pass in one or the other to the same `PumpQueue` class constructor?

Note, this also would remove the `ThreadedScheduler` class and the dependency on `schedule`, and replaces the `Timer` class with a `DelayedWorker` object.

baseplate/frameworks/queue_consumer/kombu.py
gfmio: The handler does not get notified when the server gets terminated, so this unfortunately needs to happen in the pump. Alternatively, we can also make the message handler aware of the lifecycle, but I think that would be a bit too much.

gfmio:
@pacejackson  Thank you for your recommendations. I've adapted them slightly to be generic and to merged ideas of the DelayedWorker and Timer.

I do think that either the pump worker or the queue consumer should requeue batched messages in this design. Since we didn't want to modify the QueueConsumer class, the PumpWorker seems like the only good option to me.

baseplate/lib/batched_queue.py
```diff
@@ -0,0 +1,47 @@
...
+        self._flusher.stop()
+        batch = list(self._batch)
+        self._batch = []
+        self._work_queue.put(batch)
+
+    def flush_and_return_batch(self) -> List[T]:
+        self._flusher.stop()
+        batch = list(self._batch)
+        self._batch = []
+        return batch
```
pacejackson: I think you can tidy this up a little bit by separating out `drain` and `flush` and just calling `drain` in `flush`:

```python
def _flush(self) -> None:
    batch = self.drain()
    self._work_queue.put(batch)

def drain(self) -> List[T]:
    self._flusher.stop()
    batch = list(self._batch)
    self._batch = []
    return batch
```

baseplate/lib/timer.py
```diff
@@ -0,0 +1,40 @@
...
+        self._thread = Thread(target=self._run)
+
+    def stop(self) -> None:
+        self._thread = None
+
+    def _run(self) -> None:
+        thread = self._thread
+        while self.is_running() and self._thread == thread:
+            time.sleep(self.interval.total_seconds())
+            self.action()
```
pacejackson: I think we need to check `if self.is_running() and self._thread == thread:` here as well in case it was `stopped` while it was sleeping.  I _thiiiiink_ we also might want to use `is` rather than `==` since they should be the exact same object and that's what we really care about?

```python
    def _run(self) -> None:
        thread = self._thread
        while self.is_running() and self._thread is thread:
            time.sleep(self.interval.total_seconds())
            if !self.is_running() or self._thread is not thread:
                return
            self.action()
            if not self.is_repeating:
                self.stop()
```

If you want to pull that check into a function, maybe something like:

```python
def _is_active_thread(self, thread: threading.Thread) -> bool:
    return self.is_running() and self._thread is thread
```

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: You're right, I think you're doing the right thing here but I what tripped me up is the `if not message.acknowledged:`.

These messages are all just sitting in the batch, they haven't been handed off to the handler yet so none of them should be `acknowledged`. If they _are_ in a state where they could be then we wouldn't want to do this since that would meant the batch was being processed and we wouldn't want to re-queue messages that are currently being processed.  You might want to remove the `if not message.acknowledged:`?  I'm not 100% sure about that though...

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -80,6 +87,48 @@ def stop(self) -> None:
         # `should_stop` is an attribute of `ConsumerMixin`
         self.should_stop = True
 
+    def __del__(self) -> None:
+        if not self.should_stop:
+            self.stop()
```
pacejackson: 🔕  I'm not sure if this addition is really necessary, shouldn't the `ConsumerMixin` already handle this?  It also doesn't hurt.

baseplate/lib/batched_queue.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            self._flusher.start()
+
+    def _flush(self) -> None:
+        self._flusher.stop()
+        batch = list(self._batch)
+        self._batch = []
+        self._work_queue.put(batch)
+
+    def flush_and_return_batch(self) -> List[T]:
```
pacejackson: Nit: I think a better name for this is `drain`?  To me, flush implies that you are flushing the work to the place where it is meant to go, while this is just draining the batch?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +176,61 @@ def handle(self, message: kombu.Message) -> None:
...
+    def handle(self, messages: Sequence[kombu.Message]) -> None:  # pylint: disable=arguments-differ
+        logger.info("Processing batch with %i messages", len(messages))
+
+        context = self.baseplate.make_context_object()
+        try:
+            # We place the call to ``baseplate.make_server_span`` inside the
+            # try/except block because we still want Baseplate to see and
+            # handle the error (publish it to error reporting)
+            with self.baseplate.make_server_span(context, self.name) as span:
+                span.set_tag("kind", "batch_consumer")
```
pacejackson, line 212: 🔕  maybe add a `"size"` tag that's `len(messages)`?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +176,61 @@ def handle(self, message: kombu.Message) -> None:
...
+                    if not message.acknowledged:
+                        message.requeue()
+
+            if isinstance(exc, FatalMessageHandlerError):
+                logger.info("Recieved a fatal error, terminating the server.")
+                raise
+        else:
+            for message in messages:
+                if not message.acknowledged:
+                    message.ack()
```
pacejackson, line 267: 🔕  I wonder if we should actually do this for batch messages? It might make more sense to have the handler_fn call `ack` itself rather than doing it here?  My thinking is that it makes handling errors tricky if developers are expecting the batch to be `ack`-ed at this level so they aren't doing it in their handler and they raise an exception 1/2 way through processing a batch, we would re-queue all the messages since none were `ack`-ed but we would be re-queueing a bunch that had already been processed?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +176,61 @@ def handle(self, message: kombu.Message) -> None:
...
+            # We place the call to ``baseplate.make_server_span`` inside the
+            # try/except block because we still want Baseplate to see and
+            # handle the error (publish it to error reporting)
+            with self.baseplate.make_server_span(context, self.name) as span:
+                span.set_tag("kind", "batch_consumer")
+
+                message: kombu.Message
+                for message in messages:
+                    delivery_info = message.delivery_info
+                    span.set_tag("amqp.routing_key", delivery_info.get("routing_key", ""))
```
pacejackson: I think this is just going to override the tag each time you set this, so it's just going to be set to whatever it is set to in the last message.  I'm also guessing that these values are often going to be the same for each message in a batch or at least there will be significant overlap? It might be better to accumulate all of the unique patterns of these and just set each unique one as tags like below:

```python
class _SpanTag(NamedTuple):
    routing_key: str
    consumer_tag: str
    delivery_tag: str
    exchange: str


tags: Set[_SpanTag] = set()
for message in messages:
    delivery_info = message.delivery_info
    tags.add(_SpanTag(
        routing_key=delivery_info.get("routing_key", ""),
        consumer_tag=delivery_info.get("consumer_tag", ""),
        delivery_tag=delivery_info.get("delivery_tag", ""),
        exchange=delivery_info.get("exchange", ""),
    ))

for i, tag in enumerate(tags):
    span.set_tag(f"amqp.routing_key{i}", tag.routing_key)
    span.set_tag(f"amqp.consumer_tag{i}", tag.consumer_tag)
    span.set_tag(f"amqp.delivery_tag{i}", tag.delivery_tag)
    span.set_tag(f"amqp.exchange{i}", tag.exchange)
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +345,127 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+        baseplate: Baseplate,
+        name: str,
+        connection: kombu.Connection,
+        queues: Sequence[kombu.Queue],
+        handler_fn: BatchHandler,
+        error_handler_fn: Optional[BatchErrorHandler] = None,
+        health_check_fn: Optional[HealthcheckCallback] = None,
+        serializer: Optional[KombuSerializer] = None,
+        batch_size: int = 1,
+        batch_timeout: timedelta = timedelta(0, 1, 0, 0, 0, 0, 0),
```
pacejackson: You can just write this as:

```suggestion
        batch_timeout: timedelta = timedelta(seconds=1),
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +345,127 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+            to consume from.
+        :param handler_fn: A function that will process an individual message from a queue.
+        :param error_handler_fn: A function that will be called when an error is thrown
+            while executing the `handler_fn`. This function will be responsible for calling
+            `message.ack` or `message.requeue` as it will not be automatically called by
+            `KombuMessageHandler`'s `handle` function.
+        :param health_check_fn: A `baseplate.server.queue_consumer.HealthcheckCallback`
+            function that can be used to customize your health check.
+        :param serializer: A `baseplate.clients.kombu.KombuSerializer` that should
+            be used to decode the messages you are consuming.
```
pacejackson, line 437: This is missing the `batch_size` and `batch_timeout` arguments.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +345,127 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+            queues=queues,
+            handler_fn=handler_fn,
+            error_handler_fn=error_handler_fn,
+            health_check_fn=health_check_fn,
+            serializer=serializer,
+            batch_size=batch_size,
+            batch_timeout=batch_timeout,
+        )
+
+    def build_pump_worker(self, work_queue: BatchWorkQueue) -> KombuBatchConsumerWorker:
```
pacejackson: This is just a `queue.Queue` since it's called by the "server" framework, you'll need to initialize the `BatchWorkQueue` in here and pass that to the constructor.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +345,127 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+        exchange: kombu.Exchange,
+        connection: kombu.Connection,
+        queue_name: str,
+        routing_keys: Sequence[str],
+        handler_fn: BatchHandler,
+        error_handler_fn: Optional[BatchErrorHandler] = None,
+        health_check_fn: Optional[HealthcheckCallback] = None,
+        serializer: Optional[KombuSerializer] = None,
+        batch_size: int = 1,
+        batch_timeout: timedelta = timedelta(0, 1, 0, 0, 0, 0, 0),
```
pacejackson: Same here as above, you can just write this as:

```suggestion
        batch_timeout: timedelta = timedelta(seconds=1),
```

pacejackson (changes requested):
This is looking really good!  There's still a few things to fix but I think it's just about there (or at least to the point where you can feel comfortable writing tests).

baseplate/server/queue_consumer.py
```diff
@@ -116,7 +132,7 @@ class QueueConsumerFactory(abc.ABC):
     """
 
     @abc.abstractmethod
-    def build_pump_worker(self, work_queue: queue.Queue) -> PumpWorker:
+    def build_pump_worker(self, work_queue: WriteableQueueLike) -> PumpWorker:
```
pacejackson: I don't think you actually need these changes, `build_pump_worker` is only called in the "server" and it always passes a `queue.Queue`.

baseplate/frameworks/queue_consumer/kombu.py
gfmio: The `ConsumerMixin` doesn't handle it, but yes, it should! I've moved in there.

baseplate/frameworks/queue_consumer/kombu.py
gfmio: I agree with the conceptual questions. The reason why I implemented this behaviour is that the single `KombuMessageHandler` also has the same type of `else` clause where the message it has processed is acknowledged. However, that one is actually not even checking whether the message had been acknowledged at all, so maybe we should add that check?

I think that clients should be aware that the default behaviour is that unack'ed messages will be ack'ed if processing completes successfully and that all unacknowledged messages are requeued upon failure. However, clients should also be allowed to ack, requeue and discard messages themselves, for more fine grained control.

baseplate/frameworks/queue_consumer/kombu.py
gfmio: Maybe we can print a warning if a message was already acknowledged?

```py
            for message in messages:
                if not message.acknowledged:
                    message.requeue()
                else:
                    logger.warning("Message in unprocessed batch had already been acknowledged") 
```

pacejackson: That seems reasonable... 👍

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +169,62 @@ def handle(self, message: kombu.Message) -> None:
...
+        handler_fn: BatchHandler,
+        error_handler_fn: Optional[BatchErrorHandler] = None,
+    ):
+        self.baseplate = baseplate
+        self.name = name
+        self.handler_fn = handler_fn
+        self.error_handler_fn = error_handler_fn
+
+    def handle(self, messages: Sequence[kombu.Message]) -> None:  # pylint: disable=arguments-differ
+        logger.info("Processing batch with %i messages", len(messages))
```
pacejackson: You might want to make this `debug`? A really high throughput queue consumer system will probably generate a lot of logs with this.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +339,133 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+            :py:class:`~kombu.Queue` s to consume from.
+        :param handler_fn: A function that will process an individual message from a queue.
+        :param error_handler_fn: A function that will be called when an error is thrown
+            while executing the `handler_fn`. This function will be responsible for calling
+            `message.ack` or `message.requeue` as it will not be automatically called by
+            `KombuMessageHandler`'s `handle` function.
+        :param health_check_fn: A `baseplate.server.queue_consumer.HealthcheckCallback`
+            function that can be used to customize your health check.
+        :param serializer: A `baseplate.clients.kombu.KombuSerializer` that should
+            be used to decode the messages you are consuming.
```
pacejackson, line 493: We also need a comment for batch size/timeout here.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -127,6 +169,62 @@ def handle(self, message: kombu.Message) -> None:
...
+                span.set_tag("kind", "batch_consumer")
+                span.set_tag("size", len(messages))
+
+                message: kombu.Message
+                for i, message in enumerate(messages):
+                    delivery_info = message.delivery_info
+                    span.set_tag(f"amqp.routing_key_{i}", delivery_info.get("routing_key", ""))
+                    span.set_tag(f"amqp.consumer_tag_{i}", delivery_info.get("consumer_tag", ""))
+                    span.set_tag(f"amqp.delivery_tag_{i}", delivery_info.get("delivery_tag", ""))
+                    span.set_tag(f"amqp.exchange_{i}", delivery_info.get("exchange", ""))
```
pacejackson: I'm still not sure that we want to do this, this is going to create a lot of tags (4 per message in the batch).  I would still suggest using [the example I gave in a previous comment](https://github.com/reddit/baseplate.py/pull/483#discussion_r485097887) to only have unique sets of tags, or just leave these off altogether.

tests/unit/lib/timer_tests.py
```diff
@@ -0,0 +1,82 @@
...
+from unittest.mock import Mock
+
+import pytest
+
+from baseplate.lib.timer import Timer
+
+
+@pytest.fixture
+def mock() -> Mock:
+    return Mock()
```
pacejackson: This seems unnecessary, you can just make a Mock() in the test itself.

tests/unit/lib/batched_queue_tests.py
```diff
@@ -0,0 +1,90 @@
...
+
+if TYPE_CHECKING:
+    WorkQueue = Queue[Sequence[int]]
+else:
+    WorkQueue = Queue
+
+
+@pytest.fixture
+def queue() -> WorkQueue:
+    return Queue()
```
pacejackson: This seems kinda unnecessary, it's probably clearer to just initialize the Queue() in the test method itself?

pacejackson (changes requested):

baseplate/frameworks/queue_consumer/kombu.py
gfmio: That's a valid concern. The disadvantage of this would be that we don't know which messages have which tags.

Alternatively, would it be useful to have a single comma-delimited list of tags (for all items or uniques)?

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -81,6 +88,44 @@ def stop(self) -> None:
...
+        queues: Sequence[kombu.Queue],
+        work_queue: BatchedQueue,
+        serializer: Optional[KombuSerializer] = None,
+    ) -> None:
+        self.connection = connection
+        self.queues = queues
+        self.work_queue = work_queue
+        self.serializer = serializer
+
+    def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
```
spladug, line 109: :no_bell: i know this is a kombu-ism, but the capitalized parameters really freak me out

baseplate/frameworks/queue_consumer/kombu.py
spladug: If I saw this message in my logs I'd be pretty confused about what I'm supposed to do with it. Given the conversation above and the fact we're emitting at warning level I think I'm understanding that this is indicative of something the application needs to do differently? If so, can the message give some more context about where this is happening, what it means, and what to do about it?

(This thinking is coming from many attempts at making [this error message](https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/metrics.py#L243-L248) actionable. I _hope_ we're at a good place there now. If you want to copy that pattern I can help you get the shortlink set up to point at a relevant answer in [the FAQ](https://baseplate.readthedocs.io/en/stable/guide/faq.html))

baseplate/frameworks/queue_consumer/kombu.py
spladug: I think tags are useful for their raw text content in the trace view and they're useful for dimensioning like in tagged metrics. We should probably ignore the second use case here since it'd be hard to make meaningful dimensions in metrics out of something like this because the values are all muddled together.  Having large batch sizes mean massive trace payloads is a potential subtle and confusing bottleneck. So, for traces we probably just want to optimize for human consumption. De-duping and plopping into a comma-delimited list sounds great. I bet Andrew's right that in many cases it'll end up with single-item lists anyway.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +342,136 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+        error_handler_fn: Optional[BatchErrorHandler] = None,
+        health_check_fn: Optional[HealthcheckCallback] = None,
+        serializer: Optional[KombuSerializer] = None,
+        batch_size: int = 1,
+        batch_timeout: timedelta = timedelta(seconds=1),
+    ) -> None:
+        """`KombuQueueConsumerFactory` constructor.
+
+        :param baseplate: The Baseplate set up for your consumer.
+        :param exchange: The `kombu.Exchange` that you will bind your :py:class:`~kombu.Queue` s
```
spladug:
```suggestion
        :param exchange: The `kombu.Exchange` that you will bind your :py:class:`~kombu.Queue` objects
```

that weird floating s (which i know has to be that way for reST :angry:) will look real weird in the docs

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -241,3 +342,136 @@ def build_message_handler(self) -> KombuMessageHandler:
...
+            to.
+        :param queues: List of  :py:class:`~kombu.Queue` s to consume from.
+        :param queue_name: Name for your queue.
+        :param routing_keys: List of routing keys that you will create :py:class:`~kombu.Queue` s
```
spladug: same "s" deal

baseplate/lib/batched_queue.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _flush(self) -> None:
+        batch = self.drain()
+        self._work_queue.put(batch)
+
+    def drain(self) -> List[T]:
+        self._flusher.stop()
+        batch = list(self._batch)
+        self._batch = []
+        return batch
```
spladug: This feels pretty similar to https://github.com/reddit/baseplate.py/blob/develop/baseplate/sidecars/__init__.py#L55-L81

Rather than using threads, that solution relies on the pump timing out occasionally and putting an empty object into the batch to give it a chance to process timers. Can we do that here? I'd really like to avoid the threading stuff going on in the timer if we can.

baseplate/lib/timer.py
```diff
@@ -0,0 +1,46 @@
...
+from threading import Thread
+from typing import Callable
+from typing import Optional
+
+
+class Timer:
+    """A restartable timer that execute an action after a timeout or at an interval."""
+
+    def __init__(
+        self, action: Callable[[], None], interval: timedelta, is_repeating: bool = False,
```
spladug: what's the deal with is_repeating? I don't see it used anywhere.

spladug (changes requested):
Cool! Looks solid overall. Thanks for doing this.

Biggest blocker for me is figuring out the duplication of the batch stuff.

baseplate/lib/batched_queue.py
gfmio: Yes, it's definitely similar in spirit. I've reworked it, so that the BatchedQueue does not flush automatically, but instead is called in the `on_iteration` handler.

gfmio:
Sorry for taking 7 months to add the requested changes to this PR ;-)

spladug (approved):
LGTM! Thank you!

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -72,18 +79,65 @@ def __init__(
         self.serializer = serializer
         self.kwargs = kwargs
 
-    def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
+    def get_consumers(self, consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
```
pacejackson: I think we should leave this capitalized, it's like that because that's the signature of this method in `ConsumerMixin`

https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.mixins.html

```suggestion
    def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -72,18 +79,65 @@ def __init__(
...
         args = dict(queues=self.queues, on_message=self.work_queue.put, **self.kwargs)
         if self.serializer:
             args["accept"] = [self.serializer.name]
-        return [Consumer(**args)]
+        return [consumer(**args)]
```
pacejackson:
```suggestion
        return [Consumer(**args)]
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -72,18 +79,65 @@ def __init__(
...
+        serializer: Optional[KombuSerializer] = None,
+        **kwargs: Any,
+    ) -> None:
+        self.connection = connection
+        self.queues = queues
+        self.work_queue = work_queue
+        self.serializer = serializer
+        self.kwargs = kwargs
+
+    def get_consumers(self, consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
```
pacejackson:
```suggestion
    def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
```

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -72,18 +79,65 @@ def __init__(
...
+        args = dict(queues=self.queues, on_message=self.work_queue.put, **self.kwargs)
+        if self.serializer:
+            args["accept"] = [self.serializer.name]
+        return [consumer(**args)]
```
pacejackson:
```suggestion
        return [Consumer(**args)]
```

pacejackson (approved):
Some small things around naming of arguments

gfmio:
@spladug Ready to merge ♻️

spladug:
@gfmio it looks like it's failing lint (non-lazy string formatting in a logger call)

spladug (changes requested):
just for the lint failure

gfmio:
@spladug The linter and tests are now all green 💚

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -84,6 +91,54 @@ def stop(self) -> None:
...
+                    message.requeue()
+                else:
+                    logger.warning(
+                        "%s %s %s %s %s",
+                        "KombuBatchConsumerWorker is stopping and requeuing messages in the",
+                        "unprocessed batch. A message in this unprocessed batch had already",
+                        "been acknowledged, so it is not being requeued. However, this indicates",
+                        "an inconsistency in the batch processing logic which should be",
+                        "investigated.",
+                    )
```
spladug, line 139: What's going on here? I'm guessing it was complaining about concatenating the strings with +? It should be possible to concatenate them without this formatting stuff by just putting them next to eachother with no operators:


```suggestion
                    logger.warning(
                        "KombuBatchConsumerWorker is stopping and requeuing messages in the "
                        "unprocessed batch. A message in this unprocessed batch had already "
                        "been acknowledged, so it is not being requeued. However, this indicates "
                        "an inconsistency in the batch processing logic which should be "
                        "investigated."
                    )
```

spladug (approved):

PR #488, opened 2020-09-08
Make BaseplateRequest play more nicely with superclass methods

spladug:
(Landing this after the 1.4 release because it's pretty low-level and I'd like it to bake for a while before going into a release. But it'll be really nice to have for the testing helpers I'm building out in Harold2)

[Things like Request.from_file call the current subclass's constructor
with a specific order of arguments.][1] This doesn't work if we put our
context config at the front and make it required.

By reordering like this we play nicely as a subclass and can still get
our data in when we need it.

[1]: https://github.com/Pylons/webob/blob/ac2d238f50bed7fbbf115818ab7f208b71d7bdca/src/webob/request.py#L1251-L1253

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -124,9 +124,10 @@ class StaticTrustHandler(HeaderTrustHandler):
 
 # pylint: disable=too-many-ancestors
 class BaseplateRequest(RequestContext, pyramid.request.Request):
-    def __init__(self, context_config: Dict[str, Any], environ: Dict[str, str]):
-        RequestContext.__init__(self, context_config)
-        pyramid.request.Request.__init__(self, environ)
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        context_config = kwargs.pop("context_config", None)
+        RequestContext.__init__(self, context_config=context_config)
+        pyramid.request.Request.__init__(self, *args, **kwargs)
 
 
 class RequestFactory:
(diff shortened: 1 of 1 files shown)


baseplate/frameworks/pyramid/__init__.py
```diff
@@ -124,21 +124,22 @@ def should_trust_edge_context_payload(self, request: Request) -> bool:
 
 # pylint: disable=too-many-ancestors
 class BaseplateRequest(RequestContext, pyramid.request.Request):
-    def __init__(self, context_config: Dict[str, Any], environ: Dict[str, str]):
-        RequestContext.__init__(self, context_config)
-        pyramid.request.Request.__init__(self, environ)
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
bradengroom, line 127: I'm curious why you changed the function signature. It _seems_ like the original signature would have worked.

bradengroom, line 127: I'm curious why you changed the function signature here. It _seems_ like the original signature would have worked.

bradengroom (approved):

baseplate/frameworks/pyramid/__init__.py
spladug: The `from_file` call into `cls()` expects `__init__` to have a signature like `__init__(environ: Dict[str, str], **kwargs)` but we were putting `context_config` (our thing) first which made that environ go to the wrong place and mess everything up.

PR #489, opened 2020-09-08
pyramid: Keep server span open while app_iter iterates

spladug:
The WSGI spec expects applications to return an iterable object. In the
common case, the iterable is a single-item list containing a byte string
of the full response. However, if the application wants to stream a
response back to the client (e.g. it's sending a lot of data, or it
wants to get some bytes on the wire really quickly before some database
calls finish) the iterable can take a while to finish iterating.

This makes it so that we keep the server span open while the iterable is
still iterable and only close it when we're done. This allows use of
request context even after the original view callable has returned but
as long as the response is still streaming out to the user.

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -5,6 +5,8 @@ import sys
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import Iterable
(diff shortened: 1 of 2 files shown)


spladug:
:eyeglasses: @madbook

madbook:
🐟 awesome, lgtm!

Issue #486, opened 2020-09-03
Expose and Variant have difficult method signatures


markis:
Mostly the hard part to keep track of is `name` in `variant` and `experiment_name` in `expose`.

From here: 
https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/experiments/__init__.py#L138
```python
    def variant(
        self,
        name: str,
        user: Optional[User] = None,
        bucketing_event_override: Optional[bool] = None,
        **kwargs: str,
    ) -> Optional[str]:
```
From here: 
https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/experiments/__init__.py#L233
```python
    def expose(
        self, experiment_name: str, variant_name: str, user: Optional[User] = None, **kwargs: str
    ) -> None:
```
Would it be possible to add `experiment_name` to `variant` and deprecate `name`?  Or just accept both?

spladug:
Making the naming consistent between the two calls sounds good to me. Deprecating the old one also sounds good as long as we don't break apps that are using the old name still.

Are you interested in taking this on?

markis:
Yes, I am interested in taking it on.  I wanted to ask before filing a PR.

PR #492, opened 2020-09-10
Change `name` to `experiment_name` on `Experiments.expose`

markis:
This addresses #486

Currently, the `Experiments` class has two methods that are used in tandem quite often, `variant` and `expose`. And they both have different parameters for the name of the experiment.

`variant` -> `name`
`expose` -> `experiment_name`

This PR will introduce `experiment_name` to `variant` and log a deprecation warning for `name`.

### DIFF
--- a/baseplate/lib/experiments/__init__.py
+++ b/baseplate/lib/experiments/__init__.py
@@ -140,6 +140,7 @@ class Experiments:
     @overload
     def variant(
         self,
+        *,
         name: str,
         user: Optional[User] = None,
         bucketing_event_override: Optional[bool] = None,
@@ -149,6 +150,7 @@ class Experiments:
     @overload
     def variant(
         self,
+        *,
         experiment_name: str,
         user: Optional[User] = None,
         bucketing_event_override: Optional[bool] = None,
--- a/baseplate/lib/experiments/__init__.py
+++ b/baseplate/lib/experiments/__init__.py
@@ -144,6 +144,7 @@ class Experiments:
         name: str,
         user: Optional[User] = None,
         bucketing_event_override: Optional[bool] = None,
+        **kwargs: str,
     ) -> Optional[str]:
         ...
 
@@ -154,6 +155,7 @@ class Experiments:
         experiment_name: str,
         user: Optional[User] = None,
         bucketing_event_override: Optional[bool] = None,
+        **kwargs: str,
     ) -> Optional[str]:
         ...
 
--- a/baseplate/lib/experiments/__init__.py
+++ b/baseplate/lib/experiments/__init__.py
@@ -10,6 +10,7 @@ from typing import Set
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
+from baseplate.lib import warn_deprecated
 from baseplate.lib.edge_context import User
 from baseplate.lib.events import DebugLogger
 from baseplate.lib.events import EventLogger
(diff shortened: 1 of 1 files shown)


markis:
^ rebase

markis:
@spladug Could I get :eyes:?

baseplate/lib/experiments/__init__.py
```diff
@@ -183,7 +186,13 @@ def variant(
 
         :return: Variant name if a variant is active, None otherwise.
```
spladug, line 210: might be nice to add something like:

```rst
.. versionchanged:: 1.5
    ``name`` was renamed to ``experiment_name``
```

so it's clear in the docs when this happened

baseplate/lib/experiments/__init__.py
```diff
@@ -137,9 +138,10 @@ def is_valid_experiment(self, name: str) -> bool:
 
     def variant(
```
spladug, line 162: I'm not sure if this is worth the extra syntax and maybe fighting with linters (it's been a pain in the past) but `@typing.overload` can be nice for calling out mutually exclusive variants of a function's signature to the type system.

```python
@overload
def variant(self, name: str, user: Optional[User] = None, bucketing_event_override: Optional[bool] = None) -> Optional[str]:
    ...

@overload
def variant(self, name: str, user: Optional[User] = None, bucketing_event_override: Optional[bool] = None) -> Optional[str]:
    ...

def variant(self, experiment_name=None, user=None, bucketing_event_override=None, name=None):
    r"""The actual code..."""
```

(in use here: https://github.com/reddit/baseplate.py/blob/develop/baseplate/__init__.py#L475-L524)

spladug (approved):
thanks for doing this! I appreciate the attention to detail on the order of the arguments for if they're passed positionally.  Left a couple of non-blocking things that might make docs/typing clearer.

baseplate/lib/experiments/__init__.py
markis: The overloads need to be "Keyword Only" functions in order for them to work with mypy. I think it makes sense in this case, that the overloads should be keyword only.  For typing purposes, keywords matter and ordering doesn't matter as much.

Error I got from mpy.
```
baseplate/lib/experiments/__init__.py:158: error: Overloaded function implementation does not accept all possible arguments of signature 1
```

spladug: Oh I missed the `**kwargs: str` in the original method signature. Is that what it's complaining about in the overloads? I think we want to keep that as that's how you pass arbitrary variables in, right?

spladug (changes requested):
looks great! just want to make sure the **kwargs are still cromulent on the overloads

baseplate/lib/experiments/__init__.py
markis: I thought it wasn't as necessary, because the base signature would handle kwargs and any arguments.  But maybe for correctness, it makes sense to have the kwargs in the narrowed signatures.

markis: >Is that what it's complaining about in the overloads?

It wasn't related to kwargs, it's related to the order of arguments and their names.  Because if I switch the `name` and `experiment_name` arguments back with `name` as the first argument, then it complains that the second signature is incorrect.

spladug (approved):
Thanks!

markis:
@spladug, thanks for the review and approval!  Could you merge it?

PR #493, opened 2020-09-24
Fix JSON logging of Trace ID

spladug:
We needed to ask for threadName to be in the JSON output before we could
manipulate it in the custom log formatter.

:eyeglasses: @isugimpy

### DIFF
--- a/baseplate/lib/log_formatter.py
+++ b/baseplate/lib/log_formatter.py
@@ -5,7 +5,13 @@ class CustomJsonFormatter(jsonlogger.JsonFormatter):
     def process_log_record(self, log_record: dict) -> dict:
         log_record["level"] = log_record.pop("levelname", None)
         try:
-            log_record["traceID"] = int(log_record["threadName"])
+            # see if there's an integer thread name, that's probably a traceID
+            # from the logging observer.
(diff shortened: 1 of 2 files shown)


ConorMcGee (approved):

baseplate/lib/log_formatter.py
```diff
@@ -5,7 +5,13 @@ class CustomJsonFormatter(jsonlogger.JsonFormatter):
     def process_log_record(self, log_record: dict) -> dict:
         log_record["level"] = log_record.pop("levelname", None)
         try:
-            log_record["traceID"] = int(log_record["threadName"])
+            # see if there's an integer thread name, that's probably a traceID
+            # from the logging observer.
+            trace_id = int(log_record["threadName"])
```
isugimpy, line 10: :no_bell: Why not just make this one line?  Otherwise, LGTM.

isugimpy (approved):
:fish:

baseplate/lib/log_formatter.py
spladug: just so I could add some explanation in comments for why we're doing `str(int(str))`

PR #494, opened 2020-09-28
Redis: unlock pubsub() and add connection pool runtime metrics

spladug:
Unlocking pubsub is like unlocking locks: c3143199489245144d720ef0eb221720682eeb04. We're just letting the underlying magic in execute() deal with it for us.

Runtime metrics are documented in the .rst file modified below. This should be helpful for tracking connection pool usage in prod, especially with things like pubsub in use.


bradengroom (approved):

MelissaCole:
:fish: lgtm!

PR #498, opened 2020-10-12
Add worker_kwargs

KTAtkinson:
This enabled clients to provide any keyword arguments to the kombu worker constructor. This would have been very useful when diagnosing and mitigating a mailroom outage on Friday.

### DIFF
--- a/baseplate/frameworks/queue_consumer/kombu.py
+++ b/baseplate/frameworks/queue_consumer/kombu.py
@@ -4,6 +4,7 @@ import socket
 
 from typing import Any
 from typing import Callable
+from typing import Dict
 from typing import Optional
 from typing import Sequence
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -63,14 +64,17 @@ def __init__(
...
     ):
         self.connection = connection
         self.queues = queues
         self.work_queue = work_queue
         self.serializer = serializer
+        self.kwargs = kwargs
 
     def get_consumers(self, Consumer: kombu.Consumer, channel: Channel) -> Sequence[kombu.Consumer]:
         args = dict(queues=self.queues, on_message=self.work_queue.put)
+        args.update(self.kwargs)
```
spladug: would adding `, **self.kwargs)` to the `dict()` call above do this in fewer words?

spladug (approved):
seems reasonable to me! could you put "kombu consumer" somewhere in the commit message so it's clearer what worker_kwargs are being added to please?

baseplate/frameworks/queue_consumer/kombu.py
pacejackson: That'll give you an error if "kwargs" has a value that is already being passed in explicitly while this overrides the value.

pacejackson: Both are valid IMO, just different.

pacejackson (approved):

baseplate/frameworks/queue_consumer/kombu.py
KTAtkinson: If repeat kwargs cause errors, is that going to be a problem?

PR #502, opened 2020-10-19
Document hidden capabilities of the `add_to_context` method

foreverest:
[This code](https://github.com/reddit/baseplate.py/blob/develop/baseplate/__init__.py#L219-L224) implies the `add_to_context` method can accept not only a `ContextFactory` object but also other types of arguments, which can be useful in some scenarios.

This PR:
- Updates the method's signature
- Updates the docstring to cover the undocumented capabilities
- Adds tests to cover the undocumented capabilities

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -539,21 +539,30 @@ class Baseplate:
         cfg = config.parse_config(app_config, context_spec)
         self._context_config.update(cfg)
 
-    def add_to_context(
-        self, name: str, context_factory: "baseplate.clients.ContextFactory"
-    ) -> None:
-        """Add an attribute to each request's context object.
+    def add_to_context(self, name: str, attribute_config: Any) -> None:
+        """Add an attribute or a structure of attributes to each request's context object.
 
-        On each request, the factory will be asked to create an appropriate
-        object to attach to the :py:class:`~baseplate.RequestContext`.
+        The given attribute config object can be one of the following:
+
+        * An arbitrary object to be added to the
+        :py:class:`~baseplate.RequestContext`.
+
+        * A factory with a method named ``make_object_for_context``.  On each
+        request, the factory will be asked to create an appropriate object to
+        attach to the :py:class:`~baseplate.RequestContext`.
+
+        * A dict containing arbitrary objects, factories, or other dicts.  In
+        this case, a nested object will be added to the context. Each item of
+        the dict will be processed using the same rules to become an attribute
+        of the nested object.
 
         :param name: The attribute on the context object to attach the
             created object to. This may also be used for metric/tracing
             purposes so it should be descriptive.
-        :param context_factory: A factory.
+        :param attribute_config: A configuration object.
 
         """
-        self._context_config[name] = context_factory
+        self._context_config[name] = attribute_config
 
     def make_context_object(self) -> RequestContext:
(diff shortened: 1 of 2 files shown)


spladug:
These hidden features were new powers of `configure_context()` at which point `add_to_context()` became mostly there just for the few remaining cases that `configure_context()` doesn't cover. Does it make sense to encourage these additional uses of the old method when the new one exists? Are the docs on the new one sufficient?

foreverest:
The docs on the `configure_context()` are great and I love the way the new method works. However, I believe there are use cases where the old method suits better. For example, in a personal project, I want to add a bunch of uniform objects to the context. The reasons why I'd prefer `add_to_context()` over `configure_context()` in my case are:
1. I don't want to clog up the root space of the context with all those objects, but want them to be grouped under a single root-level key.
2. I have a factory for my objects and would prefer not to complicate my code by implementing the `Parser` interface additionally.
3. The new method couples the context attributes and the corresponding keys in the standard Baseplate app config file, but my objects don't have any configuration in the standard file, which makes the use of `configure_context()` even less reasonable.

As the result, in my project, I pass a dict to `add_to_context()` and it works perfectly, except for the fact that mypy complains about mismatching the type of the argument. 

I understand that my use case might be not very common, but I still think one can love the old method for its simplicity and straightforwardness and welcome the flexibility of the new uses.

baseplate/__init__.py
```diff
@@ -539,21 +539,27 @@ def configure_context(self, *args: Any, **kwargs: Any) -> None:  # noqa: F811
...
-        """Add an attribute to each request's context object.
+    def add_to_context(self, name: str, attribute_config: Any) -> None:
+        """Add an attribute or a structure of attributes to each request's context object.
+
+        The given attribute config object can be one of the following::
+            - An arbitrary object to be added to the :py:class:`~baseplate.RequestContext`.
+
+            - A factory with a method named ``make_object_for_context``.
+            On each request, the factory will be asked to create an appropriate
+            object to attach to the :py:class:`~baseplate.RequestContext`.
```
spladug: This reST syntax is leading to some kinda funky output when rendered:

Screenshot from 2020-10-20 11-10-33

How about something like:

```suggestion
        The given attribute config object can be one of the following:        
                                                                              
        * An arbitrary object to be added to the                              
        :py:class:`~baseplate.RequestContext`.                                
                                                                              
        * A factory with a method named ``make_object_for_context``.  On each 
        request, the factory will be asked to create an appropriate object to 
        attach to the :py:class:`~baseplate.RequestContext`.                  
                                                                              
        * A dict containing arbitrary objects, factories, or other dicts.  In 
        this case, a nested object will be added to the context. Each item of 
        the dict will be processed using the same rules to become an attribute
        of the nested object.                                                 

```

(you can test this locally in OneVM with `make docs` and visiting https://baseplate-py.reddit.local)

spladug (changes requested):
Thanks for the thorough writeup! Really appreciate it. Totally agreed now and this looks great. Just one comment on the reST formatting in the docstring.

baseplate/__init__.py
foreverest: Yes, this looks much better. Thanks for the suggestion!

spladug (approved):
Thanks!

PR #504, opened 2020-10-19
Provide ability for config to be pulled from the environment.

KTAtkinson:
Provide a utility to load default configuration values from the environment. Currently Infrared uses this following bash command to manually load values from the environment:
```
mkdir -p /etc/baseplate && cat /opt/baseplate/secrets-fetcher.ini  | sed 's^VAULT_URL^'"$REDDIT_VAULT_URL"'^'  |
        sed 's^VAULT_MOUNT^'"$REDDIT_VAULT_MOUNT"'^'  > $BASEPLATE_CONFIG_PATH  &&
        python3 -m baseplate.sidecars.secrets_fetcher $BASEPLATE_CONFIG_PATH
```
It would be great to have this be a native feature of Baseplate's configuration loading.

Environment variables as defaults may also soon be used to provide default trace publishing endpoints and other variables which are environment, often Kubernetes cluster, specific.

### DIFF
--- a/baseplate/lib/config.py
+++ b/baseplate/lib/config.py
@@ -19,6 +19,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
 .. testsetup:: overview
 
+    import os
     import configparser
     from baseplate import config
     from tempfile import NamedTemporaryFile
@@ -29,6 +30,9 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     tempfile.write(b"cool")
     tempfile.flush()
     config_parser.set("app:main", "some_file", tempfile.name)
+
+    env_var_name = "BASEPLATE_DEFAULT_VALUE"
+    os.environ[env_var_name] = "default"
 
 .. doctest:: overview
 
@@ -49,6 +53,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
     ...     "optional": config.Optional(config.Integer, default=9001),
     ...     "sample_rate": config.Percent,
     ...     "interval": config.Fallback(config.Timespan, config.Integer),
+    ...     "default_from_env": config.DefaultFromEnv(config.String, env_var_name),
     ... })
 
     >>> print(cfg.simple)
@@ -62,6 +67,7 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
     >>> cfg.some_file.read()
     'cool'
+
     >>> cfg.some_file.close()
 
     >>> cfg.sample_rate
@@ -69,6 +75,9 @@ server, The ``config_parser.items(...)`` step is taken care of for you and
 
     >>> print(cfg.interval)
     0:00:30
+
+    >>> print(cfg.default_from_env)
+    example
 
 .. testcleanup:: overview
 
@@ -79,6 +88,7 @@ import base64
 import datetime
 import functools
 import grp
+import os
 import pwd
 import re
 import socket
@@ -377,6 +387,30 @@ def TupleOf(item_parser: Callable[[str], T]) -> Callable[[str], Sequence[T]]:  #
         return [item_parser(item) for item in stripped if item]
 
     return tuple_of
+
+
+def DefaultFromEnv(
+    item_parser: Callable[[str], T], default_src: str
+) -> Callable[[str], OptionalType[T]]:  # noqa: D401
+    """An option of type T or a default.
+
+    The default is sourced from an environment variable with the name ``default_src``.
+    Either the option or the default must be provided
+    """
+    default = None
+    env = os.getenv(default_src)
+    if env:
+        default = Optional(item_parser)(env)
+
+    def default_from_env(text: str) -> OptionalType[T]:
+        if text:
+            return Optional(item_parser)(text)
+        if default:
+            return default
+
+        raise ValueError("No value provided.")
+
+    return default_from_env
 
 
 def Optional(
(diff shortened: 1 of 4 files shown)


tests/unit/lib/config_tests.py
```diff
@@ -239,6 +240,41 @@ def test_tupleof_invalid(self):
             parser("a, b")
 
 
+class DefaultFromEnvTests(unittest.TestCase):
+    def __init__(self, name):
+        self.env_var_name = "DASEPLATE_DEFAULT_VALUE"
```
spladug: BASEPLATE_DEFAULT_VALUE or DASEPLATE_DEFAULT_DALUE maybe? :stuck_out_tongue:

baseplate/lib/config.py
```diff
@@ -379,6 +389,30 @@ def tuple_of(text: str) -> Sequence[T]:
     return tuple_of
 
 
+def DefaultFromEnv(
+    item_parser: Callable[[str], T], default_src: str
+) -> Callable[[str], OptionalType[T]]:  # noqa: D401
+    """An option of type T or a default.
+
+    The default is sourced from an environment variable with the name ``default_src``.
```
spladug:
```suggestion
    The default is sourced from an environment variable with the name specified in ``default_src``.
```

(just so people don't think the actual name of the env var is "default_src")

baseplate/sidecars/secrets_fetcher.py
```diff
@@ -377,17 +377,18 @@ def main() -> None:
...
             "vault": {
-                "url": config.String,
+                "url": config.DefaultFromEnv(config.String, "BASEPLATE_DEFAULT_VAULT_URL"),
                 "role": config.String,
                 "auth_type": config.Optional(
                     config.OneOf(**VaultClientFactory.auth_types()),
                     default=VaultClientFactory.auth_types()["aws"],
                 ),
-                "mount_point": config.Optional(config.String, default="aws-ec2"),
+                "mount_point": config.Optional(config.String, default=default_mount_point),
```
spladug: does this hint that maybe DefaultFromEnv should have a "fallback" default too? then it could replace the Optional here

baseplate/lib/config.py
```diff
@@ -379,6 +389,30 @@ def tuple_of(text: str) -> Sequence[T]:
...
+    if env:
+        default = Optional(item_parser)(env)
+
+    def default_from_env(text: str) -> OptionalType[T]:
+        if text:
+            return Optional(item_parser)(text)
+        if default:
+            return default
+
+        raise ValueError("No value provided.")
```
spladug:
```suggestion
        raise ValueError("no value provided")
```

(just for consistency with the general format of error messages in this module and the String config option's error messaging in particular)

baseplate/lib/config.py
```diff
@@ -379,6 +389,30 @@ def tuple_of(text: str) -> Sequence[T]:
...
+    """
+    default = None
+    env = os.getenv(default_src)
+    if env:
+        default = Optional(item_parser)(env)
+
+    def default_from_env(text: str) -> OptionalType[T]:
+        if text:
+            return Optional(item_parser)(text)
+        if default:
```
spladug: tiny nit: can you put a blank between these two separate if branches so they don't read at a distance like they're if/elif pairs please?

spladug (changes requested):
This is great, thank you! I think it makes total sense to get env vars involved here for the reasons you laid out and I like this integration with the existing system. It's very clean.  I've left a few small nit-level comments but overall it looks good. I do think it'd be worth considering adding a fallback to the DefaultFromEnv thing as that seems like it'd end up being a pretty common use case and it'd already be helpful inside this PR.

KTAtkinson:
:haircut_woman:

baseplate/sidecars/secrets_fetcher.py
```diff
@@ -381,13 +381,15 @@ def main() -> None:
...
-                "url": config.String,
+                "url": config.DefaultFromEnv(config.String, "BASEPLATE_DEFAULT_VAULT_URL"),
                 "role": config.String,
                 "auth_type": config.Optional(
                     config.OneOf(**VaultClientFactory.auth_types()),
                     default=VaultClientFactory.auth_types()["aws"],
                 ),
-                "mount_point": config.Optional(config.String, default="aws-ec2"),
+                "mount_point": config.DefaultFromEnv(
+                    config.String, "BASEPLATE_VAULT_MOUNT_POINT", "aws-ec2"
```
spladug:
```suggestion
                    config.String, "BASEPLATE_VAULT_MOUNT_POINT", fallback="aws-ec2"
```

tests/unit/lib/config_tests.py
```diff
@@ -239,6 +240,51 @@ def test_tupleof_invalid(self):
...
+        cur_value = os.getenv(self.unset_env_var_name)
+        self.set_env(self.unset_env_var_name, "")
+        self.addCleanup(self.set_env, self.unset_env_var_name, cur_value)
+
+    def set_env(self, key, value):
+        if key:
+            if value:
+                os.environ[key] = value
+            else:
+                os.environ[key] = ""
```
spladug: [patch.dict can be used for patching the environment](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict)

spladug (approved):
LGTM! Thank you! Left a couple tiny comments but feel free to merge if you'd prefer.

PR #507, opened 2020-10-28
Switch from 'embedding' to 'starting' IPython

kemitche:
Embedding IPython results in issues where import statements and other
globals changes in the shell environment don't work as expected.[1]
As an example, the following commands result in a NameError:

```ipython
In [1]: import time

In [2]: def foo():
   ...:     print(time.time())
   ...:

In [3]: foo()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-c19b6d9633cf> in <module>
----> 1 foo()

<ipython-input-2-a038e7f91c93> in foo()
      1 def foo():
----> 2     print(time.time())
      3

NameError: name 'time' is not defined
```

By switching to use "start_ipython()" instead, we get an environment
more similar to what we get by running `ipython` directly.

```ipython
In [1]: import time

In [2]: def foo():
   ...:     print(time.time())
   ...: 

In [3]: foo()
1603919279.0674398
```

[1] https://github.com/ipython/ipython/issues/62

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -376,7 +376,16 @@ def load_and_run_shell() -> None:
         # try to use IPython if possible
         from IPython import start_ipython
 
-        start_ipython(argv=[], user_ns=env)
+        try:
+            # IPython 5.x+
+            from traitlets.config.loader import Config
+        except ImportError:
+            # IPython 4 and below
+            from IPython import Config
+
+        ipython_config = Config()
+        ipython_config.TerminalInteractiveShell.banner2 = banner
+        start_ipython(argv=[], user_ns=env, config=ipython_config)
         raise SystemExit
     except ImportError:
         import code
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -374,11 +374,10 @@ def load_and_run_shell() -> None:
 
     try:
         # try to use IPython if possible
-        from IPython.terminal.embed import InteractiveShellEmbed
-        from IPython.core.interactiveshell import DummyMod
+        from IPython import start_ipython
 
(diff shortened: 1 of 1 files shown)


baseplate/server/__init__.py
```diff
@@ -374,11 +374,9 @@ def load_and_run_shell() -> None:
 
     try:
         # try to use IPython if possible
-        from IPython.terminal.embed import InteractiveShellEmbed
-        from IPython.core.interactiveshell import DummyMod
-
-        shell = InteractiveShellEmbed(banner2=banner)
```
spladug, line 380: It looks like the banner isn't getting shown.

spladug:
```
ubuntu@reddit:~/src/harold2 (harold2)$ onevm shell
Running “baseplate-shell /home/ubuntu/src/harold2/example.ini”…
INFO:raven.base.Client:Raven is not configured (logging is disabled). Please see the documentation for more information.
Python 3.8.5 (default, Jul 20 2020, 19:48:14) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: context
Out[1]: <baseplate.RequestContext at 0x7fdd587c3610>

```

so the values are there as expected but there's no message explaining how to use 'em.

spladug (changes requested):
Sweet! Thanks for doing this. I think this just needs the banner and then is good to go.

baseplate/server/__init__.py
kemitche: Good catch, let me get the banner back into place! (I'm so used to it at this point I've gloss right over it)

spladug (approved):
LGTM! Thanks so much!

PR #508, opened 2020-10-30
metrics: By default don't log metrics when there is no configured endpoint

bsimpson63:
https://github.com/reddit/baseplate.py/issues/500

Usually the metrics endpoint isn't enabled during local development, so
the debug logging of unsent metrics was very noisy.

👓 @spladug @foreverest

### DIFF
--- a/tests/unit/lib/metrics_tests.py
+++ b/tests/unit/lib/metrics_tests.py
@@ -399,9 +399,9 @@ class HistogramTests(unittest.TestCase):
 
 class MakeClientTests(unittest.TestCase):
     def test_no_endpoint(self):
-        client = metrics.make_client("namespace", None, False)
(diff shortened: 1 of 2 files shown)


tests/unit/lib/metrics_tests.py
```diff
@@ -399,9 +399,9 @@ def test_log_tagged(self):
 
 class MakeClientTests(unittest.TestCase):
     def test_no_endpoint(self):
-        client = metrics.make_client("namespace", None)
+        client = metrics.make_client("namespace", None, False)
```
spladug: can you give these the kwarg please?

spladug (approved):
LGTM! Thank you!

spladug (approved):

pacejackson (approved):

PR #509, opened 2020-11-05
add experiment global cache

suzie-su:
## Description
This is for adding global cache for experiment. Requests could be given a reference to the current global cache and hold onto that for their whole life. when the global cache is invalidated,  when file been updated,  the global cache would be a new {}. so in-flight requests keep pointing at the old one 
## side effect:
    when initial each experiment it has to read the config file first,  and clear the global cache when filewatcher reads from file. this will cause 

- in same request, the config file data will be the same as when its initialized, won't update when request in flight.
- time latency increased when initialize each experiment object.

## Note:
will update the comments later

@pacejackson @spladug

### DIFF
--- a/baseplate/lib/experiments/__init__.py
+++ b/baseplate/lib/experiments/__init__.py
@@ -7,7 +7,6 @@ from typing import Optional
 from typing import overload
 from typing import Sequence
 from typing import Set
-from typing import Tuple
 
 from baseplate import Span
 from baseplate.clients import ContextFactory
@@ -18,7 +17,7 @@ from baseplate.lib.events import DebugLogger
 from baseplate.lib.events import EventLogger
 from baseplate.lib.experiments.providers import parse_experiment
 from baseplate.lib.experiments.providers.base import Experiment
-from baseplate.lib.file_watcher import FileWatcherWithUpdatedFlag
+from baseplate.lib.file_watcher import FileWatcher
 from baseplate.lib.file_watcher import WatchedFileNotAvailableError
 
 
@@ -28,42 +27,6 @@ logger = logging.getLogger(__name__)
 class EventType(Enum):
     EXPOSE = "expose"
     BUCKET = "choose"
-
-
-class ExperimentsGlobalCache:
-    """Experiment global cache store the parsed experiment.
-
-    Every time gets the global cache, it reads the data from FileWatcher,
-    if file has been updated, it assign its cache to new empty dictionary.
-    Global_cache will be passed to each request's experiment object,
-    parsed experiment will be lazily added when each request parse its experiment
-    if it does not exist in the global_cache.
-    """
-
-    def __init__(self, filewatcher: FileWatcherWithUpdatedFlag):
-        self._filewatcher = filewatcher
-        self._global_cache: Dict[str, Optional[Experiment]] = {}
-
-    def _get_config(self) -> Optional[Dict[str, Dict[str, str]]]:
-        try:
-            config_data: Dict[str, Dict[str, str]]
-            config_data, updated = self._filewatcher.get_data()
-
-            if updated and self._global_cache:
-                self._global_cache = {}
-            return config_data
-        except WatchedFileNotAvailableError as exc:
-            logger.warning("Experiment config unavailable: %s", str(exc))
-            return None
-        except TypeError as exc:
-            logger.warning("Could not load experiment config: %s", str(exc))
-            return None
-
-    def get_cfg_and_global_cache(
-        self,
-    ) -> Tuple[Optional[Dict[str, Dict[str, str]]], Dict[str, Optional[Experiment]]]:
-        cfg = self._get_config()
-        return cfg, self._global_cache
 
 
 class ExperimentsContextFactory(ContextFactory):
@@ -91,18 +54,30 @@ class ExperimentsContextFactory(ContextFactory):
         timeout: Optional[float] = None,
         backoff: Optional[float] = None,
     ):
-        self._experiments_global_cache: ExperimentsGlobalCache = ExperimentsGlobalCache(
-            FileWatcherWithUpdatedFlag(path, json.load, timeout=timeout, backoff=backoff)
-        )
+        self._filewatcher = FileWatcher(path, json.load, timeout=timeout, backoff=backoff)
+        self._global_cache: Dict[str, Optional[Experiment]] = {}
         self._event_logger = event_logger
+        self.cfg_mtime = 0.0
 
     def make_object_for_context(self, name: str, span: Span) -> "Experiments":
-        cfg_data, global_cache = self._experiments_global_cache.get_cfg_and_global_cache()
+        config_data: Dict[str, Dict[str, str]] = {}
+        try:
+            config_data, mtime = self._filewatcher.get_data_and_mtime()
+
+            if mtime > self.cfg_mtime:
+                self.cfg_mtime = mtime
+                if self._global_cache:
+                    self._global_cache = {}
+        except WatchedFileNotAvailableError as exc:
+            logger.warning("Experiment config unavailable: %s", str(exc))
+        except TypeError as exc:
+            logger.warning("Could not load experiment config: %s", str(exc))
+
         return Experiments(
             server_span=span,
             context_name=name,
-            cfg_data=cfg_data,
-            global_cache=global_cache,
+            cfg_data=config_data,
+            global_cache=self._global_cache,
             event_logger=self._event_logger,
         )
 
@@ -121,7 +96,7 @@ class Experiments:
         self,
         server_span: Span,
         context_name: str,
-        cfg_data: Optional[Dict[str, Dict[str, str]]],
+        cfg_data: Dict[str, Dict[str, str]],
         global_cache: Dict[str, Optional[Experiment]],
         event_logger: Optional[EventLogger] = None,
     ):
@@ -136,10 +111,6 @@ class Experiments:
             self._event_logger = DebugLogger()
 
     def _get_experiment(self, name: str) -> Optional[Experiment]:
-
-        if not self._cfg_data:
-            return None
-
         if name in self._global_cache:
             return self._global_cache[name]
 
@@ -160,8 +131,6 @@ class Experiments:
 
         :return: List of all valid experiment names.
         """
-        if not self._cfg_data:
-            return []
         experiment_names = list(self._cfg_data.keys())
         return experiment_names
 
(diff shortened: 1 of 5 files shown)


baseplate/lib/file_watcher.py
```diff
@@ -174,3 +175,43 @@ def get_data(self) -> T:
             self._mtime = current_mtime
 
         return typing.cast(T, self._data)
+
+
+class FileWatcherWithUpdatedFlag(FileWatcher):
+    def get_data(self) -> Tuple[T, bool]:
```
spladug: I really like the idea of the `get_data()` function returning the "it's changed" flag! I hadn't thought of that approach to this API. That's clean. 

Would it make sense to return the mtime? It might be a little more informative for clients and also allows cache invalidation to still happen if we missed it the first time we call `get_data()`.

We can't change the signature of the method when subclassing though, as [instances of the subclass have to be able to be used anywhere instances of the superclass can](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

Another way of doing this might be to add a new method on `FileWatcher` called e.g. `get_data_and_mtime()` (or something less ugly) and move all this code in there. Then the old simple `get_data()` can just call that function and drop the mtime part before returning.

baseplate/lib/experiments/__init__.py
```diff
@@ -54,14 +91,18 @@ def __init__(
...
         backoff: Optional[float] = None,
     ):
-        self._filewatcher = FileWatcher(path, json.load, timeout=timeout, backoff=backoff)
+        self._experiments_global_cache: ExperimentsGlobalCache = ExperimentsGlobalCache(
+            FileWatcherWithUpdatedFlag(path, json.load, timeout=timeout, backoff=backoff)
+        )
         self._event_logger = event_logger
 
     def make_object_for_context(self, name: str, span: Span) -> "Experiments":
+        cfg_data, global_cache = self._experiments_global_cache.get_cfg_and_global_cache()
```
spladug: Ooo. So I was originally worried this would mean that we're no longer lazily loading the experiment config only when the current request actually tries to use experiments, but since Baseplate.py 1.0 `make_object_for_context` is only called lazily on use so this is fine and actually makes things nice and clean!

Do we actually need `ExperimentsGlobalCache`? Can we just do that logic right here in `make_object_for_context`?

baseplate/lib/experiments/__init__.py
```diff
@@ -29,6 +30,42 @@ class EventType(Enum):
...
+        try:
+            config_data: Dict[str, Dict[str, str]]
+            config_data, updated = self._filewatcher.get_data()
+
+            if updated and self._global_cache:
+                self._global_cache = {}
+            return config_data
+        except WatchedFileNotAvailableError as exc:
+            logger.warning("Experiment config unavailable: %s", str(exc))
+            return None
```
spladug: Would returning `{}` work? It looks like at least some cases down below (like `get_all_experiment_names()`) wouldn't need extra checks anymore if we do that.

spladug (changes requested):
This is great, thanks for doing this! I really like the idea of having FileWatcher.get_data() return a tuple like that. I've left some comments on ways I think will help simplify even further, but overall I think this is a solid approach and should save us a lot of object creation at runtime.

baseplate/lib/experiments/__init__.py
pacejackson: > Do we actually need ExperimentsGlobalCache? Can we just do that logic right here in make_object_for_context?

`+1`

pacejackson (changes requested):
+1 to everything neil said, this is looking good, just needs a little cleaning up.

suzie-su:
Thank you all for all the suggestions, very good point,
i cannot image I ignored the method signature should be the same.

baseplate/lib/experiments/__init__.py
```diff
@@ -55,13 +55,29 @@ def __init__(
...
 
     def make_object_for_context(self, name: str, span: Span) -> "Experiments":
+        config_data: Dict[str, Dict[str, str]] = {}
+        try:
+            config_data, mtime = self._filewatcher.get_data_and_mtime()
+
+            if mtime > self.cfg_mtime:
+                self.cfg_mtime = mtime
+                if self._global_cache:
+                    self._global_cache = {}
```
spladug:
```suggestion
                self._global_cache = {}
```

I think we want to replace the cache regardless of if it's currently empty. 

1. Request A comes in. It tries to fetch an experiment that doesn't exist. This triggers reading the experiments config into memory but doesn't put anything in the cache.
2. Request A blocks on some IO like a database call and yields to the event loop.
2. Request B comes in. It tries to do some bucketing and finds the experiment config has changed. It re-reads from disk but because the global cache is currently empty it doesn't make a new cache object. Both request A and B are now pointing at the same cache despite having different config data. 
3. Request B now blocks on IO and yields.
4. Request A wakes back up and tries to do some more bucketing. This time it buckets for an experiment that _does_ exist and it places the data it calculates into the global cache. 
5. Request A finishes up.
6. Request B wakes back up and tries to use that same experiment. It's in the cache! But oh no, it's based on the old file. We've poisoned the cache with stale data.

Replacing the cache unconditionally is probably safest.

spladug (changes requested):
just the one last thing about always invalidating the cache.

this looks great. thank you so much!

baseplate/lib/experiments/__init__.py
suzie-su: Yea That is true. for empty global cache with file been updated between 2 req fetch files.
A fetch gets GC_A, file updated, B fetch from file with GC is empty, gets GC_A.then A parse and insert into old experiment config. GC will keep the old experiment of A.
the reason i do this coz there would be another case i thought about (maybe this case has less side effect)
A reads the data from file, check mtime > 0, yield
B reads the data from file,  check mtime > 0, yield
A gets GC_A {}
if we always update GC
B gets GC_B {}
then might end up some req always gets different version of GC update to its own GC, 

how can we handle this case.

suzie-su:
just reply the comment

baseplate/lib/experiments/__init__.py
spladug: When we're running in gevent, rather than using real threads, we don't yield except at points of blocking IO.  So we should be safe from yielding happening anywhere between `get_data_and_mtime()` returning and `make_object_for_context()` returning. If we always replace the cache when the file has changed, we're sure we never have stale data in the cache. The worst case with real threading is that we overzealously invalidate the cache which just means a bit of extra CPU work but not any invalid data. I think that sounds fair?

spladug (changes requested):
that sound ok?

baseplate/lib/experiments/__init__.py
suzie-su: That makes a lot of sense. yea I totally agree and definitely did not think about this stale cache issue before.
Great catch and cannot say more thank you for this guidance and help here.
Learned a  lot from this. Really appreciate.

spladug (approved):
looks great! thank you so much!

suzie-su:
Thank you Neil for ALL the help and guidance here.

baseplate/lib/experiments/__init__.py
```diff
@@ -78,54 +93,44 @@ class Experiments:
 
     def __init__(
         self,
-        config_watcher: FileWatcher,
         server_span: Span,
         context_name: str,
+        cfg_data: Dict[str, Dict[str, str]],
+        global_cache: Dict[str, Optional[Experiment]],
```
pacejackson: Note, this is _technically_ a breaking change if anyone calls this manually, we might want to set default values for these (`None`) and just create new dicts in the constructor if they aren't set?

pacejackson (approved):
LGTM modulo than the concern about adding new arguments being technically a breaking change

baseplate/lib/experiments/__init__.py
suzie-su: but this also removed the config_watcher , for `not a breaking change` should we need to add that back as well?

suzie-su:
sorry Andrew, just wanna be more clear this. since file_watcher been removed as well. it no matter what gonna be a breaking change right? does default {} necessary?

baseplate/lib/experiments/__init__.py
spladug: Blerg. We technically haven't documented the parameters of this constructor, so could make an argument it's an internal interface, but it does seem to be getting some use: particularly in tests.

For full backwards compatibility, I guess we need to make it able to take a filewatcher and fetch the data from there if cfg_data is None. Does that sound OK? :/

pacejackson: Ya, not ideal but probably something that we want to do now and document for removal plus "warn_deprecated" maybe?

spladug: sgtm. `warn_deprecated` that it'll be removed in 2.0

suzie-su:
💇

baseplate/lib/experiments/__init__.py
```diff
@@ -81,51 +97,59 @@ def __init__(
...
+        self._cfg_data = cfg_data
+        self._global_cache = global_cache if global_cache is not None else {}
         if event_logger:
             self._event_logger = event_logger
         else:
             self._event_logger = DebugLogger()
 
-    def _get_config(self, name: str) -> Optional[Dict[str, str]]:
+    def _get_config(self) -> Dict[str, Dict[str, str]]:
+        """Warn: Deprecate in Baseplate 2.0."""
```
spladug:
```suggestion
        # Warn: Deprecate in Baseplate 2.0.
```

since this is more of an internal comment, and not something we'd want to show in public documentation, it should just be a comment rather than a docstring

baseplate/lib/experiments/__init__.py
```diff
@@ -81,51 +97,59 @@ def __init__(
         config_watcher: FileWatcher,
```
spladug, line 97: do we want to make this optional now so people can start migrating to 2.0 style before it's required?

spladug (approved):
lgtm modulo those optional things! it's a bummer how much extra code the compatibility requires, but such is life. baseplate 2.0 here we come!

pacejackson (approved):

baseplate/lib/experiments/__init__.py
suzie-su: the point is i wonder if people just initiate the object without argument 
Experiment(filewatcher, span, name)?

spladug (approved):

PR #512, opened 2020-11-10
First testing fakes, more mtimes, and authn token validator public key caching

### LINKED ISSUE #477: Cache authn service public key in auth token validation
It's not done for us by pyjwt: https://github.com/jpadilla/pyjwt/issues/204

We do need to make sure it gets invalidated when the underlying secret rotates.

spladug:
This is probably best to go commit-by-commit on.

This builds on @suzie-su's improvements to `FileWatcher`'s API in #509. The `SecretsStore` now has similar `_and_mtime` variants of most of its methods. Once that's in place we can cache the authn token public keys which saves a bit of processing time every request that we validate a token.

This also brings in the first of some testing fakes from Harold2. I did this in this PR because it makes updating the test suite for the _mtime() stuff much nicer -- we drop a lot of horrid deep mocking in favor of the simple fakes.

A [quick micro-benchmark](https://gist.github.com/spladug/3121c6664378384889cb841bb913517b) puts this at an order of magnitude improvement: from ~600μsec per token to ~80μsec.

Fixes #477.

### DIFF
--- a/baseplate/lib/edge_context.py
+++ b/baseplate/lib/edge_context.py
@@ -3,12 +3,14 @@ import re
 
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import NamedTuple
 from typing import Optional
 from typing import Set
 
 import jwt
 
+from jwt.algorithms import get_default_algorithms
 from thrift import TSerialization
 from thrift.protocol.TBinaryProtocol import TBinaryProtocolAcceleratedFactory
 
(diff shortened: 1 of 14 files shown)


pacejackson (approved):

MelissaCole:
:fish: love the testing fakes!

bradengroom (approved):
I'm excited about these testing fakes.

PR #513, opened 2020-11-10
Cache JWT key parses

ketralnis:
`jwt.decode` takes key data as a `bytes` but parsing that into an
`RSAPublicKey` isn't free. This uses `functools.lru_cache` to try to
avoid doing that work on every request

Addresses https://github.com/reddit/baseplate.py/issues/477

👓 @spladug

### DIFF
--- a/baseplate/lib/edge_context.py
+++ b/baseplate/lib/edge_context.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import re
 
@@ -50,8 +51,9 @@ class AuthenticationTokenValidator:
             return InvalidAuthenticationToken()
 
         secret = self.secrets.get_versioned("secret/authentication/public-key")
-        for public_key in secret.all_versions:
+        for public_key_data in secret.all_versions:
             try:
+                public_key = self._prepare_key(public_key_data)
                 decoded = jwt.decode(token, public_key, algorithms="RS256")
                 return ValidatedAuthenticationToken(decoded)
             except jwt.ExpiredSignatureError:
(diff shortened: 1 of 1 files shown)


baseplate/lib/edge_context.py
```diff
@@ -61,6 +63,10 @@ def validate(self, token: bytes) -> "AuthenticationToken":
 
         return InvalidAuthenticationToken()
 
+    @functools.lru_cache()
```
ketralnis, line 66: the default size is 256 which is way bigger than we need. this also keeps old key data around in memory even after we (can) know we'll never need it again

baseplate/lib/edge_context.py
```diff
@@ -61,6 +63,10 @@ def validate(self, token: bytes) -> "AuthenticationToken":
...
+    def _prepare_key(self, key_data: bytes) -> "jwt.algorithms.RSAPublicKey":
```
ketralnis, line 67: this fails mypy because the import of `jwt.algorithms.RSAPublicKey` is internally conditional on `cryptography` being installed and it looks like mypy doesn't run that way?

ketralnis:
This approach is following the wording of #477 but a more comprehensive approach might be letting the FileWatcher/SecretsFetcher setup take a function that transforms the data. That would let us only run that transform when the data changes instead of papering a cache over the top of it. That's a little harder to rig into the model we have though

spladug:
Aw crap, I'm so sorry. I should have marked this one as in progress.  I just opened #512 which addresses this using the changes @suzie-su did to FileWatcher recently.

spladug:
The lru_cache is super clever. I didn't think of that.

baseplate/lib/edge_context.py
spladug: Ah weird. Yeah, mypy must not know what to do with that conditional. The linters should have cryptography installed.

ketralnis:
No worries, killing 🔪

PR #515, opened 2020-11-11
server/queue_consumer: Handle ServerTimeout to terminate the server

bsimpson63:
Previously we were only handling `Exception`, but `ServerTimeout` inherits
from `BaseException` so it was not caught. This meant that a server timeout
would crash the pump or handler thread but not terminate the server, so the
crashed thread would never be restarted.

https://github.com/reddit/baseplate.py/issues/485

### DIFF
--- a/baseplate/server/queue_consumer.py
+++ b/baseplate/server/queue_consumer.py
@@ -23,6 +23,7 @@ from gevent.server import StreamServer
 import baseplate.lib.config
 
 from baseplate.lib.retry import RetryPolicy
+from baseplate.observers.timeout import ServerTimeout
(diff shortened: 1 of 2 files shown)


spladug (approved):
nice, thanks!

bsimpson63:
do you think it's worth adding a test for this?

pacejackson (approved):

spladug:
Yeah, it probably is. I can see it potentially being a bit hard to test since it's a confluence of two things, but if it's possible and not too messy that'd be great please.

spladug (approved):

Issue #396, opened 2020-01-23
Add user-agent headers to requests from sidecars


spladug:
And include the baseplate.py version number in it

ketralnis:
Addressed by #516

PR #516, opened 2020-11-13
Add User-Agent to HTTP requests made by sidecars

ketralnis:
This adds a `User-Agent` HTTP header to outbound requests made by the sidecars. An example looks like looks like `baseplate.py-ZipkinPublisher/1.4.1.dev22+gea38896.d20201112`. 

Addresses #396 (Add user-agent headers to requests from sidecars)

👓 @spladug

### DIFF
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -12,6 +12,7 @@ from typing import Optional
 
 import requests
 
+from baseplate import __version__ as baseplate_version
 from baseplate.lib import config
 from baseplate.lib import metrics
 from baseplate.lib.events import MAX_EVENT_SIZE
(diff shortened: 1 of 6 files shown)


ketralnis:
Something I'd like to do but isn't done here is to not say `baseplate-ZipkinPublisher-1.4` but instead say `comment-service-ZipkinPublisher (baseplate-1.4)`. I think this is an okay first pass though

baseplate/sidecars/event_publisher.py
```diff
@@ -92,6 +93,9 @@ def __init__(self, metrics_client: metrics.Client, cfg: Any):
         self.key_name = cfg.key.name
         self.key_secret = cfg.key.secret
         self.session = requests.Session()
+        self.session.headers.update(
```
spladug: :no_bell: is there any advantage to doing this with `update()` over just `headers["User-Agent"] = "..."` ?

spladug (approved):
awesome, thanks for doing this!

baseplate/sidecars/event_publisher.py
ketralnis: Probably none at all :) Will change

baseplate/sidecars/event_publisher.py
```diff
@@ -92,6 +93,9 @@ def __init__(self, metrics_client: metrics.Client, cfg: Any):
...
+            {"User-Agent": f"baseplate-{self.__class__.__name__}/{baseplate_version}"}
```
bradengroom: Should this specify that it's `baseplate.py` since we also have `baseplate.go` now?

bradengroom (approved):

PR #517, opened 2020-11-13
Add list of exceptions for raven to always filter out

### LINKED ISSUE #524: Some errors aren't useful for debugging
Some errors being sent to Sentry aren't useful from a debugging perspective. These errors should be filtered out. The errors in question include:

```
ConnectionError / ConnectionRefusedError / ConnectionResetError (if these have a nice hierarchy of exceptions and sentry's smart enough, maybe just ConnectionError would be sufficient)
HTTPError / HTTPRequestTimeout / HTTPServiceUnavailable (same)
OperationalError
TTransportException / TApplicationException / TProtocolException
MemcachedServerError
```

KTAtkinson:
Add a default list of exceptions to Baseplate Sentry error observer. This is outlined in [this design document](https://docs.google.com/document/d/1n9VCZZz0XhWXvZM2wWvsiR8oi1LNk7-AsShsrbt4B6Y/edit?usp=sharing). [This](https://github.snooguts.net/reddit/design-docs/pull/994#issuecomment-235670) outlines the list of errors that should be filtered out.

## ConnectionError / ConnectionRefusedError / ConnectionResetError
These errors I addressed by filtering out the generic `ConnectionError`

## HTTPError / HTTPRequestTimeout / HTTPServiceUnavailable
I set Sentry to filter out `requests.HTTPError`. These were the flavor of `HTTPError` which I could find in Sentry. Both `HTTPRequestTimeout` and 'HTTPServiceUnavailable` seem to be errors with stem from GraphQLs codebase. These are probably not ones we want to filter out.

## OperationalError
I couldn't find this error in Sentry to figure out where it comes from.

## TTransportException / TApplicationException / TProtocolException
I decided to filter out `TException` which includes all Thrift errors.

## MemcachedServerError
Same as above, I chose to filter out `MemcachedError`.

## RedisClusterException
This error seems to come from some the `reddit-py-cluster` library. Modern baseplate doesn't depend on this anymore.

### DIFF
--- a/baseplate/observers/sentry.py
+++ b/baseplate/observers/sentry.py
@@ -10,11 +10,6 @@ from typing import TYPE_CHECKING
 
 import raven
 
-from pymemcache.exceptions import MemcacheError
-from requests import HTTPError
-from requests import Timeout
-from thrift.Thrift import TException
-
 from baseplate import _ExcInfo
 from baseplate import BaseplateObserver
 from baseplate import RequestContext
@@ -28,18 +23,21 @@ if TYPE_CHECKING:
 
 
 ALWAYS_IGNORE_EXCEPTIONS = (
-    ConnectionError,
-    HTTPError,
-    Timeout,
-    TException,
-    MemcacheError,
+    "ConnectionError",
+    "ConnectionRefusedError",
+    "ConnectionResetError",
+    "HTTPError",
+    "TApplicationException",
+    "TProtocolException",
+    "TTransportException",
+    "MemcacheServerError",
 )
 
 
 def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -> raven.Client:
     """Configure and return a error reporter.
 
-    This expects one configuration option and can take many optional ones:
+    This expects one uration option and can take many optional ones:
 
     ``sentry.dsn``
         The DSN provided by Sentry. If blank, the reporter will discard events.
--- a/baseplate/observers/sentry.py
+++ b/baseplate/observers/sentry.py
@@ -10,6 +10,11 @@ from typing import TYPE_CHECKING
 
 import raven
 
+from pymemcache.exceptions import MemcacheError
+from requests import HTTPError
+from requests import Timeout
+from thrift.Thrift import TException
+
 from baseplate import _ExcInfo
 from baseplate import BaseplateObserver
 from baseplate import RequestContext
@@ -18,9 +23,17 @@ from baseplate import Span
 from baseplate.lib import config
 from baseplate.observers.timeout import ServerTimeout
 
-
 if TYPE_CHECKING:
     from gevent.hub import Hub
+
+
+ALWAYS_IGNORE_EXCEPTIONS = (
+    ConnectionError,
+    HTTPError,
+    Timeout,
+    TException,
+    MemcacheError,
+)
 
 
 def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -> raven.Client:
@@ -67,7 +80,12 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
                 "environment": config.Optional(config.String, default=None),
                 "include_paths": config.Optional(config.String, default=None),
                 "exclude_paths": config.Optional(config.String, default=None),
-                "ignore_exceptions": config.Optional(config.TupleOf(config.String), default=[]),
+                "ignore_exceptions": config.Optional(
+                    config.TupleOf(config.String), default=[]
+                ),  # Depricated in favor of `additional_ignore_exception
+                "additional_ignore_exceptions": config.Optional(
+                    config.TupleOf(config.String), default=[]
+                ),
                 "sample_rate": config.Optional(config.Percent, default=1),
                 "processors": config.Optional(
                     config.TupleOf(config.String),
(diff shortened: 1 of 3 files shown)


baseplate/observers/sentry.py
```diff
@@ -18,11 +23,19 @@
...
     from gevent.hub import Hub
 
 
+ALWAYS_IGNORE_EXCEPTIONS = (
+    ConnectionError,
+    HTTPError,
+    Timeout,
+    TException,
+    MemcacheError,
+)
```
spladug: It's kinda weird, but I think we want to do this with stringy names of exceptions rather than the actual exception classes right now. Unfortunately, [raven<6.1 doesn't support doing it with classes](https://github.com/reddit/baseplate.py/commit/070cc2d654fb3fcc200e67c93a880282d0a6d096) and we can't bump a minimum version until we go to v2.0 baseplate (since it's a breaking change)

baseplate/observers/sentry.py
```diff
@@ -10,6 +10,11 @@
 
 import raven
 
+from pymemcache.exceptions import MemcacheError
```
spladug: :no_bell: moot since we need to use strings anyway, but I think we'd want to be more careful about importing this as we don't require pymemcache to be installed for a service unless it's using that integration.

baseplate/observers/sentry.py
```diff
@@ -67,7 +80,12 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
                 "environment": config.Optional(config.String, default=None),
                 "include_paths": config.Optional(config.String, default=None),
                 "exclude_paths": config.Optional(config.String, default=None),
-                "ignore_exceptions": config.Optional(config.TupleOf(config.String), default=[]),
+                "ignore_exceptions": config.Optional(
+                    config.TupleOf(config.String), default=[]
+                ),  # Depricated in favor of `additional_ignore_exception
```
spladug:
```suggestion
                ),  # Deprecated in favor of additional_ignore_exception
```

If it's deprecated, should we do a [`warn_deprecated`](https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/__init__.py#L12-L20) when it's defined?

spladug (changes requested):
Thanks! Looks good. Main blocker is the stringy thing for raven<6.1 compatibility.

baseplate/observers/sentry.py
KTAtkinson: Is there a way we can capture this in the tests?

baseplate/observers/sentry.py
```diff
@@ -89,6 +106,25 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
         else:
             break
 
+    cfg_ignore_exceptions = cfg.sentry.ignore_exceptions
+    cfg_additional_ignore_exceptions = cfg.sentry.additional_ignore_exceptions
+
+    if cfg_ignore_exceptions != None:
+        warn_deprecated(
+            "'ignore_exceptions' configuration varible is depricated."
+            + "Please use 'additional_ignore_exceptions' instead.",
```
spladug:
```suggestion
            "Please use 'sentry.additional_ignore_exceptions' instead.",
```

kinda weird but python will just auto-concatenate adjacent strings like this so we don't need a `+` here

```
>>> str("a" "b")
'ab'
```

baseplate/observers/sentry.py
```diff
@@ -89,6 +106,25 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
         else:
             break
 
+    cfg_ignore_exceptions = cfg.sentry.ignore_exceptions
+    cfg_additional_ignore_exceptions = cfg.sentry.additional_ignore_exceptions
+
+    if cfg_ignore_exceptions != None:
+        warn_deprecated(
+            "'ignore_exceptions' configuration varible is depricated."
```
spladug:
```suggestion
            "'sentry.ignore_exceptions' is deprecated. "
```

baseplate/observers/sentry.py
```diff
@@ -16,17 +16,29 @@
...
+    "TTransportException",
+    "MemcacheServerError",
+)
+
+
 def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -> raven.Client:
     """Configure and return a error reporter.
 
-    This expects one configuration option and can take many optional ones:
+    This expects one uration option and can take many optional ones:
```
bradengroom: typo?

spladug:
```suggestion
    This expects one configuration option and can take many optional ones:
```

baseplate/observers/sentry.py
```diff
@@ -89,6 +106,25 @@ def error_reporter_from_config(raw_config: config.RawConfig, module_name: str) -
         else:
             break
 
+    cfg_ignore_exceptions = cfg.sentry.ignore_exceptions
+    cfg_additional_ignore_exceptions = cfg.sentry.additional_ignore_exceptions
+
+    if cfg_ignore_exceptions != None:
```
spladug:
```suggestion
    if cfg_ignore_exceptions:
```

since it defaults to `[]` would this have ever been `None`? if we do expect `None` it'd be more pythonic to write `is not None`

spladug (changes requested):
just some small nits. can this be taken out of draft mode now? thanks for doing this!

KTAtkinson:
Update PR to be 'In Review' instead of 'Draft'.

spladug (approved):
great! thank you! i'm sure sentry will breathe a sigh of relief as this rolls out :)

bradengroom (approved):

PR #526, opened 2020-11-17
Add ability for live data fetcher to create directories

### LINKED ISSUE #521: live_data_fetcher can't create directories
If the live data fetcher is told to write live data to a directory that doesn't exist it fails.

KTAtkinson:
Allow the live data fetcher to create output directories if they don't exist.

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -26,8 +26,6 @@ HEARTBEAT_INTERVAL = 300
 
 class NodeWatcher:
     def __init__(self, dest: str, owner: int, group: int, mode: int):
-        pathlib.Path(dest).parent.mkdir(parents=True, exist_ok=True)
-
         self.dest = dest
         self.owner = owner
         self.group = group
@@ -46,6 +44,7 @@ class NodeWatcher:
         # swap out the file atomically so clients watching the file never catch
         # us mid-write.
         logger.info("Updating %r", self.dest)
+        pathlib.Path(self.dest).parent.mkdir(parents=True, exist_ok=True)
         with open(self.dest + ".tmp", "wb") as tmpfile:
             if self.owner and self.group:
                 os.fchown(tmpfile.fileno(), self.owner, self.group)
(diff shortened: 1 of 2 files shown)


baseplate/sidecars/live_data_watcher.py
```diff
@@ -3,18 +3,18 @@
...
 from typing import NoReturn
 
-from kazoo.client import KazooClient
-from kazoo.protocol.states import ZnodeStat
-
 from baseplate.lib import config
 from baseplate.lib.live_data.zookeeper import zookeeper_client_from_config
 from baseplate.lib.secrets import secrets_store_from_config
+from kazoo.client import KazooClient
+from kazoo.protocol.states import ZnodeStat
```
KTAtkinson: For some reason `reorder-python-imports` wanted this change even though I didn't add these imports.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -25,6 +25,8 @@
 
 class NodeWatcher:
     def __init__(self, dest: str, owner: int, group: int, mode: int):
+        pathlib.Path(dest).parent.mkdir(parents=True, exist_ok=True)
+
```
KTAtkinson: It's kind of odd to me to have this in the `__init__`, but it seemed to work best here. It sort of makes sense that the object which writes to a file should be concerned that the destination directory exists. I also wanted to avoid doing this each time there is a change, as this needs to be done at most once per execution.

spladug: I think we might actually want to do this every time. What happens if the directory is deleted?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -3,6 +3,7 @@
 import configparser
 import logging
 import os
+import pathlib
```
spladug: might be nice to `from pathlib import Path` like you do in the tests. would look a bit nicer below.

spladug (changes requested):
nice! just the one thing. thank you!

baseplate/sidecars/live_data_watcher.py
KTAtkinson: That is true, I hadn't thought of that :+1: I'll move it.

KTAtkinson:
:haircut:

spladug (approved):
LGTM! thanks!

PR #528, opened 2020-11-18
Configure metrics endpoint and namespace through env

### LINKED ISSUE #522: trace publisher can't get configuration through the environment
It would be great if the trace publisher could get it's configuration through the environment. Generally many applications send data to the same trace publisher hostname, so there is not reason for each application to explicitly set it.

KTAtkinson:
This facilitates being able to configure the trace publisher through the environment. The metrics collector configuration is used by the trace publisher.

### DIFF
--- a/baseplate/lib/config.py
+++ b/baseplate/lib/config.py
@@ -390,7 +390,28 @@ def TupleOf(item_parser: Callable[[str], T]) -> Callable[[str], Sequence[T]]:  #
 
 
 def DefaultFromEnv(
-    item_parser: Callable[[str], T], default_src: str, fallback: OptionalType[T] = None
+    item_parser: Callable[[str], T], default_src: str, fallback: OptionalType[T] = None,
+) -> Callable[[str], OptionalType[T]]:  # noqa: D401
+    """An option of type T or a default.
+
+    The default is sourced from an environment variable with the name specified in ``default_src``.
+    If the environment variable is not set, then the fallback will be used.
+    One of the following values must be provided: fallback, default_src, or the provided configuration
+    """
+    parser = OptionalDefaultFromEnv(item_parser, default_src, fallback)
+
+    def default_from_env(text: str) -> OptionalType[T]:
+        val = parser(text)
+        if val:
+            return val
+
+        raise ValueError("Value is required.")
+
+    return default_from_env
+
+
+def OptionalDefaultFromEnv(
+    item_parser: Callable[[str], T], default_src: str, fallback: OptionalType[T] = None,
 ) -> Callable[[str], OptionalType[T]]:  # noqa: D401
     """An option of type T or a default.
 
@@ -402,11 +423,7 @@ def DefaultFromEnv(
     default = Optional(item_parser, fallback)(env)
 
     def default_from_env(text: str) -> OptionalType[T]:
-        val = Optional(item_parser, default)(text)
-        if val:
-            return val
-
-        raise ValueError("no value provided")
+        return Optional(item_parser, default)(text)
 
     return default_from_env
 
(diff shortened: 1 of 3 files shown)


baseplate/lib/config.py
```diff
@@ -390,23 +390,40 @@ def tuple_of(text: str) -> Sequence[T]:
...
         if val:
             return val
 
-        raise ValueError("no value provided")
+        raise ValueError("Value is required.")
+
+    return default_from_env
+
+
+def OptionalDefaultFromEnv(
```
spladug, line 413: Do we need this? I think we can just do it with composition: `Optional(DefaultFromEnv(...))`?

baseplate/lib/metrics.py
```diff
@@ -563,8 +563,12 @@ def metrics_client_from_config(raw_config: config.RawConfig) -> Client:
         raw_config,
         {
             "metrics": {
-                "namespace": config.Optional(config.String, default=""),
-                "endpoint": config.Optional(config.Endpoint),
+                "namespace": config.OptionalDefaultFromEnv(
```
spladug: Note that `namespace` is soft-deprecated in favor of configuring this in telegraf. In fact, when doing tagged metrics `namespace` is disallowed.

spladug (changes requested):

baseplate/lib/metrics.py
KTAtkinson: I'll remove this one then :)

baseplate/lib/config.py
KTAtkinson: The issue with using `Optional(DefaultFromEnv(...))` is that `DefaultFromEnv(...)` injects defaults. The problematic code in `Optional` is in the parser function:
```
if text:
    return item_parse(text)

return default
```

In the case of `DefaultFromEnv` we need the item parser to always parse text, not only when text is provided. We could re-write `Optional(...) like this to account for this:

```
try:
    return item_parser(text)
except ValueError:
    return default
```
There are a few downsides to this implementation. First, `ValueError` is not exclusively used to mean that a parser didn't receive a valid value, it's also used to indicate the parser received an invalid value. Also there is some odd semantics is using `Optional` with `DefaultFromEnv`, as `DefaultFromEnv` already has a fallback. This creates two layers of defaults.

What do you think is the best way forward here?

spladug: Got it. Good point. What you have is probably best then.

baseplate/lib/config.py
```diff
@@ -390,23 +390,40 @@ def tuple_of(text: str) -> Sequence[T]:
...
+    parser = OptionalDefaultFromEnv(item_parser, default_src, fallback)
 
     def default_from_env(text: str) -> OptionalType[T]:
-        val = Optional(item_parser, default)(text)
+        val = parser(text)
         if val:
             return val
 
-        raise ValueError("no value provided")
+        raise ValueError("Value is required.")
```
spladug: can this stay how it was before please? just for consistency with other error messaging in this file. (it's not a whole sentence, it gets concatenated into one)

spladug (approved):
with one capitalization nit

tests/unit/lib/config_tests.py
```diff
@@ -265,6 +265,26 @@ def test_fallback(self):
...
+    def test_use_optional_default_from_env(self):
+        parser = config.OptionalDefaultFromEnv(config.String, "BASEPLATE_DEFAULT_VALUE")
+        self.assertEqual(parser(""), "default")
+
+    def test_empty_default(self):
+        parser = config.DefaultFromEnv(config.String, "NOT_PROVIDED")
+        self.assertEqual(parser("foo"), "foo")
+
+    def test_use_provided(self):
+        parser = config.DefaultFromEnv(config.String, "BASEPALTE_DEFAULT_VALUE")
```
spladug:
```suggestion
        parser = config.DefaultFromEnv(config.String, "BASEPLATE_DEFAULT_VALUE")
```

PR #530, opened 2020-11-24
requests: pass through verify in send() kwargs

curioussavage:
This allows development clients to make local certs work. Right now it looks like many services have verify turned off, in addition to using their own implementation a requests client. but this will at least let them turn it on when they migrate to the built in client.

Would it be ok to also just do the following?

```python
verify = os.environ.get('REQUESTS_CA_BUNDLE', True)
...
"verify": kwargs.pop("verify", verify),
```

### DIFF
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -190,6 +190,7 @@ class BaseplateSession:
         send_kwargs = {
             "timeout": kwargs.pop("timeout", None),
             "allow_redirects": kwargs.pop("allow_redirects", None),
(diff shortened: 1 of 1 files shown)


spladug (approved):
I think this is probably necessary to allow a proper escape hatch for services, but I don't think we actually want to do `verify=False` in most situations do we? OneVM sets up the right bundles for you andI I imagine snoodev would be fine as well. It's super dangerous to turn off cert verification.

curioussavage:
yeah I think it is mostly because people copied an implementation of a requests baseplate client that just defaulted it to off in dev.

PR #533, opened 2020-12-08
Add pool_size and max_overflow to SQLAlchemy config parser

curioussavage:
For context: in post service after adding graphs for the pool usage metrics we saw that we were hitting our max overflow often so we started increasing pool size.

I've seen usage of these in in other services too.


spladug (approved):

spladug:
thanks!

curioussavage:
@spladug do I just merge whenever?

spladug:
yup! go for it!

PR #534, opened 2021-01-04
Spring cleaning

spladug:
This does a bunch of housecleaning, much of it breaking changes. But that's OK as we're heading towards 2.0 this release cycle!

While the HVAC library is used in a few places internally, the baseplate client doesn't appear to have any users. So that should be safe to delete. The other deprecation removal gets rid of almost all uses of `warn_deprecated` inside the library, except trace publishing because we're in flux on if we like sidecars for that purpose.

I'm also dropping support for Python 3.6 as it's going EOL at the end of this year, few services are still using it, nothing's blocking them from upgrading once they're on k8s, and 2.0 is a good time to do it.


spladug:
cc: @pacejackson fyi the deprecated queue consumer stuff is going away now!

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -309,39 +296,6 @@ def start_server_span(
...
-    To use it, add a single line to your app's section in its INI file:
-
-        [app:your_app]
-        use = egg:baseplate
-
-    """
-    return make_app(local_config)
-
-
-def pshell_setup(env: Dict[str, Any]) -> None:
```
MelissaCole, line 327: Do you want to remove the reference to `pshell_setup` in the `includeme.start_server_span` comment?

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -309,39 +296,6 @@ def start_server_span(
         config.add_request_method(start_server_span, "start_server_span")
 
 
-def paste_make_app(_: Dict[str, str], **local_config: str) -> Any:
```
MelissaCole, line 312: :knife: `"paste.app_factory": ["main = baseplate.frameworks.pyramid:paste_make_app"],` in setup.py?

MelissaCole:
:fish: looks great! :hocho: :hocho: :hocho:

baseplate/frameworks/pyramid/__init__.py
spladug: done!

baseplate/frameworks/pyramid/__init__.py
spladug: nice! tore that stuff out altogether and simplified startup as a result.

bradengroom:
🐟

PR #535, opened 2021-01-04
Make UnixUser/UnixGroup fall back to IDs if name isn't found

spladug:
This allows us to use numeric IDs in contexts that don't have proper /etc/passwd or /etc/group configuration, like docker containers. We look up by name first, and fall back to decimal UID/GID.


MelissaCole:
:fish: lgtm!

PR #536, opened 2021-01-05
Remove deprecated observer configuration functions

spladug:
configure_observers() is the one-stop-shop now! Since some observers like the server timeout mechanism only come through this method, it's best to force people to move away from the older methods when upgrading.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -12,7 +12,6 @@ from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
 from typing import Type
-from typing import TYPE_CHECKING
 
 import gevent.monkey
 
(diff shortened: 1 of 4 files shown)


bradengroom (approved):

baseplate/__init__.py
```diff
@@ -302,103 +295,7 @@ def register(self, observer: BaseplateObserver) -> None:
...
-        """
-        # pylint: disable=cyclic-import
-        from baseplate.observers.metrics_tagged import TaggedMetricsBaseplateObserver
-
-        self._metrics_client = metrics_client
-        self.register(
-            TaggedMetricsBaseplateObserver.from_config_and_client(self._app_config, metrics_client)
-        )
-
-    def configure_metrics(self, metrics_client: metrics.Client) -> None:
```
MelissaCole, line 335: `configure_metrics` is referred to in `serve.rst`

MelissaCole:
:fish: lgtm!

baseplate/__init__.py
spladug: eek, thanks! fixed

PR #540, opened 2021-01-09
Extract edge context into its own library

spladug:
I spent some time thinking about how the concept of automatically propagated headers might be extended, and ended up deciding that having a single edge context header name that baseplate knows while letting the content of it be defined elsewhere is the cleanest way to do this in our current setup. So... this just reduces baseplate's knowledge of what's inside the context to nothing and relies entirely on a single factory interface for parsing them. The rest happens in the library. This will thankfully be a pretty simple change in services using it!

https://github.com/reddit/edgecontext.py (more to be done there)


spladug:
@fishy tagging you on this to make sure you're OK with the extraction of this stuff from the main baseplate thrift file, since I imagine that has implications for baseplate.go.

fishy (approved):
As long as the new `edgecontext.thrift` file is still open source, :fish:

MelissaCole:
:fish: lgtm!

pacejackson (approved):

PR #542, opened 2021-01-13
do not disable existing loggers when reading log config with fileConfig

Tecuya:
When logging options are read from configuration using `logging.config.fileConfig`, we should pass the `disable_existing_loggers=False` option to prevent it from interfering with loggers created by users.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -134,7 +134,7 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     sentry_logger.setLevel(logging.WARNING)
 
     if config.has_logging_options:
-        logging.config.fileConfig(config.filename)
+        logging.config.fileConfig(config.filename, disable_existing_loggers=False)
 
 
 def make_listener(endpoint: EndpointConfiguration) -> socket.socket:


spladug:
Oof, yeah it seems bad if it's knocking out your module-level loggers. 

I do think we want to clear out older loggers when switching to `fileConfig` though. Otherwise the StreamHandler and stuff configured in the lines above that call will stick around and we'll get double logging, right?

For a `baseplate-serve` app, we import the application _after_ `configure_logging` is called so the only loggers configured before calling `fileConfig` should be baseplate's own. For `baseplate-script` it looks like we do the import as part of argument parsing, which means your loggers are defined before we wipe them out. Oops.

So, I'm thinking maybe the right answer is to defer import of the application until after `configure_logging()` is called in the `baseplate-script` path. How does that sound?

spladug (changes requested):
Oof, yeah it seems bad if it's knocking out your module-level loggers. 

I do think we want to clear out older loggers when switching to `fileConfig` though. Otherwise the StreamHandler and stuff configured in the lines above that call will stick around and we'll get double logging, right?

For a `baseplate-serve` app, we import the application _after_ `configure_logging` is called so the only loggers configured before calling `fileConfig` should be baseplate's own. For `baseplate-script` it looks like we do the import as part of argument parsing, which means your loggers are defined before we wipe them out. Oops.

So, I'm thinking maybe the right answer is to defer import of the application until after `configure_logging()` is called in the `baseplate-script` path. How does that sound?

Tecuya:
In the campaign management scenario, I tried it, and it didn't end up with doubled logging.. but I can see that calling `disable_existing_loggers=False` would prevent the configuration file from having as much control over the logging as it enjoys now, and so adding the parameter could potentially even break existing logging setups that rely on that behavior.. which is apparently why `disable_existing_loggers` was defaulted to `True` in the first place, for backwards compatibility, according to the docs.

Your alternative suggestion about deferring import of the application until after `configure_logging` is called sounds good to me, though I can't claim to have read the code for `baseplate-script`, but making `baseplate-script` behave more like `baseplate-server` is a win.

Tecuya:
This problem will be solved in a different way

PR #543, opened 2021-01-15
baseplate-script dont import app until after logging is configured

Tecuya:
Change `baseplate-script` so it does not import the application until after logging is configured, to prevent `logging.config.fileConfig` calls from disabling loggers that get initialized by the import.

See discussion in: https://github.com/reddit/baseplate.py/pull/542

I tested this locally and verified that with this patch in place imported loggers no longer get disabled.


spladug (approved):
Looks great, modulo the linter issue. Thank you!

spladug (approved):

PR #546, opened 2021-01-19
Rename context.trace -> context.span

### LINKED ISSUE #544: Rename context.trace -> context.span
It confuses people since it's not exclusively related to tracing, but rather the whole observability system.

spladug:
The span is for the whole observer system, not just for tracing. Even
though the span concept comes from tracing.

Fixes #544

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -715,7 +715,7 @@ class ServerSpan(LocalSpan):
     """A server span represents a request this server is handling.
 
     The server span is available on the :py:class:`~baseplate.RequestContext`
-    during requests as the ``trace`` attribute.
+    during requests as the ``span`` attribute.
 
     """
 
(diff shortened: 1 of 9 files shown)


baseplate/__init__.py
```diff
@@ -704,7 +704,7 @@ def make_child(
                 context_copy,
                 self.baseplate,
             )
-        context_copy.trace = span
+        context_copy.span = span
```
MelissaCole, line 707: Do you want to update the comment on L718?
> during requests as the ``trace`` attribute.

MelissaCole:
:fish: lgtm!

baseplate/__init__.py
spladug: thanks!

kemitche (approved):
Thanks for this!

bradengroom (approved):

PR #549, opened 2021-02-01
PyPI Updates

spladug:
This updates the project URLs so PyPI can show the right stuff for us. Shown below is current [baseplate.py on PyPI](https://pypi.org/project/baseplate/) and on the right is [the new edgecontext library on PyPI](https://pypi.org/project/reddit-edgecontext/). Baseplate.py should look more like edgecontext once this is merged.

Screenshot from 2021-02-01 15-48-25 Screenshot from 2021-02-01 15-48-50

Additionally, this setups a github action for automatic publishing to PyPI so all we have to do in the future is create a github release and we're done.

### DIFF
--- a/.github/workflows/python-publish.yaml
+++ b/.github/workflows/python-publish.yaml
@@ -1,0 +1,32 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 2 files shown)


MelissaCole (approved):
cool!

PR #554, opened 2021-02-10
Display Trace IDs as UUID

spladug:
Since Wavefront formats trace IDs this way, we should standardize to make it clearer how they mesh.


spladug:
This does not appear to be how the IDs are mapped!

PR #556, opened 2021-02-10
Rework tagged metrics to ensure proper aggregations

spladug:
This is related to a recent discussion in #baseplate-spec. Unfortunately, tagging `success` on the timers meant that we can't get a complete view of the timers across successes and failures (since you can't just sum up percentiles). So this removes that tag from the timer, adds a counter back into the mix with that success tag, and renames both to give them some space.

The untagged metrics implementation emits counters for server timeouts and for each exception returned from a client. This was done with tags in the previous tagged metrics implementation, which can't continue for the same reasons as above. As a straw argument: I'm just not reimplementing them. How do we feel about this? Does anyone actually use those metrics?

Since we're gearing up for breaking 2.0 changes, I figured it's a good time to get rid of "whitelist" here too.

### DIFF
--- a/baseplate/observers/metrics_tagged.py
+++ b/baseplate/observers/metrics_tagged.py
@@ -1,4 +1,3 @@
-from enum import Enum
 from random import random
 from typing import Any
 from typing import Dict
@@ -13,12 +12,6 @@ from baseplate import Span
 from baseplate import SpanObserver
 from baseplate.lib import config
 from baseplate.lib import metrics
-from baseplate.observers.timeout import ServerTimeout
-
-
-class Errors(Enum):
-    TIMED_OUT = "timed_out"
-    EXCEPTION = "internal_server_error"
 
 
 class TaggedMetricsBaseplateObserver(BaseplateObserver):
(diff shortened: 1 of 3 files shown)


baseplate/observers/metrics_tagged.py
```diff
@@ -52,17 +45,16 @@ def from_config_and_client(
             raw_config,
             {
                 "metrics": {
-                    "whitelist": config.Optional(
-                        config.TupleOf(config.String),
-                        default=["client", "endpoint", "success", "error"],
+                    "allowlist": config.Optional(
+                        config.TupleOf(config.String), default=["client", "endpoint"],
```
fishy: I think this approach makes it possible for service owner to remove `client` and/or `endpoint` from the allowlist (either accidentally or intentionally), thus making the metrics no longer comply to the baseplate spec. should we hardcode these 2 to be always included instead?

technically with the new code we are actually hardcoding `success` tag,

spladug: :+1: makes sense. i've added those as always allowed.

fishy (approved):

pacejackson (approved):

PR #557, opened 2021-02-10
Make trace IDs strings for flexibility

spladug:
This will allow us, once fully rolled out, to upgrade to larger trace
IDs to avoid collisions and work better with other systems.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -102,13 +102,13 @@ class TraceInfo(NamedTuple):
     """
 
     #: The ID of the whole trace. This will be the same for all downstream requests.
-    trace_id: int
+    trace_id: str
 
     #: The ID of the parent span, or None if this is the root span.
-    parent_id: Optional[int]
+    parent_id: Optional[str]
 
     #: The ID of the current span. Should be unique within a trace.
-    span_id: int
+    span_id: str
 
     #: True if this trace was selected for sampling. Will be propagated to child spans.
     sampled: Optional[bool]
@@ -124,15 +124,15 @@ class TraceInfo(NamedTuple):
         with any upstream requests.
 
         """
-        trace_id = random.getrandbits(64)
+        trace_id = str(random.getrandbits(64))
         return cls(trace_id=trace_id, parent_id=None, span_id=trace_id, sampled=None, flags=None)
 
     @classmethod
     def from_upstream(
         cls,
-        trace_id: Optional[int],
-        parent_id: Optional[int],
-        span_id: Optional[int],
+        trace_id: Optional[str],
+        parent_id: Optional[str],
+        span_id: Optional[str],
         sampled: Optional[bool],
         flags: Optional[int],
     ) -> "TraceInfo":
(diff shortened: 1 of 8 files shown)


fishy (approved):

baseplate/__init__.py
```diff
@@ -147,13 +147,13 @@ def from_upstream(
...
+        if trace_id is None:
             raise ValueError("invalid trace_id")
 
-        if span_id is None or not 0 <= span_id < 2 ** 64:
+        if span_id is None:
             raise ValueError("invalid span_id")
 
-        if parent_id is None or not 0 <= parent_id < 2 ** 64:
+        if parent_id is None:
             raise ValueError("invalid parent_id")
```
fishy, line 157: hmm I believe this part of implementation is actually slightly different from the spec.

the spec says services should pass `"0"` as the header value to explicitly state that the given id does not exist, and then the server should just generate a new one randomly (for trace id) or just treat itself's server span as a root span (for parent/span id). we might want to check against empty string and "0" here and generate random ones instead?

baseplate/clients/sqlalchemy.py
```diff
@@ -123,6 +125,9 @@ def parse(
 Parameters = Optional[Union[Dict[str, Any], Sequence[Any]]]
 
 
+SAFE_TRACE_ID = re.compile("^[A-Za-z0-9_-]+$")
```
fishy, line 128: 🔕 I like this approach better than the one discussed in the transition doc: only do the sanitization locally when needed, instead of doing that globally. Doing it locally also means we can use the proper sanitization (e.g. sql sanitization for sql, url sanitization for url, etc.)

baseplate/__init__.py
spladug: I don't see anything about `"0"` in the spec. Do you mean this?

> If valid trace headers are present on the request, they MUST be used to configure the span object. If no headers are present, or they are invalid, implementations MUST generate a new random 64-bit integer for Trace ID, set span ID to the same value, and leave parent ID as null.

That happens when the exception gets raised: 
1. https://github.com/reddit/baseplate.py/blob/develop/baseplate/frameworks/thrift/__init__.py#L91-L99
2. https://github.com/reddit/baseplate.py/blob/develop/baseplate/__init__.py#L470-L471

baseplate/clients/sqlalchemy.py
spladug: I think they're complementary. I'm not suggesting that we have this sort of validation up front, but just that we state in the spec some basic bounds on what the trace IDs can be so that we're not violating the spec by doing any sanity checks. I think we can agree trace IDs will never be 3 megabyte gifs, so there's gotta be some inbetween on validation we can do that's not strict but not completely open.

baseplate/__init__.py
fishy: I could have sworn I've seen `0` mentioned in the spec, maybe that's on an older version of the spec, or maybe an older version of the implementation of basepalte.py, or maybe my memory just failed me 😅

fishy: Oh now I remembered, this line in current code will fallback to use "0" (instead of empty string) when serializing to zipkin json and that's likely where I got that impression from: https://github.com/reddit/baseplate.py/blob/673fe736ac004a2f141ae7b25c6c47ec7af7edef/baseplate/observers/tracing.py#L282

bradengroom (approved):

baseplate/__init__.py
spladug: ah, funky!

PR #559, opened 2021-02-11
Add optional bool field "retryable" to Error in thrift definition

fishy:
The library code can then take advantage of that to make retry decisions
based on the Error returned by the server.


spladug (approved):

bradengroom (approved):

fishy:
:haircut:

~~NOTE: I run thrift compiler 0.14 to generate those files. If we prefer to keep on 0.12 for now I can do that as well.~~

spladug (approved):
LGTM, thanks!

PR #560, opened 2021-02-12
Overhaul local development and CI

spladug:
I wanted to break the dependency on OneVM since it's 1) not available in open source, and 2) slated to go away this year. We've also been finding GH Actions nice on other libraries, so I figured I'd go full bore on them here.

This also upgrades a ton of linter dependencies, some of which required
some minor cosmetic changes to the library.

See the new [CONTRIBUTING.md](https://github.com/reddit/baseplate.py/blob/eddbb86286a9828d9fd4517e29056707872954ff/CONTRIBUTING.md) for how it works.


CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+Our CI system will run a ton of checks on pull requests. You can run these
+checks locally before submitting changes as described below.
+
+A `Makefile` full of actions and a [Docker Compose] development environment are
+provided. [Install Docker Compose] if you don't already have it, then you can
+`docker-compose run` various commands to format, lint, test, and compile the
+library. For example:
+
+```console
+$ docker-compose run baseplate make test
```
bradengroom, line 14: 🔕 I think it'd be great if we also had an example of how to run a single test.

requirements-transitive.txt
```diff
@@ -0,0 +1,85 @@
+advocate==1.0.0
```
bradengroom, line 1: Would we ever consider using something like [poetry](https://python-poetry.org/)?

bradengroom (approved):

requirements-transitive.txt
spladug: Yes! Absolutely. We may have more flexibility out here, but for internal services things like our package vulnerability scanner don't support non-requirements.txt based systems right now :(

PR #561, opened 2021-02-16
Upgrade to Thrift v0.14

spladug:
Thrift 0.14 is now published as `thrift-unofficial` via https://github.com/reddit/thrift-python/

This allows us to upgrade.


fishy (approved):
the name sounds... very... unofficial :badpokerface:

pacejackson (approved):

PR #562, opened 2021-02-17
tagged metrics: Fix tagging on multi-part client names

spladug:
```python
baseplate.configure_context({
    "foo": {
        "bar": ThriftClient(...),
    },
})
```

would generate span names like `foo.bar.is_healthy` which we'd want to turn into tags as `client=foo.bar` + `endpoint=is_healthy` but instead we're doing it as `client=foo` + `endpoint=bar` which is not great!


fishy (approved):
While this fixes the immediate issue, I would still strongly prefer to remove the client name part from the span name.

bradengroom (approved):

PR #566, opened 2021-02-26
Circuit breakers

bsimpson63:
This is largely copied from reddit/reddit-service-graphql/tree/master/graphql_api/lib/circuit_breaker. That code is great and running in production in a few service.

I added docstrings and types where missing and then added the `CircuitBreakerClientWrapperFactory` interface to hopefully make it a bit easier to integrate with baseplate service.

I'm going to copy/paste this code into thing service and r2 to ensure it works as expected and isn't missing anything. Then we can consider merging this PR into baseplate.

👓 @suzie-su @pnovotnak @spladug @manishapme @jennLam


baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+    ```
+    """
+
+    def __init__(self, client_factory: ContextFactory, breaker_box: "CircuitBreakerBox"):
+        self.client_factory = client_factory
+        self.breaker_box = breaker_box
+
+    def make_object_for_context(self, name: str, span: Span) -> Any:
+        client = self.client_factory.make_object_for_context(name, span)
+
```
suzie-su, line 40: extra line?

baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+def breaker_box_from_config(
+    app_config: config.RawConfig, name: str, prefix: str = "breaker.",
+) -> CircuitBreakerBox:
+    """Make a CircuitBreakerBox from a configuration dictionary."""
+    # TODO: fix default handling here. if these are not set
+    # they will be None and passed through to the Breaker() constructor
+    # which will override the defaults set in Breaker()
+    assert prefix.endswith(".")
+    parser = config.SpecParser(
```
suzie-su, line 153: what is difference between `config.parse_config` vs `config.SpecParser`

baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+    ```
+    breakable_exceptions = (...)  # exceptions indicating the service is unhealthy
+    with context.breaker_wrapped_client.breaker_context("identifier", breakable_exceptions) as svc:
+        svc.get_something()
+    ```
+    """
+
+    def __init__(self, client_factory: ContextFactory, breaker_box: "CircuitBreakerBox"):
+        self.client_factory = client_factory
+        self.breaker_box = breaker_box
```
suzie-su, line 36: so in this way, people will have to initialize the breaker_box in. __init__ file and pass it in to factory?

suzie-su (commented):
good refactor 👍 
just some questions.

baseplate/lib/circuit_breaker/breaker.py
```diff
@@ -0,0 +1,127 @@
...
+        :param success: Whether the attempt was a success (not a failure).
+        """
+        # This breaker has already tripped, so ignore the "late" registrations
+        if self.state == BreakerState.TRIPPED:
+            return
+
+        if not success:
+            self.failures += 1
+
+        if self._is_bucket_full and not self.results_bucket[0]:
```
pnovotnak, line 82: 🔕 can we grab [this comment](https://github.com/reddit/baseplate.py/pull/565/files#diff-e79733da473c929b08e27ae53c7fe982df8bf0253a46c7e9540f2800db9ace27R83) or put something like it here?

baseplate/lib/circuit_breaker/breaker.py
```diff
@@ -0,0 +1,127 @@
...
+    is successful the breaker moves to the WORKING state. If the next attempt is a failure the
+    breaker moves back to the TRIPPED state.
+
+    :param name: full name/path of the circuit breaker
+    :param samples: number of previous results used to calculate the trip failure ratio
+    :param trip_failure_percent: the minimum ratio of sampled failed results to trip the breaker
+    :param trip_for: how long to remain tripped before resetting the breaker
+    :param fuzz_ratio: how much to randomly add/subtract to the trip_for time
+    """
+
```
pnovotnak, line 40: 🔕 My linter complains about variables initialized outside init, feel free to disregard

```suggestion

failures: int
```

pnovotnak (approved):
All in all I think this is good and I like the ergonomics. Stoked to have this in baseplate! Thank you for putting in the time, Brian

baseplate/lib/circuit_breaker/factory.py
bsimpson63: yeah exactly. that's similar to how we do other clients in baseplate.py, and will make sure that all workers/greenlets share the same breakers.

baseplate/lib/circuit_breaker/breaker.py
bsimpson63: yeah that's a little funky that it's only initialized in `reset()`(which is called from the constructor), but it does prevent some duplication. I'm leaning towards leaving it as is because this is copied exactly from the gql implementation.

baseplate/lib/circuit_breaker/factory.py
bsimpson63: https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/config.py#L625-L635

`parse_config` is essentially `SpecParser` (returned by `Parser.from_spec`) and then `parser.parse("", app_config)`, so it seems it doesn't let you specify the prefix.

bsimpson63:
Note from Manisha:

only request is some docs that show how you set it up in the tests in the docstring somewhere. like add this to your init.py file

```
breaker_box = breaker_box_from_config(
            app_config={
                "brkr.samples": "4",
                "brkr.trip_failure_ratio": "0.75",
                "brkr.trip_for": "1 minute",
                "brkr.fuzz_ratio": "0.1",
            }
wrapped_client_factory = CircuitBreakerClientWrapperFactory(
            client_factory, self.breaker_box
        )
baseplate.add_to_context("wrapped_client", wrapped_client_factory)
```

baseplate/lib/circuit_breaker/factory.py
suzie-su: but in the parser.parse{"", {prefix: }} can set the `prefix` in the dict right

suzie-su (approved):

baseplate/lib/circuit_breaker/breaker.py
spladug: it'd be good to add the type annotation just so everything knows it can expect the attribute to be there

baseplate/lib/circuit_breaker/breaker.py
```diff
@@ -0,0 +1,127 @@
...
+from math import ceil
+from random import random
+from typing import Deque
+
+
+class BreakerState(Enum):
+    WORKING = "working"
+    TRIPPED = "tripped"
+    # trip immediately after failure
+    TESTING = "testing"
```
spladug, line 14: given the sequence of states described below, it's a bit weird that testing is after tripped in this enum

baseplate/lib/circuit_breaker/breaker.py
```diff
@@ -0,0 +1,127 @@
...
+        name: str,
+        samples: int = 20,
+        trip_failure_ratio: float = 0.5,
+        trip_for: timedelta = timedelta(minutes=1),
+        fuzz_ratio: float = 0.1,
+    ):
+        self.name = name
+        self.samples = samples
+        self.results_bucket: Deque = deque([], self.samples)
+        self.tripped_until: datetime = datetime.utcnow()
```
spladug, line 55: :no_bell: would `time.monotonic()` be better for this? much less baggage and no worries about the clock shifting under us

also happens to be a bit faster

```
$ python3 -m timeit -s 'import datetime' -- 'datetime.datetime.utcnow()'
2000000 loops, best of 5: 150 nsec per loop

$ python3 -m timeit -s 'import time' -- 'time.monotonic()'
5000000 loops, best of 5: 60.8 nsec per loop
```

baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+from baseplate import Span
+from baseplate.clients import ContextFactory
+from baseplate.lib import config
+from baseplate.lib.circuit_breaker.breaker import Breaker
+from baseplate.lib.circuit_breaker.breaker import BreakerState
+from baseplate.lib.circuit_breaker.errors import BreakerTrippedError
+
+
+logger = logging.getLogger(__name__)
+
```
spladug, line 20: It'd be nice to add a `config.Parser` subclass like https://github.com/reddit/baseplate.py/blob/develop/baseplate/clients/thrift.py#L24-L44 here so this circuit breaker can be used with `baseplate.configure_context()`

tests/integration/circuit_breaker_tests.py
```diff
@@ -0,0 +1,192 @@
...
+        with pytest.raises(ValueError):
+            make_call(self.server_span, self.context, "d")
+
+        self.client.get_something.assert_has_calls(
+            [mock.call("a"), mock.call("b"), mock.call("c"), mock.call("d")]
+        )
+
+        breaker = self.breaker_box.breaker_box["get_something"]
+        assert breaker.name == "test_breaker.get_something"
+        assert breaker.results_bucket == deque([True, True, False, False])
```
spladug, line 108: the dequeue is an implementation detail, right? if we assert the exact internal state of the breaker like this it means we have to change all these tests if we change the implementation at all. it'd be less brittle to only test the externally visible state.

spladug (changes requested):
super cool! i've left a few comments. also, it'd be good to add `docs/api/baseplate/lib/circuit_breaker.rst` to make this show up in docs. that's a great place to include example usage and any exposition on how it works. see the `docs/api/baseplate/clients/` docs for some good examples.

baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+
+class CircuitBreakerClientWrapperFactory(ContextFactory):
```
curioussavage, line 22: Should you also be exposing a `report_runtime_metrics` that calls one on the `client_factory` if it exists? Or is the idea that you would still attach the bare client factory to the context?

baseplate/lib/circuit_breaker/factory.py
```diff
@@ -0,0 +1,164 @@
...
+        self.client_factory = client_factory
+        self.breaker_box = breaker_box
+
+    def make_object_for_context(self, name: str, span: Span) -> Any:
+        client = self.client_factory.make_object_for_context(name, span)
+
+        return CircuitBreakerWrappedClient(span, self.breaker_box, client)
+
+
+class CircuitBreakerWrappedClient:
```
curioussavage, line 44: Either way may be ideally out of scope for this PR but it could be nice to have a `ThriftCircuitBreakerWrappedClient` that acts like a proxy class and just decorates the thrift interface functions with the `breaker_context` logic.

curioussavage: I wrote a little implementation in our codebase for this, works fine. I wonder if a patch to integrate a circuit breaker with the thrift client would be accepted after this merges.

Also, since retry logic can cause N requests to be sent before it lets the last exception bubble up to the breaker any approach not integrated with the thrift client doesn't really have an accurate picture of failures.

manishapme:
🏃‍♂️

PR #573, opened 2021-03-08
Add rediscluster support

FranGM:
Add a `ClusterRedisClient` next to the existing `RedisClient`, allowing us to connect to Redis clusters.

We also provide a `MonitoredClusterRedisConnection` and `MonitoredClusterRedisPipeline` as analogues for `MonitoredRedisConnection` and `MonitoredRedisPipeline`.


baseplate/clients/redis.py
```diff
@@ -4,6 +4,7 @@
 from typing import Optional
 
 import redis
+import rediscluster
```
spladug: This would make anyone using the current redis client suddenly have to install `redis-py-cluster` as well just to continue using the other client.  Can you move the new cluster stuff into its own module in the clients folder? Like `baseplate/clients/redis_cluster.py` perhaps?

baseplate/clients/redis.py
```diff
@@ -253,3 +254,173 @@ def close(self) -> None:
...
+    * ``max_connections``: an integer maximum number of connections in the pool
+    * ``timeout``: how long to wait for sockets to connect. e.g.
+        ``200 milliseconds`` (:py:func:`~baseplate.lib.config.Timespan`)
+    """
+
+    assert prefix.endswith(".")
+
+    parser = config.SpecParser(
+        {
+            "startup_nodes": config.TupleOf(config.String),
```
spladug: If we use the `Endpoint` configuration parser:

```suggestion
            "startup_nodes": config.TupleOf(config.Endpoint),
```

baseplate/clients/redis.py
```diff
@@ -253,3 +254,173 @@ def close(self) -> None:
...
+            "timeout": config.Optional(config.Timespan, default=None),
+        }
+    )
+
+    options = parser.parse(prefix[:-1], app_config)
+
+    nodes = [
+        {"host": node[0], "port": node[1]}
+        for node in [startup_node.split(":") for startup_node in options.startup_nodes]
+    ]
```
spladug: It can handle the host/port parsing for us.

```suggestion
    nodes = [{"host": endpoint.address.host, "port": endpoint.address.port} for endpoint in options.startup_nodes]
```

(note `endpoint.address.port` is an integer now, is that OK for redis-py-cluster? or is it really expecting a string in that port value?)

baseplate/clients/redis.py
```diff
@@ -253,3 +254,173 @@ def close(self) -> None:
...
+    This connection acts like :py:class:`rediscluster.Redis` except that all
+    operations are automatically wrapped with diagnostic collection.
+
+    The interface is the same as that class except for the
+    :py:meth:`~baseplate.clients.redis.MonitoredClusterRedisConnection.pipeline`
+    method.
+
+    """
+
+    # TODO: Add all args below
```
spladug: ?

docker-compose.yml
```diff
@@ -25,3 +28,29 @@ services:
     image: "redis:4.0.9"
   zookeeper:
     image: "zookeeper:3.4.10"
+  redis-cluster-node-0:
```
spladug: wow, this is a lot of stuff. do we actually need a three node cluster to test this out? or can we mock it more simply?

tests/integration/redis_tests.py
```diff
@@ -6,9 +6,13 @@
 except ImportError:
     raise unittest.SkipTest("redis-py is not installed")
 
+try:
```
spladug: likewise on splitting up the redis/redis_cluster modules here please

spladug (changes requested):
This looks great! Thank you! I've left a few small comments in there, but overall it looks like it'll do the trick nicely. A few blockers:

* splitting up the files so we don't cause import issues for existing users
* adding a page to the docs here: https://github.com/reddit/baseplate.py/tree/develop/docs/api/baseplate/clients (you can do `make docs` in dev to see the output)
* adding an appropriate "extra" to setup.py to specify supported version ranges: https://github.com/reddit/baseplate.py/blob/develop/setup.py#L14
* some lint failures (you can do `make lint` and `make fmt` in dev https://github.com/reddit/baseplate.py/blob/develop/CONTRIBUTING.md)

docker-compose.yml
konradreiche: One idea would be to use
```
grokzen/redis-cluster
```
which gives us a Redis cluster with a single Docker dependency. It is not recommended for production use but perfect for testing environments. I will say, Redis cluster has idiosyncrasies that are worth having tests run against a real instance but that depends entirely on the tests.

spladug: we're not actually implementing a full redis cluster client here, thankfully, but just wrapping an existing one. do you think we need to have behavioral tests here in baseplate beyond ensuring that we generate spans?

FranGM:
I think I've addressed all the comments. I've split redis into two clients and added a page to the docs.
After more testing I ended up needing to work around a quirk or two of the upstream library but nothing too major.

docker-compose.yml
FranGM: There were a few cases where I was thinking more testing would be useful but ultimately any test I could think of would just end up testing the underlying `redis-py-cluster` library, so didn't add anything beyond what the redis client already has (I did end up using the `redis-cluster` container for the integration tests though, it was also pretty handy for manual testing)

baseplate/clients/redis_cluster.py
```diff
@@ -0,0 +1,225 @@
+import random
+
+from typing import Any
+from typing import Dict
+
+import rediscluster  # type: ignore
```
spladug: is this because `rediscluster` isn't typed? if so, you can get rid of the pragmas here in code and pop something like this into setup.cfg: https://github.com/reddit/baseplate.py/blob/develop/setup.cfg#L72-L73

baseplate/clients/redis_cluster.py
```diff
@@ -0,0 +1,225 @@
...
+    def get_node_by_slot(self, slot: int, read_command: bool = False) -> Dict[str, Any]:
+        """
+        Get a random node from the slot, including master
+        """
+        nodes_in_slot = self.nodes.slots[slot]
+        if read_command:
+            random_index = random.randrange(1, len(nodes_in_slot))
+            return nodes_in_slot[random_index]
+
+        return nodes_in_slot[0]
```
spladug: Am I understanding right that the first node is always the primary and the rest are replicas? If so, would something like this make that a little more self-explanatory?

```suggestion
        primary, *replicas = self.nodes.slots[slot]
        if read_command:
            return random.choice(replicas)
        return primary
```

Edge case check: is there ever a situation where we'd have zero replicas?

baseplate/clients/redis_cluster.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def cluster_pool_from_config(
+    app_config: config.RawConfig, prefix: str = "rediscluster.", **kwargs: Any
+) -> rediscluster.ClusterConnectionPool:
+    """Make a ClusterConnectionPool from a configuration dictionary.
+
+    The keys useful to :py:func:`clusterpool_from_config` should be prefixed, e.g.
```
spladug:
```suggestion
    The keys useful to :py:func:`cluster_pool_from_config` should be prefixed, e.g.
```

baseplate/clients/redis_cluster.py
```diff
@@ -0,0 +1,225 @@
...
+            "skip_full_coverage_check": config.Optional(config.Boolean, default=True),
+            "nodemanager_follow_cluster": config.Optional(config.Boolean, default=None),
+            "decode_responses": config.Optional(config.Boolean, default=True),
+        }
+    )
+
+    options = parser.parse(prefix[:-1], app_config)
+
+    # We're explicitly setting a default here because of https://github.com/Grokzen/redis-py-cluster/issues/435
+    kwargs.setdefault("max_connections", options.max_connections)
```
spladug, line 278: yikes! thanks for catching this and documenting it!

spladug: but it looks like the default in our config is also `None` so would we just be causing the same bug?

tests/integration/redis_tests.py
```diff
@@ -8,7 +8,6 @@
 
 from baseplate.clients.redis import RedisClient
 from baseplate import Baseplate
-
```
spladug: gasp!

spladug (changes requested):
looks great! only blocker is the question about our default causing the infinite loop as well. otherwise, just some stylistic questions. thank you!

baseplate/clients/redis_cluster.py
FranGM: oh good catch! I think I unset the default for some tests and forgot to change it back, I'll fix it

baseplate/clients/redis_cluster.py
FranGM: actually it's definitely possible, yes, let me rework it so it can handle all cases

baseplate/clients/redis_cluster.py
FranGM: oh that's great , that's much cleaner

spladug (approved):
LGTM!

baseplate/clients/redis_cluster.py
```diff
@@ -21,16 +21,14 @@ class ClusterWithReadReplicasBlockingConnectionPool(rediscluster.ClusterBlocking
...
+        if read_command:
+            return random.choice(self.nodes.slots[slot])
 
-        if replicas and read_command:
-            return random.choice(replicas)
-
-        # Either this isn't a read command or there aren't any replicas
-        return primary
+        # This isn't a read command, so return the primary
+        return self.nodes.slots[slot]
```
spladug: does this not return all of them rather than just the primary?

FranGM: ah yes, and I'm very annoyed at myself for not having added a test for that method. I'll fix it and look at adding a test.

spladug (approved):

FranGM:
@spladug based on our testing I added a requested feature to the client which is the ability to track hot keys in the cluster. 

Another change that's pending is that I need to bump the version of `redis-py-cluster` to `2.1.3` but need to wait until that is released first. We'll need that version since it will include https://github.com/Grokzen/redis-py-cluster/pull/450

docs/api/baseplate/clients/redis_cluster.rst
```diff
@@ -0,0 +1,92 @@
...
+
+The following metrics are reported:
+
+``runtime.pool.size``
+   The size limit for the connection pool.
+``runtime.pool.in_use``
+   How many connections have been established and are currently checked out and
+   being used.
+
+.. versionadded:: 2.1
```
spladug: can this go up at the top, like right before the `Example` heading? I've only put the versionadded stuff down here on some other clients where the runtime metrics were the thing that was new.

docs/api/baseplate/clients/redis_cluster.rst
```diff
@@ -0,0 +1,92 @@
+``baseplate.clients.redis_cluster``
```
spladug, line 1: should the hot key tracking stuff be mentioned somewhere in this page?

spladug (approved):
looks great! just a couple notes on the docs. once that's done, we can merge whenever that new release is cut upstream.

docs/api/baseplate/clients/redis_cluster.rst
```diff
@@ -0,0 +1,145 @@
...
+   being used.
+
+
+Hot Key Tracking
+----------------
+
+Optionally, the client can help track key usage across the Redis cluster to
+help you identify if you have "hot" keys (keys that are read from or
+written to much more frequently than other keys). This is particularly useful
+in clusters with the `noeviction` set as the eviction policy, since Redis
```
spladug:
```suggestion
in clusters with ``noeviction`` set as the eviction policy, since Redis
```

(single ` are used for references to other rST things, you have to use double to get inline code formatting)

spladug (approved):

PR #574, opened 2021-03-09
Add deadline-budget to thrift client headers

ethverie:
Add a `Deadline-Budget` to thrift client headers to allow for deadline propagation.

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -170,16 +170,16 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                         prot.trans.set_header(b"Flags", str(span.flags).encode())
 
                     min_timeout = None
-                    is_time_retry_policy = (
-                        retry_policy is TimeBudgetRetryPolicy
-                        or retry_policy is ExponentialBackoffRetryPolicy
+                    is_time_retry_policy = isinstance(retry_policy, TimeBudgetRetryPolicy) or (
+                        isinstance(retry_policy, ExponentialBackoffRetryPolicy)
+                        and isinstance(retry_policy.subpolicy, TimeBudgetRetryPolicy)
                     )
                     if is_time_retry_policy:
                         min_timeout = value
                     if self.pool.timeout:
                         if not min_timeout or self.pool.timeout < min_timeout:
                             min_timeout = self.pool.timeout
-                    if min_timeout:
+                    if min_timeout and min_timeout > 0:
                         prot.trans.set_header(
                             b"Deadline-Budget", str(int(ceil(min_timeout * 1000))).encode()
                         )
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -146,7 +146,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
         trace_name = f"{self.namespace}.{name}"
         last_error = None
 
-        for value in self.retry_policy:
+        for time_remaining in self.retry_policy:
             span = self.server_span.make_child(trace_name)
             span.start()
 
@@ -167,11 +167,13 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                     if span.flags:
                         prot.trans.set_header(b"Flags", str(span.flags).encode())
 
-                    min_timeout = value or None
+                    min_timeout = time_remaining
                     if self.pool.timeout:
                         if not min_timeout or self.pool.timeout < min_timeout:
                             min_timeout = self.pool.timeout
                     if min_timeout and min_timeout > 0:
+                        # min_timeout is in float seconds, we are converting to int milliseconds
+                        # rounding up here.
                         prot.trans.set_header(
                             b"Deadline-Budget", str(int(ceil(min_timeout * 1000))).encode()
                         )
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -17,9 +17,7 @@ from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
 from baseplate.lib import metrics
-from baseplate.lib.retry import ExponentialBackoffRetryPolicy
 from baseplate.lib.retry import RetryPolicy
-from baseplate.lib.retry import TimeBudgetRetryPolicy
 from baseplate.lib.thrift_pool import thrift_pool_from_config
 from baseplate.lib.thrift_pool import ThriftConnectionPool
 
@@ -148,7 +146,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
         trace_name = f"{self.namespace}.{name}"
         last_error = None
 
-        for retry_policy, value in self.retry_policy:
+        for value in self.retry_policy:
             span = self.server_span.make_child(trace_name)
             span.start()
 
@@ -169,13 +167,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                     if span.flags:
                         prot.trans.set_header(b"Flags", str(span.flags).encode())
 
-                    min_timeout = None
-                    is_time_retry_policy = isinstance(retry_policy, TimeBudgetRetryPolicy) or (
-                        isinstance(retry_policy, ExponentialBackoffRetryPolicy)
-                        and isinstance(retry_policy.subpolicy, TimeBudgetRetryPolicy)
-                    )
-                    if is_time_retry_policy:
-                        min_timeout = value
+                    min_timeout = value or None
                     if self.pool.timeout:
                         if not min_timeout or self.pool.timeout < min_timeout:
                             min_timeout = self.pool.timeout
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -16,7 +16,9 @@ from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
 from baseplate.lib import metrics
+from baseplate.lib.retry import ExponentialBackoffRetryPolicy
 from baseplate.lib.retry import RetryPolicy
+from baseplate.lib.retry import TimeBudgetRetryPolicy
 from baseplate.lib.thrift_pool import thrift_pool_from_config
 from baseplate.lib.thrift_pool import ThriftConnectionPool
 
@@ -145,7 +147,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
         trace_name = f"{self.namespace}.{name}"
         last_error = None
 
-        for _ in self.retry_policy:
+        for retry_policy, value in self.retry_policy:
             span = self.server_span.make_child(trace_name)
             span.start()
 
@@ -160,6 +162,8 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                     prot.trans.set_header(b"Trace", str(span.trace_id).encode())
                     prot.trans.set_header(b"Parent", str(span.parent_id).encode())
                     prot.trans.set_header(b"Span", str(span.id).encode())
+                    if isinstance(retry_policy, (TimeBudgetRetryPolicy, ExponentialBackoffRetryPolicy)):
+                        prot.trans.set_header(b"Deadline-Budget", str(value).encode())
                     if span.sampled is not None:
                         sampled = "1" if span.sampled else "0"
                         prot.trans.set_header(b"Sampled", sampled.encode())
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -2,6 +2,7 @@ import contextlib
 import inspect
 import sys
 
+from math import ceil
 from typing import Any
 from typing import Callable
 from typing import Iterator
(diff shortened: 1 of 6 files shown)


fishy:
~NOTE:~

1. ~This is a breaking change as it changed all the `RetryPolicy` implementations to yield two values: it added the class of the actual policy so it's easier for the user to get the meaning of the value (currently the value yielded can mean different things depending on the policy)~
2. ~For some reason unknown it didn't trigger CI.~

baseplate/clients/thrift.py
```diff
@@ -166,6 +169,21 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
+                    is_time_retry_policy = (
+                        retry_policy is TimeBudgetRetryPolicy
+                        or retry_policy is ExponentialBackoffRetryPolicy
+                    )
+                    if is_time_retry_policy:
+                        min_timeout = value
+                    if self.pool.timeout:
+                        if not min_timeout or self.pool.timeout < min_timeout:
+                            min_timeout = self.pool.timeout
+                    if min_timeout:
```
fishy: nit: I don't think it's possible but we probably want to check that `min_timeout` is not negative? In baseplate.go we checked if it's less than 1ms (which includes negative values) and give it a minimal of 1ms (https://github.com/reddit/baseplate.go/blob/08f6b3c8e11a95dcc6a99d26d8b961acd6c08615/thriftbp/client_middlewares.go#L234). the `ceil` part handled the case that the timeout is between 0 and 1ms, but doesn't handle the case if it's negative.

baseplate/clients/thrift.py
```diff
@@ -2,6 +2,7 @@
 import inspect
 import sys
 
+from math import ceil
```
fishy, line 5: I think this import should be on the first group?

baseplate/lib/retry.py
```diff
@@ -21,21 +23,22 @@ class RetryPolicy:
...
+        yielded instead.
 
         The iterable will raise :py:exc:`StopIteration` once the operation
         should not be retried any further.
 
         """
         raise NotImplementedError
 
-    def __iter__(self) -> Iterator[Optional[float]]:
+    def __iter__(self) -> Iterator[Tuple[Type["RetryPolicy"], Optional[float]]]:
```
fishy: This one is `Type["RetryPolicy"]` while others are `Type[RetryPolicy]` (without quotation mark).

tests/integration/thrift_tests.py
```diff
@@ -224,6 +227,31 @@ def example(self, context):
         self.assertEqual(handler.server_span.sampled, False)
         self.assertTrue(client_result)
 
+    def test_budget_header(self):
+        """Test that the bedget header is set in the headers if the client sets it."""
```
fishy: typo: `s/bedget/budget/`

tests/integration/thrift_tests.py
```diff
@@ -72,13 +72,16 @@ def raw_thrift_client(endpoint, client_spec):
...
-    )
+def baseplate_thrift_client(
+    endpoint, client_spec, client_span_observer=None, budget_header_test=False,
+):
+    app_config = {
+        "baseplate.service_name": "fancy test client",
+        "example_service.endpoint": str(endpoint),
+    }
+    if budget_header_test:
+        app_config["example_service.timeout"] = "1000 seconds"
```
fishy: it's probably better to also pass `1000 seconds` part in as an arg of `baseplate_thrift_client`.

e.g. make the arg actually `timeout=None`, then here we can do

```python
if timeout
  app_config["example_service.timeout"] = timeout
```

instead.

tests/integration/thrift_tests.py
```diff
@@ -406,6 +434,46 @@ def example(self, context):
 
         assert handler.edge_context == FakeEdgeContextFactory.DECODED_CONTEXT
 
+    def test_budget_header_pool_timeout(self):
+        """Test that the budget header is set in the headers if the client sets it."""
```
fishy: You didn't really set the pool timeout in this test, so the name should be something like `test_timeout_budget_from_retry_policy` and the doc should be something like "tests that the budget header is set from the retry policy".

tests/integration/thrift_tests.py
```diff
@@ -406,6 +434,46 @@ def example(self, context):
...
+        span_observer = mock.Mock(spec=SpanObserver)
+        with serve_thrift(handler, TestService) as server:
+            with baseplate_thrift_client(server.endpoint, TestService, span_observer) as context:
+                with context.example_service.retrying(attempts=3, budget=budget) as svc:
+                    svc.example()
+
+        assert handler.context.headers.get(b"Deadline-Budget").decode() == str(1000)
+
+    def test_budget_header_retry_timeout(self):
+        """Test that the budget header is set in the headers if the client sets it."""
```
fishy: this test actually tests when retry timeout is smaller than the pool timeout we pick the retry one correctly.

we need another test that when retry timeout is larger than the pool timeout we pick the pool one correctly. we probably need another one that only set the pool timeout and not the retry timeout.

also for retry timeout there can be the one from exponential backoff and the one from budget, we probably need separated tests for them as well.

baseplate/lib/retry.py
```diff
@@ -124,14 +126,15 @@ def __init__(self, policy: RetryPolicy, base: float):
...
-        for attempt, time_remaining in enumerate(self.subpolicy):
+    def yield_attempts(self) -> Iterator[Tuple["RetryPolicy", Optional[float]]]:
+        for attempt, (return_policy, value) in enumerate(self.subpolicy):
+            time_remaining = None
             if attempt > 0:
                 delay = self.base * 2.0 ** (attempt - 1.0)
-                if time_remaining:
-                    delay = min(delay, time_remaining)
-                    time_remaining -= delay
+                if value and isinstance(return_policy, TimeBudgetRetryPolicy):
```
fishy: note: this fixes a bug in the current code.

in the current code, we always assume that `ExponentialBackoffRetryPolicy` is wrapping a `TimeBudgetRetryPolicy` by just assuming the iterator value is `time_remaining`. But that's actually not always true. Based on the implementation of `RetryPolicy.new`, if we set `backoff` but not `budget` it would be `ExponentialBackoffRetryPolicy` wrapping other subpolicies and the value from iterator is not really `time_remaining`.

cc @spladug

baseplate/lib/message_queue.py
```diff
@@ -72,7 +72,7 @@ def get(self, timeout: Optional[float] = None) -> bytes:
             duration of the call.
 
         """
-        for time_remaining in RetryPolicy.new(budget=timeout):
+        for _, time_remaining in RetryPolicy.new(budget=timeout):
```
fishy: 🔕 this is likely also a bug in the current code that the value is not always time_remaining, but I think it's out of the scope of this PR.

baseplate/lib/message_queue.py
```diff
@@ -92,7 +92,7 @@ def put(self, message: bytes, timeout: Optional[float] = None) -> None:
...
```
fishy: 🔕 here as well

baseplate/lib/retry.py
```diff
@@ -124,14 +126,15 @@ def __init__(self, policy: RetryPolicy, base: float):
         self.subpolicy = policy
         self.base = base
 
-    def yield_attempts(self) -> Iterator[Optional[float]]:
-        for attempt, time_remaining in enumerate(self.subpolicy):
+    def yield_attempts(self) -> Iterator[Tuple["RetryPolicy", Optional[float]]]:
+        for attempt, (return_policy, value) in enumerate(self.subpolicy):
```
fishy: nit: this probably should be `retry_policy` instead of `return_policy`? the name "return_policy" sounds weird.

fishy (approved):

baseplate/lib/retry.py
ethverie: ah, i meant to say `retry_policy` here oops

baseplate/clients/thrift.py
spladug: It's good where it is. We're configured to separate `import` and `from ... import` into two blocks for each of the three groups of imports.

baseplate/lib/message_queue.py
spladug: what are you seeing instead of the time remaining or None?

baseplate/clients/thrift.py
```diff
@@ -166,6 +169,21 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                     if span.flags:
                         prot.trans.set_header(b"Flags", str(span.flags).encode())
 
+                    min_timeout = None
+                    is_time_retry_policy = isinstance(retry_policy, TimeBudgetRetryPolicy) or (
```
spladug: The retry policies are an internal implementation detail here. It's a bit weird to expose them directly.  If there's a time_remaining value, would we ever _not_ want to propagate that as a deadline? It sounds like y'all ran into some other number coming out of that iterator which is surprising to me. What kind of data did you see there instead of a time or None?

spladug (changes requested):
cool! makes a lot of sense to propagate the remaining time on the retry policy to services we're calling. I'm a little confused about what's going on in the retry code though. what kinds of values are you seeing that aren't time remaining?

baseplate/clients/thrift.py
fishy: Oh you are right. I read the implementation of `MaximumAttemptsRetryPolicy` wrong. I thought it emits number of retries.

So yes it's either None or time remaining, and those are not bugs 😅

spladug: heh, it took me too a lot of staring at that code to realize it was still returning time. i think there's an opportunity for readability improvement there.

spladug: https://github.com/reddit/baseplate.py/pull/577

baseplate/clients/thrift.py
```diff
@@ -145,7 +146,7 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
         trace_name = f"{self.namespace}.{name}"
         last_error = None
 
-        for _ in self.retry_policy:
+        for value in self.retry_policy:
```
fishy: since the value emitted by `self.retry_policy` is either time_remaining or none, let's just call it `time_remaining` instead of `value` for better readability.

baseplate/clients/thrift.py
```diff
@@ -166,6 +167,15 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                     if span.flags:
                         prot.trans.set_header(b"Flags", str(span.flags).encode())
 
+                    min_timeout = value or None
```
fishy: the `or None` part is actually unnecessary since the value emitted from retry_policu is either time remaining or none, so here it can be just `min_timeout = value` (or `time_remaining` if you renamed it as suggested above)

baseplate/clients/thrift.py
```diff
@@ -166,6 +167,15 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
+                    if self.pool.timeout:
+                        if not min_timeout or self.pool.timeout < min_timeout:
+                            min_timeout = self.pool.timeout
+                    if min_timeout and min_timeout > 0:
+                        prot.trans.set_header(
+                            b"Deadline-Budget", str(int(ceil(min_timeout * 1000))).encode()
```
fishy, line 178: the whole `str(int(ceil(min_timeout * 1000))).encode()` part could be hard to follow for future engineers reading the code, so it's probably worth adding a comment to explain that we are converting it to milliseconds rounding up.

tests/integration/thrift_tests.py
```diff
@@ -224,6 +227,31 @@ def example(self, context):
...
+
+        handler = Handler()
+
+        with serve_thrift(handler, TestService) as server:
+            with raw_thrift_client(server.endpoint, TestService) as client:
+                transport = client._oprot.trans
+                transport.set_header(b"Deadline-Budget", budget.encode())
+                client_result = client.example()
+
+        self.assertIsNotNone(handler.server_span)
```
fishy: nit: since this test is called `test_budget_header`, this assert sounds unrelated.

tests/integration/thrift_tests.py
```diff
@@ -406,6 +434,70 @@ def example(self, context):
...
+        class Handler(TestService.Iface):
+            def example(self, context):
+                self.context = context
+                return True
+
+        handler = Handler()
+
+        span_observer = mock.Mock(spec=SpanObserver)
+        with serve_thrift(handler, TestService) as server:
+            with baseplate_thrift_client(server.endpoint, TestService, span_observer) as context:
```
fishy: this test always takes me a few looks to figure out why do we define the budget as `100.0` above and assert that the budget equals to `str(1000)`. my suggestions are:

1. Here explicitly add `timeout = "1 second"` to `baseplate_thrift_client` so we are not relying on the default value
2. rename the `budget` variable on line 439 to `pool_timeout_seconds`.

tests/integration/thrift_tests.py
```diff
@@ -406,6 +434,70 @@ def example(self, context):
...
+                with context.example_service.retrying(attempts=3, budget=budget) as svc:
+                    svc.example()
+
+        assert handler.context.headers.get(b"Deadline-Budget").decode() == str(1000)
+
+    def test_budget_header_retry_timeout(self):
+        """Test that the budget header is set in the headers with the retry timeout."""
+        budget = 100.0
```
fishy: 🔕 this is really nitpicky but maybe use `0.1` instead of `100.0` so we are actually testing 100ms instead of 100s (which would be too unrealistic compare to real code). You would also need to change the assert on line 475 to `str(int(budget *1000))` instead.

tests/integration/thrift_tests.py
```diff
@@ -406,6 +433,80 @@ def example(self, context):
...
+            with baseplate_thrift_client(
+                server.endpoint, TestService, span_observer, timeout="1000 seconds"
+            ) as context:
+                with context.example_service.retrying(
+                    attempts=3, budget=retry_timeout_seconds, backoff=backoff
+                ) as svc:
+                    svc.example()
+
+        assert handler.context.headers.get(b"Deadline-Budget").decode() == str(
+            int(retry_timeout_seconds) * 1000
```
fishy: slight nit: `int(retry_timeout_seconds * 1000)` would be better.

tests/unit/lib/retry_tests.py
```diff
@@ -117,4 +117,4 @@ def test_budget_overrides_backoff(self, sleep, time):
 
         time.return_value = 1
         with self.assertRaises(StopIteration):
-            next(retries)
+            next(retries)[1]
```
fishy: I think this is a leftover change from the retry policy change and should be reverted?

tests/integration/thrift_tests.py
```diff
@@ -406,6 +433,80 @@ def example(self, context):
...
+        with serve_thrift(handler, TestService) as server:
+            with baseplate_thrift_client(
+                server.endpoint, TestService, span_observer, timeout="1 second"
+            ) as context:
+                with context.example_service.retrying(
+                    attempts=3, budget=retry_timeout_seconds
+                ) as svc:
+                    svc.example()
+
+        assert handler.context.headers.get(b"Deadline-Budget").decode() == str(1000)
```
fishy, line 458: nit: maybe add a comment to explain that this `1000` is coming from `"1 second"` above.

spladug (approved):
LGTM! thank you!

PR #576, opened 2021-03-16
Sidecar stability improvements

fishy:
When the sidecar crashes (unhandled exceptions) it will cause the whole
service to be restarted by k8s, so this change improves sidecar
stability to reduce crashes:

1. In metrics client, add swallow_network_errors option to log and
   swallow the network errors instead of raising them.
2. In trace publishing sidecar, do not raise client errors, but also
   skip retries for those errors.

### DIFF
--- a/baseplate/lib/metrics.py
+++ b/baseplate/lib/metrics.py
@@ -112,7 +112,10 @@ class NullTransport(Transport):
 class RawTransport(Transport):
     """A transport which sends messages on a socket."""
 
-    def __init__(self, endpoint: config.EndpointConfiguration):
+    def __init__(
+        self, endpoint: config.EndpointConfiguration, swallow_network_errors: bool = False,
+    ):
+        self.swallow_network_errors = swallow_network_errors
         self.socket = socket.socket(endpoint.family, socket.SOCK_DGRAM)
         self.socket.connect(endpoint.address)
 
@@ -120,6 +123,9 @@ class RawTransport(Transport):
         try:
             self.socket.sendall(serialized_metric)
         except socket.error as exc:
+            if self.swallow_network_errors:
+                logger.exception("Failed to send to metrics collector")
+                return
             if exc.errno == errno.EMSGSIZE:
                 raise MessageTooBigTransportError(len(serialized_metric))
             raise TransportError(exc)
@@ -517,13 +523,18 @@ class Gauge:
 
 
 def make_client(
-    namespace: str, endpoint: config.EndpointConfiguration, log_if_unconfigured: bool
+    namespace: str,
+    endpoint: config.EndpointConfiguration,
+    log_if_unconfigured: bool,
+    swallow_network_errors: bool = False,
 ) -> Client:
     """Return a configured client.
 
     :param namespace: The root key to prefix all metrics with.
     :param endpoint: The endpoint to send metrics to or :py:data:`None`.  If
         :py:data:`None`, the returned client will discard all metrics.
+    :param swallow_network_errors: Swallow (log) network errors during sending
+        to metrics collector.
     :return: A configured client.
 
     .. seealso:: :py:func:`baseplate.lib.metrics.metrics_client_from_config`.
@@ -532,7 +543,7 @@ def make_client(
     transport: Transport
 
     if endpoint:
-        transport = RawTransport(endpoint)
+        transport = RawTransport(endpoint, swallow_network_errors=swallow_network_errors)
     else:
         transport = NullTransport(log_if_unconfigured)
     return Client(transport, namespace)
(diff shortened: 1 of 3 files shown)


baseplate/sidecars/trace_publisher.py
```diff
@@ -99,9 +99,9 @@ def publish(self, payload: SerializedBatch) -> None:
                 response = getattr(exc, "response", None)
                 if response is not None:
                     logger.exception("HTTP Request failed. Error: %s", response.text)
-                    # If client error, crash
+                    # If client error, skip retries
```
spladug, line 102: :no_bell: I almost asked for a metric, but checked the collapsed diff above and we've already got one. So that's good and means we can make sure people alert on the `error.http` counter to be sure their traces aren't getting quietly dropped. :+1:

baseplate/lib/metrics.py
```diff
@@ -112,17 +112,25 @@ def flush(self) -> None:
...
-                raise MessageTooBigTransportError(len(serialized_metric))
-            raise TransportError(exc)
+            try:
+                self.socket.sendall(serialized_metric)
+            except socket.error as exc:
+                if exc.errno == errno.EMSGSIZE:
+                    raise MessageTooBigTransportError(len(serialized_metric))
+                raise TransportError(exc)
+        except Exception:
+            if not self.swallow_network_errors:
```
spladug: the configurable is "swallow _network_ errors" but we've added a new try/except here that's far broader than network errors. was there a non-socket.error exception you were running into in practice that made you want the broader catch? or would it work to add this `if` around the `raise TransportError` above? i'm just worried about swallowing true programming errors here rather than purely operational stuff

spladug (changes requested):
sweet, thanks. just one question!

baseplate/lib/metrics.py
fishy: good point. will do.

fishy: Done.

baseplate/sidecars/trace_publisher.py
fishy: Do you think ~a separated~ an additional counter for client errors would be useful?

baseplate/lib/metrics.py
```diff
@@ -112,14 +112,20 @@ def flush(self) -> None:
...
 
     def send(self, serialized_metric: bytes) -> None:
         try:
             self.socket.sendall(serialized_metric)
         except socket.error as exc:
+            if self.swallow_network_errors:
+                logger.exception("Failed to send to metrics collector")
+                return
             if exc.errno == errno.EMSGSIZE:
                 raise MessageTooBigTransportError(len(serialized_metric))
```
spladug, line 130: I'd hate to lose the special error messaging around this one as it's something people have asked a lot about in the past. Since we catch-and-swallow this exception when we flush a batch, could the swallowing check here come after this re-raise? One small risk is we might get an EMSGSIZE false positive in a sidecar and it'd still cause a crash, but I think that's generally pretty infrequent?

https://github.com/reddit/baseplate.py/blob/729452d6246332cd894cc4cc7107b7ff51615531/baseplate/lib/metrics.py#L251

fishy: I think that should be safe enough for sidecar use case, but it kind of contradicts to the "swallow_network_errors" name as when people uses that config in other environment they might expect these errors are also swallowed?

I don't feel strongly either way.

fishy: but I do agree that the error message of that exception would be very useful, so if we just add another layer of try...catch (but only catch on `MessageTooBigTransportError` and `TransportError`), do you consider that too ugly? 😅

spladug: hrm, yeah, true. I think the only times this error should legitimately happen would be from using a batch and putting too much in it, in which case the error would still be swallowed. But if we really do get false positives with any frequency then it kinda sucks to not swallow it, like you say. Let's go with not swallowing that one at this layer just because I think it's far more common to have the batch case happen and that's caused a lot of support requests so having the extra messaging is helpful.

fishy: Do you like https://github.com/reddit/baseplate.py/pull/576/commits/92e0f900420962fc6a4ba3c49ae01503aea25bae better?

baseplate/sidecars/trace_publisher.py
spladug: hm, maybe. "our fault" vs "their fault" errors? can't hurt i suppose

baseplate/lib/metrics.py
spladug: hmm. the more I think about it, I think the commit before was fine. We should generally only be doing the error swallowing in sidecars, not main services, and services are where the special error messaging matters. Going back to that OK? sorry :(

spladug: probably also want to hard-code swallowing network errors when setting up the metrics client in sidecars? so we don't even need to make everyone change configs https://github.com/reddit/baseplate.py/blob/develop/baseplate/sidecars/trace_publisher.py#L170

fishy: I... am not sure that's feasible to do.

So to do that I think the best way is to duplicate `metrics_client_from_config` into another version `metrics_client_from_config_for_sidecars` and remove `swallow_network_errors` from the config (and hardcode it in `metrics_client_from_config_for_sidecars`)? Otherwise we would need to get a client from config first, then access its transport to change the `swallow_network_errors` field to True.

baseplate/sidecars/trace_publisher.py
fishy: Added `errors.http.client` counter.

spladug (approved):
awesome, looks great. thank you!

PR #578, opened 2021-03-24
Add deadline-budget to thrift servers

ethverie:
Following the changes from https://github.com/reddit/baseplate.py/pull/574, this adds `Deadline-Budget` server side so it can use the value that is passed from the client header.

### DIFF
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -103,10 +103,11 @@ def baseplateify_processor(
             if edge_context_factory:
                 context.edge_context = edge_context_factory.from_upstream(edge_payload)
 
-            deadline_budget = headers.get(b"Deadline-Budget", None)
-            context.deadline_budget = (
-                float(deadline_budget.decode()) / 1000 if deadline_budget else None
-            )
+            try:
+                raw_deadline_budget = headers[b"Deadline-Budget"].decode()
+                context.deadline_budget = float(raw_deadline_budget) / 1000
+            except (KeyError, ValueError):
+                context.deadline_budget = None
 
             span = baseplate.make_server_span(context, name=fn_name, trace_info=trace_info)
 
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -103,6 +103,11 @@ def baseplateify_processor(
             if edge_context_factory:
                 context.edge_context = edge_context_factory.from_upstream(edge_payload)
 
+            deadline_budget = headers.get(b"Deadline-Budget", None)
+            context.deadline_budget = (
+                float(deadline_budget.decode()) / 1000 if deadline_budget else None
+            )
+
             span = baseplate.make_server_span(context, name=fn_name, trace_info=trace_info)
 
             try:
--- a/baseplate/observers/timeout.py
+++ b/baseplate/observers/timeout.py
@@ -47,10 +47,11 @@ class TimeoutBaseplateObserver(BaseplateObserver):
         min_timeout = None
         if timeout is not config.InfiniteTimespan:
             min_timeout = timeout.total_seconds()
+
         try:
             deadline_budget = context.deadline_budget
             if deadline_budget:
-                if not min_timeout or deadline_budget < timeout.total_seconds():
+                if not min_timeout or deadline_budget < min_timeout:
                     min_timeout = deadline_budget
         except AttributeError:
             # no deadline budget in request header
--- a/baseplate/observers/timeout.py
+++ b/baseplate/observers/timeout.py
@@ -43,10 +43,6 @@ class TimeoutBaseplateObserver(BaseplateObserver):
 
     def on_server_span_created(self, context: RequestContext, server_span: ServerSpan) -> None:
         timeout = self.config.by_endpoint.get(server_span.name, self.config.default)
-        try:
-            deadline_budget = context.deadline_budget
-        except AttributeError:
-            deadline_budget = None
 
         min_timeout = None
         if timeout is not config.InfiniteTimespan:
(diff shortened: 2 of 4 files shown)


baseplate/observers/timeout.py
```diff
@@ -43,10 +43,20 @@ def __init__(self, timeout_config: config.ConfigNamespace):
 
     def on_server_span_created(self, context: RequestContext, server_span: ServerSpan) -> None:
         timeout = self.config.by_endpoint.get(server_span.name, self.config.default)
+        try:
```
fishy: since this is only used to compare to existing `min_timeout`, I'd prefer you to move this after getting min_timeout from the config, something like this:

```py
  min_timeout = None
  if timeout is not config.InfiniteTimespan:
    min_timeout = timeout.total_seconds()
  try:
    deadline_budget = context.deadline_budget
    if deadline_budget:
      if not min_timeout or deadline_budget < timeout.total_seconds():
        min_timeout = deadline_budget
  except AttributeError:
    # no deadline_budget from the request, do nothing
    pass
```

tests/integration/timeout_tests.py
```diff
@@ -23,18 +27,39 @@ def test_default_timeout():
...
     with baseplate.make_server_span(context, "test"):
         with pytest.raises(ServerTimeout):
             gevent.sleep(1)
 
+    # tests long deadline_budget and server timeout doesn't timeout
     context = baseplate.make_context_object()
+    baseplate.add_to_context("deadline_budget", 999999999999)
     with baseplate.make_server_span(context, "test"):
-        gevent.sleep(0)  # shouldn't time out since it's so fast!
+        gevent.sleep(1)  # shouldn't time out since timeouts are so long
```
fishy: This means this test would take at least 1 second to run, which is in general not a good idea. Try avoid long sleeps in tests, use something like `gevent.sleep(0.01)` and a deadline_budget higher than 10 millisecond (for example, "1000") instead.

(I think the other sleeps with a shorter timeout would be better because they will timeout before sleep for 1 full second.)

tests/integration/timeout_tests.py
```diff
@@ -23,18 +27,39 @@ def test_default_timeout():
...
+        with pytest.raises(ServerTimeout):
+            gevent.sleep(1)
 
-    observer = TimeoutBaseplateObserver.from_config(
-        {"server_timeout.default": "1 hour", "server_timeout.by_endpoint.test": "5 milliseconds"}
-    )
-    baseplate.register(observer)
+    context = baseplate.make_context_object()
+    with baseplate.make_server_span(context, "test"):
+        gevent.sleep(0)  #
```
fishy: what does this one do?

ethverie: this test was originally here before my changes!

baseplate/observers/timeout.py
```diff
@@ -43,10 +43,25 @@ def __init__(self, timeout_config: config.ConfigNamespace):
...
+            deadline_budget = context.deadline_budget
+        except AttributeError:
+            deadline_budget = None
```
fishy: This part is no longer needed.

tests/integration/thrift_tests.py
```diff
@@ -455,7 +462,7 @@ def example(self, context):
                     svc.example()
 
         # this should be 1 second (1000 ms)
-        assert handler.context.headers.get(b"Deadline-Budget").decode() == str(1000)
+        assert handler.context.deadline_budget == 1.0  # pool timeout seconds
```
fishy: In general using `==` on floats will cause issues and we usually use something like https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual instead. Same for other `==` asserts.

fishy (approved):

baseplate/frameworks/thrift/__init__.py
```diff
@@ -103,6 +103,11 @@ def call_processor_with_span_context(
             if edge_context_factory:
                 context.edge_context = edge_context_factory.from_upstream(edge_payload)
 
+            deadline_budget = headers.get(b"Deadline-Budget", None)
+            context.deadline_budget = (
+                float(deadline_budget.decode()) / 1000 if deadline_budget else None
+            )
```
spladug:
```suggestion
            try:
                raw_deadline_budget = headers[b"Deadline-Budget"].decode()
                context.deadline_budget = float(raw_deadline_budget) / 1000
            except (KeyError, ValueError):
                context.deadline_budget = None
```

It'd be good to do some error handling so we can't accidentally end up crashing apps by sending e.g `Deadline-Budget: hahahah`.  Just for example purposes I restructured this as a big try/except since I think making the code less hedgy  (`.get()` etc) makes it a bit clearer. But feel free to do it elsehow if this doesn't make sense.

baseplate/observers/timeout.py
```diff
@@ -43,10 +43,21 @@ def __init__(self, timeout_config: config.ConfigNamespace):
...
     def on_server_span_created(self, context: RequestContext, server_span: ServerSpan) -> None:
         timeout = self.config.by_endpoint.get(server_span.name, self.config.default)
+
+        min_timeout = None
         if timeout is not config.InfiniteTimespan:
-            observer = TimeoutServerSpanObserver(
-                server_span, timeout.total_seconds(), self.config.debug
-            )
+            min_timeout = timeout.total_seconds()
+        try:
```
spladug, line 51: nit: can you put a blank line before this so it's disconnected from the preceding `if` block please?

baseplate/observers/timeout.py
```diff
@@ -43,10 +43,21 @@ def __init__(self, timeout_config: config.ConfigNamespace):
...
+            deadline_budget = context.deadline_budget
+            if deadline_budget:
+                if not min_timeout or deadline_budget < timeout.total_seconds():
```
spladug:
```suggestion
                if not min_timeout or deadline_budget < min_timeout:
```

spladug (changes requested):
cool! it's really neat to see this all come together. only blocker I see is some error handling in case of bad incoming headers but otherwise it's just style nits. awesome.

baseplate/frameworks/thrift/__init__.py
ethverie: this is a great point, thanks!

spladug (approved):
looks great, thank you!

PR #579, opened 2021-03-30
Issue 575 fix no content exceptions

danielduke:
This is a fairly hacky fix, but it does resolve the immediate issue.

AFAICT, the root issue is that the app_iter does not have the custom content in it so when it builds the response body it renders nothing. However, the default app_iter does is able to get that information somehow when it builds the response body from the app_iter.

A better solution might be to include the response object in the custom app_iter and checking for custom properties on the response once we're done iterating over the app_iter.

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -2,6 +2,7 @@ import base64
 import logging
 import sys
 
+from itertools import chain
 from typing import Any
 from typing import Callable
 from typing import Dict
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -82,7 +82,13 @@ def _make_baseplate_tween(
             if request.span:
                 request.span.set_tag("http.status_code", response.status_code)
                 if hasattr(response, "explanation") and response.explanation:
-                    response.app_iter = SpanFinishingAppIterWrapper(request.span, chain(response.app_iter, [str.encode(response._status), str.encode(response.explanation)]))
+                    response.app_iter = SpanFinishingAppIterWrapper(
+                        request.span,
+                        chain(
+                            response.app_iter,
+                            [str.encode(response._status), str.encode(response.explanation)],
+                        ),
+                    )
                 else:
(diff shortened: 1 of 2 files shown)


baseplate/frameworks/pyramid/__init__.py
```diff
@@ -81,7 +82,16 @@ def baseplate_tween(request: Request) -> Response:
         else:
             if request.span:
                 request.span.set_tag("http.status_code", response.status_code)
-                response.app_iter = SpanFinishingAppIterWrapper(request.span, response.app_iter)
+                if hasattr(response, "explanation") and response.explanation:
+                    response.app_iter = SpanFinishingAppIterWrapper(
+                        request.span,
+                        chain(
+                            response.app_iter,
+                            [str.encode(response._status), str.encode(response.explanation)],
```
spladug, line 90: why `str.encode(...)` vs `foo.encode()`?

spladug (changes requested):
Thanks for taking a crack at this!  My understanding is this sort of issue could happen to any response object that modifies its own body/app_iter during the `__call__` phase which would mean that this fix sorta helps with HTTPException subclasses but not other cases this can happen in. Maybe that's sufficient for our use cases, but I want to understand a bit better before committing to targeted fixes.

Does chaining the app_iter like you've done actually produce the same sort of body that we expect from these responses? It looks like we'd just send the status code and message but none of the rest of that payload?

baseplate/frameworks/pyramid/__init__.py
danielduke: It looks like this may be only a test issue, but in this line https://github.com/Pylons/webtest/blob/master/webtest/lint.py#L296
its checking to make sure that everything in the app_iter is a binary type which fails if we don't convert to binary when adding to the app_iter.

We might be able to make it more generic by doing this `[bytes(response._status, 'utf-8'), bytes(response.explanation, 'utf-8')]`

RyanBalfanz:
@maeivysea Should this be closed now that you've addressed the issue elsewhere?

maeivysea:
@RyanBalfanz Thank you!! Yes, this can be closed. [Issue 575](https://github.com/reddit/baseplate.py/issues/575) was closed by PR #739

PR #584, opened 2021-04-14
Don't emit JSON logs to TTYs

spladug:
It's super annoying to try and read the JSON and we can reasonably
assume it's a human watching when stdout is a TTY.


MelissaCole (approved):
Nice :) I was almost getting used to parsing the json!
:fish: after mypy is happy again

PR #586, opened 2021-04-27
Add runtime metrics for memcache pools

curioussavage:
I Monkey patched this in post service and have been running it for a while now. Didn't notice any negative impacts so I figure it could be useful for any other memcached users.

Exceptions like [this](https://sentry.reddit.ue1.snooguts.net/sentry/posts/issues/2667031/) are what motivated me to start tracking memcache pool utilization.


spladug (approved):
love it, thank you!

PR #588, opened 2021-05-03
Add timeouts to HTTP requests for DVP-92

kylelemons:
This is a probable fix for secrets-fetcher sidecars sometimes hanging
and causing spinnaker deploys to fail.  The Requests library by default
will not apply a timeout, and so will happily wait forever if there is
no HTTP response.


spladug (changes requested):
LGTM, just the lint stylecheck stuff and this should be good to go

kylelemons:
Lint errors fixed, thanks @spladug !

spladug (approved):

PR #589, opened 2021-05-08
Update docstring to use read_file

alisaifee:
# Description
`configparser.readfp` has been deprecated and it is recommended to use `read_file` instead. This simply updates the docstrings/doctests  to use the  recommended method.

[Reference](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.readfp) 

Tested locally using:
```
 docker-compose run baseplate make test
```


spladug (approved):
thanks!

PR #590, opened 2021-05-12
Support publishing v2j batches in the event publisher sidecar.

aearly:


baseplate/sidecars/event_publisher.py
```diff
@@ -181,7 +192,7 @@ def publish_events() -> None:
         {
             "collector": {
                 "hostname": config.String,
-                "version": config.Optional(config.Integer, default=1),
+                "version": config.Optional(config.String, default="2"),
```
aearly, line 195: This is a small functional change with the default value, but it seemed the default value would always cause an error (perhaps that is intended?).

baseplate/sidecars/event_publisher.py
```diff
@@ -86,10 +86,21 @@ def reset(self) -> None:
         self._size = len(self._header) + len(self._end)
 
 
+class V2JBatch(V2Batch):
```
spladug, line 89: could you add a test here please? https://github.com/reddit/baseplate.py/blob/develop/tests/unit/lib/events/publisher_tests.py#L40

just something simple to exercise this code and make sure it doesn't regress in the future

spladug (changes requested):
LGTM modulo a test. thanks!

baseplate/sidecars/event_publisher.py
spladug: oops, a bad default from the days when v1 still existed. this looks good, thanks.

spladug (approved):
LGTM!

PR #591, opened 2021-05-12
Add a drain period to server shutdown

spladug:
When a baseplate-serve process receives a SIGTERM, it puts the server
into a graceful shutdown state immediately. New requests are no longer
accepted and the process continues handling in-flight requests until
they're all complete or stop_timeout seconds have elapsed.

Kubernetes won't mark the pod as TERMINATING until the server has
completed its graceful shutdown or has spent long enough shutting down
that it fails to respond to a liveness healthcheck and is forceably shut
down.

During this intervening time, new requests will still be routed to the
pod but the server will not be listening for them. So they get dropped
on the floor.

To prevent this, we add a drain_time period that happens before the
graceful shutdown is kicked off. If configured, a SIGTERM will cause
baseplate-serve to set a global flag indicating shutdown has begun and
then it will wait the specified time until beginning the actual graceful
shutdown. This gives the application a chance to deliberately fail
READINESS healthchecks during that grace period and get taken out of
rotation so that once it starts graceful shutdown it should not be
getting any new requests.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -15,9 +15,11 @@ import signal
 import socket
 import sys
 import threading
+import time
 import traceback
 import warnings
 
+from dataclasses import dataclass
 from types import FrameType
 from typing import Any
 from typing import Callable
@@ -34,12 +36,23 @@ from gevent.server import StreamServer
 from baseplate import Baseplate
 from baseplate.lib.config import Endpoint
 from baseplate.lib.config import EndpointConfiguration
+from baseplate.lib.config import Optional as OptionalConfig
+from baseplate.lib.config import parse_config
+from baseplate.lib.config import Timespan
 from baseplate.lib.log_formatter import CustomJsonFormatter
 from baseplate.server import einhorn
 from baseplate.server import reloader
 
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ServerState:
+    shutting_down: bool = False
+
+
+SERVER_STATE = ServerState()
 
 
 def parse_args(args: Sequence[str]) -> argparse.Namespace:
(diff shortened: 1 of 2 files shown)


baseplate/server/__init__.py
```diff
@@ -246,13 +261,20 @@ def load_app_and_run_server() -> None:
...
-    logger.info("Listening on %s, PID:%s", listener.getsockname(), os.getpid())
+    logger.info("Listening on %s", listener.getsockname())
     server.start()
     try:
         shutdown_event.wait()
-        logger.info("Finally stopping server, PID:%s", os.getpid())
+
+        SERVER_STATE.shutting_down = True
+
+        if cfg.drain_time:
```
jerroydmoore, line 271: IMO this should be `if cfg.drain_time and cfg.drain_time != 0`, so that we can update our cookiecutter INI file to include the field with a explanatory comment. something like:

```
# drain_time should be greater than k8s' (failureThreshold * periodSeconds)
# see <<insert path to k8s config>>
drain_time = 0
```

spladug: we can also just have 

```
# drain time should be blah blah...
drain_time =
```

in the config and this'll work as-is

spladug: (or put a value in the cookiecutter that works for our infra and be done with it :)

docs/cli/serve.rst
```diff
@@ -126,6 +126,42 @@ An example command line::
...
+.. code-block:: py
+
+   def is_healthy(self, context, healthcheck_request):
+       if healthcheck_request.probe == IsHealthyProbe.LIVENESS:
+           return True
+       elif healthcheck_request.probe == IsHealthyProbe.READINESS:
+           if SERVER_STATE.shutting_down:
+               return False
+           return True
+       return True
```
bradengroom, line 162: Should we create a cookie cutter PR to do this?

bradengroom (approved):

docs/cli/serve.rst
spladug: very yes!

PR #594, opened 2021-05-17
Make ServerState.shutting_down into an enum

spladug:
This is more extensible and should've been done all along.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -5,6 +5,7 @@ This command serves your application from the given configuration file.
 """
 import argparse
 import configparser
+import enum
 import fcntl
 import gc
 import importlib
@@ -20,6 +21,7 @@ import traceback
 import warnings
 
 from dataclasses import dataclass
+from enum import Enum
 from types import FrameType
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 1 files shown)


spladug:
:eyeglasses: @mattknox

baseplate/server/__init__.py
```diff
@@ -47,9 +50,19 @@
 logger = logging.getLogger(__name__)
 
 
+class ServerLifecycle(Enum):
+    RUNNING = enum.auto()
+    SHUTTING_DOWN = enum.auto()
```
mattknox, line 55: This is great!  Is the ordering sensitive?  (It almost certainly doesn't matter unless you're serializing stuff, which is hard to imagine)

baseplate/server/__init__.py
```diff
@@ -266,7 +279,7 @@ def load_app_and_run_server() -> None:
     try:
         shutdown_event.wait()
 
-        SERVER_STATE.shutting_down = True
+        SERVER_STATE.state = ServerLifecycle.SHUTTING_DOWN
```
mattknox, line 282: Lovely!

mattknox (approved):
My name is Matt Knox, and I approve this change.

baseplate/server/__init__.py
spladug: yeah, it's ordering sensitive. if we added a new one inbetween it'd mess up the latter but this state should only ever make sense in-process

PR #601, opened 2021-07-02
Add environment variable interpolation to config parser

spladug:
This is intended to allow us to inject some configuration from
environment variables while keeping the config file explicit about where
these values are coming from.

You can do something like this in config:

```ini
foo = $BAR
```

and `foo` will be populated with the value of the environment variable `$BAR` at runtime.

This is applied to `baseplate-serve`, `baseplate-script`, and the sidecars.

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -27,6 +27,8 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
+from typing import Mapping
+from typing import MutableMapping
 from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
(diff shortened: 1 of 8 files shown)


baseplate/server/__init__.py
```diff
@@ -109,6 +111,18 @@ def parse_args(args: Sequence[str]) -> argparse.Namespace:
     return parser.parse_args(args)
 
 
+class EnvironmentInterpolation(configparser.Interpolation):
+    def before_get(
```
spladug, line 115: cribbing off typeshed's annotations for this API https://github.com/python/typeshed/blob/master/stdlib/configparser.pyi#L41

MelissaCole (approved):
lgtm!

KTAtkinson (approved):

kylelemons (approved):

PR #603, opened 2021-07-18
Require redis-py-cluster version 2.1.3

jerroydmoore:
See diff: https://github.com/Grokzen/redis-py-cluster/compare/2.1.2...2.1.3

Essentially, one of our services is losing connection raising a `ConnectionError`. Calling `connection.disconnect()` throws another exception, as `connection` is set to None.


jerroydmoore:
the failing tests are unrelated to the change I made. 🤔

FranGM (approved):
Change LGTM, but note that I still wouldn't recommend running against an unpatched upstream `redis-py-cluster` until the following PRs have made it into a release:
https://github.com/Grokzen/redis-py-cluster/pull/473
https://github.com/Grokzen/redis-py-cluster/pull/472
https://github.com/Grokzen/redis-py-cluster/pull/471

spladug (approved):
nit: This isn't pinning to 2.1.3 but requiring _at least_ 2.1.3. Might be good to update the commit message accordingly.

Do we intend to do this every time there's a bug fix in the upstream? These requirements are more about compatibility with baseplate than being comprehensive of all known issues (at least for the other libraries) but I do understand it's useful to have the bounds if we know it's completely unusable.

jerroydmoore:
@FranGM , did you for the lib to run a custom version with the patches?
@spladug updated the PR title! Interesting question! IMHO if we can, we should update the required versions for libs in baseplate. It will force service owners to update their deps, which they should be doing regularly. Also, if there are operational issues with the required version of libraries, then having an opinionated baseplate-py would be more helpful than hurtful.

spladug:
Agreed about operational issues. But sometimes version bumps mean _new_ issues, so it's something we should be sorta careful about. In this case I think it's totally reasonable since there are known blocking issues. Thanks for doing this!

jerroydmoore:
Closing this PR since we're going to be using a patched version with backported bug fixes

PR #604, opened 2021-07-22
Conform zipkin traceId, spanId, parentId specs

praxist:
Using v1 API defined in https://github.com/openzipkin/zipkin-api/blob/7692ca7be4dc3be9225db550d60c4d30e6e9ec59/zipkin-api.yaml

traceId to 32 character lowercase hex string
spanId to 16 character lowercase hex string
parentId to be absent for root span

Currently required changes to prove out a grafana-agent + tempo tracing architecture.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -124,8 +124,9 @@ class TraceInfo(NamedTuple):
         with any upstream requests.
 
         """
-        trace_id = str(random.getrandbits(64))
-        return cls(trace_id=trace_id, parent_id=None, span_id=trace_id, sampled=None, flags=None)
+        trace_id = "{0:0{1}x}".format(random.getrandbits(128), 32)
+        span_id = "{0:0{1}x}".format(random.getrandbits(64), 16)
+        return cls(trace_id=trace_id, parent_id=None, span_id=span_id, sampled=None, flags=None)
 
     @classmethod
     def from_upstream(
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -279,7 +279,9 @@ class TraceSpanObserver(SpanObserver):
             "binaryAnnotations": binary_annotations,
         }
 
-        span["parentId"] = self.span.parent_id or 0
+        if self.span.parent_id:
+            span["parentId"] = self.span.parent_id
+
         return span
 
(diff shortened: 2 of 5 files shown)


baseplate/__init__.py
```diff
@@ -124,8 +124,9 @@ def new(cls) -> "TraceInfo":
         with any upstream requests.
 
         """
-        trace_id = str(random.getrandbits(64))
-        return cls(trace_id=trace_id, parent_id=None, span_id=trace_id, sampled=None, flags=None)
+        trace_id = "{0:0{1}x}".format(random.getrandbits(128), 32)
```
praxist, line 127: padded string with leading 0's to reach 32 characters - necessary in case hex string comes up short.

fishy:
This needs to be opt-in and default to false, as the older version of baseplate.py/go library will reject hex ids.

We already implemented the opt-in in .go, see the comment there: https://pkg.go.dev/github.com/reddit/baseplate.go/tracing#TracerConfig.UseUUID

baseplate/__init__.py
spladug:
```suggestion
        trace_id = f"{random.getrandbits(128):032x}"
```

Parameterizing the 32 makes it quite hard to read the format string,  and using f-strings makes this a little easier to understand as well.

baseplate/__init__.py
```diff
@@ -124,8 +124,9 @@ def new(cls) -> "TraceInfo":
...
+        span_id = "{0:0{1}x}".format(random.getrandbits(64), 16)
```
spladug, line 128:
```suggestion
        span_id = f"{random.getrandbits(64):016x}"
```

baseplate/observers/tracing.py
```diff
@@ -279,7 +279,9 @@ def _to_span_obj(
             "binaryAnnotations": binary_annotations,
         }
 
-        span["parentId"] = self.span.parent_id or 0
```
spladug, line 282: Hm. This means we don't send a parentId at all if it's undefined. Is that right? No default?

spladug (changes requested):
Thanks! It sounds like we're going ahead with the hex format so this looks good but as @fishy pointed out we'll need this to be opt-in for now.

PR #605, opened 2021-07-28
redis_cluster: add moar timeouts

jerroydmoore:
- Add support for the options socket_timeout and socket_connect_timeout.
- Set socket_timeout and socket_connection_timeout to timeout's value,
  if and only if, these weren't already set.

### DIFF
--- a/baseplate/clients/redis_cluster.py
+++ b/baseplate/clients/redis_cluster.py
@@ -262,6 +262,8 @@ def cluster_pool_from_config(
             "url": config.String,
             "max_connections": config.Optional(config.Integer, default=50),
             "max_connections_per_node": config.Optional(config.Boolean, default=False),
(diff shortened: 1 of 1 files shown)


FranGM (approved):

bradengroom (changes requested):
Can you add docs in the function docstring? This helps generate docs [here](https://baseplate.readthedocs.io/en/stable/api/baseplate/clients/redis_cluster.html?#baseplate.clients.redis_cluster.cluster_pool_from_config).

spladug (changes requested):
lgtm modulo docs like @bradengroom pointed out. especially important to document that `timeout` will propagate to the other two settings if not explicitly set.

bradengroom (approved):

spladug (approved):
ty!

kylelemons (approved):

PR #607, opened 2021-08-12
Disable local var reporting to Sentry by default

kylelemons:


baseplate/observers/sentry.py
```diff
@@ -87,6 +87,8 @@ def init_sentry_client_from_config(raw_config: config.RawConfig, **kwargs: Any)
     ignore_errors.extend(ALWAYS_IGNORE_ERRORS)
     ignore_errors.extend(cfg.sentry.ignore_errors)
     kwargs.setdefault("ignore_errors", ignore_errors)
+    
+    kwargs.setdefault("with_locals", false)
```
spladug:
```suggestion
    kwargs.setdefault("with_locals", False)
```

spladug (changes requested):

baseplate/observers/sentry.py
kylelemons: oops, eeevery time

spladug (approved):

pacejackson (approved):

PR #609, opened 2021-08-19
Run tests on Python 3.9

spladug:
Since we're using Python 3.9 in a few places already it's important our test suite runs against it. Pylint needed an upgrade to be able to handle some changes in the language. That's the bulk of this PR. I took the opportunity to bump mypy as well but it only affected some pragma comments so there's no functional change.

### DIFF
--- a/.github/workflows/python-package.yaml
+++ b/.github/workflows/python-package.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
 
     container:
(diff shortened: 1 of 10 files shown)


fishy (approved):

.github/workflows/python-package.yaml
```diff
@@ -37,6 +37,8 @@ jobs:
             ${{ runner.os }}-pip-
 
       - name: Install dependencies
+        env:
+          CASS_DRIVER_NO_EXTENSIONS: theytaketoolongtobuild
```
MelissaCole, line 41: :smile:

MelissaCole (approved):
lgtm!

PR #610, opened 2021-08-23
Add baseplate-shell logging for audit purposes

sp3nx0r:
This PR aims to add audit logging (or what is at least possible) for the `baseplate-shell`functionality. 

* IPython
~~So IPython doesn't have an API for capturing input, but does have a history logging functionality that can be forced via CLI. This was opted over the messiness of dealing with IPython profiles (especially in the container space). This allows realtime streaming of the commands to stdout for shipping upstream. There's no option to prefix this to help parse the logs, but any non-JSON error logs should be treated as human interacted so that should be sufficient. It does initialize with a `# IPython log file` entry~~

PR now approaches this by monkeypatching `IPython.logger.log_write()` to be able to handle log formatting and also disable the ability to stop logs.

* code.InteractiveConsole
InteractiveConsole, because it's dependent on `readline` for it's history functionality, can leverage a common pattern of capturing the history and writing it out upon session close. This does have a draw back of a long running sessions or an ungraceful shutdown dropping the logs. It might be more preferable to overriding `readline`'s stdin/stdout behaviors...

Note: also added tab completion as part of the InteractiveConsole setup

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -542,4 +542,4 @@ class LoggedInteractiveConsole(code.InteractiveConsole):
         prompt = f"<{self.pri}>1 {timestamp} {self.hostname} baseplate-shell {self.pid} {message_id} {structured} {message}"
         with open(self.output_file, "w") as f:
             print(prompt, file=f)
-        self.output_file.flush()
+            f.flush()
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -469,7 +469,6 @@ def load_and_run_shell() -> None:
             """
         ]
         ipython_config.TerminalInteractiveShell.banner2 = banner
-        # ipython_config.InteractiveShell.logfile = console_logpath
         ipython_config.LoggingMagics.quiet = True
         start_ipython(argv=[], user_ns=env, config=ipython_config)
         raise SystemExit
@@ -504,12 +503,12 @@ def _is_containerized() -> bool:
     """Determine if we're running in a container based on cgroup awareness for various container runtimes."""
     path = "/proc/self/cgroup"
     in_container = ["kubepods", "docker", "containerd"]
-    if (
-        os.path.exists("/.dockerenv")
-        or os.path.isfile(path)
-        and any(in_container in line for line in open(path))
-    ):
+    if os.path.exists("/.dockerenv"):
         return True
+    elif os.path.isfile(path):
+        for c in in_container:
+            if (c in line for line in open(path)):
+                return True
     return False
 
 
@@ -526,8 +525,8 @@ class LoggedInteractiveConsole(code.InteractiveConsole):
     def __del__(self):
         self.output_file.close()
 
-    def raw_input(self, _prompt=""):
-        data = input(_prompt)
+    def raw_input(self, prompt=""):
+        data = input(prompt)
         print(f"{datetime.now()} {os.getpid()} - {data}", file=self.output_file)
         self.output_file.flush()
         return data
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -39,7 +39,6 @@ from typing import TextIO
 from typing import Tuple
 
 from gevent.server import StreamServer
-from psutil import Process
 
 from baseplate import Baseplate
 from baseplate.lib import warn_deprecated
@@ -179,7 +178,7 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     root_logger.addHandler(handler)
 
     # add PID 1 stdout logging if we're containerized and not running under an init system
-    if _is_containerized() and Process(1) not in Process(os.getpid()).parents():
+    if _is_containerized() and not _has_PID1_parent():
         file_handler = logging.FileHandler("/proc/1/fd/1", mode="w")
         file_handler.setFormatter(formatter)
         root_logger.addHandler(file_handler)
@@ -516,6 +515,18 @@ def _is_containerized() -> bool:
         pass
 
     return False
+
+
+def _has_PID1_parent() -> bool:
+    """Determine parent PIDs up the tree until PID 1 or 0 is reached, do this natively"""
+    parent_pid = os.getppid()
+    while parent_pid > 1:
+        with open(f"/proc/{parent_pid}/status", "r") as proc_status:
+            for line in proc_status.readlines():
+                if line.startswith("PPid:"):
+                    parent_pid = int((line.replace("PPid:", "")))
+                    break
+    return True if parent_pid == 1 else False
 
 
 class LoggedInteractiveConsole(code.InteractiveConsole):
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -179,7 +179,7 @@ def configure_logging(config: Configuration, debug: bool) -> None:
 
     # add PID 1 stdout logging if we're containerized and not running under init system
     if _is_containerized() and os.getppid() != 1:
-        file_handler = logging.FileHandler("/proc/1/fd/1", mode='w')
+        file_handler = logging.FileHandler("/proc/1/fd/1", mode="w")
         file_handler.setFormatter(formatter)
         root_logger.addHandler(file_handler)
 
@@ -537,7 +537,9 @@ class LoggedInteractiveConsole(code.InteractiveConsole):
         self.log_event(message=data, message_id="CEXC")
         return data
 
-    def log_event(self, message: str, message_id: Optional[str] = "-", structured: Optional[str] = "-"):
+    def log_event(
+        self, message: str, message_id: Optional[str] = "-", structured: Optional[str] = "-"
+    ):
         """Generate an RFC 5424 compliant syslog format."""
         timestamp = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
         prompt = f"<{self.pri}>1 {timestamp} {self.hostname} baseplate-shell {self.pid} {message_id} {structured} {message}"
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -4,6 +4,7 @@ This command serves your application from the given configuration file.
 
 """
 import argparse
+import atexit
 import configparser
 import enum
 import fcntl
@@ -22,6 +23,7 @@ import warnings
 
 from dataclasses import dataclass
 from enum import Enum
+from rlcompleter import Completer
 from types import FrameType
 from typing import Any
 from typing import Callable
@@ -415,6 +417,8 @@ def load_and_run_shell() -> None:
         setup = _load_factory(config.shell["setup"])
         setup(env, env_banner)
 
+    configure_logging(config, False)
+
     # generate banner text
     banner = "Available Objects:\n"
     for var in sorted(env_banner.keys()):
@@ -441,14 +445,31 @@ def load_and_run_shell() -> None:
         newbanner = f"Baseplate Interactive Shell\nPython {sys.version}\n\n"
         banner = newbanner + banner
 
-        # import this just for its side-effects (of enabling nice keyboard
-        # movement while editing text)
         try:
-            import readline
+            _interactive_console_setup(env)
 
-            del readline
         except ImportError:
             pass
 
         shell = code.InteractiveConsole(locals=env)
         shell.interact(banner)
+
+
+def _interactive_console_setup(env: Dict) -> None:
+    # setup some quality of life console interactions
+    import readline
+    readline.set_completer(Completer(env).complete)
+    readline.parse_and_bind("tab: complete")
+
+    # create audit log output for executed commands
+    pid_1_path = "/proc/1/fd/1"
+    # check if running in a containerized environment
+    if (os.environ.get('KUBERNETES_SERVICE_HOST', False) and os.access(pid_1_path, os.W_OK)):
+        log_output = os.path.abspath(pid_1_path)
+    else:
+        log_output = os.path.expanduser("~/.shell_history")
+
+    def save_console_history(history_path=log_output) -> None:
+        readline.write_history_file(history_path)
+
+    atexit.register(save_console_history)
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -177,7 +177,8 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     root_logger.setLevel(logging_level)
     root_logger.addHandler(handler)
 
-    if _is_containerized():
+    # add PID 1 stdout logging if we're containerized and not running under init system
+    if _is_containerized() and os.getppid() != 1:
         file_handler = logging.FileHandler("/proc/1/fd/1")
         file_handler.setFormatter(formatter)
         root_logger.addHandler(file_handler)
@@ -466,6 +467,7 @@ def load_and_run_shell() -> None:
                     self.logfile.flush()
             ip.logger.logstop = None
             ip.logger.log_write = partial(log_write, ip.logger)
+            ip.logger.log_write("Start IPython logging\\n")
             """
         ]
         ipython_config.TerminalInteractiveShell.banner2 = banner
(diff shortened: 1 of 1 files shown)


baseplate/server/__init__.py
```diff
@@ -415,11 +422,15 @@ def load_and_run_shell() -> None:
         setup = _load_factory(config.shell["setup"])
         setup(env, env_banner)
 
+    configure_logging(config, False)
```
MelissaCole: Should this be `configure_logging(config, debug=args.debug)` to enable debug logging when requested?

baseplate/server/__init__.py
```diff
@@ -433,22 +444,43 @@ def load_and_run_shell() -> None:
...
+def _InteractiveConsole_setup(env: Dict, console_logpath: str) -> None:
+    # Setup some quality of life console interactions
+    import readline
+
+    readline.set_completer(Completer(env).complete)
+    readline.parse_and_bind("tab: complete")
+
+    # Define audit logging with readline history
+    def save_console_history(history_path=console_logpath) -> None:
+        readline.write_history_file(history_path)
```
MelissaCole: Do you want to `set_history_length` so it doesn't grow unruly?

MelissaCole: :no_bell: It seems like if you don't `read_history_file` then `write_history_file` will overwrite the file. Is that ok for auditing purposes?

MelissaCole (changes requested):
lgtm, just a few questions

kylelemons:
I don't know enough about baseplate.py, so I'll defer to @spladug here

baseplate/server/__init__.py
sp3nx0r: yup, thought `args` didn't exist in that function, but it's loaded up above, so fixing this

baseplate/server/__init__.py
sp3nx0r: I'd like to see an unlimited history size, but it's probably more prudent to set a high upper bounds. Unlikely that a human would have _that_ many commands in a history. 

As for `read_history_file`, I opted to not include that because this output should be streamed out and not rely on local disk storage. Unlikely that a concurrency issue would arise if running not in a containerized environment.

MelissaCole (approved):
lgtm! (besides the linting error)

baseplate/server/__init__.py
```diff
@@ -174,6 +176,11 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     root_logger.setLevel(logging_level)
     root_logger.addHandler(handler)
 
+    if os.getpid() != 1:
```
spladug: What are we testing for here? Would `dumb-init` confuse this check?

baseplate/server/__init__.py
```diff
@@ -433,22 +444,43 @@ def load_and_run_shell() -> None:
...
 
-            del readline
         except ImportError:
             pass
 
         shell = code.InteractiveConsole(locals=env)
         shell.interact(banner)
+
+
+def _shell_commands_log_path() -> str:
```
spladug:
```suggestion
def _get_shell_log_path() -> str:
```

It'd be clearer to stick to imperative names for the functions.

baseplate/server/__init__.py
```diff
@@ -433,22 +444,43 @@ def load_and_run_shell() -> None:
...
+    # Define path for console log output
+    pid_1_path = "/proc/1/fd/1"
+    # check if running in a containerized environment
+    if os.getenv("KUBERNETES_SERVICE_HOST") and os.access(pid_1_path, os.W_OK):
+        return os.path.abspath(pid_1_path)
+    return os.path.abspath("/var/log/.shell_history")
+
+
+def _InteractiveConsole_setup(env: Dict, console_logpath: str) -> None:
```
spladug:
```suggestion
def _set_up_interactive_console(env: Dict, console_logpath: str) -> None:
```

spladug: why pull this out into its own function? just to keep things shorter in the main function?

baseplate/server/__init__.py
```diff
@@ -433,22 +444,43 @@ def load_and_run_shell() -> None:
 
         ipython_config = Config()
         ipython_config.TerminalInteractiveShell.banner2 = banner
-        start_ipython(argv=[], user_ns=env, config=ipython_config)
+        start_ipython(argv=[f"--logfile=/{console_logpath}"], user_ns=env, config=ipython_config)
```
spladug: fake passing a command line argument feels a bit brittle. is there no direct way to configure this?

baseplate/server/__init__.py
```diff
@@ -433,22 +444,43 @@ def load_and_run_shell() -> None:
...
             pass
 
         shell = code.InteractiveConsole(locals=env)
         shell.interact(banner)
+
+
+def _shell_commands_log_path() -> str:
+    # Define path for console log output
+    pid_1_path = "/proc/1/fd/1"
+    # check if running in a containerized environment
```
spladug: > containerized environment

Is pretty general, but we're specifically checking if we're in k8s below. Do we care about docker-compose etc? Or just k8s? If the latter, maybe we should call that out specifically?

spladug (changes requested):
Cool! I'm a bit unsure of the specifics of PID and envvar checks going on here so a bit more explanation in the code would be nice. I think this is a good step in the right direction but definitely not foolproof. 

> Note: also added tab completion as part of the InteractiveConsole setup

Can you make this a separate commit please? If we had to revert one but not the other it's easier that way. Also makes it easier to call this stuff out in release notes.

> This does have a draw back of a long running sessions or an ungraceful shutdown dropping the logs

That's great that IPython does real-time streaming. There's no way to force that with readline? I suspect we'll miss a lot of logs if not.

baseplate/server/__init__.py
sp3nx0r: this is now abstracted out to an `_is_containerized()` functional call which looks like the preferred way to capture this. if there's additional runtimes we need to add, it's easy to do so

baseplate/server/__init__.py
sp3nx0r: looks like there is, added this to the IPython Config variable

baseplate/server/__init__.py
sp3nx0r: Replaced this with the `_is_containerized()` check as well, as this wasn't a graceful check. This code now purposefully writes loglines also to PID 1 stdout if containerized, so we can capture script/shell executions

sp3nx0r:
> That's great that IPython does real-time streaming. There's no way to force that with readline? I suspect we'll miss a lot of logs if not.

Agreed, and figured out how we can do this now by overriding `code.InteractiveConsole.raw_input()` and adding logging code there. Opted for a non `logger` approach because otherwise things get duplicated in the terminal and it looks a bit rough. 

Looking into how we might be able to do that with IPython

baseplate/server/__init__.py
sp3nx0r: ended up scrapping the `readline` approach, so resolving this

sp3nx0r:
> Looking into how we might be able to do that with IPython

https://github.com/reddit/baseplate.py/pull/610/commits/f00bc712e2762b9f7da2c78a0547f775386cc668 adds a monkeypatch of IPython's logger to provide logging. There's no graceful way to use the in-built logging logic

baseplate/server/__init__.py
```diff
@@ -432,23 +444,90 @@ def load_and_run_shell() -> None:
...
+                        odata = u'\\n'.join([u'#[Out]# %s' % s
+                                        for s in data.splitlines()])
+                        write(u'%s\\n' % odata)
+                    self.logfile.flush()
+            ip.logger.logstop = None
+            ip.logger.log_write = partial(log_write, ip.logger)
+            """
+        ]
         ipython_config.TerminalInteractiveShell.banner2 = banner
+        # ipython_config.InteractiveShell.logfile = console_logpath
```
spladug: ?

baseplate/server/__init__.py
```diff
@@ -432,23 +444,90 @@ def load_and_run_shell() -> None:
...
+
+
+def _is_containerized() -> bool:
+    """Determine if we're running in a container based on cgroup awareness for various container runtimes."""
+    path = "/proc/self/cgroup"
+    in_container = ["kubepods", "docker", "containerd"]
+    if (
+        os.path.exists("/.dockerenv")
+        or os.path.isfile(path)
+        and any(in_container in line for line in open(path))
```
spladug: Does this work in testing? I don't think anything but a string can be on the left side of `in` here.

baseplate/server/__init__.py
```diff
@@ -432,23 +444,90 @@ def load_and_run_shell() -> None:
...
+
+
+def _get_shell_log_path() -> str:
+    """Determine where to write shell audit logs."""
+    # Define path for console log output
+    if _is_containerized():
+        # write to PID 1 stdout for log aggregation
+        return os.path.abspath("/proc/1/fd/1")
+    # otherwise write to a local file
+    return os.path.abspath("/var/log/.shell_history")
```
spladug: What are the abspath calls for? These are both already hard-coded absolute paths.

baseplate/server/__init__.py
```diff
@@ -432,23 +444,90 @@ def load_and_run_shell() -> None:
             from IPython import Config
 
         ipython_config = Config()
+        ipython_config.InteractiveShellApp.exec_lines = [
```
spladug, line 448: _yikes_. haha. is there no way to pass in a function or something since we're code and not a config file?

baseplate/server/__init__.py
```diff
@@ -174,6 +177,11 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     root_logger.setLevel(logging_level)
     root_logger.addHandler(handler)
 
+    if _is_containerized():
```
spladug: Is this really something we want production services to do? Aren't they already going to be outputting to their own stdout? Would it make more sense to push this into the load_and_run_shell function only?

spladug (changes requested):
Thanks for the updates. Let me know if you need any help getting set up for testing this out.

baseplate/server/__init__.py
sp3nx0r: Yes, it's admittedly bananas. There is an option to pass in a file via `InteractiveShellApp.exec_files` but resolving the absolute path might be tricky here (since we'd need to spelunk into the pip installed directory). This is essentially running a string (either via `lines` or `files`) with IPython interpreter before the shell gets dropped into.

baseplate/server/__init__.py
sp3nx0r: snuck in during testing, this is extraneous now that it's loading in via `ip.magic('logstart {console_logpath}')`. removing

baseplate/server/__init__.py
sp3nx0r: It worked only because the `/.dockerenv` was true and short-circuiting. Will fix

baseplate/server/__init__.py
sp3nx0r: What probably makes more sense is to check if we're running under parent PID 1. If we're not under `dumb-init` then we'd want to add the stdout logging. Addressed in 499353c

baseplate/server/__init__.py
```diff
@@ -432,23 +445,90 @@ def load_and_run_shell() -> None:
...
+def _is_containerized() -> bool:
+    """Determine if we're running in a container based on cgroup awareness for various container runtimes."""
+    path = "/proc/self/cgroup"
+    in_container = ["kubepods", "docker", "containerd"]
+    if os.path.exists("/.dockerenv"):
+        return True
+    elif os.path.isfile(path):
+        for c in in_container:
+            if (c in line for line in open(path)):
+                return True
```
spladug: I think this could be a little clearer like this:

```suggestion
    if os.path.exists("/.dockerenv"):
        return True
    
    try:
        with open("/proc/self/cgroup") as my_cgroups_file:
            my_cgroups = my_cgroups_file.read()
            
            for hint in ["kubepods", "docker", "containerd"]:
                if hint in my_cgroups:
                    return True
    except IOError:
        pass
```

A couple of advantages:

* we only read the file in once
* rather than checking if the file exists and then trying to read it as two separate things, we just try to read it and move on if that fails

baseplate/server/__init__.py
```diff
@@ -174,6 +177,12 @@ def configure_logging(config: Configuration, debug: bool) -> None:
     root_logger.setLevel(logging_level)
     root_logger.addHandler(handler)
 
+    # add PID 1 stdout logging if we're containerized and not running under init system
+    if _is_containerized() and os.getppid() != 1:
```
spladug: In my testing, this doesn't seem to cause double logging. But I don't understand why. If we're running directly under dumb-init, we should have inherited stdout from that parent process and so writing to stdout is the same as writing to the container's stdout. That'd mean that writing directly to it in addition should cause double logging shouldn't it? Since we're not wiping out our console handler right?

baseplate/server/__init__.py
```diff
@@ -432,23 +445,90 @@ def load_and_run_shell() -> None:
...
+            ip = get_ipython()
+            from functools import partial
+            ip.magic('logstart {console_logpath}')
+            def log_write(self, data, kind="input"):
+                import time, os
+                if self.log_active and data:
+                    write = self.logfile.write
+                    if kind=='input':
+                        if self.timestamp:
+                            write(time.strftime('# %a, %d %b %Y %H:%M:%S\\n', time.localtime()))
```
spladug: It might be nice to add more to the log line prefix. It's not clear after the first message where the log is coming from. Even just saying "shell" somewhere in there would be useful (but if we can get the user that shelled in that'd be even better).

spladug (changes requested):
cool! i did some testing and this is all looking good. blockers are only these:

* understanding of why it's not double logging for direct children of dumb-init
* fixing the linter errors (the suggested change to is_containerized should help with some of that)

thank you and apologies for the delayed review!

baseplate/server/__init__.py
sp3nx0r: adopted in 60bd099b82bb5149eb64be2629b8b488bd46b290

baseplate/server/__init__.py
sp3nx0r: agreed, and actually took the chance to format modern syslog format across both. this'll persist the pid and hostname across the log lines. also removed the `self.timestamp` code which is an ipython setting, but we're not going to honor it. 

unfortunately we can't get the human user that shelled in in k8s space because of the abstraction with k8s api server. but this'll be correlation done between kube api server logs and these stdout logs

baseplate/server/__init__.py
sp3nx0r: This works because we're checking the parent pid. For dumb-init spawned services, the ppid will always be 1. This line then catches anyone doing a shell/script outside of the PID1 process tree.

spladug: Ahhh I totally missed that second "p" in `getppid`.  Thanks.

So if we're running under dumb-init, but we forgot to put `exec` in `container-run.sh`, we'd be a child of a bash process that itself is a child of dumb-init, right? Do we need to walk the tree up? 

Maybe to improve readability it'd be useful to extract the ppid into a nicely named variable so it's obvious what's going on? e.g. `parent_pid = os.getppid()`.  (Or maybe I'm just the only one that can't spell haha)

spladug (approved):
cool! lgtm. one comment/question on the parent PID stuff but that's non-blocking.

sp3nx0r:
> So if we're running under dumb-init, but we forgot to put exec in container-run.sh, we'd be a child of a bash process that itself is a child of dumb-init, right? Do we need to walk the tree up?

Ah yup, we should run it up the tree. Added some logic for this with 5975b0b (assuming we can use psutil)

baseplate/server/__init__.py
```diff
@@ -432,23 +445,111 @@ def load_and_run_shell() -> None:
...
+def _has_PID1_parent() -> bool:
+    """Determine parent PIDs up the tree until PID 1 or 0 is reached, do this natively"""
+    parent_pid = os.getppid()
+    while parent_pid > 1:
+        with open(f"/proc/{parent_pid}/status", "r") as proc_status:
+            for line in proc_status.readlines():
+                if line.startswith("PPid:"):
+                    parent_pid = int((line.replace("PPid:", "")))
+                    break
+    return bool(parent_pid)
```
sp3nx0r, line 529: @spladug here's the proposed path using `/proc/PID/status` to walk upwards natively. will merge this next week

PR #611, opened 2021-08-25
Swallow and log exceptions in event_publisher sidecar

fishy:
In the same spirit of https://github.com/reddit/baseplate.py/pull/576,
we don't want sidecars to crash and bring down the whole pod, so swallow
and log the exceptions instead of crash.

### DIFF
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -225,7 +225,10 @@ def publish_events() -> None:
             batcher.add(message)
         except BatchFull:
             serialized = batcher.serialize()
-            publisher.publish(serialized)
+            try:
+                publisher.publish(serialized)
+            except Exception:
+                logger.exception("Events publishing failed.")
             batcher.reset()
             batcher.add(message)
 


baseplate/sidecars/event_publisher.py
```diff
@@ -225,7 +225,10 @@ def publish_events() -> None:
             batcher.add(message)
         except BatchFull:
             serialized = batcher.serialize()
-            publisher.publish(serialized)
+            try:
+                publisher.publish(serialized)
+            except Exception:
```
fishy, line 230: I know the blanket `except Exception` is usually considered anti-pattern in python, but I think we should make an exception here.

The alternative is to list all the possible exceptions could be thrown from `publish` and list them here, but that comes with the issue that it's error-prone, as people might add new exceptions in the future to `publish` and forgot to update the list here.

bradengroom (approved):

fishy:
The failed test on py3.9 doesn't seem related to my change. Is the test flaky? 🤔

spladug (approved):
just for my understanding, what kind of crashes were you seeing?

fishy:
@spladug https://github.com/reddit/baseplate.py/blob/f9222f534070d82bcb78101bd951bbe227ff3a0c/baseplate/sidecars/event_publisher.py#L161, but the code could also reraise on Line 151 (also the comment before Line 151 doesn't make sense to me?)

spladug:
do you know what the underlying error causing the retries to run out though? 

> (also the comment before Line 151 doesn't make sense to me?)

it's only raising the exception if the status is not 422, which indicates a validation error. they did this because invalid events should be discarded, not retried. https://github.com/reddit/baseplate.py/pull/450

fishy:
@spladug the pods are recycled and our log ingestor's search feature is not that great, so it's hard to find the log now. but the root cause was that we are sending too many v2 events, so the issue was either rate limiting or size too big or something like that.

spladug:
cool, thanks. just wanted to make sure there wasn't something else to fix too.

PR #613, opened 2021-08-26
Prevent tracing SidecarRecorder from trying to send traces that are t…

curioussavage:
I have a downstream service sending large traces and It's breaking an endpoint. The comment in this method states that it's called in the request/response path so given that trying to send messages that are too large raises it looks like we should return here.

### DIFF
--- a/baseplate/observers/tracing.py
+++ b/baseplate/observers/tracing.py
@@ -563,6 +563,7 @@ class SidecarRecorder(Recorder):
                 MAX_SPAN_SIZE,
                 len(serialized_str),
             )
+            return
         try:
             self.queue.put(serialized_str, timeout=0)
         except TimedOutError:


baseplate/observers/tracing.py
```diff
@@ -563,6 +563,7 @@ def send(self, span: TraceSpanObserver) -> None:
                 MAX_SPAN_SIZE,
                 len(serialized_str),
             )
+            return
```
kylelemons, line 566: 🔕 I don't know enough about this, maybe @spladug knows... can we truncate traces without compromising them? (I assume not, since hopefully they are some structured format)

curioussavage:
Looks like [this](https://github.com/osvenskan/posix_ipc/blob/38c7d500e445d15b799ea6e66c046032d805b050/posix_ipc_module.c#L1633) is the cause of the exception in the library. 

sample trace:
```
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1281, in _execute_context
    self.dispatch.after_cursor_execute(
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/event/attr.py", line 322, in __call__
    fn(*args, **kw)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/clients/sqlalchemy.py", line 213, in on_after_execute
    conn.info["span"].finish()
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/__init__.py", line 628, in finish
    observer.on_finish(exc_info)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/observers/tracing.py", line 227, in on_finish
    self.recorder.send(self)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/observers/tracing.py", line 567, in send
    self.queue.put(serialized_str, timeout=0)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/lib/message_queue.py", line 97, in put
    return self.queue.send(message=message)
ValueError: The message is too long

The above exception was the direct cause of the following exception:

  File "/src/post/post_thrift/PostService.py", line 2059, in process_get_user_poll_votes
    result.success = self._handler.get_user_poll_votes(args.request)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/frameworks/thrift/__init__.py", line 38, in call_with_context
    result = handler_fn(self.context, *args, **kwargs)
  File "/src/post/lib/validators.py", line 128, in execute
    return fn(self, context, request)
  File "/src/post/handlers/polls.py", line 249, in get_user_poll_votes
    polls = get_polls_from_cache(
  File "/src/post/lib/poll_utils.py", line 34, in get_polls_from_cache
    return get_from_cache(
  File "/src/post/lib/cache.py", line 49, in get_from_cache
    lookup_res = lookup_fn(still_need)
  File "/src/post/lib/poll_utils.py", line 14, in get_polls
    postgres.query(Polls)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3346, in all
    return list(self)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3508, in __iter__
    return self._execute_and_instances(context)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3533, in _execute_and_instances
    result = conn.execute(querycontext.statement, self._params)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1011, in execute
    return meth(self, multiparams, params)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1124, in _execute_clauseelement
    ret = self._execute_context(
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1316, in _execute_context
    self._handle_dbapi_exception(
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1508, in _handle_dbapi_exception
    util.raise_(newraise, with_traceback=exc_info[2], from_=e)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1281, in _execute_context
    self.dispatch.after_cursor_execute(
  File "/usr/local/lib/venv/lib/python3.8/site-packages/sqlalchemy/event/attr.py", line 322, in __call__
    fn(*args, **kw)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/clients/sqlalchemy.py", line 213, in on_after_execute
    conn.info["span"].finish()
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/__init__.py", line 628, in finish
    observer.on_finish(exc_info)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/observers/tracing.py", line 227, in on_finish
    self.recorder.send(self)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/observers/tracing.py", line 567, in send
    self.queue.put(serialized_str, timeout=0)
  File "/usr/local/lib/venv/lib/python3.8/site-packages/baseplate/lib/message_queue.py", line 97, in put
    return self.queue.send(message=message)
AssertionError
```

baseplate/observers/tracing.py
curioussavage: I had the same thought. Just wanted to get a basic fix out there first. For now I'm just disabling tracing for my service which is not ideal.

spladug:
> downstream service sending large traces

your trace sidecar should only be handling traces from your service. are you sure your service isn't doing way too much work in one request?

baseplate/observers/tracing.py
spladug: right, at this point it's a json blob so truncating would make it invalid. we could cut out some spans though.

curioussavage:
> your trace sidecar should only be handling traces from your service

Ah, I think I made a wrong assumption about that because I had no value for sampling rate in my config. I thought that would turn tracing off but when it didn't I assumed we just report them when we see them from downstream too. I see baseplate has a default though.

The endpoint just makes a 1-2 db queries and a memcached lookup. But it's used by export service I think which does a lot of work I think.

spladug:
> Ah, I think I made a wrong assumption about that because I had no value for sampling rate in my config. I thought that would turn tracing off but when it didn't I assumed we just report them when we see them from downstream too. I see baseplate has a default though.

Sample rate determines how often you'll start a new trace if you're at the root. For most services, it's meaningless since someone else is calling them and the sample rate has already been determined. But you're just getting a flag from the service calling you that says "this request should be sampled." That service itself sends its own spans and you only send your spans. So if you're hitting size limits, that's almost certainly because of something creating a lot of spans inside the same request in your service.

spladug (approved):
I think the patch looks good but we've also got some root cause stuff to figure out.

Can you shorten the commit message a little bit so it fits please? it's getting cut off right now

kylelemons:
🏃‍♂️ since it sounds like you guys have this in hand and my approval would just be a rubber stamp :)

PR #614, opened 2021-08-26
Redis client changes: Add retries

FranGM:


baseplate/clients/redis.py
```diff
@@ -108,12 +170,134 @@ def report_runtime_metrics(self, batch: metrics.Client) -> None:
...
+        MAX_RETRIES = 10
+        MAX_RETRY_WAIT_MS = 50
+
+        cmd_list = [args[0] for args, _ in commands]
+        is_read_only = all(cmd in RETRIABLE_COMMANDS for cmd in cmd_list)
+
+        while retries < MAX_RETRIES:
+            try:
+                return super()._execute_pipeline(connection, commands, raise_on_error)
+            except Exception:
```
bradengroom, line 213: I think I'd be hesitant to catch all exceptions here. You probably only want to retry certain failures, right? For example, if I have a bug in my code and I try to `SUNION` on a non-set filed, I'll get an error like the following:
```
127.0.0.1:6379> SET foo 5
OK
127.0.0.1:6379> SUNION foo
(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

That's not going to work no matter how many times I retry.

bradengroom (changes requested):
I'm concerned that this is pretty different from how we do retries elsewhere in baseplate.py and doesn't leave the dev with much control over how retries are performed. What if I want a certain retry policy in one location, but a totally different retry policy in a different location. The `MAX_RETRIES` and `MAX_RETRY_WAIT_MS` values are not configurable.

I think I would prefer a solution that doesn't really take over like this, but instead makes it easier for devs to add retries when and how they want. Today, baseplate supports this with a `retrying` context manager on thrift clients. So you can do something like this:
````
with context.my_service.retrying(attempts=3, budget=5, backoff=.5) as svc:
    svc.some_method()
````

These inputs to `retrying` are used to create a [`RetryPolicy`](https://baseplate.readthedocs.io/en/stable/api/baseplate/lib/retry.html).

It looks like we currently don't support the same `retrying` context manager on redis clients. What if we did that instead?

File: baseplate/clients/redis.py (hunk 1 above)
Comment (FranGM) [reply baseplate/clients/redis.py:213]: that command won't get retried, because it's not considered a retriable command. I'm happy to switch that to just catching the same exceptions we filter for the non-pipeline case, but it feels silly because the client will only raise the first exception within the pipeline, so filtering doesn't do much for you.

**Comment by FranGM** [comment]
Sorry, didn't get to add more context after creating this PR. One of the main motivations behind this PR is to not require any changes to existing code or just minimal changes, and to still be as close as possible to a vanilla Redis client.

Adding a new context manager to help configure retry behaviour might make this difficult, not to mention that `pipeline` is already a context manager, which can make things confusing. (Also note that for every pipeline execution we can individually enable or disable retries, so we could also tweak number of retries, etc)

File: baseplate/clients/redis.py (hunk 1 above)
Comment (spladug) [reply baseplate/clients/redis.py:213]: The problem with catching `Exception` is that if there's a legit bug (e.g. out-of-bounds indexing an array) we catch that too. For library code, that's almost never the right answer. So we should definitely find a more tightly scoped catch here. It doesn't have to check subtypes like in the other path but not catching the base Exception is important.

File: baseplate/clients/redis.py [hunk 2]
```diff
@@ -108,12 +170,134 @@ def report_runtime_metrics(self, batch: metrics.Client) -> None:
...
+                    return True
+
+        return False
+
+    def execute_command(self, *args: Any, **options: Any) -> Any:
+        """Execute a command and return a parsed response."""
+        retries = 0
+        MAX_RETRIES = 25
+        MAX_RETRY_WAIT_MS = 25
+        while retries < MAX_RETRIES:
```
spladug, line 284: fwiw there's a helper in baseplate for this sorta logic https://baseplate.readthedocs.io/en/stable/api/baseplate/lib/retry.html

baseplate/clients/redis.py
```diff
@@ -253,3 +461,33 @@ def close(self) -> None:
         and dequeue as the actions will recreate the queue)
         """
         self.client.delete(self.queue)
+
+
+class RetryStats:
```
spladug, line 466: since all the methods are static/class level, is there any value in this being a class?

spladug (changes requested):
Can you explain a bit more why we want transparent retries without requiring code changes to get them? I thought we were generally shying away from that kind of magic so it feels weird to add it here.

MelissaCole:
:running_woman: looks like others have this handled

FranGM:
I'm going to close this for now as I think we're going to want to revisit the general approach not just in terms of retries, but also with the concept of having a shared connection pool across the whole context

PR #615, opened 2021-08-26
sqlalchemy: limit statement tag length

curioussavage:
related to https://github.com/reddit/baseplate.py/pull/613

I logged some spans and saw that the ORM was generating a statement that grows with the length of a list of ids. I need to change that query but it still seems like it could be a good idea to limit what is allowed.

We could also log a warning in when binary annotations are created on the span with really long strings too.

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -191,6 +191,8 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
 
         trace_name = "{}.{}".format(context_name, "execute")
         span = server_span.make_child(trace_name)
+        if len(statement) > 1024:
+            statement = statement[:1024] + "..."
         span.set_tag("statement", statement)
         span.start()
 
--- a/requirements-transitive.txt
+++ b/requirements-transitive.txt
@@ -79,7 +79,6 @@ SQLAlchemy==1.4.12
 thrift-unofficial==0.14.1
 toml==0.10.2
 translationstring==1.4
-typed-ast==1.4.2
 typing-extensions==3.7.4.3
 urllib3==1.26.6
 venusian==3.0.0


baseplate/clients/sqlalchemy.py
```diff
@@ -191,6 +191,8 @@ def on_before_execute(
 
         trace_name = "{}.{}".format(context_name, "execute")
         span = server_span.make_child(trace_name)
+        if len(statement) > 100:
+            statement = statement[:100]
         span.set_tag("statement", statement)
```
spladug:
```suggestion
        span.set_tag("statement", statement[:100])
```

It's safe to just truncate without checking. `"foo"[:100] == "foo"`.

spladug (changes requested):
Cool! That makes sense. I think 100 might be a bit intense though. I have regular queries in my app that are 400~ characters of just SQL -- not even the parameters. How about 1024 as a limit?

baseplate/clients/sqlalchemy.py
kylelemons: 🔕 on the other hand, that's magical and someone in the future might not know about that behavior and be confused.

You might also want to say `statement = statement[:100] + "..."` when you're truncating so it's clear that some of the query is missing.   In the case where it happens to truncate it, say, right before the WHERE clause, having this could really aid in keeping confusion down.

kylelemons (approved):

baseplate/clients/sqlalchemy.py
spladug: yeah fair point about the ellipsis to make it clear it was truncated (or even `[AND %d MORE BYTES]` or something)

curioussavage:
I'm sorry but I just realized This PR got left hanging. Fixed up with the suggestions.

curioussavage:
tried upgrading typed-ast after CI failed. That didn't work so I added a commit removing the line pinning the version so PIP could resolve it.

spladug (approved):
lgtm, thanks! it looks like a bunch of extra commits made it onto the branch though. can you make sure to clean those out before merging please?

PR #618, opened 2021-08-31
Thrift connection pools can have fractional timeouts

markis:
As the title states, thrift connection pools can have fractional timeouts, which can be represented as floats.  They don't have to just be ints.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -24,6 +24,7 @@ from typing import Any
 from typing import Generator
 from typing import Optional
 from typing import TYPE_CHECKING
+from typing import Union
 
 from thrift.protocol import THeaderProtocol
 from thrift.protocol.TProtocol import TProtocolBase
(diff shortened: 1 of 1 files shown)


kylelemons (approved):
Seems fine to me, but cc'ing neil in case he can think of other downstream effects or other places this might need to reflect this

baseplate/lib/thrift_pool.py
```diff
@@ -152,7 +153,7 @@ def __init__(
         endpoint: config.EndpointConfiguration,
         size: int = 10,
         max_age: int = 120,
-        timeout: int = 1,
+        timeout: Union[int, float] = 1,
```
bradengroom: nit: I think `float` is sufficient here and you don't actually need to union these.

```python
def foo(x: float) -> None:
    print(x)

foo(1)
foo(1.0)
```

```console
$ mypy foo.py
Success: no issues found in 1 source file
```

bradengroom (approved):

spladug (approved):

MelissaCole (approved):

PR #619, opened 2021-08-31
event publisher sidecar: allow http endpoint

fishy:
We are working on a project to have a "proxy" between sidecar and event
collector to reduce the volume of events send to event collector. Since
this "proxy" is entirely internal facing, making it available under http
instead of https is much easier (we don't have to setup nginx-ingress
and a domain name for it).

Thus, adding a scheme config, and default to https, to allow http
endpoints.


kylelemons (approved):

spladug (approved):

spladug:
thanks for the thorough explanation of why!

PR #622, opened 2021-09-26
Extend support for environment variable expansion

alisaifee:
# Description
Use the [expandvars](https://github.com/sayanarijit/expandvars/) library to support different bash parameter expansion methods.

This allows use cases such as providing defaults, offsets and indirection.

To retain compatibility any error in expansion (whether due to missing variable, or invalid syntax) will not raise but simply return the config value as declared in the config file - while logging a warning.

`pytest-check` was added to development dependencies to use multiple assertions per test case

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -39,6 +39,8 @@ from typing import Sequence
 from typing import TextIO
 from typing import Tuple
 
+from expandvars import expandvars
+from expandvars import ExpandvarsException
 from gevent.server import StreamServer
 
 from baseplate import Baseplate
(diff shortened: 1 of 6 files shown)


kylelemons (changes requested):
This needs to be discussed as a a baseplate.spec change first.  The environment substitution is intended to handle some very specific cases where the value must come from the Kubernetes Downward-Facing API, it is not intended to be used for situational overrides, etc.  Among other things, this is likely to severely complicate the cross-language compatibility story.

spladug (changes requested):
cool! thanks for submitting this. let's chat it through in the spec and make sure it fits the goal of our config systems and it's something we want to support across all languages.

MelissaCole:
:running_woman: to clear my salon

alisaifee:
I'lll close this for now since this doesn't appear to be inline with the spec.

PR #623, opened 2021-10-01
Fix healthcheck failure message

bradengroom:
The current healthcheck message does not interpolate the probe variable.
````
$ baseplate-healthcheck --probe readiness thrift localhost:9091
<snip>
AssertionError: service indicated unhealthiness in probe {probe}
````


spladug (approved):
thanks!

PR #625, opened 2021-10-16
Extend SecretsStore to support Vault CSI

praxist:
- SecretsStore reworked to support multiple filewatchers given a directory instead of a secret file
- Added parsing classes to separate some of the Vault secret format from the SecretsStore, since format is different between the Vault secrets fetcher and Vault CSI
- More information on vault-csi: https://www.vaultproject.io/docs/platform/k8s/csi
- Add deprecation notice for Vault-specific functions in SecretsStore

Reddit-only design link: https://docs.google.com/document/d/1nQ8Fhh-WQ4egtIMG_FZzW-dvqAEObsp-onKrQzEn9CA/edit#
This change should be completely transparent for services sticking with secrets fetchers. For those wanting to use vault-csi, they will have to do the following:
- .ini change to `secrets.path = some_directory`
- .ini change to `secrets.provider = vault_csi`
- change references of secret paths in code to just the secret name itself.

Note for testing: I moved the existing test suite to its own file and duplicated the tests in the context of SecretsStore using a dir with multiple secrets files vs a single file - I think this provides better coverage though I'm open to collapsing it back down to one file.

### DIFF
--- a/baseplate/lib/secrets.py
+++ b/baseplate/lib/secrets.py
@@ -3,7 +3,6 @@ import base64
 import binascii
 import json
 import logging
-import os
 
 from pathlib import Path
 from typing import Any
@@ -400,13 +399,12 @@ class DirectorySecretsStore(SecretsStore):
     ):  # pylint: disable=super-init-not-called
         self.parser = parser
         self._filewatchers = {}
-        for root, _, files in os.walk(path):
-            for f in files:
-                p = Path(root, f)
-                file_path = str(p.relative_to(path))
-                self._filewatchers[file_path] = FileWatcher(
-                    file_path, json.load, timeout=timeout, backoff=backoff
-                )
+        root = Path(path)
+        for p in root.glob("**"):
+            file_path = str(p.relative_to(path))
+            self._filewatchers[file_path] = FileWatcher(
+                file_path, json.load, timeout=timeout, backoff=backoff
+            )
 
     def get_vault_url(self) -> str:
         raise NotImplementedError
--- a/baseplate/lib/secrets.py
+++ b/baseplate/lib/secrets.py
@@ -5,16 +5,18 @@ import json
 import logging
 import os
 
+from pathlib import Path
 from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import Iterator
 from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
-from typing import Type
 
 from baseplate import Span
 from baseplate.clients import ContextFactory
+from baseplate.lib import cached_property
 from baseplate.lib import config
 from baseplate.lib import warn_deprecated
 from baseplate.lib.file_watcher import FileWatcher
@@ -118,25 +120,19 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
     raise CorruptSecretError(path, f"unknown encoding: {encoding!r}")
 
 
-class SecretParser:
-    @staticmethod
-    def get_secret(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
-        raise NotImplementedError
+SecretParser = Callable[[Dict[str, Any], str], Dict[str, str]]
 
 
-class VaultFileParser(SecretParser):
-    @staticmethod
-    def get_secret(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
-        try:
-            return data["secrets"][secret_path]
-        except KeyError:
-            raise SecretNotFoundError(secret_path)
+def parse_secrets_fetcher(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
+    try:
+        return data["secrets"][secret_path]
+    except KeyError:
+        raise SecretNotFoundError(secret_path)
 
 
-class VaultCSIParser(SecretParser):
-    @staticmethod
-    def get_secret(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
-        return data["data"]
+# pylint: disable=method-hidden,unused-argument
+def parse_vault_csi(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
+    return data["data"]
 
 
 class SecretsStore(ContextFactory):
@@ -156,32 +152,9 @@ class SecretsStore(ContextFactory):
         path: str,
         timeout: Optional[int] = None,
         backoff: Optional[float] = None,
-        parser: Optional[Type[SecretParser]] = None,
+        parser: Optional[SecretParser] = None,
     ):
-        self._filewatchers = {}
-        self.path = path
-        self.path_type = "dir" if os.path.isdir(path) else "file"
-        self.parser = parser or VaultFileParser
-        if self.path_type == "dir":
-            for root, _, files in os.walk(path):
-                for f in files:
-                    file_path = os.path.relpath(os.path.join(root, f), path)
-
-                    self._filewatchers[file_path] = FileWatcher(
-                        file_path, json.load, timeout=timeout, backoff=backoff
-                    )
-        else:
-            self._filewatchers[path] = FileWatcher(
-                path, json.load, timeout=timeout, backoff=backoff
-            )
-
-    def _get_data(self, filename: str) -> Tuple[Any, float]:
-        try:
-            return self._filewatchers[filename].get_data_and_mtime()
-        except KeyError:
-            raise SecretNotFoundError(filename)
-        except WatchedFileNotAvailableError as exc:
-            raise SecretsNotAvailableError(exc)
+        raise NotImplementedError
 
     def get_raw(self, path: str) -> Dict[str, str]:
         """Return a dictionary of key/value pairs for the given secret path.
@@ -249,29 +222,6 @@ class SecretsStore(ContextFactory):
         """
         return self.get_versioned_and_mtime(path)[0]
 
-    def get_vault_url(self) -> str:
-        """Return the URL for accessing Vault directly."""
-        if self.path_type == "dir":
-            raise NotImplementedError
-
-        warn_deprecated("get_vault_url is deprecated and will be removed in the future.")
-        data, _ = self._get_data(self.path)
-        return data["vault"]["url"]
-
-    def get_vault_token(self) -> str:
-        """Return a Vault authentication token.
-
-        The token will have policies attached based on the current EC2 server's
-        Vault role. This is only necessary if talking directly to Vault.
-
-        """
-        if self.path_type == "dir":
-            raise NotImplementedError
-
-        warn_deprecated("get_vault_token is deprecated and will be removed in the future.")
-        data, _ = self._get_data(self.path)
-        return data["vault"]["token"]
-
     def get_raw_and_mtime(self, secret_path: str) -> Tuple[Dict[str, str], float]:
         """Return raw secret and modification time.
 
@@ -283,12 +233,7 @@ class SecretsStore(ContextFactory):
         .. versionadded:: 1.5
 
         """
-        if self.path_type == "file":
-            data, mtime = self._get_data(self.path)
-        else:
-            data, mtime = self._get_data(secret_path)
-
-        return self.parser.get_secret(data, secret_path), mtime
+        raise NotImplementedError
 
     def get_credentials_and_mtime(self, path: str) -> Tuple[CredentialSecret, float]:
         """Return credentials secret and modification time.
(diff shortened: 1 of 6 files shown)


baseplate/lib/secrets.py
```diff
@@ -116,24 +118,68 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+        timeout: Optional[int] = None,
+        backoff: Optional[float] = None,
+        parser: Optional[Type[SecretParser]] = None,
+    ):
+        self._filewatchers = {}
+        self.path = path
+        self.path_type = "dir" if os.path.isdir(path) else "file"
+        self.parser = parser or VaultFileParser
+        if self.path_type == "dir":
+            for root, _, files in os.walk(path):
```
spladug: Are there ever going to be subdirectories? or just a flat directory of files? In the latter case it'd be simpler to use `os.listdir` or `Path(...).iterdir()`.

Do we need to worry about new files being created later on? (i.e. new secrets) If so we'd want to do this directory listing operation more than once.

baseplate/lib/secrets.py
```diff
@@ -116,24 +118,68 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+    def __init__(
+        self,
+        path: str,
+        timeout: Optional[int] = None,
+        backoff: Optional[float] = None,
+        parser: Optional[Type[SecretParser]] = None,
+    ):
+        self._filewatchers = {}
+        self.path = path
+        self.path_type = "dir" if os.path.isdir(path) else "file"
```
spladug: This is a pretty big smell: we shouldn't have to branch about this in all the places we do below. But instead the `secrets_store_from_config` should return one of two concrete implementations of SecretsStore depending on which kind of secrets storage we're dealing with.

baseplate/lib/secrets.py
```diff
@@ -116,24 +118,68 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+        self.parser = parser or VaultFileParser
+        if self.path_type == "dir":
+            for root, _, files in os.walk(path):
+                for f in files:
+                    file_path = os.path.relpath(os.path.join(root, f), path)
```
spladug: [pathlib](https://docs.python.org/3/library/pathlib.html) makes this kind of path manipulation much easier on the eye.

baseplate/lib/secrets.py
```diff
@@ -205,7 +251,11 @@ def get_versioned(self, path: str) -> VersionedSecret:
 
     def get_vault_url(self) -> str:
         """Return the URL for accessing Vault directly."""
-        data, _ = self._get_data()
+        if self.path_type == "dir":
+            raise NotImplementedError
+
+        warn_deprecated("get_vault_url is deprecated and will be removed in the future.")
```
spladug:
```suggestion
        warn_deprecated("get_vault_url is deprecated and will be removed in v3.0.")
```

baseplate/lib/secrets.py
```diff
@@ -339,21 +393,22 @@ def make_object_for_context(self, name: str, span: Span) -> "SecretsStore":
...
 
-    def __init__(self, filewatcher: FileWatcher):  # pylint: disable=super-init-not-called
-        self._filewatcher = filewatcher
+    def __init__(
+        self, filewatchers: Dict[str, FileWatcher]
+    ):  # pylint: disable=super-init-not-called
+        self._filewatchers = filewatchers
+        self._cached_data: Dict[str, Tuple[Dict, float]] = {}
 
-    @cached_property
```
spladug: note: this is going to be a performance regression for the single-file use case. `@cached_property` replaces itself with the cached data after the first lookup so we do fewer dict lookups when re-using the property.

baseplate/lib/secrets.py
```diff
@@ -339,21 +393,22 @@ def make_object_for_context(self, name: str, span: Span) -> "SecretsStore":
...
-    def _data(self) -> Tuple[Any, float]:
-        return super()._get_data()
-
-    def _get_data(self) -> Tuple[Dict, float]:
-        return self._data
+    def _get_data(self, filename: str) -> Tuple[Dict, float]:
+        if not self._cached_data[filename]:
+            self._cached_data[filename] = super()._get_data(filename)
+        return self._cached_data[filename]
```
spladug: won't this raise a KeyError if the data's not cached? 

something like this would be more efficient in lookups since we don't hit the dictionary multiple times:

```suggestion
        try:
            result = self._cached_data[filename]
        except KeyError:
            result = super()._get_data(filename)
            self._cached_data[filename] = result
        return result
```

(though note this is still slower than `@cached_property` for the single file case)

baseplate/lib/secrets.py
```diff
@@ -116,24 +118,68 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+        try:
+            return data["secrets"][secret_path]
+        except KeyError:
+            raise SecretNotFoundError(secret_path)
+
+
+class VaultCSIParser(SecretParser):
+    @staticmethod
+    def get_secret(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
+        return data["data"]
```
spladug: These don't need to be classes. It can just be a function:

```suggestion
SecretParser = Callable[[Dict[str, Any], str], Dict[str, str]]


def parse_secrets_fetcher(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
    try:
        return data["secrets"][secret_path]
    except KeyError:
        raise SecretNotFoundError(secret_path)


def parse_vault_csi(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
    return data["data"]
```

baseplate/lib/secrets.py
spladug: we'll also need to keep an eye on the multi-file applications' performance. now they'll need to do multiple stat calls per request to make sure they're fresh when before it was just one.

we have a couple options should it prove too slow: cache for longer, or switch to inotify.

spladug (changes requested):
Cool! Thanks for doing this. I think the biggest thing is it'd really be best not to be branching in every function like that but instead use separate classes for each implementation. We'll have to keep an eye on performance and consider our options if checking all those mtimes is too slow. Excited to see the CSI provider replace the secrets fetcher soon!

kylelemons (changes requested):
It looks like spladug has this review, but from my perspective I think it would be useful for the PR / commit to summarize the high-level considerations that need to be made in the transition.  Notably, this will need to be mimicked in the Go and TypeScript baseplate libraries, so having a concise description of the changes are that motivate the change would be useful so that the implementation doesn't have to be the canonical documentation of how it works.

baseplate/lib/secrets.py
praxist: yep, there can be subdirectories if the objectName has `/`s in them. e.g. trying to use `secret/blah` instead of `blah` would create an additional directory `secret`.

praxist: second question - new files wouldn't be created unless a change is made to the deploy config to add additional objects.

baseplate/lib/secrets.py
praxist: decided to use `.get()` rather than the try/except, I believe that's slightly more performant.

praxist:
Split the SecretsStore logic into 2 subclasses representing the file/directory implementations.

baseplate/lib/secrets.py
```diff
@@ -116,26 +120,41 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+
+
+def parse_secrets_fetcher(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
+    try:
+        return data["secrets"][secret_path]
+    except KeyError:
+        raise SecretNotFoundError(secret_path)
+
+
+# pylint: disable=method-hidden,unused-argument
```
spladug: is `method-hidden` still necessary?

baseplate/lib/secrets.py
```diff
@@ -116,26 +120,41 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
...
+def parse_vault_csi(data: Dict[str, Any], secret_path: str = "") -> Dict[str, str]:
+    return data["data"]
+
+
 class SecretsStore(ContextFactory):
```
spladug, line 137: I agree that making the base class into an interface would be ideal, but unfortunately we've got a lot of users that are directly instantiating this (mostly in test suites). So this is a breaking change. For now, we can make the FileSecretsStore the base class and make the swap out in v3.0. Does that sound OK?

baseplate/lib/secrets.py
```diff
@@ -339,10 +338,133 @@ def make_object_for_context(self, name: str, span: Span) -> "SecretsStore":
...
+
+    def __init__(
+        self,
+        path: str,
+        timeout: Optional[int] = None,
+        backoff: Optional[float] = None,
+        parser: Optional[SecretParser] = None,
+    ):  # pylint: disable=super-init-not-called
+        self._filewatchers = {}
+        self.parser = parser or parse_secrets_fetcher
```
spladug: This could probably just be a mandatory parameter. Or if not, maybe it'd make more sense to default to `parse_vault_csi` since that's more likely to be used with this subclass than the other parser.

baseplate/lib/secrets.py
```diff
@@ -356,6 +478,22 @@ def _get_data(self) -> Tuple[Dict, float]:
...
+    def __init__(
+        self, filewatchers: Dict[str, FileWatcher]
+    ):  # pylint: disable=super-init-not-called
+        self._filewatchers = filewatchers
+        self._cached_data: Dict[str, Tuple[Dict, float]] = {}
+
+    def _get_data(self, filename: str) -> Tuple[Dict, float]:
+        if not self._cached_data.get(filename):
+            self._cached_data[filename] = super()._get_data(filename)
+        return self._cached_data[filename]
```
spladug: This is still looking up the value in the hash twice for each entry (the get() and the []). Is there a reason to prefer this over the try/except method I commented previously?

baseplate/lib/secrets.py
```diff
@@ -398,4 +539,12 @@ def secrets_store_from_config(
         backoff = None
 
     # pylint: disable=maybe-no-member
-    return SecretsStore(options.path, timeout=timeout, backoff=backoff)
+    if options.provider == "vault":
+        parser = parse_secrets_fetcher
+    elif options.provider == "vault_csi":
+        parser = parse_vault_csi
+
+    if os.path.isdir(options.path):
```
spladug: Iiiinteresting. I'm curious why we'd want to do this separately from the branching on provider? It seems like it might lead to easier accidental misconfiguration.

baseplate/lib/secrets.py
```diff
@@ -376,8 +514,10 @@ def secrets_store_from_config(
         to "secrets."
     :param backoff: retry backoff time for secrets file watcher. Defaults to
         None, which is mapped to DEFAULT_FILEWATCHER_BACKOFF.
+    :param provider: The secrets provider. Defaults to 'vault'.
```
spladug: maybe good to document the alternative too?

spladug (changes requested):
awesome! final stretch. this is looking really good. just a couple final things.

baseplate/lib/secrets.py
praxist: set to mandatory param

baseplate/lib/secrets.py
praxist: you're right - switched to try/except.

baseplate/lib/secrets.py
praxist: was originally trying to decouple the path/file secrets store from any vault references, but that might have been unnecessary.

baseplate/lib/secrets.py
praxist: ok, collapsed FileSecretsStore back to SecretsStore (hope I'm assuming correctly that's what you meant?)

baseplate/lib/secrets.py
```diff
@@ -356,6 +380,88 @@ def _get_data(self) -> Tuple[Dict, float]:
...
+    ):  # pylint: disable=super-init-not-called
+        self.parser = parser
+        self._filewatchers = {}
+        for root, _, files in os.walk(path):
+            for f in files:
+                p = Path(root, f)
+                file_path = str(p.relative_to(path))
+                self._filewatchers[file_path] = FileWatcher(
+                    file_path, json.load, timeout=timeout, backoff=backoff
+                )
```
spladug: :no_bell: this can be fully written with pathlib to simplify a bit

```suggestion
        root = Path(path)
        for p in root.glob("**"):
            file_path = str(p.relative_to(root))
            self._filewatchers[file_path] = FileWatcher(
                file_path, json.load, timeout=timeout, backoff=backoff
            )
```

spladug (approved):
looks great. thank you!

baseplate/lib/secrets.py
praxist: oh yeah, that looks much nicer, awesome

praxist:
@spladug / @kylelemons ready to merge - will squash with the following commit message:
```
Extend SecretsStore to support Vault CSI

This change adds an additional way to load secrets besides the existing secrets fetcher:
- When vault.provider is set to vault_csi, vault.path will be assumed to be a directory containing individual files representing secrets, rather than a single file containing every secret.
- Vault CSI loads secrets automatically into these files on pod startup, so no additional sidecar is required.
- Secrets are written to files in a different JSON format
- Each secrets file has its own filewatcher representation with per-request caching behavior preserved

Implementation details:
- Add DirectorySecretsStore subclass of SecretsStore supporting multiple filewatchers
- Duplicate SecretsStore unit tests in the context of DirectorySecretsStore
- Modify secrets_store_from_config to return SecretsStore or DirectorySecretsStore based off vault.provider
- Split out vault-specific file parsing logic into functions outside of SecretsStore
- Add deprecation notice for get_vault_url, get_vault_token in SecretsStore
```

kylelemons (approved):

PR #627, opened 2021-11-01
Explicitly support Python 3.10

spladug:
https://www.python.org/downloads/release/python-3100/


spladug:
This is currently blocked on the Thrift C extension not supporting Py3.10 yet.

spladug:
Closing for now in favor of https://github.com/reddit/baseplate.py/pull/638 but we'll want to land the last commit from here once we're ready to support 3.10.

PR #628, opened 2021-11-01
Update greenlet version to support py3.10

alisaifee:
# Description
To support python 3.10 greenlet needs to be updated to atleast 1.1.0.
[1.1.2](https://greenlet.readthedocs.io/en/latest/changes.html#id1) seems somehow karmic for baseplate 2.2.3


spladug:
i've got a bunch of necessary changes piling up over here https://github.com/reddit/baseplate.py/pull/627

PR #629, opened 2021-11-29
redis: Return checked-out connections to connection pool at end of span

spladug:
I'm opening this as a draft but will be testing it out in a real service before making it a real PR.

This copies the [pattern we use in the sqlalchemy client](https://github.com/reddit/baseplate.py/blob/b734323259c2ff6fdccbe5af17048ed22c0284c0/baseplate/clients/sqlalchemy.py#L258-L265) to make sure that we return checked-out connections back to the pool at end of request. Hopefully this reduces connection pool growth.

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -12,6 +12,8 @@ except ImportError:
     from redis.client import Pipeline
 
 from baseplate import Span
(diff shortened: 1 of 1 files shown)


spladug:
In testing this on Harold, this did not appear to do anything.

PR #630, opened 2021-12-03
Fix annoying extra IPython exception in shell tracebacks

spladug:
This is happening because Python's adding the original "IPython isn't
installed" exception's context to any later exceptions because we're
running the fallback shell in the except: branch. By doing it outside,
we avoid this.


kylelemons (approved):
oooo yes

pacejackson (approved):

bradengroom (approved):

PR #632, opened 2021-12-08
Add prometheus exporter to baseplate-serve

spladug:
We're not adding observers or anything else yet, but this will allow
services to start adding Prometheus metrics incrementally.


bradengroom (approved):

MelissaCole (approved):
lgtm!

baseplate/server/prometheus.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger(__name__)
+PROMETHEUS_EXPORTER_ADDRESS = Endpoint("0.0.0.0:6060")
+
+
+def export_metrics(environ: "WSGIEnvironment", start_response: "StartResponse") -> Iterable[bytes]:
+    registry = CollectorRegistry()
+    multiprocess.MultiProcessCollector(registry)
+    data = generate_latest(registry)
+
+    if environ["PATH_INFO"] != "/metrics":
```
JessicaGreben, line 42: should we put this  path checking`if` block before the code on ln 42-44 to skip doing that work unnecessarily?

JessicaGreben (approved):

docs/cli/serve.rst
```diff
@@ -186,6 +186,37 @@ Note that Einhorn will exit if you send it a ``SIGUSR1``. You can instead open u
...
+
+To enable the exporter, install the ``prometheus-client`` package from PyPI.
+
+.. code-block:: console
+
+   $ pip install prometheus-client
+
+The ``PROMETHEUS_MULTIPROC_DIR`` environment variable must be set to the path
+to an extant writeable directory. This is handled automatically in the official
+base Docker images.
```
JessicaGreben, line 207: Is there a PR to add this env var  `PROMETHEUS_MULTIPROC_DIR` to the official base Docker image?  I don't actually know where the official base docker image lives.

JessicaGreben:
The limitations of the multiprocess mode are a bummer. Sounds like the automatic metrics for process, platform, and gc aren't included with multiprocess :( I guess we will need to implement those ourselves.
Also I'm not quite sure in what circumstances this happens "Registering metrics to a registry later used by a MultiProcessCollector may cause duplicate metrics to be exported", will look into it more.
Ref: https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn.

docs/cli/serve.rst
spladug: yup! it's at reddit/docker-baseplate-py#57 on our github enterprise

baseplate/server/prometheus.py
spladug: good callout. fixed

spladug:
> Also I'm not quite sure in what circumstances this happens "Registering metrics to a registry later used by a MultiProcessCollector may cause duplicate metrics to be exported", will look into it more.

Yeah, that's a super confusing sentence. My read is that's why we make a custom registry in the `export_metrics` app -- if we'd given it the global registry that's presumably being used for other metrics then it would have picked those metrics up in addition to the ones it's pulling off disk (which would _also_ include the metrics we got from the reused registry)

> Sounds like the automatic metrics for process, platform, and gc aren't included with multiprocess :( I guess we will need to implement those ourselves.

for sure. luckily we already had all that in baseplate.py https://github.com/reddit/baseplate.py/blob/develop/baseplate/server/runtime_monitor.py so we can just make sure that works with the prometheus stuff and we're good to go

PR #635, opened 2021-12-14
sqlalchemy: Don't truncate the executed statement, just the tag

spladug:
This broke in d16351f575bb7cc2c3df0986107e8768fefde3e7. Long queries
were having the actually executed SQL be truncated as well which
obviously leads to syntax errors with the appended ellipsis.


curioussavage (approved):
Looks fine to me. Thanks for fixing that.

PR #636, opened 2021-12-16
ThriftConnectionPool: catch possible timeouts in _release method

curioussavage:
I believe the `prot.trans.isOpen` call ends up yielding to the hub which means a timeout exception can be raised here resulting in a connection slot getting counted as "used" forever.

Assuming I'm correct I'm not necessarily saying this is the best solution. Maybe it makes sense to remove the `isOpen` call here and rely on the code to detect stale connections after claiming one.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -203,10 +203,14 @@ class ThriftConnectionPool:
         return False
 
     def _release(self, prot: Optional[TProtocolBase]) -> None:
-        if prot and prot.trans.isOpen():
-            self.pool.put(prot)
-        else:
+        try:
+            if prot and prot.trans.isOpen():
+                self.pool.put(prot)
+            else:
+                self.pool.put(None)
+        except:  # noqa: E722
             self.pool.put(None)
+            raise
 
     @contextlib.contextmanager
     def connection(self) -> Generator[TProtocolBase, None, None]:
--- a/tests/unit/lib/thrift_pool_tests.py
+++ b/tests/unit/lib/thrift_pool_tests.py
@@ -277,6 +277,24 @@ class ThriftConnectionPoolTests(unittest.TestCase):
 
     @mock.patch("baseplate.lib.thrift_pool._make_transport")
     @mock.patch("time.time")
+    def test_context_server_timeout_in_release(self, mock_time, mock_make_transport):
+        mock_time.return_value = 123
+
+        self.mock_queue.get.return_value = None
+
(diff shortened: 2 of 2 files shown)


curioussavage:
Oops. I was intending to open this as Draft for now. We can close for now if discussion is preferred somewhere else.

baseplate/lib/thrift_pool.py
```diff
@@ -203,10 +203,14 @@ def _is_stale(self, prot: TProtocolBase) -> bool:
...
-            self.pool.put(prot)
-        else:
+        try:
+            if prot and prot.trans.isOpen():
+                self.pool.put(prot)
+            else:
+                self.pool.put(None)
+        except:  # noqa: E722
             self.pool.put(None)
+            raise
```
spladug, line 213: How about something like this?

```suggestion
        to_return = None
        try:
            if prot and prot.trans.isOpen():
                to_return = prot
        finally:
            self.pool.put(to_return)
```

I'm a little worried we'd accidentally double-release in some situations if we just put an extra `put()` in the error path.

spladug (changes requested):
yikes, good catch! thanks for writing this up and getting a test for it too. i put a potential alternative fix here, what do you think of that?

baseplate/lib/thrift_pool.py
curioussavage: Yep that looks better.

curioussavage:
I did some more testing and poking around to confirm and I think I was actually wrong. I see in the [ThriftSocket code](https://github.com/apache/thrift/blob/master/lib/py/src/transport/TSocket.py#L90) that the socket is put in non blocking mode first. 

I also used the following code to log whenever execution switched to the event loop from the current greenlet.

```
class GeventWaitLogger():
    def __enter__(self):
        # patch gevent
        current = gevent.getcurrent()
        def switch_out():
            for line in traceback.format_stack():
                print(line.strip())
        current.switch_out = switch_out

    def __exit__(self, type, value, traceback):
        current = gevent.getcurrent()
        current.switch_out = None
```
 Here is a [link to the results](https://gist.github.snooguts.net/adam/3fc72249b52cabcf4c82d984d9909b7a), the highlighted line is the only place `_release` shows up in the stack traces. It only sleeps in order to claim a lock after it has already put the socket back in the underlying list. So you are totally right about the potential to call put twice in my code. Maybe we can safely close this PR then?

spladug:
Ah ok, yeah I think you're right.

pacejackson:
🏃‍♂️

PR #639, opened 2022-01-12
Fix Thrift pool connection leak when a ServerTimeout is raised while releasing a connection

RoganMurley:
_(disclaimer: I haven't 100% confirmed that this is the root cause of our connection leak, but think that it is likely. Would appreciate any insights here!)_

If a `ServerTimeout` exception is raised while the Thrift Pool is releasing a connection, the connection is not properly released back into the pool. This causes a connection leak as the pool size is permanently decreased until restart, eventually resulting in service instability.

Given the difficulty of properly managing these connections, it might make sense to introduce some kind of self-healing mechanism.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -36,6 +36,7 @@ from thrift.transport.TTransport import TTransportException
 
 from baseplate.lib import config
 from baseplate.lib.retry import RetryPolicy
+from baseplate.observers.timeout import ServerTimeout
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


baseplate/lib/thrift_pool.py
```diff
@@ -203,10 +204,16 @@ def _is_stale(self, prot: TProtocolBase) -> bool:
         return False
 
     def _release(self, prot: Optional[TProtocolBase]) -> None:
-        if prot and prot.trans.isOpen():
```
RoganMurley, line 206: I'm assuming that `isOpen` can context switch, still trying to verify that.

RoganMurley: Looks like it actually peeks at message bytes on the socket to check if open, could very well context switch.

https://github.com/apache/thrift/blob/master/lib/py/src/transport/TSocket.py#L76

RoganMurley:
Oh, there's already a PR for this that was closed: #636.  👀

RoganMurley:
Closing this given analysis by @curioussavage in the other PR. I think I found the connection leak elsewhere, so I will be raising another PR.

PR #641, opened 2022-01-14
Log the inner exception that TTransportException wraps

RoganMurley:
I am debugging a Thrift connection issue and am seeing lots of `TTransportException` errors with the message `unexpected exception`. These `TTransportException` errors [wrap an inner exception](https://github.com/apache/thrift/blob/66d897667c451ef6560d89b979b7001c57a3eda6/lib/py/src/transport/TSocket.py#L164) containing the actual socket error, but this is never logged because it isn't shown by `TTransportException.__str__`.

This PR checks for an inner exception when logging a `TTransportException` in the Thrift client and logs it if it is present.


MelissaCole (approved):
lgtm!

PR #643, opened 2022-01-18
Fix Thrift pool connection leak

RoganMurley:
If a `ServerTimeout` exception is raised as a Thrift connection is being released back into the pool then the connection is never released. This results in a connection leak and eventually causes service instability until the service is restarted. I have [stack traces](https://gist.github.snooguts.net/rogan-murley/110ec99c68ea78be3b5b13f67edbab7b) and [connection pool usage graphs](https://gist.github.snooguts.net/rogan-murley/0071fb116b924c74138a5f57fdf1083b) showing this behavior.

Screenshot 2022-01-18 at 17 58 46

> Server timeout error spike (blue) followed by stepwise Thrift pool usage increase consistent with a connection leak (purple)

The above stack trace was surprising because gevent queues should provide atomicity: a context switch should not happen during a non-blocking `.put()`. However from stack traces I can see that the Thrift pool does not use a gevent queue, it actually uses the standard library queue. It appears that queues (other than `SimpleQueue`) are [not ever monkey patched by gevent](https://github.com/gevent/gevent/issues/1248#issuecomment-430469639), so you need to explicitly import gevent queues if you want to use them.

The standard library queue works fine with gevent as it is built ontop of threads (which themselves are patched by gevent), but has worse performance and doesn't provide atomicity guarantees. In this PR I have swapped the Thrift pool to use the gevent queue. This should prevent context switching during the `.put()` and so fix the connection leak.

There are other uses of standard library queues rather than gevent queues in baseplate.py which should also be reviewed.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -16,7 +16,6 @@ A basic example of usage::
 """
 import contextlib
 import logging
-import queue
 import socket
 import time
 
@@ -33,6 +32,8 @@ from thrift.Thrift import TApplicationException
 from thrift.Thrift import TException
 from thrift.transport.TSocket import TSocket
 from thrift.transport.TTransport import TTransportException
+
+import gevent.queue as queue
 
 from baseplate.lib import config
 from baseplate.lib.retry import RetryPolicy
--- a/tests/integration/thrift_tests.py
+++ b/tests/integration/thrift_tests.py
@@ -8,6 +8,8 @@ from unittest import mock
 import gevent.monkey
 import pytest
 
+from gevent.util import _FailedToSwitch
+from gevent.util import assert_switches
 from thrift.Thrift import TApplicationException
 
 from baseplate import Baseplate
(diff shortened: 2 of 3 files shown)


FranGM:
```
There are other uses of standard library queues rather than gevent queues in baseplate.py which should also be reviewed.
```
This is definitely the case for the redis (and redis-cluster) client as well, to the point where we internally ask service owners to patch the queue themselves.

MelissaCole (approved):

curioussavage:
I think we should add a test case using gevent.util's `assert_switches` to prove this now works fine.

example:
```
>>> from baseplate.lib.thrift_pool import ThriftConnectionPool
>>> old = ThriftConnectionPool._release
>>> from gevent.util import assert_switches
>>> def newrel(self, prot):
...     with assert_switches(max_blocking_time=1.0):
...         old(self, prot)
...
>>> ThriftConnectionPool._release = newrel
```

RoganMurley:
Thanks @curioussavage, I've added a test based on your code.

I avoided doing this initially because I had trouble with gevent "not working properly" in tests previously. In retrospect this was just queues not being patched as this PR addresses!

PR #645, opened 2022-01-20
Add queuing option to Thrift pool

SuperQ:
Add option to allow passing different queue methods to ThriftConnectionPool.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -24,7 +24,7 @@ from typing import Generator
 from typing import Optional
 from typing import TYPE_CHECKING
 
-from gevent import queue
+from gevent import queue as gevent_queue
 from thrift.protocol import THeaderProtocol
 from thrift.protocol.TProtocol import TProtocolBase
 from thrift.protocol.TProtocol import TProtocolException
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
 
     ProtocolPool = std_lib_queue.Queue[TProtocolBase]  # pylint: disable=unsubscriptable-object
 else:
-    ProtocolPool = queue.Queue
+    ProtocolPool = gevent_queue.Queue
 
 
 def _make_transport(endpoint: config.EndpointConfiguration) -> TSocket:
@@ -139,6 +139,7 @@ class ThriftConnectionPool:
     :param protocol_factory: The factory to use for creating protocols from
         transports. This is useful for talking to services that don't support
         THeaderProtocol.
+    :param queue_cls: A stdlib compatible queue class.
 
     All exceptions raised by this class derive from
     :py:exc:`~thrift.transport.TTransport.TTransportException`.
@@ -157,6 +158,7 @@ class ThriftConnectionPool:
         timeout: float = 1,
         max_connection_attempts: int = 3,
         protocol_factory: TProtocolFactory = _DEFAULT_PROTOCOL_FACTORY,
+        queue_cls: ProtocolPool = gevent_queue.LifoQueue(),
     ):
         self.endpoint = endpoint
         self.max_age = max_age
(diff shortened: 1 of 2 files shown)


baseplate/lib/thrift_pool.py
```diff
@@ -114,6 +116,8 @@ def thrift_pool_from_config(
         kwargs.setdefault("max_connection_attempts", options.max_connection_attempts)
     if options.max_retries is not None:
         raise Exception("max_retries was renamed to max_connection_attempts")
+    if options.queue is not None:
+        kwargs.setdefault("queue", "lifo")
```
MelissaCole: What about defining the default in the config parser and pulling it from there for the kwargs update?
`kwargs.setdefault("queue", options.queue)`

baseplate/lib/thrift_pool.py
```diff
@@ -165,7 +169,11 @@ def __init__(
         self.protocol_factory = protocol_factory
 
         self.size = size
-        self.pool: ProtocolPool = queue.LifoQueue()
+        if config.Queue == "fifo":
```
MelissaCole: `config.Queue` -> `config.queue`
```
baseplate/lib/thrift_pool.py:172:11: E1101: Module 'baseplate.lib.config' has no 'Queue' member (no-member)
```

MelissaCole (changes requested):
lgtm! Just a couple of tiny fixes

baseplate/lib/thrift_pool.py
SuperQ: Sure, I was following the style of the other options. This makes more sense to me.

MelissaCole (approved):
lgtm!

baseplate/lib/thrift_pool.py
```diff
@@ -165,7 +168,11 @@ def __init__(
         self.protocol_factory = protocol_factory
 
         self.size = size
-        self.pool: ProtocolPool = queue.LifoQueue()
+        if config.queue == "fifo":
```
MelissaCole: `config.queue` doesn't exist so I think it wants to be passed in through init's kwargs

MelissaCole (approved):
lgtm!

baseplate/lib/thrift_pool.py
```diff
@@ -157,6 +162,7 @@ def __init__(
         timeout: float = 1,
         max_connection_attempts: int = 3,
         protocol_factory: TProtocolFactory = _DEFAULT_PROTOCOL_FACTORY,
+        queue: str = "lifo",
```
spladug: I think it'd be a clearer API and more flexible to take a queue class object here rather than a string. 

```suggestion
        queue: Type = gevent_queue.LifoQueue,
```

baseplate/lib/thrift_pool.py
```diff
@@ -100,6 +101,7 @@ def thrift_pool_from_config(
             "timeout": config.Optional(config.Timespan, default=config.Timespan("1 second")),
             "max_connection_attempts": config.Optional(config.Integer),
             "max_retries": config.Optional(config.Integer),
+            "queue": config.Optional(config.String, default="lifo"),
```
spladug: Is this actually something we want to vary via configuration? i.e. it's changing from one environment to the next? I think we'd probably just be OK with the queue being a hard-coded aspect of the application.

spladug (changes requested):

baseplate/lib/thrift_pool.py
SuperQ: I thought about that, but my Python is bad and didn't know if that was a recommended way to do things. :smile:

baseplate/lib/thrift_pool.py
SuperQ: My thought was that since this is per-pool, it might make sense to allow changing only one downstream pool

spladug: That's important, for sure. It will still be configurable per-pool if we leave it out of config parsing. In the wire-up function it'd look like:

```python
baseplate.configure_context({
    "foo_service": ThriftClient(FooService.Client),  # uses the default LifoQueue
    "bar_service": ThriftClient(BarService.Client, queue=FifoQueue),  # overridden for just this pool
})
```

(this will just work right now because `ThriftClient` passes any `**kwargs` straight through to `thrift_pool_from_config` so you won't need to add any more plumbing)

SuperQ: Sure, that seems ok to me.

spladug (approved):

PR #646, opened 2022-01-26
Default to gevent queue for Redis connection pool

FranGM:
This change adapts https://github.com/reddit/baseplate.py/pull/643 to the Redis connection pool.

We've already been recommending that they use this queue for a while, by passing `queue_class=gevent.queue.LifoQueue` when creating their connection pool, so this just changes the default.

Note that the redis-cluster client would in theory also benefit from this but that change would require internally patching `redis-py-cluster`, which is probably not worth doing given that library has already been deprecated in favour of `redis-py`

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -11,6 +11,8 @@ try:
 except ImportError:
     from redis.client import Pipeline
 
+from gevent import queue
+
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
@@ -56,6 +58,9 @@ def pool_from_config(
         kwargs.setdefault("socket_connect_timeout", options.socket_connect_timeout.total_seconds())
     if options.socket_timeout is not None:
         kwargs.setdefault("socket_timeout", options.socket_timeout.total_seconds())
+
+    if "queue_class" not in kwargs:
+        kwargs["queue_class"] = queue.LifoQueue
 
     return redis.BlockingConnectionPool.from_url(options.url, **kwargs)
 


MelissaCole (approved):

spladug:
Wow, good catch. TIL about that queue not getting patched.

I wonder if we want to just patch it ourselves? Would we ever want the stdlib queue?

FranGM:
My only apprehension with patching it ourselves is that I don't think we know why the gevent people just didn't patch everything to begin with. I don't think there's a case where you'd want the stdlib queue but maybe there is? 🤷 

That said, patching it ourselves is compelling because right now we need to be aware of all present and future queue uses and make sure we use the right one, and I don't love the idea

spladug:
https://github.com/gevent/gevent/issues/1248#issuecomment-430613010

baseplate/clients/redis.py
```diff
@@ -57,6 +59,9 @@ def pool_from_config(
     if options.socket_timeout is not None:
         kwargs.setdefault("socket_timeout", options.socket_timeout.total_seconds())
 
+    if "queue_class" not in kwargs:
+        kwargs["queue_class"] = queue.LifoQueue
```
SuperQ, line 63: It may be worth defaulting to `queue.Queue`. I'm adding this as an option for Thrift in https://github.com/reddit/baseplate.py/pull/645.

My analysis of traffic says that LIFO is very bad for load-balancing.

spladug: Since this isn't the redis-cluster client, does load balancing come into play here? Usually there's only a single primary node that we want to talk to anyway so the downsides of maxing out connection pool sizes early on may not be worth it.

SuperQ: Hmmm, good question. Even with non-cluster, there are still read-only replica options. So it may still be relevant to distribute the load for read requests.

RoganMurley:
@FranGM I think this PR will be made redundant if we merge #656

FranGM:
As usual @RoganMurley comes in clutch, going to close this PR since his change will take care of business.

PR #648, opened 2022-02-08
make sentry with_locals configurable

aokolish:
Currently, this is defaulted to false but is not configurable. This
commit enables users to configure with_locals to suit their needs.


MelissaCole (approved):

kylelemons (changes requested):
I don't think `with_locals` can be used safely.

Developers do not think "Oh, this variable contains PII, so I should make sure no exceptions propagate through this function".

I am opposed to the inclusion of this, but SRE is responsible for the response if PII leaks into sentry, so defer to their judgment.

aokolish:
actually, after some consideration...I think we can get by without enabling this setting. I'll close this for now.

PR #649, opened 2022-02-11
Add a contains function

nsheaps:
As part of reddit/devprod#33, I need this function in order to check if something is on the queue before I queue it to help prevent duplicates (I understand that this will scan, and will not prevent duplicates for things that have already been pulled off the queue).

Talking with @spladug about actually upgrading redis.py throughout our repos but didn't want that to be a blocker here since this has caused a few incidents.

Thanks @MelissaCole for the advice!

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -232,6 +232,25 @@ class MessageQueue:
 
         return message
 
+    def contains(
+        self, message: bytes
+    ) -> bool:
+        """Scan the queue for an instance of the message, returns True if found
+        False otherwise.
+
+        NOTE: THIS SCANS. IT IS NOT O(1). ONLY USE THIS WITH QUEUE SIZES YOU
+        EXPECT A SCAN TO BE OKAY ON.
+
+        ref: https://redis.io/commands/lpos
+
+        :param message: will be typecast to a string upon storage and will come
+               out of the queue as a string regardless of what type they are
+               when passed into this method.
+        """
+        # TODO using execute_command because our python lib for redis is out of date.
+        # When that gets upgraded (larger scope, change to use lpos directly)
+        return self.client.execute_command('LPOS', self.queue, message) is not None
+
     def put(  # pylint: disable=unused-argument
         self, message: bytes, timeout: Optional[float] = None
     ) -> None:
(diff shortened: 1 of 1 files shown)


baseplate/clients/redis.py
```diff
@@ -232,6 +232,23 @@ def get(self, timeout: Optional[float] = None) -> bytes:
...
+        """Scan the queue for an instance of the message, returns True if found
+        False otherwise.
+
+        NOTE: THIS SCANS. IT IS NOT O(1). ONLY USE THIS WITH QUEUE SIZES YOU
+        EXPECT A SCAN TO BE OKAY ON.
+
+        ref: https://redis.io/commands/lpos
+
+        :param message: will be typecast to a string upon storage and will come
+               out of the queue as a string regardless of what type they are
```
kylelemons, line 245: > will come out of the queue as a string

I don't quite grok "will come out of the queue" here.   I think what this means to say is that "whatever in the queue is assumed to be a string, and the message will be coerced into a string as well for comparison" or something, but the words here don't quite line up for me

kylelemons (approved):

baseplate/clients/redis.py
nsheaps: FWIW thats the same verbage as the `put` function

baseplate/clients/redis.py
```diff
@@ -232,6 +232,23 @@ def get(self, timeout: Optional[float] = None) -> bytes:
...
+               when passed into this method.
+        """
+        # TODO using execute_command because our python lib for redis is out of date.
+        # When that gets upgraded (larger scope, change to use lpos directly)
```
spladug:
```suggestion
        # When that gets upgraded (larger scope) change to use lpos directly
```

spladug (approved):
the harold stuff can be unblocked by just using LPOS directly on the client if we want. i'm a little skeptical on checking that a json-serialized payload exactly matches though so we will want to be careful there.

nsheaps:
@spladug I had the same thought, double checked the thing that gets queued and the message is simple enough. It will _not_ dedupe multiple updates that have different topics, which I think we want. This breaks if the message includes a timestamp so it's not perfect.

nsheaps:
Not feasible at the moment due to redis version dependency.

PR #652, opened 2022-03-14
Add default prometheus metrics to clients

krav:
This adds Prometheus client metrics equivalent to the existing statsd ones.

## Caveat
I've added an identifying name to each of the client classes that produce metrics, defaulted in each case except for the thrift client where it can be inferred from the class. In the case of instantiating two clients of the same class the last to be instantiated would otherwise be the only one to report metrics. However if clients are dynamically created at runtime this can cause labels to disappear and reappear in the prometheus server on restart.

### DIFF
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -8,6 +8,7 @@ from typing import Sequence
 from typing import Tuple
 from typing import Union
 
+from prometheus_client import Gauge
 from pymemcache.client.base import PooledClient
 
 from baseplate import Span
@@ -112,7 +113,7 @@ class MemcacheClient(config.Parser):
             serializer=self.serializer,
             deserializer=self.deserializer,
         )
-        return MemcacheContextFactory(pool)
+        return MemcacheContextFactory(pool, key_path)
 
 
 class MemcacheContextFactory(ContextFactory):
@@ -129,8 +130,34 @@ class MemcacheContextFactory(ContextFactory):
 
     """
 
-    def __init__(self, pooled_client: PooledClient):
+    PROM_PREFIX = "bp_memcached_pool"
+    PROM_LABELS = ["pool"]
+
+    promTotalConnections = Gauge(
+        f"{PROM_PREFIX}_size",
+        "Maximum size of this pool",
+        PROM_LABELS,
+    )
+
+    promUsedConnections = Gauge(
+        f"{PROM_PREFIX}_in_use",
+        "Number of connections in this pool currently in use",
+        PROM_LABELS,
+    )
+
+    promFreeConnections = Gauge(
+        f"{PROM_PREFIX}_free",
+        "Number of free connections in this pool",
+        PROM_LABELS,
+    )
+
+    def __init__(self, pooled_client: PooledClient, name: str = "memcache"):
         self.pooled_client = pooled_client
+
+        pool = self.pooled_client.client_pool
+        self.promTotalConnections.labels(name).set_function(lambda: pool.max_size)
+        self.promFreeConnections.labels(name).set_function(lambda: len(pool.free))
+        self.promUsedConnections.labels(name).set_function(lambda: len(pool.used))
 
     def report_memcache_runtime_metrics(self, batch: metrics.Client) -> None:
         pool = self.pooled_client.client_pool
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -133,31 +133,31 @@ class MemcacheContextFactory(ContextFactory):
     PROM_PREFIX = "bp_memcached_pool"
     PROM_LABELS = ["pool"]
 
-    promTotalConnections = Gauge(
+    prom_total_connections = Gauge(
         f"{PROM_PREFIX}_size",
         "Maximum size of this pool",
         PROM_LABELS,
     )
 
-    promUsedConnections = Gauge(
+    prom_used_connections = Gauge(
         f"{PROM_PREFIX}_in_use",
         "Number of connections in this pool currently in use",
         PROM_LABELS,
     )
 
-    promFreeConnections = Gauge(
+    prom_free_connections = Gauge(
         f"{PROM_PREFIX}_free",
         "Number of free connections in this pool",
         PROM_LABELS,
     )
 
-    def __init__(self, pooled_client: PooledClient, name: str = "memcache"):
+    def __init__(self, pooled_client: PooledClient, name: str = "default"):
         self.pooled_client = pooled_client
 
         pool = self.pooled_client.client_pool
-        self.promTotalConnections.labels(name).set_function(lambda: pool.max_size)
-        self.promFreeConnections.labels(name).set_function(lambda: len(pool.free))
-        self.promUsedConnections.labels(name).set_function(lambda: len(pool.used))
+        self.prom_total_connections.labels(name).set_function(lambda: pool.max_size)
+        self.prom_free_connections.labels(name).set_function(lambda: len(pool.free))
+        self.prom_used_connections.labels(name).set_function(lambda: len(pool.used))
 
     def report_memcache_runtime_metrics(self, batch: metrics.Client) -> None:
         pool = self.pooled_client.client_pool
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -134,8 +134,8 @@ class MemcacheContextFactory(ContextFactory):
     PROM_LABELS = ["pool"]
 
     pool_size_gauge = Gauge(
-        f"{PROM_PREFIX}_size",
-        "Maximum size of this pool",
+        f"{PROM_PREFIX}_max_size",
+        "Maximum number of connections allowed in this pool",
         PROM_LABELS,
     )
 
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -133,20 +133,20 @@ class MemcacheContextFactory(ContextFactory):
     PROM_PREFIX = "bp_memcached_pool"
     PROM_LABELS = ["pool"]
 
-    prom_total_connections = Gauge(
+    pool_size_gauge = Gauge(
         f"{PROM_PREFIX}_size",
         "Maximum size of this pool",
         PROM_LABELS,
     )
 
-    prom_used_connections = Gauge(
-        f"{PROM_PREFIX}_in_use",
+    used_connections_gauge = Gauge(
+        f"{PROM_PREFIX}_active_connections",
         "Number of connections in this pool currently in use",
         PROM_LABELS,
     )
 
-    prom_free_connections = Gauge(
-        f"{PROM_PREFIX}_free",
+    free_connections_gauge = Gauge(
+        f"{PROM_PREFIX}_free_connections",
         "Number of free connections in this pool",
         PROM_LABELS,
     )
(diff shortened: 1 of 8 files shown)


baseplate/clients/thrift.py
```diff
@@ -69,6 +70,21 @@ class ThriftContextFactory(ContextFactory):
...
+        f"{PROM_PREFIX}_size",
+        "Number of connections in this thrift pool",
+        PROM_LABELS,
+    )
+
+    promUsedConnections = Gauge(
+        f"{PROM_PREFIX}_in_use",
+        "Number of connections currently in use in this thrift pool",
+        PROM_LABELS,
+    )
```
fishy: 🔕 In Baseplate.go the equivalent (more or less, this is not really 100% mapped between go and py code) of these 2 gauges [are](https://github.com/reddit/baseplate.go/blob/ca159a8a3540e5154f5c6f7cde55c2f1ccdc7ae9/thriftbp/prometheus.go#L203-L215): `thriftbp_client_pool_allocated_clients` and `thriftbp_client_pool_active_connections` with `thrift_slug` label.

these are not core metrics defined in baseplate spec and do not have to be the same between go and py (as the implementation details differ), but just for your reference as it's likely still beneficial to make them as consistent as possible.

baseplate/clients/memcache/__init__.py
```diff
@@ -129,9 +130,35 @@ class MemcacheContextFactory(ContextFactory):
...
+        PROM_LABELS,
+    )
+
+    def __init__(self, pooled_client: PooledClient, name: str = "memcache"):
         self.pooled_client = pooled_client
 
+        pool = self.pooled_client.client_pool
+        self.promTotalConnections.labels(name).set_function(lambda: pool.max_size)
+        self.promFreeConnections.labels(name).set_function(lambda: len(pool.free))
+        self.promUsedConnections.labels(name).set_function(lambda: len(pool.used))
```
JessicaGreben: why are the metrics set in the init func instead of in the report_memcache_runtime_metrics method?

JessicaGreben: do we need the label "pool=memcache" when the metric name is "bp_memcached_pool"?

baseplate/clients/thrift.py
bjk-reddit: Yes, it would be useful to determine if these metrics are equivalent enough to the go versions. Having them use the same prefix would be nice, as it would make comparing / alerting / dashboards easier.

JessicaGreben (changes requested):
It might be nice to add some tests for these prom metrics.

baseplate/clients/memcache/__init__.py
bjk-reddit: I think the `pool` label is there for if a service is connected to multiple memcached services.

For example a service could be connected to `pool="stalecache"` and `pool="thing"`

krav: Yes, that’s what I tried to explain in the caveat section above. If you were to add the metric to the registry when creating the class, that is you call ‘.labels’ there, then that will be shared for each instance of the class regardless of what it connects to.

krav: This is seemingly a problem with the statsd metric too.

krav: > why are the metrics set in the init func instead of in the report_memcache_runtime_metrics method?

Because that function is called periodically to push statsd data, while the prometheus metrics are made on demand.

baseplate/clients/memcache/__init__.py
```diff
@@ -129,9 +130,35 @@ class MemcacheContextFactory(ContextFactory):
...
+
     def report_memcache_runtime_metrics(self, batch: metrics.Client) -> None:
```
JessicaGreben, line 162: i wonder why this memcache method is `report_memcache_runtime_metrics` but all the other clients the equivalent method is `report_runtime_metrics`. I wonder why the difference.

baseplate/clients/memcache/__init__.py
JessicaGreben: for prom "made on demand" meaning a new instance of the class is initialized?

baseplate/clients/memcache/__init__.py
krav: Hm, this might be a bug unless it's called outside of the baseplate. It's not referenced anywhere, while those with the latter name are called from https://github.com/reddit/baseplate.py/blob/48bcb0b099608aa8b8eecbf97cb929537d68753d/baseplate/__init__.py#L525

baseplate/clients/memcache/__init__.py
krav: No, when you get `/metrics` the function in `set_function` is called to create the number.

baseplate/clients/memcache/__init__.py
krav: Heh https://github.snooguts.net/reddit/reddit-service-post/blob/b0dbc37074a4299fdb916eaddd6b6cb1fe1fd568/post/__init__.py#L132

baseplate/clients/thrift.py
krav: I did look at the redis implementation and they're unfortunately very different, but none of the others. Thrift is a likely candidate to be similar.

baseplate/clients/memcache/__init__.py
JessicaGreben: lol. i see. thanks for the explanation

baseplate/clients/memcache/__init__.py
JessicaGreben: i see. yes thats great.

JessicaGreben (approved):
This PR looks good. Optionally we can add a few tests if you want. Also we can make the thrift metrics more similar to baseplate.go. Thanks for this!

tests/unit/clients/memcache_tests.py
```diff
@@ -54,6 +54,16 @@ def test_nodelay(self):
...
 
+    def test_metrics(self):
+        max_pool_size = "123"
+        ctx = MemcacheContextFactory(
+            pool_from_config(
+                {"memcache.endpoint": "localhost:1234", "memcache.max_pool_size": max_pool_size}
+            )
+        )
+        metric = ctx.promTotalConnections.collect()
+        self.assertEqual(metric[0].samples[0].value, float(max_pool_size))
```
JessicaGreben: would it be useful to assert the labels are set correctly?

baseplate/clients/memcache/__init__.py
```diff
@@ -129,9 +130,35 @@ class MemcacheContextFactory(ContextFactory):
 
     """
 
-    def __init__(self, pooled_client: PooledClient):
+    PROM_PREFIX = "bp_memcached_pool"
+    PROM_LABELS = ["pool"]
+
+    promTotalConnections = Gauge(
```
MelissaCole: Could these metrics follow pep8 naming by using snakecase? `prom_total_connections`, `prom_used_connections`, etc

baseplate/clients/memcache/__init__.py
```diff
@@ -129,9 +130,35 @@ class MemcacheContextFactory(ContextFactory):
...
+        PROM_LABELS,
+    )
+
+    promFreeConnections = Gauge(
+        f"{PROM_PREFIX}_free",
+        "Number of free connections in this pool",
+        PROM_LABELS,
+    )
+
+    def __init__(self, pooled_client: PooledClient, name: str = "memcache"):
```
MelissaCole: :no_bell: Would `None` or `""` work as a default since the metric name already has "memcache" in it?

baseplate/clients/redis.py
```diff
@@ -92,9 +94,35 @@ class RedisContextFactory(ContextFactory):
 
     """
 
-    def __init__(self, connection_pool: redis.ConnectionPool):
+    PROM_PREFIX = "bp_redis_pool"
+    PROM_LABELS = ["pool"]
+
+    totalConnections = Gauge(
+        f"{PROM_PREFIX}_connections",
```
MelissaCole: Thoughts on making this (and clustered redis) `f"{PROM_PREFIX}_size"` for consistency with memcached and sqlalchemy implementations?

baseplate/clients/sqlalchemy.py
```diff
@@ -155,12 +156,46 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
+    )
+
+    promCheckedInConnections = Gauge(
+        f"{PROM_PREFIX}_checked_in",
+        "Number of available, checked in, connections in this pool",
+        PROM_LABELS,
+    )
+
+    promCheckedOutConnections = Gauge(
+        f"{PROM_PREFIX}_checked_out",
```
MelissaCole: Could this be `in_use` to be consistent with thrift and memcached and `free` for `checked_in`?

bjk-reddit: We usually say `active` for connection pools.

MelissaCole (dismissed):
lgtm with some questions about naming conventions and consistency

baseplate/clients/memcache/__init__.py
krav: Leaving the label empty removes which can make the queries a bit more difficult. We can rename it "default" maybe?

baseplate/clients/memcache/__init__.py
krav: Totally, fixed!

baseplate/clients/redis.py
krav: I wonder if we shouldn't do it the other way around, since it describes the contents of the pool.

The redis client in baseplate.go seems to have `connections_total`, `connections_idle`. Having something end in `_total` signals that it's a Counter instead of a Gauge, so I'd keep that out - but switching the other two.

baseplate/clients/sqlalchemy.py
krav: Yeah, named it like this to be in line with the library vocabulary, but it does not make intuitive sense to me either. Changed it to active/idle.

baseplate/clients/redis_cluster.py
```diff
@@ -346,9 +347,33 @@ class ClusterRedisContextFactory(ContextFactory):
     :param connection_pool: A connection pool.
     """
 
-    def __init__(self, connection_pool: rediscluster.ClusterConnectionPool):
+    PROM_PREFIX = "bp_redis_cluster_pool"
+    PROM_LABELS = ["pool"]
+
+    promTotalConnections = Gauge(
```
nsheaps: Python best practices are snake case for var names

I'd also consider dropping prom from here to match the patterns you established above (change the below as well)

baseplate/clients/thrift.py
```diff
@@ -69,6 +70,21 @@ class ThriftContextFactory(ContextFactory):
 
     """
 
+    PROM_PREFIX = "bp_thrift_pool"
+    PROM_LABELS = ["client_cls"]
+
+    prom_total_connections = Gauge(
+        f"{PROM_PREFIX}_size",
```
nsheaps: You were going to change this to total right?

baseplate/clients/thrift.py
```diff
@@ -69,6 +70,21 @@ class ThriftContextFactory(ContextFactory):
...
+        "Number of connections in this thrift pool",
+        PROM_LABELS,
+    )
+
+    prom_used_connections = Gauge(
+        f"{PROM_PREFIX}_in_use",
```
nsheaps: "active" for consistency?

nsheaps (changes requested):
Few updates that you've already discussed still need to be made. Thanks for tackling this!

baseplate/clients/thrift.py
krav: `total` has a special meaning in Prometheus to denote counters, changed it to `connections` so it has a unit.

baseplate/clients/thrift.py
krav: Done!

baseplate/clients/redis_cluster.py
krav: Fixed

nsheaps (approved):

nsheaps (approved):

PR #655, opened 2022-03-23
Better integrate with baseplate.Error

pacejackson:
* add an option to 'baseplateify_processor' to automatically convert
   unhandled errors to baseplate.Error(INTERNAL_SERVER_ERROR)
 * Record 5xx baseplate.Errors as failures

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -20,6 +20,7 @@ from baseplate.lib import metrics
 from baseplate.lib.retry import RetryPolicy
 from baseplate.lib.thrift_pool import thrift_pool_from_config
 from baseplate.lib.thrift_pool import ThriftConnectionPool
+from baseplate.thrift.ttypes import Error
 
 
 class ThriftClient(config.Parser):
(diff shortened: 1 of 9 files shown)


baseplate/frameworks/thrift/__init__.py
```diff
@@ -41,13 +50,25 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 # should be expected in the protocol
                 span.finish(exc_info=sys.exc_info())
                 raise
+            except Error as exc:
+                # mark 5xx errors as failures since those are still "unexpected"
+                if exc.code / 100 == 5:
```
bradengroom: I think this will only work for 500, right?
```
>>> 504 / 100 == 5
False
```

I think we want integer division:
```
>>> 504 // 100 == 5
True
```

baseplate/clients/thrift.py
```diff
@@ -199,6 +200,14 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                 # should be expected in the protocol. this is an error!
                 span.finish(exc_info=sys.exc_info())
                 raise
+            except Error as exc:
+                # a 5xx error is an unexpected exception but not 5xx are
+                # not.
+                if exc.code / 100 == 5:
```
bradengroom: Same comment about about integer division

bradengroom (changes requested):
small comment

baseplate/frameworks/thrift/__init__.py
nanassito: also would be easier to read as `if 500 <= exc.code < 600:`

MelissaCole (approved):
lgtm!

bradengroom (approved):

PR #656, opened 2022-03-28
Fix connection leaks in dependencies by monkey patching gevent queues

RoganMurley:
Connection pools implemented using standard library queues [can leak connections when a gevent timeout exception is raised](https://github.com/gevent/gevent/issues/1875). To avoid this gevent queues should be used rather than standard library queues.

We fixed this in [baseplate.py Thrift pools](https://github.com/reddit/baseplate.py/pull/643), but there are still many such issues in dependencies. In some cases [we can explicitly use a gevent queue](https://github.com/reddit/baseplate.py/pull/646), but in other cases we have to [monkey patch](https://github.com/celery/kombu/blob/master/kombu/resource.py).

Previously we did not monkey patch because it is unclear why gevent `patch_queue` and `patch_all` does not patch by default. I've [raised an issue asking the gevent team about this](https://github.com/gevent/gevent/issues/1875), and also started canarying such a change in a Reddit service that was experiencing Kombu connection leaks. The service hasn't experienced any issues with the change so I'd like to merge it upstream.

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -16,6 +16,7 @@ A basic example of usage::
 """
 import contextlib
 import logging
+import queue
 import socket
 import time
 
@@ -24,7 +25,6 @@ from typing import Generator
 from typing import Optional
 from typing import TYPE_CHECKING
 
-from gevent import queue as gevent_queue
 from thrift.protocol import THeaderProtocol
 from thrift.protocol.TProtocol import TProtocolBase
 from thrift.protocol.TProtocol import TProtocolException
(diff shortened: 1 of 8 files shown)


baseplate/lib/thrift_pool.py
```diff
@@ -24,7 +25,6 @@
 from typing import Optional
 from typing import TYPE_CHECKING
 
-from gevent import queue as gevent_queue
```
RoganMurley, line 27: Undoing explicitly using a gevent queue, because the standard library queue should be patched.

https://github.com/reddit/baseplate.py/pull/643

baseplate/server/monkey.py
```diff
@@ -0,0 +1,13 @@
+from gevent.monkey import patch_module
```
RoganMurley, line 1: Was unsure where to put this file.

MelissaCole (approved):
lgtm!

bradengroom (approved):

PR #657, opened 2022-03-28
Add default Prometheus metrics for Thrift server

JessicaGreben:
This original [PR](https://github.com/reddit/baseplate.py/pull/647) was getting too big, so this PR splits out one part. There will be approx. 4 PRs split out in total:
* 1st PR (this PR) adds default Prometheus metrics for Thrift server
* 2nd PR will do the same for Thrift Client
* 3rd PR will add default Prometheus metrics for HTTP server
* 4th PR will do the same for HTTP client

Background:
We are adding support to export Prometheus metrics from baseplate.py services.  As of baseplate.py [v2.3.0](https://github.com/reddit/baseplate.py/releases/tag/v2.3.0), `baseplate-serve` also runs a server that exports Prometheus metrics on endpoint `/metrics` on port 6060. 

This PR adds the default Prometheus metrics for Thrift servers. The default Thrift metrics to be exported are defined in the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#thrift).

Testing:
I make a test thrift app from baseplate-cookiecutter and tested out this new Prometheus code. Here are the steps I used to test:
1) create thrift service with baseplate-cookiecutter (make sure baseplate v2.3.0 or greater) so that prom metrics are exported on port 6060 by default.
2) make sure `prometheus-client==0.12.0` is in the requirements.txt file
3) set env var `PROMETHEUS_MULTIPROC_DIR` to a directory where metrics from multiprocess python can be written  (this is already included in prod [dockerfile](https://sourcegraph.build.ue1.snooguts.net/github.snooguts.net/reddit/docker-baseplate-py/-/blob/2/Dockerfile?L25:5))
4) either set `metrics.tagging` or `metrics.namespace` in the service's ini config file.
5) build thrift with `make thrift` then run `baseplate-serve example.ini`
6) use the test-client to hit the `is_healthy` endpoint
7) check the prometheus metrics that were created:
```
$ curl localhost:6060/metrics
# HELP thrift_server_active_requests Multiprocess metric
# TYPE thrift_server_active_requests gauge
thrift_server_active_requests{pid="23318",thrift_method="is_healthy"} 0.0
# HELP thrift_server_latency_seconds Multiprocess metric
# TYPE thrift_server_latency_seconds histogram
thrift_server_latency_seconds_sum{thrift_method="is_healthy",thrift_success="true"} 0.041624628
thrift_server_latency_seconds_bucket{le="0.0001",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.00025",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.000625",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.0015625",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.00390625",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.009765625",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.0244140625",thrift_method="is_healthy",thrift_success="true"} 0.0
thrift_server_latency_seconds_bucket{le="0.06103515625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="0.152587890625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="0.3814697265625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="0.95367431640625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="2.384185791015625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="5.9604644775390625",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="14.901161193847656",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_bucket{le="+Inf",thrift_method="is_healthy",thrift_success="true"} 1.0
thrift_server_latency_seconds_count{thrift_method="is_healthy",thrift_success="true"} 1.0
# HELP thrift_server_requests_total Multiprocess metric
# TYPE thrift_server_requests_total counter
thrift_server_requests_total{thrift_baseplate_status="",thrift_baseplate_status_code="",thrift_exception_type="",thrift_method="is_healthy",thrift_success="true"} 1.0
```

It appears the `HELP` text is clobbered in multiprocess mode, [issue](https://github.com/prometheus/client_python/issues/211).

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -203,7 +203,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
             except Error as exc:
                 # a 5xx error is an unexpected exception but not 5xx are
                 # not.
-                if exc.code // 100 == 5:
+                if 500 <= exc.code < 600:
                     span.finish(exc_info=sys.exc_info())
                 else:
                     span.finish()
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -20,6 +20,7 @@ from baseplate.lib import metrics
 from baseplate.lib.retry import RetryPolicy
 from baseplate.lib.thrift_pool import thrift_pool_from_config
 from baseplate.lib.thrift_pool import ThriftConnectionPool
+from baseplate.thrift.ttypes import Error
 
 
 class ThriftClient(config.Parser):
@@ -198,6 +199,14 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                 # these are subclasses of TException but aren't ones that
                 # should be expected in the protocol. this is an error!
                 span.finish(exc_info=sys.exc_info())
+                raise
+            except Error as exc:
+                # a 5xx error is an unexpected exception but not 5xx are
+                # not.
+                if exc.code // 100 == 5:
+                    span.finish(exc_info=sys.exc_info())
+                else:
+                    span.finish()
                 raise
             except TException:
                 # this is an expected exception, as defined in the IDL
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -237,6 +237,7 @@ class BaseplateConfigurator:
             name=request.matched_route.name,
             trace_info=trace_info,
         )
+        span.set_tag("protocol", "http")
         span.set_tag("http.url", request.url)
         span.set_tag("http.method", request.method)
         span.set_tag("peer.ipv4", request.remote_addr)
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -64,9 +64,8 @@ class _ContextAwareHandler:
                 else:
                     span.finish()
                 raise
-            except TException:
-                exc_info = sys.exc_info()
-                span.set_tag("exception_type", exc_info[1].__class__.__name__)
+            except TException as e:
+                span.set_tag("exception_type", type(e).__name__)
                 span.set_tag("success", "false")
                 # this is an expected exception, as defined in the IDL
                 span.finish()
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -18,6 +18,8 @@ from baseplate import Baseplate
 from baseplate import RequestContext
 from baseplate import TraceInfo
 from baseplate.lib.edgecontext import EdgeContextFactory
+from baseplate.thrift.ttypes import Error as bp_error
+from baseplate.thrift.ttypes import ErrorCode
 
 
 class _ContextAwareHandler:
@@ -33,6 +35,7 @@ class _ContextAwareHandler:
             handler_fn = getattr(self.handler, fn_name)
 
             span = self.context.span
+            span.set_tag("thrift.method", fn_name)
             try:
                 span.start()
                 result = handler_fn(self.context, *args, **kwargs)
@@ -42,6 +45,11 @@ class _ContextAwareHandler:
                 span.finish(exc_info=sys.exc_info())
                 raise
             except TException:
+                name, code, status = processException(sys.exc_info())
+                span.set_tag("exception_type", name)
+                span.set_tag("thrift.status_code", code)
+                span.set_tag("thrift.status", status)
+                span.set_tag("success", "false")
                 # this is an expected exception, as defined in the IDL
                 span.finish()
                 raise
@@ -55,6 +63,27 @@ class _ContextAwareHandler:
                 return result
 
         return call_with_context
+
+
+def processException(exc_info):
+    """
+    processException attempts to get additional information from the
+    exception info. If the exception is a baseplate thrift Error type
+    then the status code and status is also returned.
+    The code is the numeric value from a baseplate.Error, for example: "404".
+    The status is the human-readable status, for example: "NOT_FOUND".
+    """
+    exc_class = exc_info[0]
+    exc = exc_info[1]
+    code, status = "", ""
+    name = exc_class.__name__
+
+    if issubclass(exc_class, bp_error):
+        code = exc.code
+        c = ErrorCode()
+        status = c._VALUES_TO_NAMES.get(code, "")
+
+    return name, code, status
 
 
 def baseplateify_processor(
@@ -110,7 +139,7 @@ def baseplateify_processor(
                 context.deadline_budget = None
 
             span = baseplate.make_server_span(context, name=fn_name, trace_info=trace_info)
-
+            span.set_tag("protocol", "thrift")
             try:
                 service_name = headers[b"User-Agent"].decode()
             except (KeyError, UnicodeDecodeError):
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -18,15 +18,22 @@ from baseplate import Baseplate
 from baseplate import RequestContext
 from baseplate import TraceInfo
 from baseplate.lib.edgecontext import EdgeContextFactory
-from baseplate.thrift.ttypes import Error as bp_error
+from baseplate.thrift.ttypes import Error
 from baseplate.thrift.ttypes import ErrorCode
 
 
 class _ContextAwareHandler:
-    def __init__(self, handler: Any, context: RequestContext, logger: Logger):
+    def __init__(
+        self,
+        handler: Any,
+        context: RequestContext,
+        logger: Logger,
+        convert_to_baseplate_error: bool,
+    ):
         self.handler = handler
         self.context = context
         self.logger = logger
+        self.convert_to_baseplate_error = convert_to_baseplate_error
 
     def __getattr__(self, fn_name: str) -> Callable[..., Any]:
         def call_with_context(*args: Any, **kwargs: Any) -> Any:
@@ -44,11 +51,24 @@ class _ContextAwareHandler:
                 # should be expected in the protocol
                 span.finish(exc_info=sys.exc_info())
                 raise
-            except TException:
-                name, code, status = processException(sys.exc_info())
-                span.set_tag("exception_type", name)
+            except Error as exc:
+                exc_info = sys.exc_info()
+                code = exc_info[1].code
+                c = ErrorCode()
+                status = c._VALUES_TO_NAMES.get(code, "")
+                span.set_tag("exception_type", exc_info[0].__name__)
                 span.set_tag("thrift.status_code", code)
                 span.set_tag("thrift.status", status)
+                span.set_tag("success", "false")
+                # mark 5xx errors as failures since those are still "unexpected"
+                if exc.code // 100 == 5:
+                    span.finish(exc_info=sys.exc_info())
+                else:
+                    span.finish()
+                raise
+            except TException:
+                exc_info = sys.exc_info()
+                span.set_tag("exception_type", exc_info[0].__name__)
                 span.set_tag("success", "false")
                 # this is an expected exception, as defined in the IDL
                 span.finish()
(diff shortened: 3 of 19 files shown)


baseplate/server/__init__.py
```diff
@@ -296,13 +275,9 @@ def load_app_and_run_server() -> None:
...
-    try:
-        # pylint: disable=cyclic-import
-        from baseplate.server.prometheus import start_prometheus_exporter
-    except ImportError:
-        logger.debug("Prometheus exporter not available. pip install prometheus-client to enable.")
-    else:
-        start_prometheus_exporter()
+    from baseplate.server.prometheus import start_prometheus_exporter
+
+    start_prometheus_exporter()
```
nsheaps: I'm not totally sold on requiring prometheus on everything that uses baseplate libs. Was this discussed somewhere?

nsheaps (commented):
Since this is one of my first baseplate PRs, can we hop on a call and go over it together? Sorry for the delay

JessicaGreben:
@nsheaps Thanks for the help with the review. Yes jumping on a call would be great. I will DM you to set it up.

baseplate/server/__init__.py
JessicaGreben: Since the company is migrating to Prometheus from Wavefront, we want to add the same functionality that there is for the statsd exporter that ships metrics to Wavefront.  The baseplate.spec documents what the baseplate frameworks should implement and there is a section for Prometheus metrics: https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md

baseplate/clients/thrift.py
```diff
@@ -203,7 +203,7 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
             except Error as exc:
                 # a 5xx error is an unexpected exception but not 5xx are
                 # not.
```
krav, line 205: 🙃

baseplate/frameworks/thrift/__init__.py
```diff
@@ -51,13 +52,22 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
...
+                span.set_tag("success", "false")
                 # mark 5xx errors as failures since those are still "unexpected"
-                if exc.code // 100 == 5:
+                if 500 <= exc.code < 600:
                     span.finish(exc_info=sys.exc_info())
                 else:
                     span.finish()
                 raise
             except TException:
+                exc_info = sys.exc_info()
```
krav: I think this could be 
```suggestion
   except TExecption as e:
   span.set_tag("exception_type", type(e).__name__)
```

baseplate/observers/prometheus.py
```diff
@@ -0,0 +1,123 @@
...
+        return f"{self.protocol}_server"
+
+    def get_labels(self) -> Dict[str, str]:
+        return self.tags
+
+    @property
+    def protocol(self) -> str:
+        return self.tags.get("protocol", "unknown")
+
+    def set_metrics_by_protocol(self) -> None:
```
krav, line 59: I've been staring myself a bit blind trying to figure this out, so someone more familiar with baseplate should verify this. I think we're breaking the abstraction here. We should implement a Prometheus Span observer, but it should only be used for shipping the results of span traces to prometheus. The actual metrics specific to thrift or http should either be implemented as spans in the server (e.g. making a child span in the thrift server when handling a request), or - my favorite - just directly in the thrift or http server.

JessicaGreben: i like this idea. i would love to implement this differently. i like the idea of doing the metrics directly in the server. less discuss more

baseplate/server/__init__.py
nsheaps: [discussed offline] Lets swap this back to the try block

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -237,6 +237,7 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
             name=request.matched_route.name,
             trace_info=trace_info,
         )
+        span.set_tag("protocol", "http")
```
nsheaps, line 240: 🔕 does this need a https distinction?

nsheaps: Answer: no

nsheaps (approved):
Spoke offline and got a full rundown of how this is working, future expectations and the like.

Few points to document:
1) For reddit's purposes, this is more than fine, as we can make a solid call as to "we support thrift, HTTP and grpc" but the mechanism for "if span protocol is HTTP" could likely be improved if it needed to be more flexible
2) Similarly, the span observers that are called from the switch statement are explicit due to the slight disparities in metric names between the traced protocol. Would be great to make a generic generator that you pass in the protocol and get one made for you (or a generic one that looks at the protocol and adjusts the metric names as necessary)
3) splitting the 4 implementation tasks was very helpful for review, thank you
4) GRPC is coming, but not in this PR, even in a "not implemented capacity" (like how HTTP is done [here](https://github.com/reddit/baseplate.py/pull/657/files#diff-90aecd525ed0f2fdafdd61f9e5a067264e6d1a748cb5e1328f211348daefa921R63)
5) Some improvements we discussed are likely not work the effort for perfection considering the long term trend towards go
6) Because of the explicit requirement of metrics collection, we finally decided that prometheus client is required if metrics are turned on, rather than an optional import that logs a warning

baseplate/server/__init__.py
```diff
@@ -296,13 +275,12 @@ def load_app_and_run_server() -> None:
     if einhorn.is_worker():
         einhorn.ack_startup()
 
-    try:
-        # pylint: disable=cyclic-import
+    if "metrics.tagging" in config.app or "metrics.namespace" in config.app:
```
JessicaGreben, line 278: @nsheaps A follow up from our previous discussions about replacing this `try/except` block with checking the config for metrics being enabled:

It turns out there are two different ways to enable the current statsd metrics (and you cannot set both at the same time):  
1. `metrics.tagging` enables the tagged metrics (described [here](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/metrics-tagged.md))
2. `metrics.namespace` enables non-tagged metrics (described [here](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/metrics.md)).

Searching in sourcegraph, it looks like there are approx. 82 repos that set `metrics.namespace` and 20 repos that set `metric.tagging`.  For reference, here are the sourcegraph queries I ran:
```
repo:reddit-service-* repo:contains.file(requirements.txt) file:ini metrics.tagging select:repo
repo:reddit-service-* repo:contains.file(requirements.txt) file:ini metrics.namespace select:repo
```

Since the goal is to enable Prometheus metrics alongside the current statsd metrics (until wavefront is turned off), then we need to enable Prometheus when either `metrics.tagging` or `metrics.namespace` are set.

Another consideration is to create a new config option specifically for Prometheus.  It would be nice to have a separate config for the Prometheus metrics to keep it untangled from the current metric configs options, however there wouldn't be a way to turn on Prometheus automatically without updating the `.ini` file which would make upgrades not backwards compatible.  But I'm open to a discussion either way.

baseplate/server/__init__.py
JessicaGreben: @nsheaps A follow up from our previous discussions about replacing this `try/except` block with checking the config for metrics being enabled:

It turns out there are two different ways to enable the current statsd metrics (and you cannot set both at the same time):  
1. `metrics.tagging` enables the tagged metrics (described [here](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/metrics-tagged.md))
2. `metrics.namespace` enables non-tagged metrics (described [here](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/metrics.md)).

Searching in sourcegraph, it looks like there are approx. 82 repos that set `metrics.namespace` and 20 repos that set `metric.tagging`.  For reference, here are the sourcegraph queries I ran:
```
repo:reddit-service-* repo:contains.file(requirements.txt) file:ini metrics.tagging select:repo
repo:reddit-service-* repo:contains.file(requirements.txt) file:ini metrics.namespace select:repo
```

Since the goal is to enable Prometheus metrics alongside the current statsd metrics (until wavefront is turned off), then we need to enable Prometheus when either `metrics.tagging` or `metrics.namespace` are set.

Another consideration is to create a new config option specifically for Prometheus.  It would be nice to have a separate config for the Prometheus metrics to keep it untangled from the current metric configs options, however there wouldn't be a way to turn on Prometheus automatically without updating the `.ini` file which would make upgrades not backwards compatible.  But I'm open to a discussion either way.

baseplate/frameworks/thrift/__init__.py
```diff
@@ -51,13 +52,21 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
...
             except Error as exc:
+                c = ErrorCode()
+                status = c._VALUES_TO_NAMES.get(exc.code, "")
+                span.set_tag("exception_type", "Error")
+                span.set_tag("thrift.status_code", exc.code)
+                span.set_tag("thrift.status", status)
+                span.set_tag("success", "false")
                 # mark 5xx errors as failures since those are still "unexpected"
-                if exc.code // 100 == 5:
+                if 500 <= exc.code < 600:
```
nsheaps, line 62: 🔕 what other errors come through this code path? Should they maybe be split out into specific exceptions?

baseplate/frameworks/thrift/__init__.py
```diff
@@ -51,13 +52,21 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
...
                     span.finish(exc_info=sys.exc_info())
                 else:
                     span.finish()
                 raise
-            except TException:
+            except TException as e:
+                span.set_tag("exception_type", type(e).__name__)
+                span.set_tag("success", "false")
```
nsheaps, line 69: 🔕 should there be a `span.set_tag("success", "true")` somewhere?

baseplate/server/__init__.py
nsheaps: Thanks for all the info!

I think the way you have it implemented is fine, and easiest to upgrade to without requiring a change to configs.

I do see at least one project that actually has both of those configs, a difference between development and production.

I think a modified config section might be warranted in a future version, addressed separately, as the configurations for both metrics.namespace and metrics.tagging seem somewhat specific on where to send the metrics to, rather than prometheus' poll-based metrics. We can definitely address the implementation of it separately, but I'd picture the config to be somewhat pluggable, such as the below. The reason this didn't come up is the prometheus observer doesn't actually have a config at the moment, though I could imagine adding things like tls configurations. Rolling this out can be done more easily while not blocking this progress using a batch change later


```ini
metrics.enabled = true # used to be metrics.tagging, more understandable
metrics.whitelist = success, error, endpoint, client

# statsd/telegraf config
metrics.statsd.enabled = true
metrics.statsd.namespace = mine
metrics.statsd.endpoint = telegraf:8125
metrics.statsd.swallow_network_errors = true

# prometheus config
metrics.prometheus.enabled = true
```

nsheaps (approved):

baseplate/frameworks/thrift/__init__.py
JessicaGreben: when `span.finish()` gets called its calls the `on_finish` method in the Prometheus observer and `span.set_tag("success", "true")` gets set if there are no exceptions.

baseplate/frameworks/thrift/__init__.py
JessicaGreben: `sys.exc_info()`  will indicate what the exception is. I'm not sure what you mean by "split out"?

PR #658, opened 2022-03-29
Support loading encrypted contents from S3 bucket for live_data_watch

SiddharthManoj:
### Summary

We are adding support in Baseplate to be able to fetch contents from a file rather than just reading the data from the ZK Node. This specific PR allows teams to provide the necessary information to retrieve their **encrypted** file stored in an S3 Bucket.

### Why?

The Experiments team (at least) specifically needs this. Our ZK Node where we store Experiments Config data is slowly approaching its 1MB cap. We need to scale so instead we are now storing our config data in S3 File Storage. 

In the future, other teams can also begin using this if they prefer to not have to write all their data to the ZK node and instead reference a pointer to their data (or some other logic).

### How does it work?

Instead of the zk source containing ALL the data, it could instead come in a JSON blob as follows (specific example for S3):

```
data = {
    live_data_watcher_load_type: "S3"
    bucket_name: str
    file_key: str
    sse_key: str
    region_name:str
}
```

In this PR, we specifically add logic to check the contents of the zk source and if it's a JSON with key 'live_data_watcher_load_type' and value 'S3', then we know we can go and retrieve that config data from S3. We also make sure that the logic only handles encrypted files for security reasons.

If we scan the contents of the zk source and realize that the JSON key we're expecting is not there, we know to just write directly to the destination as we did before **(backwards compatible)**

In the future, other teams can support other types (not just 'S3')...but will have to define their own keys necessary to pull from the data storage (e.g. GCS)

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -38,7 +38,8 @@ class NodeWatcher:
         self.mode = mode
 
     @staticmethod
-    def get_encrypted_json_from_s3(bucket_name: str, file_key: str, region_name: str, sse_key: str
+    def get_encrypted_json_from_s3(
+        bucket_name: str, file_key: str, region_name: str, sse_key: str
     ) -> Optional[dict]:
         s3_client = boto3.client(
             "s3",
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -70,7 +70,7 @@ class NodeWatcher:
         return data
 
     @staticmethod
-    def get_data_to_write(json_data: dict, data: bytes) -> Optional[bytes]:
+    def get_data_to_write(json_data: dict, data: Optional[bytes]) -> Optional[bytes]:
         # Check if we have a JSON in a special format containing:
         # data = {
         #    "live_data_watcher_load_type": str
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -6,16 +6,15 @@ import logging
 import os
 import sys
 import time
-import boto3
 
 from pathlib import Path
 from typing import Any
 from typing import NoReturn
 from typing import Optional
 
+import boto3
 
 from botocore.client import ClientError
-
 from kazoo.client import KazooClient
 from kazoo.protocol.states import ZnodeStat
 
@@ -39,8 +38,7 @@ class NodeWatcher:
         self.mode = mode
 
     @staticmethod
-    def get_encrypted_json_from_s3(
-        self, bucket_name:str, file_key: str, region_name:str, sse_key: str
+    def get_encrypted_json_from_s3(bucket_name: str, file_key: str, region_name: str, sse_key: str
     ) -> Optional[dict]:
         s3_client = boto3.client(
             "s3",
@@ -57,7 +55,6 @@ class NodeWatcher:
         try:
             s3_object = s3_client.get_object(**kwargs)
             data = s3_object["Body"].read()
-
         except ClientError as error:
             logger.exception(
                 "Failed to load from S3. Received error code %s: %s",
@@ -80,16 +77,18 @@ class NodeWatcher:
         # downloading the contents of encrypted files uploaded to S3.
         if json_data.get("live_data_watcher_load_type") == "S3":
             # Only write the data if we actually managed to fetch its contents.
-            bucket_name=json_data.get("bucket_name")
-            file_key=json_data.get("file_key"),
-            sse_key=json_data.get("sse_key")
+            bucket_name = json_data.get("bucket_name")
+            file_key = (json_data.get("file_key"),)
+            sse_key = json_data.get("sse_key")
             region_name = json_data.get("region_name")
             # We require all of these keys to properly read from S3.
             if bucket_name is None or file_key is None or sse_key is None or region_name is None:
                 logger.debug("Missing data in live data watch zk node to read from S3.")
                 return None
             # If we have all the correct keys, attempt to read the config from S3.
-            json_data = NodeWatcher.get_encrypted_json_from_s3(bucket_name=bucket_name, file_key=file_key, region_name=region_name, sse_key=sse_key)
+            json_data = NodeWatcher.get_encrypted_json_from_s3(
+                bucket_name=bucket_name, file_key=file_key, region_name=region_name, sse_key=sse_key
+            )
             return json_data
         else:
             return data
@@ -125,7 +124,9 @@ class NodeWatcher:
                 if data_to_write is not None:
                     tmpfile.write(data_to_write)
                 else:
-                    logger.warning("No data written to destination node. Something is likely misconfigured.")
+                    logger.warning(
+                        "No data written to destination node. Something is likely misconfigured."
+                    )
         os.rename(self.dest + ".tmp", self.dest)
 
 
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -40,7 +40,7 @@ class NodeWatcher:
     @staticmethod
     def get_encrypted_json_from_s3(
         bucket_name: str, file_key: str, region_name: str, sse_key: str
-    ) -> Optional[dict]:
+    ) -> Optional[bytes]:
         s3_client = boto3.client(
             "s3",
             region_name=region_name,
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -95,7 +95,7 @@ class NodeWatcher:
             )
         return data
 
-    def handle_empty_data(self):
+    def handle_empty_data(self) -> None:
         # the data node does not exist
         try:
             logger.info("Removing %r; watched node deleted.", self.dest)
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -5,10 +5,14 @@ import logging
 import os
 import sys
 import time
+import json
+from tkinter import E
+import requests
 
 from pathlib import Path
 from typing import Any
 from typing import NoReturn
+from typing import Optional
 
 from kazoo.client import KazooClient
 from kazoo.protocol.states import ZnodeStat
@@ -32,6 +36,41 @@ class NodeWatcher:
         self.group = group
         self.mode = mode
 
+    @staticmethod
+    def fetch_data_from_url(url: str) -> Optional[str]:
+        data = None
+        try:
+            # Fetch the data from the url as bytes.
+            data = requests.get(url).content
+        except requests.exceptions.RequestException as e:
+            logger.exception(e)
+            return None
+        return data
+
+    @staticmethod
+    def get_data_to_write(json_data: dict, data: bytes) -> Optional[bytes]:
+        # Check if we have a JSON in this special format:
+        # data = {
+        #    "live_data_watcher_load_type": str
+        #    "data": str
+        #    "md5_hashed_data": str
+        # }
+        # If the load type is 'http', this format is an indication that we support
+        # downloading the contents of files uploaded to S3, GCP, etc when provided 
+        # with an accessible URL.
+        if json_data.get("live_data_watcher_load_type") == "http":
+            # Only write the data if we actually managed to fetch its contents.
+            url = json_data.get("data")
+            if url is None:
+                logger.exception("No url found in zk source JSON node.")
+                return None
+            url_data = NodeWatcher.fetch_data_from_url(url)
+            if url_data is None:
+                return None
+            return url_data
+        else:
+            return data
+
     def on_change(self, data: bytes, _znode_stat: ZnodeStat) -> None:
         if data is None:
             # the data node does not exist
@@ -50,8 +89,20 @@ class NodeWatcher:
             if self.owner and self.group:
                 os.fchown(tmpfile.fileno(), self.owner, self.group)
             os.fchmod(tmpfile.fileno(), self.mode)
-
-            tmpfile.write(data)
+            try:
+                json_data = json.loads(data.decode('UTF-8'))
+            except json.decoder.JSONDecodeError:
+                # If JSON fails to decode, still write the bytes data since
+                # we don't necessarily know if the the contents of the znode
+                # had to be valid JSON anyways.
+                tmpfile.write(data)
+            else:
+                # If no exceptions, we have valid JSON, and can parse accordingly.
+                data_to_write = NodeWatcher.get_data_to_write(json_data, data)
+                if data_to_write is not None:
+                    tmpfile.write(data_to_write)
+                else:
+                    logger.warning("No data written to destination node.")
         os.rename(self.dest + ".tmp", self.dest)
 
 
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -80,7 +80,7 @@ class NodeWatcher:
         if json_data.get("live_data_watcher_load_type") == "S3":
             # Only write the data if we actually managed to fetch its contents.
             bucket_name = json_data.get("bucket_name")
-            file_key = (json_data.get("file_key"),)
+            file_key = json_data.get("file_key")
             sse_key = json_data.get("sse_key")
             region_name = json_data.get("region_name")
             # We require all of these keys to properly read from S3.
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -88,10 +88,10 @@ class NodeWatcher:
                 logger.debug("Missing data in live data watch zk node to read from S3.")
                 return None
             # If we have all the correct keys, attempt to read the config from S3.
-            json_data = NodeWatcher.get_encrypted_json_from_s3(
+            s3_data = NodeWatcher.get_encrypted_json_from_s3(
                 bucket_name=bucket_name, file_key=file_key, region_name=region_name, sse_key=sse_key
             )
-            return json_data
+            return s3_data
         else:
             return data
 
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -1,18 +1,18 @@
 """Watch nodes in ZooKeeper and sync their contents to disk on change."""
 import argparse
 import configparser
+import json
 import logging
 import os
 import sys
 import time
-import json
-from tkinter import E
-import requests
 
 from pathlib import Path
 from typing import Any
 from typing import NoReturn
 from typing import Optional
+
+import requests
 
 from kazoo.client import KazooClient
 from kazoo.protocol.states import ZnodeStat
@@ -56,7 +56,7 @@ class NodeWatcher:
         #    "md5_hashed_data": str
         # }
         # If the load type is 'http', this format is an indication that we support
-        # downloading the contents of files uploaded to S3, GCP, etc when provided 
+        # downloading the contents of files uploaded to S3, GCP, etc when provided
         # with an accessible URL.
         if json_data.get("live_data_watcher_load_type") == "http":
             # Only write the data if we actually managed to fetch its contents.
@@ -90,7 +90,7 @@ class NodeWatcher:
                 os.fchown(tmpfile.fileno(), self.owner, self.group)
             os.fchmod(tmpfile.fileno(), self.mode)
             try:
-                json_data = json.loads(data.decode('UTF-8'))
+                json_data = json.loads(data.decode("UTF-8"))
             except json.decoder.JSONDecodeError:
                 # If JSON fails to decode, still write the bytes data since
                 # we don't necessarily know if the the contents of the znode
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -41,7 +41,7 @@ class NodeWatcher:
     @staticmethod
     def get_encrypted_json_from_s3(
         self, bucket_name:str, file_key: str, region_name:str, sse_key: str
-    ) -> tuple[Optional[dict], dict]:
+    ) -> Optional[dict]:
         s3_client = boto3.client(
             "s3",
             region_name=region_name,
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -56,7 +56,7 @@ class NodeWatcher:
         #    "md5_hashed_data": str
         # }
         # If the load type is 'http', this format is an indication that we support
-        # downloading the contents of files uploaded to S3, GCP, etc when provided
+        # downloading the contents of files uploaded to S3, GCS, etc when provided
         # with an accessible URL.
         if json_data.get("live_data_watcher_load_type") == "http":
             # Only write the data if we actually managed to fetch its contents.
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -46,7 +46,7 @@ class NodeWatcher:
             "s3",
             region_name=region_name,
         )
-        json_data = None
+        data: Optional[bytes] = None
         # The S3 data must be Server Side Encrypted and we should have the decryption key.
         kwargs = {
             "Bucket": bucket_name,
@@ -57,8 +57,7 @@ class NodeWatcher:
         try:
             s3_object = s3_client.get_object(**kwargs)
             data = s3_object["Body"].read()
-            if data:
-                json_data = json.loads(data.decode("utf-8"))
+
         except ClientError as error:
             logger.exception(
                 "Failed to load from S3. Received error code %s: %s",
@@ -69,10 +68,7 @@ class NodeWatcher:
         except ValueError as error:
             logger.exception(error)
             return None
-        except json.decoder.JSONDecodeError as error:
-            logger.exception(error)
-            return None
-        return json_data
+        return data
 
     @staticmethod
     def get_data_to_write(json_data: dict, data: bytes) -> Optional[bytes]:
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -102,9 +102,6 @@ class NodeWatcher:
         # the data node does not exist
         try:
             logger.info("Removing %r; watched node deleted.", self.dest)
-            logger.warning(
-                "No data written to destination file. Something is likely misconfigured."
-            )
             os.unlink(self.dest)
         except OSError as exc:
             logger.debug("%s: couldn't unlink: %s", self.dest, exc)
@@ -133,7 +130,9 @@ class NodeWatcher:
                 # If no exceptions, we have valid JSON, and can parse accordingly.
                 data = NodeWatcher.get_data_to_write(json_data, data)
                 if data is None:
-                    self.handle_empty_data()
+                    logger.warning(
+                        "No data written to destination file. Something is likely misconfigured."
+                    )
                     return
                 tmpfile.write(data)
         os.rename(self.dest + ".tmp", self.dest)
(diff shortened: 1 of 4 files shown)


baseplate/sidecars/live_data_watcher.py
```diff
@@ -50,8 +89,20 @@ def on_change(self, data: bytes, _znode_stat: ZnodeStat) -> None:
...
+                # we don't necessarily know if the the contents of the znode
+                # had to be valid JSON anyways.
+                tmpfile.write(data)
+            else:
+                # If no exceptions, we have valid JSON, and can parse accordingly.
+                data_to_write = NodeWatcher.get_data_to_write(json_data, data)
+                if data_to_write is not None:
+                    tmpfile.write(data_to_write)
+                else:
+                    logger.warning("No data written to destination node.")
```
SiddharthManoj: This would only occur in the current scenario if `live_data_watcher_load_type` was `http` but there was no URL or the data from the URL was not fetchable.

Otherwise, it'll always fallback to `tmpfile.write(data)` where data is just whatever was in the ZK source.

SiddharthManoj:
@fishy if you wanna get an eye on this as well.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +36,41 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+            if url is None:
+                logger.exception("No url found in zk source JSON node.")
+                return None
+            url_data = NodeWatcher.fetch_data_from_url(url)
+            if url_data is None:
+                return None
+            return url_data
+        else:
+            return data
+
```
KTAtkinson: This code appears to imply that configuration files would be available over HTTP unauthenticated. This makes me pretty uncomfortable for two reasons:

1. We have no way of authenticating that the server is trusted
2. We have little or no auditability into who accessed configuration and what response they received

Generally, I suggest treating configuration like source code, so it should be equally as locked down.

Personally, I would suggest integrating with EC2 and GCP directly. This would allow us to bootstap with their auth methods. It also gives us access to their audit logs if we need to verify the integrity of the configurations. Both of these platforms also have ways that you can watch for changes  on objects, which is probably preferred to polling,

KTAtkinson (changes requested):

baseplate/sidecars/live_data_watcher.py
fishy: I think the implication is that auth should be handled by service account so that you don't have to do anything special in the code, instead of "just make them publicly readable".

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +36,41 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        #    "md5_hashed_data": str
+        # }
+        # If the load type is 'http', this format is an indication that we support
+        # downloading the contents of files uploaded to S3, GCS, etc when provided
+        # with an accessible URL.
+        if json_data.get("live_data_watcher_load_type") == "http":
+            # Only write the data if we actually managed to fetch its contents.
+            url = json_data.get("data")
+            if url is None:
+                logger.exception("No url found in zk source JSON node.")
```
fishy: this is not in a try..exception so there's no exception in the context, thus we should not be using `logger.exception`?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +36,41 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+            return None
+        return data
+
+    @staticmethod
+    def get_data_to_write(json_data: dict, data: bytes) -> Optional[bytes]:
+        # Check if we have a JSON in this special format:
+        # data = {
+        #    "live_data_watcher_load_type": str
+        #    "data": str
+        #    "md5_hashed_data": str
```
fishy: why using md5 in 2022? also I don't see this actually checked in the code

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +36,41 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        except requests.exceptions.RequestException as e:
+            logger.exception(e)
+            return None
+        return data
+
+    @staticmethod
+    def get_data_to_write(json_data: dict, data: bytes) -> Optional[bytes]:
+        # Check if we have a JSON in this special format:
+        # data = {
+        #    "live_data_watcher_load_type": str
```
fishy, line 79: I think we actually want to support a `raw` (or whatever it's called) type in addition to `http`.

the case is that if some service actually need to sync a json file down via live data fetcher with `"live_data_watcher_load_type": "http"` field, they can wrap it again with `raw` and put the whole json blob into the top level `data` field.

baseplate/sidecars/live_data_watcher.py
KTAtkinson: I would expect that if auth was being delegated to a service account we would be using a client to connect that would deal with auth. For example, the S3 client would add auth details to a request.

Generally I would agree that application owners shouldn't need to be directly concerned with authentication to access live configurations.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: @fishy I agree. This is why I included the `live_data_watcher_load_type` key so that anyone in the future could add something like `raw`. Did you want me to support this in this PR itself? I can add that if you want. I just wanted to limit the scope to what the Experiments team needed (url), but have no issue with adding it.

baseplate/sidecars/live_data_watcher.py
fishy: service account auth should be handled automatically by the aws sdk.

but that's a good point. using `requests.get` directly will not get you the correct service account auth, you do need to use the actual aws sdk (boto) s3 client to get that.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: I should probably remove this from the comment here in the code, but the point of the `md5_hashed_data` is just for clients to know that it's a new version of the ZK file. Otherwise the contents of the ZK file will always look identical even when we re-write to it which makes debugging quite a bit more difficult. I don't think baseplate will ever care about this field.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Technically, the S3 client **could** provide auth details to a request using presigned_urls for example, but that could be dangerous if anyone happens to find the URL (they'd have access to our entire config)..

fishy: I would suggest that instead of `"live_data_watcher_load_type": "http"` we use `"s3"` instead and use boto to download the file.

SiddharthManoj: Actually I see what you both are saying.

@KTAtkinson maybe instead of `requests.get` we can actually just use the `boto3` s3 client directly. This would mean we probably want to specify the `live_data_watcher_load_type` as `S3` and NOT `url`. This way it's more secure and explicit that we're using S3 client and you can only fetch if the client has explicit access.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Updated to new approach.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Updated to new approach.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Updated with new secure flow.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        s3_client = boto3.client(
+            "s3",
+            region_name=region_name,
+        )
+        data: Optional[bytes] = None
+        # The S3 data must be Server Side Encrypted and we should have the decryption key.
+        kwargs = {
+            "Bucket": bucket_name,
+            "Key": file_key,
+            "SSECustomerKey": sse_key,
```
fishy, line 53: looking at boto's doc it's not 100% clear to me whether this "key" is supposed to be the whole secret or a "key" to look up the secret from some storage.

also looking at https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html, it seems that using SSE means the decryption actually happens on server side, so the data transferred between the wire is the already decrypted data (but of course the transfer is still encrypted by tls). I'm not sure if that's the preferred way to do that, maybe we would prefer to manage the encryption/decryption entirely on the client side?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        if json_data.get("live_data_watcher_load_type") == "S3":
+            # Only write the data if we actually managed to fetch its contents.
+            bucket_name = json_data.get("bucket_name")
+            file_key = json_data.get("file_key")
+            sse_key = json_data.get("sse_key")
+            region_name = json_data.get("region_name")
+            # We require all of these keys to properly read from S3.
+            if bucket_name is None or file_key is None or sse_key is None or region_name is None:
+                logger.debug("Missing data in live data watch zk node to read from S3.")
+                return None
```
fishy, line 94: instead of `return None`, should we fallback to just return the raw data to the caller?

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: To your first question - I verified this in our service code and it's the whole secret. In experiment-config service we upload to our bucket with this encryption so the file stored there is completely SSE encrypted.

In the console:

Screen Shot 2022-04-08 at 10 31 06 AM

I also verified that trying to decrypt this using a faulty key will not work. I received AccessDenied Error.

fishy: cc @sp3nx0r on this thread

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        # downloading the contents of encrypted files uploaded to S3.
+        if json_data.get("live_data_watcher_load_type") == "S3":
+            # Only write the data if we actually managed to fetch its contents.
+            bucket_name = json_data.get("bucket_name")
+            file_key = json_data.get("file_key")
+            sse_key = json_data.get("sse_key")
+            region_name = json_data.get("region_name")
+            # We require all of these keys to properly read from S3.
+            if bucket_name is None or file_key is None or sse_key is None or region_name is None:
+                logger.debug("Missing data in live data watch zk node to read from S3.")
```
sp3nx0r: this feels like a failure mode that should throw error instead of debug

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Also errors when trying to download it (since it's encrypted):

Screen Shot 2022-04-08 at 10 52 34 AM

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Felt weird for me to do that since we specifically expected it to be in this S3 JSON format.

If we return back this format as raw data, it seems buggy to me. I felt like doing "nothing" was better than setting the tmpfile to something that didn't make sense. Doing nothing would at least preserve the data from before?

Let me know what you think

fishy: this is basically how confident do you think this is a json file intended to be comply to this spec just wrote wrong, vs. this is some file totally not intended to be used this way just happened to have `"live_data_watcher_load_type": "S3"` in it so `json.loads` thought it's valid json and treated it this way.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Agree. Done

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Got it. I feel like "live_data_watcher_load_type": "S3" being in a JSON by coincidence feels extremely unlikely so it feels like it would be just an incorrectly written spec.

I don't feel strongly about this though - if you feel like that's possible, I think returning the raw data is fine 👯 

Let me know what you prefer overall

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        # If the load type is 'S3', this format is an indication that we support
+        # downloading the contents of encrypted files uploaded to S3.
+        if json_data.get("live_data_watcher_load_type") == "S3":
+            # Only write the data if we actually managed to fetch its contents.
+            bucket_name = json_data.get("bucket_name")
+            file_key = json_data.get("file_key")
+            sse_key = json_data.get("sse_key")
+            region_name = json_data.get("region_name")
+            # We require all of these keys to properly read from S3.
+            if bucket_name is None or file_key is None or sse_key is None or region_name is None:
```
TylerLubeck: It's a preference thing, but I'd be partial to
```py
try:
    bucket_name = json_data['bucket_name']
    file_key = json_data['file_key']
    # also the others
except KeyError:
    logger.exception("Improper configuration found while attempting to load live data from S3")
    return None
```

It's a bit cleaner, it's more "pythonic" (if we wanna fight about that term), and gives us clearer error messaging - the `logger.exception` will show the stack trace which will include the name of the missing key

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
         self.group = group
         self.mode = mode
 
+    @staticmethod
+    def get_encrypted_json_from_s3(
```
TylerLubeck: This isn't `get_encrypted_json` it's `get_encrypted_data`

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+            if bucket_name is None or file_key is None or sse_key is None or region_name is None:
+                logger.error("Missing data in live data watch zk node to read from S3.")
+                return None
+            # If we have all the correct keys, attempt to read the config from S3.
+            s3_data = NodeWatcher.get_encrypted_json_from_s3(
+                bucket_name=bucket_name, file_key=file_key, region_name=region_name, sse_key=sse_key
+            )
+            return s3_data
+        else:
+            return data
```
TylerLubeck: Can we instead do this?
```
if json_data.get(....):
    # <snip>
    data = s3_data
return data
```

baseplate/sidecars/live_data_watcher.py
```diff
@@ -50,8 +113,22 @@ def on_change(self, data: bytes, _znode_stat: ZnodeStat) -> None:
...
+                # had to be valid JSON anyways.
+                tmpfile.write(data)
+            else:
+                # If no exceptions, we have valid JSON, and can parse accordingly.
+                data_to_write = NodeWatcher.get_data_to_write(json_data, data)
+                if data_to_write is not None:
+                    tmpfile.write(data_to_write)
+                else:
+                    logger.warning(
+                        "No data written to destination node. Something is likely misconfigured."
```
TylerLubeck: nit: `s/node/file`

baseplate/sidecars/live_data_watcher.py
TylerLubeck: IMO either way we're gonna find an edge case that triggers the bug in one direction or the other - I personally agree with Sid that we should error out and make it clear that something has to change rather than silently passing data through

baseplate/sidecars/live_data_watcher.py
```diff
@@ -9,7 +10,11 @@
 from pathlib import Path
 from typing import Any
 from typing import NoReturn
+from typing import Optional
 
+import boto3  # type: ignore
```
fishy, line 15: why do we need `type: ignore` here?

requirements-transitive.txt
```diff
@@ -11,6 +11,7 @@ aspy.refactor-imports==2.1.1
 attrs==20.3.0
 Babel==2.9.1
 beautifulsoup4==4.9.3
+boto3==1.21.20
```
fishy: boto3 should be a direct dependency, not a transitive one? you are importing it directly.

requirements-transitive.txt
```diff
@@ -36,6 +37,7 @@ kombu==4.6.11
 lazy-object-proxy==1.4.3
 MarkupSafe==1.1.1
 mccabe==0.6.1
+moto==3.1.4
```
fishy, line 39: same for this one, but maybe we put test dependencies here?

fishy (commented):
the code looks good to me, left some minor comments, also I'm wondering whether @sp3nx0r would prefer to use SSE directly or manage the encryption/decryption client side.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Lint was complaining for some reason....

baseplate/sidecars/live_data_watcher.py:15: error: Skipping analyzing "boto3": found module but no type hints or library stubs

requirements-transitive.txt
SiddharthManoj: Yeah, lets keep test dependency here

requirements-transitive.txt
SiddharthManoj: I'm not really sure what the difference is. I noticed redis was imported in requirements-transitive.txt for example. Does it matter? For now I moved it to requirements.txt, but it seems to be a much smaller file. Looks like it's used for style/python stuff rather than libraries?

SiddharthManoj:
> the code looks good to me, left some minor comments, also I'm wondering whether @sp3nx0r would prefer to use SSE directly or manage the encryption/decryption client side.

By client side do you mean have every service handle this? SSE Encryption as it is in the PR uses a customer key provided by the Experiments client. It's generated fresh each time so technically it's client generated every time. The only **think** makes the most sense as it's centralized to all the listeners who probably only care about the final decrypted data

fishy:
> > the code looks good to me, left some minor comments, also I'm wondering whether @sp3nx0r would prefer to use SSE directly or manage the encryption/decryption client side.
> 
> By client side do you mean have every service handle this? SSE Encryption as it is in the PR uses a customer key provided by the Experiments client. It's generated fresh each time so technically it's client generated every time. The only **think** makes the most sense as it's centralized to all the listeners who probably only care about the final decrypted data

no I mean we handle it inside live-data-fetcher.

SiddharthManoj:
> > > the code looks good to me, left some minor comments, also I'm wondering whether @sp3nx0r would prefer to use SSE directly or manage the encryption/decryption client side.
> > 
> > 
> > By client side do you mean have every service handle this? SSE Encryption as it is in the PR uses a customer key provided by the Experiments client. It's generated fresh each time so technically it's client generated every time. The only **think** makes the most sense as it's centralized to all the listeners who probably only care about the final decrypted data
> 
> no I mean we handle it inside live-data-fetcher.

Oh got it. I think using the SSE directly had some issues when I tried it. Not 100% sure if it works, might have some IAM related problems too.

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -1,20 +1,74 @@
...
+        dest = self.output_dir.joinpath("data.txt")
+        inst = NodeWatcher(str(dest), os.getuid(), os.getgid(), 777)
+        new_content = b'{"live_data_watcher_load_type":"S3","bucket_name":"test_bucket","file_key":"test_file_key","sse_key":"test_decryption_key","region_name":"us-east-1"}'
+        expected_content = b'{"foo_encrypted": "bar_encrypted"}'
+        inst.on_change(new_content, None)
+        self.assertEqual(expected_content, dest.read_bytes())
+        self.assertEqual(dest.owner(), pwd.getpwuid(os.getuid()).pw_name)
+        self.assertEqual(dest.group(), grp.getgrgid(os.getgid()).gr_name)
+
+    def test_s3_load_type_missing_keys_on_change(self):
```
mrlevitas: this test specifically checks if "sse_key" is missing, maybe "test_s3_load_type_missing_sse_key_on_change" would be a more accurate name here?

alternatively, you could put all required keys into an array, loop through them, remove the given key from "new_content" hash, and assert failure to generalize this test for all keys.

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -1,20 +1,74 @@
...
+            inst = NodeWatcher(str(dest), os.getuid(), os.getgid(), 777)
+            new_content = b'{"live_data_watcher_load_type":"S3","bucket_name":"test_bucket","file_key":"test_file_key","region_name":"us-east-1"}'
+            expected_content = b""
+            inst.on_change(new_content, None)
+        self.assertEqual(expected_content, dest.read_bytes())
+        self.assertEqual(dest.owner(), pwd.getpwuid(os.getuid()).pw_name)
+        self.assertEqual(dest.group(), grp.getgrgid(os.getgid()).gr_name)
+        self.assertIn(
+            "WARNING:%s:No data written to destination file. Something is likely misconfigured."
+            % (logger.name),
```
mrlevitas, line 66: will it also print the exception log ("Improper configuration found while attempting to load live data from S3.") from `get_data_to_write()`?
Could also assert that.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+            s3_object = s3_client.get_object(**kwargs)
+            # Returns bytes.
+            data = s3_object["Body"].read()
+        except ClientError as error:
+            logger.exception(
+                "Failed to load from S3. Received error code %s: %s",
+                error.response["Error"]["Code"],
+                error.response["Error"]["Message"],
+            )
+            return None
```
KTAtkinson, line 66: Why does this log the error and return none, rather than raising an error?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+        #    "live_data_watcher_load_type": str
+        # }
+        # If the load type is 'S3', this format is an indication that we support
+        # downloading the contents of encrypted files uploaded to S3.
+        if json_data.get("live_data_watcher_load_type") == "S3":
+            try:
+                bucket_name = json_data["bucket_name"]
+                file_key = json_data["file_key"]
+                sse_key = json_data["sse_key"]
+                region_name = json_data["region_name"]
```
KTAtkinson, line 88: Is there any way that we can surface which one of these calls is failing in the `except` block?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -32,6 +37,64 @@ def __init__(self, dest: str, owner: int, group: int, mode: int):
...
+            except KeyError:
+                # We require all of these keys to properly read from S3.
+                logger.exception(
+                    "Improper configuration found while attempting to load live data from S3."
+                )
+                return None
+            # If we have all the correct keys, attempt to read the config from S3.
+            data = NodeWatcher.get_encrypted_data_from_s3(
```
KTAtkinson, line 96: As written, this sometimes returns None. How does the client react if it expects experiment data and finds none.

KTAtkinson (changes requested):
General observation, currently Zookeeper nodes hold the live data. When the node gets updated Zookeeper notifies all clients and they fetch the new data. Now clients will only be notified if the source of the data changes in Zookeeper. How are we ensuring that if the data gets updated in S3 it gets distributed to clients?

tests/unit/sidecars/live_data_watcher_tests.py
SiddharthManoj: Actually, i tried to do this and realized it might be difficult since in the logs, it prints the full stack trace:

`ERROR:baseplate.sidecars.live_data_watcher:Improper configuration found while attempting to load live data from S3.\nTraceback (most recent call last):\n  File "/src/baseplate/sidecars/live_data_watcher.py", line 84, in get_data_to_write\n    sse_key = json_data["sse_key"]\nKeyError: \'sse_key\''`

So we'd have to check for that full string in the test

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: logger.exception naturally will surface it. example, if it's the sse_key, it'd look like this:

`ERROR:baseplate.sidecars.live_data_watcher:Improper configuration found while attempting to load live data from S3.\nTraceback (most recent call last):\n  File "/src/baseplate/sidecars/live_data_watcher.py", line 84, in get_data_to_write\n    sse_key = json_data["sse_key"]\nKeyError: \'sse_key\''`

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: I felt that logging and error and returning None would be a more graceful way to handle this in case something went wrong. Wouldn't raising an error be a more nuclear option in this case? It seemed valid to me that if this S3 flow fails, just return None back to the on_change() function so that we don't completely crash out

SiddharthManoj:
> General observation, currently Zookeeper nodes hold the live data. When the node gets updated Zookeeper notifies all clients and they fetch the new data. Now clients will only be notified if the source of the data changes in Zookeeper. How are we ensuring that if the data gets updated in S3 it gets distributed to clients?

The ZK live data watcher should handle this for us out of the box. From experiment-config service, we write to the ZK file with the expected dictionary every time an experiment config is saved + uploaded to S3.

https://github.snooguts.net/reddit/reddit-service-experiment-config/blob/main/experiment/models/experiment.py#L218

Flow is as follows:

1. User updates an experiment
2. experiment-config service uploads the new config to S3
3. experiment-config service writes to ZK with the new decryption key + S3 bucket information
4. Step 3 triggers the live data watch
5. live_data_watcher pulls down the S3 file based on the info it sees in the ZK node (decrypts it) + writes it to the destination
6. Clients read from destination

baseplate/sidecars/live_data_watcher.py
sp3nx0r: alternative would be [client side encryption](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html). Since it's still TLS wrapped, the pull and decrypt security steps should be sufficient for what we're trying to protect here.

sp3nx0r (approved):
using S3's SSE for this exchange seems fine since the goal here is to just store it securely and not necessarily address the secure transit. since these would always be internal AWS services pulling this config data down, we'd have bigger problems if the HTTPS call is intercepted/hijacked.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: I re-used the existing logic we have for None handing now:

https://github.com/reddit/baseplate.py/blob/develop/baseplate/sidecars/live_data_watcher.py#L36

PTAL

KTAtkinson: That logic says if the content is `None` stop watching the Node. It the behavior we want?

KTAtkinson (changes requested):

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: This was the logic Baseplate was using before if None was passed into on_change(). Do we want to handle it differently?

I'm open to handling it differently as well. Another option is just to do nothing

SiddharthManoj: Actually, I changed it back. We no longer stop watching the Node if `data is None` in the new flow.

PTAL

KTAtkinson (approved):

PR #659, opened 2022-04-14
Use SequentialGeventHandler handler with KazooClient

RoganMurley:
Kazoo uses the [SequentialThreadingHandler](https://github.com/python-zk/kazoo/blob/master/kazoo/handlers/threading.py) by default. As described in the docs, this handler should not be used with gevent:

```
warning::
    Do not use :class:`SequentialThreadingHandler` with applications
    using asynchronous event loops (like gevent). Use the
    :class:`~kazoo.handlers.gevent.SequentialGeventHandler` instead.
```

Since we started patching [gevent queues](https://github.com/reddit/baseplate.py/pull/656) this handler has stopped working. We need to move to the `SequentialGeventHandler`.

### DIFF
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -2,6 +2,7 @@
 from typing import Optional
 
 from kazoo.client import KazooClient
+from kazoo.handlers.gevent import SequentialGeventHandler
 
 from baseplate.lib import config
 from baseplate.lib.secrets import SecretsStore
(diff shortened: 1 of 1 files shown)


KTAtkinson (approved):

KTAtkinson (changes requested):
Actually we probably should roll back the original change and then merge a commit which contains your patch and the original change. See #660

KTAtkinson (approved):

PR #660, opened 2022-04-14
Revert "Fix connection leaks in dependencies by monkey patching gevent queues"

KTAtkinson:
Reverts reddit/baseplate.py#656

#656 Caused users of the Kazoo client to experience errors when trying to use the `task_done` method. Reverting so that we can cut a release that is in a working state.


maeivysea (approved):

KTAtkinson:
This is no longer needed, #659 adds a fix that should solve the current issue.

This was incorrect sorry!

PR #663, opened 2022-04-27
feat(prometheus) Implement HTTP client metrics

Xaelias:
Testing this should be as simple as making sure you configure an internal / external requests client in your context as needed, and then making requests with them.

```
baseplate.configure_context(
    {
        "external": ExternalRequestsClient(client_name="ipinfo.io"),
        "internal": InternalRequestsClient(),
    },
    ...
)
```

```
with baseplate.server_context("test") as context:
    context.external.get('https://ipinfo.io/ip')
    context.external.get('https://ipinfo.io/json')
    context.external.get('https://ipinfo.io/error')
```
And any internal call you want to set up on your end.

You should see something similar to this on `/metrics`:
```
❯ curl -s localhost:6060/metrics
# HELP http_client_latency_seconds Multiprocess metric
# TYPE http_client_latency_seconds histogram
http_client_latency_seconds_sum{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true"} 0.066207297
http_client_latency_seconds_sum{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true"} 0.020237509
http_client_latency_seconds_sum{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false"} 0.018597764
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0001"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.00025"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.000625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0015625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.00390625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.009765625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0244140625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.06103515625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.152587890625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.3814697265625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.95367431640625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="2.384185791015625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="5.9604644775390625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="14.901161193847656"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true",le="+Inf"} 1.0
http_client_latency_seconds_count{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",http_success="true"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0001"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.00025"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.000625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0015625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.00390625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.009765625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.0244140625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.06103515625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.152587890625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.3814697265625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="0.95367431640625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="2.384185791015625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="5.9604644775390625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="14.901161193847656"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true",le="+Inf"} 1.0
http_client_latency_seconds_count{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",http_success="true"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.0001"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.00025"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.000625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.0015625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.00390625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.009765625"} 0.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.0244140625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.06103515625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.152587890625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.3814697265625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="0.95367431640625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="2.384185791015625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="5.9604644775390625"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="14.901161193847656"} 1.0
http_client_latency_seconds_bucket{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false",le="+Inf"} 1.0
http_client_latency_seconds_count{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",http_success="false"} 1.0
# HELP http_client_requests_total Multiprocess metric
# TYPE http_client_requests_total counter
http_client_requests_total{http_endpoint="/ip",http_method="GET",http_response_code="200",http_slug="ipinfo.io",http_success="true"} 1.0
http_client_requests_total{http_endpoint="/json",http_method="GET",http_response_code="200",http_slug="ipinfo.io",http_success="true"} 1.0
http_client_requests_total{http_endpoint="/error",http_method="GET",http_response_code="404",http_slug="ipinfo.io",http_success="false"} 1.0
# HELP http_client_active_requests Multiprocess metric
# TYPE http_client_active_requests gauge
http_client_active_requests{http_endpoint="/ip",http_method="GET",http_slug="ipinfo.io",pid="41861"} 0.0
http_client_active_requests{http_endpoint="/json",http_method="GET",http_slug="ipinfo.io",pid="41861"} 0.0
http_client_active_requests{http_endpoint="/error",http_method="GET",http_slug="ipinfo.io",pid="41861"} 0.0
```

### DIFF
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -1,5 +1,6 @@
 import base64
 import ipaddress
+import sys
 
 from typing import Any
 from typing import Optional
@@ -204,7 +205,14 @@ class BaseplateSession:
 
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        with self.span.make_child(f"{self.name}.request") as span:
+        # we cannot use a context manager anymore to create the span here as we need all the tags
+        # to be set properly before the span can start for prometheus to pick them up
+        # this is why we need to call `self.span.make_child(...)` manually and subsequently:
+        # - span.start()
+        # - span.finish()
+        span = self.span.make_child(f"{self.name}.request")
+        try:
+            span.set_tag("protocol", "http")
             span.set_tag("http.method", request.method)
             span.set_tag("http.url", request.url)
             # if a client_name was specified, use that for the prometheus http_slug otherwise
@@ -212,6 +220,7 @@ class BaseplateSession:
             span.set_tag(
                 "http.slug", self.client_name if self.client_name is not None else self.name
             )
+            span.start()
 
             self._add_span_context(span, request)
 
@@ -228,7 +237,11 @@ class BaseplateSession:
             http_status_code = response.status_code
             span.set_tag("http.status_code", http_status_code)
             span.set_tag("http.success", str(200 <= http_status_code < 400).lower())
-        return response
+            span.finish()
+            return response
+        except Exception as e:
+            span.finish(exc_info=sys.exc_info())
+            raise e
 
 
 class InternalBaseplateSession(BaseplateSession):
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -1,6 +1,5 @@
 import base64
 import ipaddress
-import sys
 
 from typing import Any
 from typing import Optional
@@ -205,23 +204,13 @@ class BaseplateSession:
 
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        # we cannot use a context manager anymore to create the span here as we need all the tags
-        # to be set properly before the span can start for prometheus to pick them up
-        # this is why we need to call `self.span.make_child(...)` manually and subsequently:
-        # - span.start()
-        # - span.finish()
-        span = self.span.make_child(f"{self.name}.request")
-        try:
-            span.set_tag("protocol", "http")
-            span.set_tag("http.method", request.method)
-            span.set_tag("http.url", request.url)
-            # if a client_name was specified, use that for the prometheus http_slug otherwise
-            # default back to the name of the current requests client in the context
-            span.set_tag(
-                "http.slug", self.client_name if self.client_name is not None else self.name
-            )
-            span.start()
-
+        tags = {
+            "protocol": "http",
+            "http.method": request.method,
+            "http.url": request.url,
+            "http.slug": self.client_name if self.client_name is not None else self.name,
+        }
+        with self.span.make_child(f"{self.name}.request").with_tags(tags) as span:
             self._add_span_context(span, request)
 
             # we cannot re-use the same session every time because sessions re-use the same
@@ -237,11 +226,7 @@ class BaseplateSession:
             http_status_code = response.status_code
             span.set_tag("http.status_code", http_status_code)
             span.set_tag("http.success", str(200 <= http_status_code < 400).lower())
-            span.finish()
-            return response
-        except Exception as e:
-            span.finish(exc_info=sys.exc_info())
-            raise e
+        return response
 
 
 class InternalBaseplateSession(BaseplateSession):
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -225,7 +225,6 @@ class BaseplateSession:
 
             http_status_code = response.status_code
             span.set_tag("http.status_code", http_status_code)
-            span.set_tag("http.success", str(200 <= http_status_code < 400).lower())
         return response
 
 
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -2,6 +2,7 @@ import base64
 import ipaddress
 
 from typing import Any
+from typing import Optional
 from typing import Type
 from typing import Union
 
@@ -107,10 +108,13 @@ class BaseplateSession:
 
     """
 
-    def __init__(self, adapter: HTTPAdapter, name: str, span: Span) -> None:
+    def __init__(
+        self, adapter: HTTPAdapter, name: str, span: Span, client_name: Optional[str] = None
+    ) -> None:
         self.adapter = adapter
         self.name = name
         self.span = span
+        self.client_name = client_name
 
     def delete(self, url: str, **kwargs: Any) -> Response:
         """Send a DELETE request.
@@ -203,6 +207,11 @@ class BaseplateSession:
         with self.span.make_child(f"{self.name}.request") as span:
             span.set_tag("http.method", request.method)
             span.set_tag("http.url", request.url)
+            # if a client_name was specified, use that for the prometheus http_slug otherwise
+            # default back to the name of the current requests client in the context
+            span.set_tag(
+                "http.slug", self.client_name if self.client_name is not None else self.name
+            )
 
             self._add_span_context(span, request)
 
@@ -216,7 +225,9 @@ class BaseplateSession:
             session.mount("https://", self.adapter)
             response = session.send(request, **kwargs)
 
-            span.set_tag("http.status_code", response.status_code)
+            http_status_code = response.status_code
+            span.set_tag("http.status_code", http_status_code)
+            span.set_tag("http.success", str(200 <= http_status_code < 400).lower())
         return response
 
 
@@ -257,15 +268,23 @@ class RequestsContextFactory(ContextFactory):
         :py:func:`http_adapter_from_config`.
     :param session_cls: The type for the actual session object to put on the
         request context.
+    :param client_name: Custom name to be emitted under the http_slug label
+        for prometheus metrics. Defaults back to session_cls.name if None
 
     """
 
-    def __init__(self, adapter: HTTPAdapter, session_cls: Type[BaseplateSession]) -> None:
+    def __init__(
+        self,
+        adapter: HTTPAdapter,
+        session_cls: Type[BaseplateSession],
+        client_name: Optional[str] = None,
+    ) -> None:
         self.adapter = adapter
         self.session_cls = session_cls
+        self.client_name = client_name
 
     def make_object_for_context(self, name: str, span: Span) -> BaseplateSession:
-        return self.session_cls(self.adapter, name, span)
+        return self.session_cls(self.adapter, name, span, client_name=self.client_name)
 
 
 class InternalRequestsClient(config.Parser):
@@ -288,9 +307,13 @@ class InternalRequestsClient(config.Parser):
 
     See :py:func:`http_adapter_from_config` for available configuration settings.
 
+    :param client_name: Custom name to be emitted under the http_slug label
+        for prometheus metrics. Defaults back to session_cls.name if None
+
     """
 
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(self, client_name: Optional[str] = None, **kwargs: Any) -> None:
+        self.client_name = client_name
         self.kwargs = kwargs
 
         if "validator" in kwargs:
@@ -317,7 +340,9 @@ class InternalRequestsClient(config.Parser):
         adapter = http_adapter_from_config(
             raw_config, prefix=f"{key_path}.", validator=validator, **self.kwargs
         )
-        return RequestsContextFactory(adapter, session_cls=InternalBaseplateSession)
+        return RequestsContextFactory(
+            adapter, session_cls=InternalBaseplateSession, client_name=self.client_name
+        )
 
 
 class ExternalRequestsClient(config.Parser):
@@ -332,11 +357,17 @@ class ExternalRequestsClient(config.Parser):
 
     See :py:func:`http_adapter_from_config` for available configuration settings.
 
+    :param client_name: Custom name to be emitted under the http_slug label
+        for prometheus metrics. Defaults back to session_cls.name if None
+
     """
 
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(self, client_name: Optional[str] = None, **kwargs: Any) -> None:
+        self.client_name = client_name
         self.kwargs = kwargs
 
     def parse(self, key_path: str, raw_config: config.RawConfig) -> RequestsContextFactory:
         adapter = http_adapter_from_config(raw_config, f"{key_path}.", **self.kwargs)
-        return RequestsContextFactory(adapter, session_cls=BaseplateSession)
+        return RequestsContextFactory(
+            adapter, session_cls=BaseplateSession, client_name=self.client_name
+        )
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -74,14 +74,14 @@ class PrometheusHTTPClientMetrics:
     def latency_seconds_metric(self, tags: Dict[str, Any]) -> Histogram:
         return http_client_latency_seconds.labels(
             http_method=tags.get("http.method", ""),
-            http_success=tags.get("http.success", ""),
+            http_success=getHTTPSuccessLabel(int(tags.get("http.status_code", "0"))),
             http_slug=tags.get("http.slug", ""),
         )
 
     def requests_total_metric(self, tags: Dict[str, Any]) -> Counter:
         return http_client_requests_total.labels(
             http_method=tags.get("http.method", ""),
-            http_success=tags.get("http.success", ""),
+            http_success=getHTTPSuccessLabel(int(tags.get("http.status_code", "0"))),
             http_response_code=tags.get("http.status_code", ""),
             http_slug=tags.get("http.slug", ""),
         )
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -46,17 +46,23 @@ http_client_active_requests_labels = [
     "http_slug",
 ]
 
+# Latency histogram of HTTP calls made by clients
+# buckets are defined above (from 100µs to ~14.9s)
 http_client_latency_seconds = Histogram(
     "http_client_latency_seconds",
     "Description of histogram",
     http_client_latency_labels,
     buckets=default_buckets,
 )
+
+# Counter counting total HTTP requests started by a given client
 http_client_requests_total = Counter(
     "http_client_requests_total",
     "Description of counter",
     http_client_requests_total_labels,
 )
+
+# Gauge showing current number of active requests by a given client
 http_client_active_requests = Gauge(
     "http_client_active_requests",
     "Description of gauge",
@@ -104,18 +110,13 @@ class PrometheusHTTPClientMetrics:
 
 def getHTTPEndpointHandler(httpURL: str) -> str:
     """
-    hhtp://localhost:9090/user/123
+    http://localhost:9090/user/123
     Path to identify the endpoint handler, may be empty.
     Ref: https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http
-    MUST NOT include the per-request value of "request fields":
-    For example, /subreddits/{subreddit_id}/posts/{post_id}, not /subreddits/1234/posts/5678.
     """
     from urllib3.util import parse_url
 
     _, _, _, _, path, _, _ = parse_url(httpURL)
-    # TODO: remove request fields. Whats the best way to do this.
-    # there is a way to get this representation from pyramid
-    # ref: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.url
     return path if path is not None else ""
 
 
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -49,7 +49,7 @@ http_client_latency_seconds = Histogram(
     "http_client_latency_seconds",
     "Latency histogram of HTTP calls made by clients",
     http_client_latency_labels,
-    buckets=default_buckets,
+    buckets=default_latency_buckets,
 )
 
 # Counter counting total HTTP requests started by a given client
@@ -71,14 +71,14 @@ class PrometheusHTTPClientMetrics:
     def __init__(self) -> None:
         pass
 
-    def latency_seconds_metric(self, tags: Dict) -> Histogram:
+    def latency_seconds_metric(self, tags: Dict[str, Any]) -> Histogram:
         return http_client_latency_seconds.labels(
             http_method=tags.get("http.method", ""),
             http_success=tags.get("http.success", ""),
             http_slug=tags.get("http.slug", ""),
         )
 
-    def requests_total_metric(self, tags: Dict) -> Counter:
+    def requests_total_metric(self, tags: Dict[str, Any]) -> Counter:
         return http_client_requests_total.labels(
             http_method=tags.get("http.method", ""),
             http_success=tags.get("http.success", ""),
@@ -86,7 +86,7 @@ class PrometheusHTTPClientMetrics:
             http_slug=tags.get("http.slug", ""),
         )
 
-    def active_requests_metric(self, tags: Dict) -> Gauge:
+    def active_requests_metric(self, tags: Dict[str, Any]) -> Gauge:
         return http_client_active_requests.labels(
             http_method=tags.get("http.method", ""),
             http_slug=tags.get("http.slug", ""),
@@ -231,14 +231,14 @@ class PrometheusHTTPServerMetrics:
     def __init__(self) -> None:
         pass
 
-    def latency_seconds_metric(self, tags: Dict[str, str]) -> Any:
+    def latency_seconds_metric(self, tags: Dict[str, Any]) -> Histogram:
         return http_server_latency_seconds.labels(
             http_method=tags.get("http.method", ""),
             http_endpoint=tags.get("http.route", ""),
             http_success=getHTTPSuccessLabel(int(tags.get("http.status_code", "0"))),
         )
 
-    def requests_total_metric(self, tags: Dict[str, str]) -> Any:
+    def requests_total_metric(self, tags: Dict[str, Any]) -> Counter:
         return http_server_requests_total.labels(
             http_method=tags.get("http.method", ""),
             http_endpoint=tags.get("http.route", ""),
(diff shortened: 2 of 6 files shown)


maeivysea:
Hey Xaelias! Took a brief look (currently on a plane for a team summit so wifi is spotty). Could you add a PR description with any details, testing instructions, etc? Thanks!

maeivysea:
Hey! Took a brief look (currently on a plane for a team summit so wifi is spotty). Could you add a PR description with any details, testing instructions, etc? Thanks!

maeivysea:
Sorry for all the duplicated comments! United wifi :shakes-fist:

maeivysea:
Hey! Took a brief look (currently on a plane for a team summit so wifi is spotty). Could you add a PR description with any details, testing instructions, etc? Thanks!

maeivysea:
Hey! Took a brief look (currently on a plane for a team summit so wifi is spotty). Could you add a PR description with any details, testing instructions, etc? Thanks!

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+    "http_response_code",
+]
+http_server_active_requests_labels = [
+    "http_method",
+    "http_endpoint",
+]
+
+http_server_latency_seconds = Histogram(
+    "http_server_latency_seconds",
+    "Description of histogram",
```
mterwill: Would you mind adding descriptions for all of these?

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+    http_server_histogram_labels,
+    buckets=default_buckets,
+)
+http_server_request_size_bytes = Histogram(
+    "http_server_request_size_bytes",
+    "Description of histogram",
+    http_server_histogram_labels,
+    buckets=default_buckets,
```
mterwill: These buckets are not suitable for measuring request/response sizes in bytes:

```python3
>>> default_buckets
[0.0001, 0.00025, 0.000625, 0.0015625, 0.00390625, 0.009765625, 0.0244140625, 0.06103515625, 0.152587890625, 0.3814697265625, 0.95367431640625, 2.384185791015625, 5.9604644775390625, 14.901161193847656]
```

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+        )
+
+    def get_latency_seconds_metric(self) -> Histogram:
+        return http_client_latency_seconds
+
+    def get_requests_total_metric(self) -> Counter:
+        return http_client_requests_total
+
+    def get_active_requests_metric(self) -> Gauge:
+        return http_client_active_requests
```
mterwill, line 102: How are these intended to be used vs. the methods above?

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+    "Description of gauge",
+    http_client_active_requests_labels,
+)
+
+
+class PrometheusHTTPClientMetrics:
+    def __init__(self) -> None:
+        pass
+
+    def latency_seconds_metric(self, tags: Dict) -> Histogram:
```
mterwill: nit: I see this follows a pattern established below for Thrift metrics, but I'm curious for context as to why `tags` are a `Dict` with keys subtly different from the label names themselves... Keyword arguments matching the names of the labels exactly would be auto-completable by an IDE and introduce less cognitive overhead. (Not suggesting to change, especially given this isn't something most people will need to interact with.)

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+    hhtp://localhost:9090/user/123
+    Path to identify the endpoint handler, may be empty.
+    Ref: https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http
+    MUST NOT include the per-request value of "request fields":
+    For example, /subreddits/{subreddit_id}/posts/{post_id}, not /subreddits/1234/posts/5678.
+    """
+    from urllib3.util import parse_url
+
+    _, _, _, _, path, _, _ = parse_url(httpURL)
+    # TODO: remove request fields. Whats the best way to do this.
```
mterwill: Since these metrics wrap a generic HTTP client, they would have no semantic understanding of what's being called (i.e. what would correspond to a request field). Would be worth confirming with the author of the internal spec to ensure we're ok with this cardinality or if we'd rather opt for something with a lower cardinality like hostname, etc. (may already be captured by 'slug' - which is also an overloaded term 😞).

baseplate/lib/prometheus_metrics.py
Xaelias: sure, can do :-)

baseplate/lib/prometheus_metrics.py
Xaelias: That... is not wrong xD I'll see what we use if anything in baseplate.go or try and find best practices.

baseplate/lib/prometheus_metrics.py
Xaelias: that part is really more for the http server side of things
I just happened to copy/paste it because it worked, and in the end we should end up with something that does this on the server side.
But you're right, on the client side this doesn't apply and will just be a regular path.

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,146 @@
...
+    def get_requests_total_metric(self) -> Counter:
+        return http_client_requests_total
+
+    def get_active_requests_metric(self) -> Gauge:
+        return http_client_active_requests
+
+
+def getHTTPEndpointHandler(httpURL: str) -> str:
+    """
+    hhtp://localhost:9090/user/123
```
KTAtkinson:
```suggestion
    http://localhost:9090/user/123
```

baseplate/lib/prometheus_metrics.py
KTAtkinson: @MelissaCole might know more about how we could do this. I'll add her to the review.

KTAtkinson (changes requested):

Xaelias:
@KTAtkinson I believe the person who is doing the HTTP server bit of this actually has a solution.
I haven't checked that it worked as expected but it should. I would direct Melissa over there at this point, I will just work around the solution (either rebase if the server part is merged first, or we can fix it when the server part gets merged)

baseplate/observers/prometheus.py
```diff
@@ -115,7 +119,80 @@ def on_child_span_created(self, span: Span) -> None:
...
+        if self.start_time is not None:
+            elapsed_ns = time.perf_counter_ns() - self.start_time
+            self.metrics.latency_seconds_metric(self.tags).observe(
+                elapsed_ns / NANOSECONDS_PER_SECOND
+            )
+
+    def on_child_span_created(self, span: Span) -> None:
+        observer: Optional[SpanObserver] = None
+        if isinstance(span, LocalSpan):
+            observer = PrometheusLocalSpanObserver()
```
mterwill: Why is it important to differentiate local spans here?

Xaelias: I haven't been able to get somebody to explain to me (in a way I understood 😅 ) what LocalSpans are used for, and which metrics we should expose for them if any. Cullen was going to setup a meeting about that I think.
So this comes from the original implementation I was working from, to answer your question :-)

mterwill: Gotcha – there's a bit of info [here](https://baseplate.readthedocs.io/en/stable/api/baseplate/index.html#baseplate.ServerSpan.make_child) that I found useful, not sure if you've seen that already.

Not sure if we need/want to instrument them by default, but should we at least implement the `on_child_span_created` function so HTTP calls nested in local spans are instrumented?

e.g. I _think_ something like this would not record metrics for the HTTP call since `PrometheusLocalSpanObserver` doesn't attach a new observer.

```python
with baseplate.server_context("test") as context:
   with context.span.make_child('my super expensive operation', local=True) as span:
      sleep(5)
      span.context.myclient.get('http://example.org')
```

baseplate/lib/prometheus_metrics.py
Xaelias: Actually this doesn't apply to my PR (I imported the server part of this by mistake... somehow). I'll forward to Kris though.

baseplate/observers/prometheus.py
```diff
@@ -115,7 +119,80 @@ def on_child_span_created(self, span: Span) -> None:
...
+        else:
+            observer = PrometheusClientSpanObserver()
+
+        observer.on_set_tag("protocol", self.protocol)
```
mterwill: I see this in the server observer code above but it seems odd to me that we'd ever be invoking the observer hook directly. Shouldn't this be called indirectly by `baseplate.clients.requests. BaseplateSession` when it invokes `span.set_tag`? If we call it in the observer, could we incorrectly propagate the protocol – e.g. a thrift RPC comes in, then calls out to an external HTTP API. The server span would be thrift protocol and the client span would be HTTP.

baseplate/lib/prometheus_metrics.py
Xaelias: The bare one should mostly not be used I guess?
It's potentially useful in tests (where you get to access the underlying object and access all the metrics at once, independent of there labels), or if you're trying to do something very custom like exposing these with labels coming from something else than `tags`?

baseplate/lib/prometheus_metrics.py
Xaelias: I think things like `http_endpoint` and `http_status` etc. are just commonly found in prometheus land.

And the tags were already defined with different names in baseplate (like `http.url`).
So it's sort of a translation layer?

baseplate/clients/requests.py
```diff
@@ -200,9 +205,22 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
 
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        with self.span.make_child(f"{self.name}.request") as span:
+        # we cannot use a context manager anymore to create the span here as we need all the tags
+        # to be set properly before the span can start for prometheus to pick them up
+        # this is why we need to call `self.span.make_child(...)` manually and subsequently:
+        # - span.start()
+        # - span.finish()
```
mterwill: I wonder if it'd be worth adding something along these lines so that we could leverage the context manager...

```diff
diff --git a/baseplate/__init__.py b/baseplate/__init__.py
index 54e7492..4e0d0e3 100644
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -592,6 +592,11 @@ class Span:
         for observer in self.observers:
             observer.on_start()
 
+    def with_tags(self, tags: Dict[str,Any]):
+        for k, v in tags:
+            self.set_tag(k, v)
+        return self
+
     def set_tag(self, key: str, value: Any) -> None:
         """Set a tag on the span.
 
```

(no idea if that's idiomatic python)

baseplate/lib/prometheus_metrics.py
mterwill: Test use case makes sense, I see that's how we're doing it for Thrift. Would usually prefer an implementation detail like this be hidden or named explicitly `get_active_requests_metric[_for_testing]` but since this isn't intended to be used directly by most developers I think it's fine 👍

tests/integration/requests_tests.py
```diff
@@ -157,3 +170,91 @@ def test_external_client_doesnt_send_headers(http_server):
...
+    # observer = TestPrometheusBaseplateObserver()
+    observer = PrometheusBaseplateObserver()
+    logger.info(f"Created observer. [type={observer.__class__.__name__}]")
+    baseplate.register(observer)
+    logger.info("Registered observer with baseplate.")
+
+    # we need to clear metrics between test runs from the parametrize annotations
+    REGISTRY._names_to_collectors["http_client_requests_total"].clear()
+    REGISTRY._names_to_collectors["http_client_latency_seconds"].clear()
+    REGISTRY._names_to_collectors["http_client_active_requests"].clear()
```
mterwill, line 199: While we have the getter methods (https://github.com/reddit/baseplate.py/pull/663#discussion_r864111819), might as well use them?

Xaelias: I don't have access to the Metrics class here.
The way we achieve this in other tests is by by creating test classes where we override a lot of things to be able to keep track of the child spans and then access their `.metrics` class variable.
Which IMO is gross, and not warranted here, since I can just query the registry.
It also lets me ensure that we don't have some rogue metrics created that we lose track of.

baseplate/clients/requests.py
Xaelias: Yeah it's similar to the builder pattern. I'm just unfamiliar enough with context managers that I didn't think to try and do it.
It seem to be working just fine though.

baseplate/__init__.py
```diff
@@ -670,6 +670,11 @@ def make_child(
         """Return a child Span whose parent is this Span."""
         raise NotImplementedError
 
+    def with_tags(self, tags: Dict[str, Any]) -> "Span":
+        for k, v in tags.items():
+            self.set_tag(k, v)
+        return self
```
mterwill: Thanks for adding this 🙏  If it's not too much of a pain what do you think about pulling this change out into its own PR + adding a test (looks like others are in `tests/unit/core_tests.py`)?

Xaelias: https://github.com/reddit/baseplate.py/pull/667

baseplate/lib/prometheus_metrics.py
```diff
@@ -17,6 +18,82 @@
...
+http_client_active_requests_labels = [
+    "http_method",
+    "http_slug",
+]
+
+# Latency histogram of HTTP calls made by clients
+# buckets are defined above (from 100µs to ~14.9s)
+http_client_latency_seconds = Histogram(
+    "http_client_latency_seconds",
+    "Description of histogram",
```
KTAtkinson: Should this be a description of this histogram?

baseplate/observers/prometheus.py
```diff
@@ -109,13 +113,84 @@ def on_child_span_created(self, span: Span) -> None:
...
+    @property
+    def protocol(self) -> str:
+        return self.tags.get("protocol", "unknown")
+
+    def set_metrics_by_protocol(self) -> None:
+        if self.protocol == "http":
+            self.metrics = PrometheusHTTPClientMetrics()
+        elif self.protocol == "thrift":
+            # self.metrics = PrometheusThriftClientMetrics()
+            logger.warning("PrometheusThriftClientMetrics not implemented.")
```
KTAtkinson: Will this spam thift application logs when prometheus metrics aren't setup yet? How often will this message be logged?

docs/api/baseplate/clients/requests.rst
```diff
@@ -31,6 +31,11 @@ declaration to your context configuration::
       }
    )
 
+You can specify an optional `client_name` in any requests client. When
+prometheus metrics are emitted, this name will be used for the label
+`http_slug`. If `client_name` is not manually specified, we default back
+to the name of the client within the context (in this case, `foo` and `bar`).
+
```
KTAtkinson:
```suggestion
Prometheus metrics include the `http_slug` label, by default the label
value is set to the key name provided in the context. In the provided code 
example the  `http_slug` labels will have the values `foo` and `bar`. If you 
would like to change the value of this label you can provide the 
`client_name` keyword argument when you create the requests client. 
```

KTAtkinson (changes requested):
I'm a little concerned that this is going to increase the log volume of Thrift applications. Would you mind describing when each of the log condition happens and how often it happens in the server lifecycle?

baseplate/observers/prometheus.py
Xaelias: Yes technically but we will merge everything together and cut a release from there. So I don't believe this should be a problem.

Xaelias: Updated them to debug since I was in there

KTAtkinson (approved):

baseplate/observers/prometheus.py
Xaelias: Thanks for bringing that up. It would technically have been fixed when merging the local span implementation... but this is probably better to implement here.

mterwill: Thanks for adding this 👍

baseplate/clients/requests.py
```diff
@@ -216,7 +223,9 @@ def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
             session.mount("https://", self.adapter)
             response = session.send(request, **kwargs)
 
-            span.set_tag("http.status_code", response.status_code)
+            http_status_code = response.status_code
+            span.set_tag("http.status_code", http_status_code)
+            span.set_tag("http.success", str(200 <= http_status_code < 400).lower())
```
mterwill: For consistency, let's mirror the HTTP server metrics and infer success from the status code rather than setting it as a tag here.

https://github.com/reddit/baseplate.py/blob/a94f21f7d4d229757572d2f2fc714b543f9a386e/baseplate/lib/prometheus_metrics.py#L203-L207

baseplate/observers/prometheus.py
```diff
@@ -121,15 +127,98 @@ def on_child_span_created(self, span: Span) -> None:
...
+        if self.protocol == "http":
+            self.metrics = PrometheusHTTPClientMetrics()
+        elif self.protocol == "thrift":
+            # self.metrics = PrometheusThriftClientMetrics()
+            logger.debug("PrometheusThriftClientMetrics not implemented.")
+        else:
+            logger.warning(
+                "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",
+                self.protocol,
+            )
```
mterwill: This log and the ones below probably don't warrant a 'warning' level, especially given that multiple are emitted on every span – perhaps 'debug' instead?

baseplate/observers/prometheus.py
```diff
@@ -121,15 +127,98 @@ def on_child_span_created(self, span: Span) -> None:
...
+        if self.metrics is None:
+            logger.warning(
+                "No metrics set for Prometheus metric collection. Metrics will not be exported correctly."
+            )
+            return
+
+        self.tags["success"] = "true"
+        if exc_info is not None:
+            self.tags["exception_type"] = exc_info[1].__class__.__name__
+            self.tags["success"] = "false"
```
mterwill: Where is this used?

baseplate/observers/prometheus.py
```diff
@@ -121,15 +127,98 @@ def on_child_span_created(self, span: Span) -> None:
...
+        span.register(observer)
 
 
 class PrometheusLocalSpanObserver(SpanObserver):
     def __init__(self) -> None:
         logger.debug("PrometheusLocalSpanObserver not implemented")
+
+    # Proper implementation for PrometheusLocalSpanObserver will come in a future PR
+    # In the meantime, we need this logic in place to be able to emit http client metrics
+    # when running inside a local span
```
mterwill, line 222: 👍

docs/api/baseplate/clients/requests.rst
```diff
@@ -31,6 +31,12 @@ declaration to your context configuration::
       }
    )
 
+Prometheus metrics include the `http_slug` label, by default the label
+value is set to the key name provided in the context. In the provided code
```
mterwill:
```suggestion
value is set to the key name provided in the context. In the above code
```

mterwill (dismissed):
A few small comments then I think this is looking good!

Aside: I think a lot of the complexity of this implementation could have been sidestepped by using [hooks](https://docs.python-requests.org/en/latest/user/advanced/#event-hooks) instead of the Baseplate observer/span system. This approach does keep consistency between protocols (thrift/http) and client/server though, for better or worse.

baseplate/observers/prometheus.py
Xaelias: Huh I had switched them all back to debug. Might have been lost during the rebase...

baseplate/clients/requests.py
Xaelias: I kinda disagree. The success is not a prometheus thing. Either the request is successful or it's not. No need to do this multiple times per request (for each metric) when the client should be responsible to tell us whether the query is successful or not.

baseplate/observers/prometheus.py
Xaelias: Oh nvm I'm confused.
No these are bad. They shouldn't happen. We should always have a valid protocol as far as I know.

baseplate/observers/prometheus.py
Xaelias: I'm not 100% sure. We seem to have it everywhere...

baseplate/clients/requests.py
mterwill: Then can you add tests that cover this tag is set correctly please? `getHTTPSuccessLabel` is already tested.

baseplate/observers/prometheus.py
mterwill: If we don't need it, let's remove it.

krav: We use this for Thrift (only), and it shares this code.

maeivysea (approved):

nsheaps (approved):

baseplate/lib/prometheus_metrics.py
```diff
@@ -26,6 +26,82 @@
...
+        )
+
+    def get_latency_seconds_metric(self) -> Histogram:
+        return http_client_latency_seconds
+
+    def get_requests_total_metric(self) -> Counter:
+        return http_client_requests_total
+
+    def get_active_requests_metric(self) -> Gauge:
+        return http_client_active_requests
```
nsheaps, line 102: I don't disagree with @mterwill 's comment here since I had the same comment on the others. Can we add some pydoc in the next change that says these are deprecated? We can worry about removing them later, but lets encourage folks to start using the other ones.

PR #665, opened 2022-05-02
Add prometheus HTTP server metrics

krav:
This is ~most of the work of~ OBS-49 ~what's missing is `http_server_request_size_bytes` and `http_server_response_size_bytes` histograms which will come in its own PR~. 

## 💸 TL;DR

This adds HTTP server metrics for Prometheus. 

## 📜 Details
Goals are detailed in https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md

To manually test create a pyramid endpoint with a route, e.g `configurator.add_route("hello_world", "hellworld/{world}/edit", request_method="GET")`.

Expected output on `:6060/metrics`: 
```
# HELP http_server_active_requests Multiprocess metric
# TYPE http_server_active_requests gauge
http_server_active_requests{http_endpoint="/",http_method="GET",pid="83477"} 0.0
[snip snip snip]
/edit",http_method="GET",http_success="true",le="2.384185791015625"} 2.0
http_server_latency_seconds_bucket{http_endpoint="hellworld/{world}/edit",http_method="GET",http_success="true",le="5.9604644775390625"} 2.0
http_server_latency_seconds_bucket{http_endpoint="hellworld/{world}/edit",http_method="GET",http_success="true",le="14.901161193847656"} 2.0
http_server_latency_seconds_bucket{http_endpoint="hellworld/{world}/edit",http_method="GET",http_success="true",le="+Inf"} 2.0
http_server_latency_seconds_count{http_endpoint="hellworld/{world}/edit",http_method="GET",http_success="true"} 2.0
```

## ✅ Checks


baseplate/lib/prometheus_metrics.py
```diff
@@ -88,3 +89,106 @@ def get_requests_total_metric(self) -> Counter:
...
+    "http_response_code",
+]
+http_server_active_requests_labels = [
+    "http_method",
+    "http_endpoint",
+]
+
+http_server_latency_seconds = Histogram(
+    "http_server_latency_seconds",
+    "Description of histogram",
```
KTAtkinson: Should this be a description of this histogram?

baseplate/lib/prometheus_metrics.py
```diff
@@ -88,3 +89,106 @@ def get_requests_total_metric(self) -> Counter:
...
+    "Description of gauge",
+    http_server_active_requests_labels,
+)
+
+
+class PrometheusHTTPServerMetrics:
+    def __init__(self) -> None:
+        pass
+
+    def latency_seconds_metric(self, tags: Dict[str, str]) -> Any:
```
KTAtkinson, line 158: In what context will these functions be used?

KTAtkinson (changes requested):

baseplate/lib/prometheus_metrics.py
krav: They're called from `observers/prometheus.py` in `on_finish`. It's not in this PR because it was already merged with the Thrift implementation. 

If we end up keeping this abstraction, we should in the future make it clear that this is implementing an interface.

baseplate/lib/prometheus_metrics.py
krav: It should indeed :sweat_smile:

baseplate/lib/prometheus_metrics.py
```diff
@@ -88,3 +89,106 @@ def get_requests_total_metric(self) -> Counter:
...
+    "Size of incoming requests in bytes",
+    http_server_histogram_labels,
+    buckets=default_buckets,
+)
+http_server_response_size_bytes = Histogram(
+    "http_server_response_size_bytes",
+    "Size of outgoing responses in bytes",
+    http_server_histogram_labels,
+    buckets=default_buckets,
+)
```
mterwill: Mentioned in #663 but these buckets aren't suitable for measuring request/response size in bytes

krav: Indeed, they're not being called at the moment. Taking them out entirely and will put them back in a separate PR.

baseplate/observers/prometheus.py
```diff
@@ -60,7 +61,7 @@ def set_metrics_by_protocol(self) -> None:
         if self.protocol == "thrift":
             self.metrics = PrometheusThriftServerMetrics()
         elif self.protocol == "http":
-            logger.warning("HTTP Prometheus metrics not implemented")
+            self.metrics = PrometheusHTTPServerMetrics()
```
mterwill, line 64: Possible to add a couple of tests? Otherwise looking good to me 👍

krav: Added a unit test, an integration test is coming: https://github.com/krav/baseplate.py/compare/k/http-server-metrics...krav:k/http-server-request-metrics?expand=1

baseplate/lib/prometheus_metrics.py
```diff
@@ -88,3 +89,81 @@ def get_requests_total_metric(self) -> Counter:
...
+
+    def get_active_requests_metric(self) -> Any:
+        return http_server_active_requests
+
+
+def getHTTPSuccessLabel(httpStatusCode: int) -> str:
+    """
+    The HTTP success label is "true" if the status code is 2xx or 3xx, "false" otherwise.
+    """
+    return str(httpStatusCode >= 200 < 400).lower()
```
mterwill: Could you add a unit test for this too please?

baseplate/lib/prometheus_metrics.py
KTAtkinson: Can you provide a code sample? This naming just seems really odd to me and I'm curious if we can come up with a better name.

KTAtkinson (changes requested):

baseplate/lib/prometheus_metrics.py
krav: Thanks, that was wrong :sweat_smile: Changed it without thinking because the linter didn't like it being two expressions. Added tests.

baseplate/observers/prometheus.py
krav: Actually, just putting that in this PR since it's complete.

krav:
@KTAtkinson 
> Can you provide a code sample? This naming just seems really odd to me and I'm curious if we can come up with a better name.

It's called from here https://github.com/reddit/baseplate.py/blob/dbf57a5333343df70f9ed90cad2533fd81618e89/baseplate/observers/prometheus.py#L101-L103
and named that way because that's what the Thrift part is called, which shares this interface and is already merged https://github.com/reddit/baseplate.py/blob/192bd4c0bbf2052f5f416c135e506e4ab04667c6/baseplate/lib/prometheus_metrics.py#L57-L62

The name of the metric "\<protocol>_{server,client}_latency_seconds" is idiomatic Prometheus naming (it's the latency, and the unit is seconds). The thinking behind this function name is that it fetches that particular metric from the registry, but I agree that it ends up looking kinda weird.

baseplate/lib/prometheus_metrics.py
```diff
@@ -15,9 +15,16 @@
...
-default_buckets = [start * factor ** i for i in range(count)]
+default_latency_buckets = [start * factor**i for i in range(count)]
 
 
+default_size_start = 8
+default_size_factor = 2
+default_size_count = 20
+default_size_buckets = [
+    default_size_start * default_size_factor**i for i in range(default_size_count)
+]
```
mterwill: Worth adding a comment like above saying what these end up being?

tests/unit/observers/prometheus_tests.py
```diff
@@ -95,3 +111,8 @@ def test_create_server_span(self):
         mock_server_span.set_tag("protocol", "thrift")
 
         self.assertEqual(mock_server_span.register.call_count, 1)
+
+    def test_http_success_label(self):
+        self.assertEqual("true", getHTTPSuccessLabel(222))
+        self.assertEqual("false", getHTTPSuccessLabel(111))
+        self.assertEqual("false", getHTTPSuccessLabel(418))
```
mterwill: We should probably test the edge cases here too... i.e. 200 and 400

mterwill (approved):
Thanks for working through this with me!

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -239,8 +241,10 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         )
         span.set_tag("protocol", "http")
         span.set_tag("http.url", request.url)
+        span.set_tag("http.route", request.matched_route.pattern)
         span.set_tag("http.method", request.method)
         span.set_tag("peer.ipv4", request.remote_addr)
+        span.set_tag("http.request_length", request.content_length)
```
mterwill, line 247: I think we might need to handle a case where this is `None`:

```
  File "/usr/local/Cellar/python@3.9/3.9.12/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/usr/local/Cellar/python@3.9/3.9.12/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 183, in finish_response
    for data in self.result:
  File "/Users/matt.terwilliger/src/baseplate.py/baseplate/frameworks/pyramid/__init__.py", line 60, in __next__
    self.span.finish()
  File "/Users/matt.terwilliger/src/baseplate.py/baseplate/__init__.py", line 644, in finish
    observer.on_finish(exc_info)
  File "/Users/matt.terwilliger/src/baseplate.py/baseplate/observers/prometheus.py", line 113, in on_finish
    self.metrics.request_size_bytes_metric(self.tags).observe(
  File "/usr/local/lib/python3.9/site-packages/prometheus_client/metrics.py", line 597, in observe
    self._sum.inc(amount)
  File "/usr/local/lib/python3.9/site-packages/prometheus_client/values.py", line 22, in inc
    self._value += amount
TypeError: unsupported operand type(s) for +=: 'float' and 'NoneType'
```

in testing with a `GET` request that has no body

krav: Oof, `({"abc": None}).get("abc", 0)` really is None 😵‍💫 . Fixed.

Can you share how you got this to throw? I thought an empty body was covered by the test, so I want to try and fix that.

tests/unit/observers/prometheus_tests.py
krav: Added!

baseplate/frameworks/pyramid/__init__.py
mterwill: I noticed it when I started up a real server and tried to make a `GET` request (which doesn't have the `Content-Length` header set). Guessing the behavior `webtest` differs subtly from a real WSGI server 😕

mterwill (approved):

krav:
@KTAtkinson I'm blocked from merging since you're marked as requesting changes

KTAtkinson (approved):

PR #666, opened 2022-05-03
Add an s3fetcher extras component

TylerLubeck:
Optionally enable the s3 live-data-fetcher base on package installation, and configure an extras package to explicitly install the package

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ extras_require = {
     "redis-py-cluster": ["redis-py-cluster>=2.1.2,<3.0.0"],
     "refcycle": ["objgraph>=3.0,<4.0"],
     "requests": ["advocate>=1.0.0,<2.0"],
+    "s3fetcher": ["boto3>=1.18"],
     "sentry": ["sentry-sdk>=0.19,<1.0"],
     "sql": ["sqlalchemy>=1.1.0,<2.0"],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -16,6 +16,7 @@
     "redis-py-cluster": ["redis-py-cluster>=2.1.2,<3.0.0"],
     "refcycle": ["objgraph>=3.0,<4.0"],
     "requests": ["advocate>=1.0.0,<2.0"],
+    "s3fetcher": ["boto3>=1.18"],
```
TylerLubeck, line 19: ngl this is a bit arbitrary. We're using pretty basic s3 features and I scanned back through the changelog to see if there were notable events. We tested with 1.20

SiddharthManoj: nit: do we wanna call this `s3-fetcher` instead?

SiddharthManoj (commented):
Do you need to add an ignore_missing_imports here (example):

https://github.com/reddit/baseplate.py/commit/81c477ffddb64b880bec6373822055e96cbb4479#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52

nsheaps (approved):
I have general questions about pulling from S3 as part of the live data watcher but nothing specifically about this is introducing bad stuff.

PR #667, opened 2022-05-03
feat(Span): Allow span to be created with tags in a context manager

Xaelias:
This is relevant to https://github.com/reddit/baseplate.py/pull/663.

As the title says, the goal is to be able to define Spans with a list of tags before the Span itself is started.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -670,6 +670,20 @@ class Span:
         """Return a child Span whose parent is this Span."""
         raise NotImplementedError
 
+    def with_tags(self, tags: Dict[str, Any]) -> "Span":
+        """Declare a set of tags to be added to a span before starting it in the context manager.
+
+        Can be used as follow:
+        tags = {...}
+        with self.span.make_child("...").with_tags(tags) as span:
+            ...
+
+        :param tags: Dict of tags to be set on the span at creation time.
+        `"""
+        for k, v in tags.items():
+            self.set_tag(k, v)
+        return self
+
 
 class ParentSpanAlreadyFinishedError(Exception):
     def __init__(self) -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/core_tests.py
```diff
@@ -152,6 +152,23 @@ def test_context(self):
...
+
+        tags = {
+            "k1": "v1",
+            "k2": "v2",
+        }
+        with span.with_tags(tags) as span:
+            self.assertEqual(mock_observer.on_start.call_count, 1)
+            mock_observer.on_set_tag.assert_has_calls(
+                [mock.call(k, v) for k, v in tags.items()], any_order=True
+            )
```
mterwill: 🔕 nit: I usually tend towards being explicit with this kind of stuff in tests, accidentally checked that `None == None` one too many times 🤷‍♂️ 😅 

```suggestion
            mock_observer.on_set_tag.assert_has_calls([
                mock.call("k1", "v1"),
                mock.call("k2", "v2"),
            ])
```

mterwill (approved):
Thank you!

tests/unit/core_tests.py
Xaelias: Updated :-) Thanks for calling me out on my laziness :p

mterwill (approved):

baseplate/__init__.py
```diff
@@ -670,6 +670,20 @@ def make_child(
         """Return a child Span whose parent is this Span."""
         raise NotImplementedError
 
+    def with_tags(self, tags: Dict[str, Any]) -> "Span":
+        """This allows us to set tags in a Span before starting it in a context manager.
```
KTAtkinson: Please write documentation in an authoritative voice:
```suggestion
        """Declare a set of tags to be added to a span before starting it in the context manager.
```

KTAtkinson (changes requested):

KTAtkinson (approved):

PR #668, opened 2022-05-05
Make Kazoo work without gevent (for example in sidecars)

RoganMurley:
## 💸 TL;DR
KazooClient needs to be initialized differently depending on whether we are using gevent or not. We don't use gevent in sidecars so need to make sure KazooClient works without gevent.

## 📜 Details
In #659 we updated Kazoo to use `SequentialGeventHandler` for compatibility with gevent. Unfortunately this handler silently fails if we are not using gevent.

Sidecars do _not_ use gevent as they don't use entrypoint scripts that patch (such as `baseplate-serve`), instead being ran directly (`python -m baseplate.sidecars.live_data_watcher some_config.ini`). I'm not sure if this is intentional or an oversight.

This PR detects if gevent patching has occurred and initializes KazooClient accordingly.

* This change has been tested on an internal service that was experiencing the above issue when updating to the latest baseplate version.

## ✅ Checks

### DIFF
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -1,4 +1,5 @@
 """Helpers for interacting with ZooKeeper."""
+import time
 from typing import Optional
 
 import gevent
@@ -8,6 +9,7 @@ from kazoo.handlers.gevent import SequentialGeventHandler
 
 from baseplate.lib import config
 from baseplate.lib.secrets import SecretsStore
+from baseplate.server.monkey import is_gevent_patched
 
 
 def zookeeper_client_from_config(
@@ -56,12 +58,22 @@ def zookeeper_client_from_config(
         credentials = secrets.get_simple(path)
         auth_data.append(("digest", credentials.decode("utf8")))
 
+    # Kazoo requires different parameters depending on whether
+    # we are using gevent or not.
+    if is_gevent_patched():
+        handler = SequentialGeventHandler()
+        sleep_func = gevent.sleep
+    else:
+        handler = None
+        sleep_func = time.sleep
+
+
     return KazooClient(
         cfg.hosts,
         timeout=cfg.timeout.total_seconds(),
         auth_data=auth_data,
         read_only=read_only,
-        handler=SequentialGeventHandler(),
+        handler=handler,
         # this retry policy tells Kazoo how often it should attempt connections
         # to ZooKeeper from its worker thread/greenlet. when the connection is
         # lost during normal operation (i.e. after it was first established)
@@ -83,6 +95,6 @@ def zookeeper_client_from_config(
             backoff=2,  # exponential backoff
             max_jitter=1,  # maximum amount to jitter sleeptimes
             max_delay=60,  # never wait longer than this
-            sleep_func=gevent.sleep,
+            sleep_func=sleep_func,
         ),
     )
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -1,5 +1,6 @@
 """Helpers for interacting with ZooKeeper."""
 import time
+
 from typing import Optional
 
 import gevent
@@ -66,7 +67,6 @@ def zookeeper_client_from_config(
     else:
         handler = None
         sleep_func = time.sleep
-
 
     return KazooClient(
         cfg.hosts,
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -3,8 +3,6 @@ import time
 
 from typing import Optional
 
-import gevent
-
 from kazoo.client import KazooClient
 from kazoo.handlers.gevent import SequentialGeventHandler
 
@@ -59,14 +57,11 @@ def zookeeper_client_from_config(
         credentials = secrets.get_simple(path)
         auth_data.append(("digest", credentials.decode("utf8")))
 
-    # Kazoo requires different parameters depending on whether
-    # we are using gevent or not.
+    # Kazoo needs to use a different handler with gevent.
     if gevent_is_patched():
         handler = SequentialGeventHandler()
-        sleep_func = gevent.sleep
     else:
         handler = None
-        sleep_func = time.sleep
 
     return KazooClient(
         cfg.hosts,
@@ -95,6 +90,6 @@ def zookeeper_client_from_config(
             backoff=2,  # exponential backoff
             max_jitter=1,  # maximum amount to jitter sleeptimes
             max_delay=60,  # never wait longer than this
-            sleep_func=sleep_func,
+            sleep_func=time.sleep,
         ),
     )
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -10,7 +10,7 @@ from kazoo.handlers.gevent import SequentialGeventHandler
 
 from baseplate.lib import config
 from baseplate.lib.secrets import SecretsStore
-from baseplate.server.monkey import is_gevent_patched
+from baseplate.server.monkey import gevent_is_patched
 
 
 def zookeeper_client_from_config(
@@ -61,7 +61,7 @@ def zookeeper_client_from_config(
 
     # Kazoo requires different parameters depending on whether
     # we are using gevent or not.
-    if is_gevent_patched():
+    if gevent_is_patched():
         handler = SequentialGeventHandler()
         sleep_func = gevent.sleep
     else:
--- a/baseplate/lib/live_data/zookeeper.py
+++ b/baseplate/lib/live_data/zookeeper.py
@@ -1,6 +1,4 @@
 """Helpers for interacting with ZooKeeper."""
-import time
-
 from typing import Optional
 
 from kazoo.client import KazooClient
@@ -90,6 +88,5 @@ def zookeeper_client_from_config(
             backoff=2,  # exponential backoff
             max_jitter=1,  # maximum amount to jitter sleeptimes
             max_delay=60,  # never wait longer than this
-            sleep_func=time.sleep,
         ),
     )
--- a/baseplate/server/monkey.py
+++ b/baseplate/server/monkey.py
@@ -13,7 +13,6 @@ def patch_stdlib_queues() -> None:
     monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
 
 
-def is_gevent_patched():
-    """Has any gevent patching has occurred?
-    """
+def is_gevent_patched() -> bool:
+    """Has any gevent patching has occurred?"""
     return monkey.saved
--- a/baseplate/server/monkey.py
+++ b/baseplate/server/monkey.py
@@ -13,6 +13,6 @@ def patch_stdlib_queues() -> None:
     monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
 
 
-def is_gevent_patched() -> bool:
+def gevent_is_patched() -> bool:
     """Has any gevent patching has occurred?"""
-    return monkey.saved
+    return bool(monkey.saved)
--- a/baseplate/server/monkey.py
+++ b/baseplate/server/monkey.py
@@ -14,5 +14,5 @@ def patch_stdlib_queues() -> None:
 
 
 def gevent_is_patched() -> bool:
-    """Has any gevent patching has occurred?"""
+    """Returns true if gevent is patched, otherwise returns false."""
     return bool(monkey.saved)
--- a/baseplate/server/monkey.py
+++ b/baseplate/server/monkey.py
@@ -1,4 +1,4 @@
-from gevent.monkey import patch_module
+from gevent import monkey
 
 
 def patch_stdlib_queues() -> None:
@@ -10,4 +10,10 @@ def patch_stdlib_queues() -> None:
     import queue
     import gevent.queue
 
-    patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+    monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+
+
+def is_gevent_patched():
+    """Has any gevent patching has occurred?
+    """
+    return monkey.saved
--- a/tests/integration/live_data/zookeeper_tests.py
+++ b/tests/integration/live_data/zookeeper_tests.py
@@ -63,7 +63,7 @@ class ZooKeeperHandlerWithPatchingTests(unittest.TestCase):
         assert isinstance(client.handler, SequentialThreadingHandler)
 
     @mock.patch("baseplate.lib.live_data.zookeeper.gevent_is_patched", return_value=True)
-    def test_create_client_uses_gevent_handler_when_gevent_patched(self):
+    def test_create_client_uses_gevent_handler_when_gevent_patched(self, mock_gevent_is_patched):
         patched_default_values = tuple(
             gevent.sleep if value is time.sleep else value
             for value in KazooRetry.__init__.__defaults__
--- a/tests/integration/live_data/zookeeper_tests.py
+++ b/tests/integration/live_data/zookeeper_tests.py
@@ -3,6 +3,8 @@ import unittest
 
 from unittest import mock
 
+import gevent.socket
+
 try:
     from kazoo.exceptions import NoNodeError
     from kazoo.handlers.gevent import SequentialGeventHandler
@@ -12,8 +14,6 @@ except ImportError:
 
 from baseplate.lib.live_data.zookeeper import zookeeper_client_from_config
 from baseplate.lib.secrets import SecretsStore
-
-import gevent
 
 from .. import get_endpoint_or_skip_container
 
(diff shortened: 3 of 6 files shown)


SiddharthManoj:
I verified that this resolves the issue of the ZK source file not triggering [live updates](https://github.com/reddit/baseplate.py/blob/develop/baseplate/sidecars/live_data_watcher.py#L118) (`on_change`). I verified locally by doing the following:

Creds to @TylerLubeck for tips on how to spin up a local testing env as follows:

1. Make a venv

2. Inside the venv 
`pip3 install -e .[zookeeper,s3fetcher]`

3. make a config file (example below) live-data.ini in the directory:

```
[app:main]
factory = helloworld:make_wsgi_app

metrics.tagging = true
metrics.log_if_unconfigured = true

db.url = sqlite:///

[live-data]
zookeeper.hosts = localhost:2181

nodes.a.source = /data-three
nodes.a.dest = /tmp/data-three

[server:main]
```

4. run `docker run --name zk -p 2181:2181 zookeeper`

5. create a znode `docker exec zk zkCli.sh create /data`

6. update the znode (do this repeatedly) `docker exec zk zkCli.sh set /data new-data-or-whatever-you-want`

7. Change permissions for secrets:  `sudo mkdir -p /var/local;  sudo chown -R ${USER}:staff /var/local; echo "{}" > /var/local/secrets.json`

8. Run the live-data-watcher `python -m baseplate.sidecars.live_data_watcher live-data.ini`

SiddharthManoj (approved):

baseplate/lib/live_data/zookeeper.py
```diff
@@ -56,12 +59,21 @@ def zookeeper_client_from_config(
         credentials = secrets.get_simple(path)
         auth_data.append(("digest", credentials.decode("utf8")))
 
+    # Kazoo requires different parameters depending on whether
+    # we are using gevent or not.
+    if is_gevent_patched():
```
KTAtkinson: This should probably be called `gevent_is_patched`

baseplate/server/monkey.py
```diff
@@ -10,4 +10,9 @@ def patch_stdlib_queues() -> None:
     import queue
     import gevent.queue
 
-    patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+    monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+
+
+def is_gevent_patched() -> bool:
+    """Has any gevent patching has occurred?"""
```
KTAtkinson: Please also describe what the return value is for this function.

KTAtkinson (changes requested):
The PR description says that you intend to use this for sidecars so that they don't try to uses gevent, but I don't see any code related to updating how the sidecars work. Is this intended?

baseplate/server/monkey.py
RoganMurley: I've added a cast to make sure that it's actually a `bool`.

Do I need to add a return type to the docstring even though I have a `bool` type annotation on the function?

RoganMurley:
> The PR description says that you intend to use this for sidecars so that they don't try to uses gevent, but I don't see any code related to updating how the sidecars work. Is this intended?

To clarify, sidecars currently do **not** use gevent because they aren't run with an entrypoint script that does gevent patching ([see docs](https://baseplate.readthedocs.io/en/stable/api/baseplate/lib/live_data.html#watcher-daemon)). This surprised me and I think that there's a discussion to be had about whether to use gevent with them, but that would be a bigger change.

This PR detects if gevent patching has occurred and chooses the correct Kazoo handler accordingly. This means that the KazooClient can be safely used in gevent environments (`baseplate-serve`, `baseplate-script`) and non-gevent environments (`sidecars`).

baseplate/lib/live_data/zookeeper.py
```diff
@@ -56,12 +59,21 @@ def zookeeper_client_from_config(
...
         auth_data.append(("digest", credentials.decode("utf8")))
 
+    # Kazoo requires different parameters depending on whether
+    # we are using gevent or not.
+    if gevent_is_patched():
+        handler = SequentialGeventHandler()
+        sleep_func = gevent.sleep
+    else:
+        handler = None
+        sleep_func = time.sleep
```
nsheaps: If gevent does it's [patching this becomes gevent.sleep anyway](https://www.gevent.org/api/gevent.monkey.html#gevent.monkey.patch_time). Do we need to explicitly use one or the other here or can we just use time.sleep always?

baseplate/lib/live_data/zookeeper.py
```diff
@@ -56,12 +59,21 @@ def zookeeper_client_from_config(
         credentials = secrets.get_simple(path)
         auth_data.append(("digest", credentials.decode("utf8")))
 
+    # Kazoo requires different parameters depending on whether
+    # we are using gevent or not.
+    if gevent_is_patched():
+        handler = SequentialGeventHandler()
```
nsheaps, line 60: This feels like something we either want or don't want all the time regardless of if gevent has patched items

baseplate/server/monkey.py
```diff
@@ -10,4 +10,9 @@ def patch_stdlib_queues() -> None:
     import queue
     import gevent.queue
 
-    patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+    monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+
+
+def gevent_is_patched() -> bool:
+    """Has any gevent patching has occurred?"""
+    return bool(monkey.saved)
```
nsheaps, line 18: **any** doesn't necessarily mean that it has patched what we want. What exactly does "use gevent and this works or don't and it silently fails" relate back to the usage of?

nsheaps (changes requested):
If patching occurs then in theory you shouldn't need to check, you can just use the native handlers.

That said, I feel like we should be committing to gevent because that's what we want, or fall back on native handling because we want patching to replace it when it happens.

I think I need a bit more info on what the silent failure mode you described in the PR description is before I have enough context to thoughtfully approve this.

baseplate/lib/live_data/zookeeper.py
RoganMurley: You're right, this can be removed!

I added it because of a failing test, but in retrospect that was because the sleep function hadn't been patched in that test.

baseplate/lib/live_data/zookeeper.py
RoganMurley: Kazoo intentionally eschews monkey patching, see the documentation below. It expects you to explicitly choose the handler yourself if you use gevent.

https://kazoo.readthedocs.io/en/latest/async_usage.html#connection-handling
```
Kazoo doesn’t rely on gevents/eventlet monkey patching and requires that you pass in the appropriate handler, the default handler is SequentialThreadingHandler
```

Kazoo docs have pretty strong guidance about not using `SequentialThreadingHandler` with gevent.

https://kazoo.readthedocs.io/en/2.6.1/api/handlers/threading.html?highlight=sequentialthreadinghandler#kazoo-sasl-handlers-threading
```
Do not use SequentialThreadingHandler with applications using asynchronous event loops (like gevent). Use the SequentialGeventHandler instead.
```

My understanding from reading the [code](https://kazoo.readthedocs.io/en/2.6.1/_modules/kazoo_sasl/handlers/threading.html#SequentialThreadingHandler) and [GitHub issues](https://github.com/python-zk/kazoo/issues/3#issuecomment-7102481) is that the Zookeeper C client library is not patched by gevent and so [spawns a separate thread](https://gevent.narkive.com/Vn0vZP9R/is-it-possible-to-execute-code-in-main-thread-when-called-from-another-thread). However, this is not well documented and we haven't been _knowingly_ hurt by it.

`SequentialThreadingHandler` _did_ seem to work okay before we patched gevent queues in #659, where we found that the handler used queue methods not available on gevent queues. This brought `SequentialGeventHandler` to our attention, which was designed to work with gevent queues from the start.

I think using an implementation built for explicitly for gevent rather than a monkey patched one is a good thing. Not only will we avoid issues with the C lib that we don't even know about, gevent code can have different semantics and performance characteristics.

baseplate/server/monkey.py
RoganMurley: I'm not actually sure... it's likely the case that an unpatched operation is yielded to that never yields back to gevent because it's not patched to do so. This makes it difficult to point to a single unpatched lib.

I'm not too worried about supporting beyond `patch_all` given that we never patch granularly outside of tests.

RoganMurley:
> If patching occurs then in theory you shouldn't need to check, you can just use the native handlers.
> 
> That said, I feel like we should be committing to gevent because that's what we want, or fall back on native handling because we want patching to replace it when it happens.
> 
> I think I need a bit more info on what the silent failure mode you described in the PR description is before I have enough context to thoughtfully approve this.

Thanks for your feedback, I have addressed some of your comments inline.

I think that committing to gevent is probably the right thing to do in the long term and would be interested in doing that migration. However, I've been bitten enough by weird gevent patching to not want to do it as part of this bugfix. I think we should do a controlled prerelease of any such change, canarying a gevent patching entrypoint before general release.

tests/unit/server/monkey_tests.py
```diff
@@ -2,15 +2,24 @@
...
 
+import gevent.monkey
 import gevent.queue
 
+from baseplate.server.monkey import gevent_is_patched
 from baseplate.server.monkey import patch_stdlib_queues
 
 
 class MonkeyPatchTests(unittest.TestCase):
+    def tearDown(self):
```
RoganMurley, line 13: I moved this cleanup to the teardown because otherwise if this test fails it can interfere with other tests.

baseplate/lib/live_data/zookeeper.py
```diff
@@ -83,6 +90,6 @@ def zookeeper_client_from_config(
             backoff=2,  # exponential backoff
             max_jitter=1,  # maximum amount to jitter sleeptimes
             max_delay=60,  # never wait longer than this
-            sleep_func=gevent.sleep,
+            sleep_func=time.sleep,
```
RoganMurley: We could just rely on the default value, but passing in time.sleep makes mocking much easier.

RoganMurley:
@nsheaps I've pushed some changes to rely on the patching `time.sleep`. I had to fiddle with the tests a bit, the comments should explain why 😓

tests/integration/live_data/zookeeper_tests.py
```diff
@@ -48,3 +53,37 @@ def test_create_client_with_credentials(self):
...
+        client = zookeeper_client_from_config(
+            secrets, {"zookeeper.hosts": "%s:%d" % zookeeper_endpoint.address}
+        )
+        assert isinstance(client.handler, SequentialThreadingHandler)
+
+
+class ZooKeeperGeventTests(unittest.TestCase):
+    def setUp(self):
+        # We patch `socket` just to make sure that the gevent handler is chosen.
+        # Nothing is special about `socket` in particular.
```
nsheaps: If we don't care about socket in particular, I would suggest patching the same way baseplate does, aka patch_all + baseplate.monkey.patch_queue

tests/integration/live_data/zookeeper_tests.py
```diff
@@ -48,3 +53,37 @@ def test_create_client_with_credentials(self):
 
         secrets.get_simple.assert_called_with("secret/zk-user")
         self.assertEqual(list(client.auth_data), [("digest", "myzkuser:hunter2")])
+
+    def test_create_client_uses_correct_handler(self):
```
nsheaps: 🔕  by the existence of the other test class, I think it makes sense to move it to the other one and drop the word "correct" as both are correct depending on the use case. `test_client_uses_gevent_handler_when_queue_patched` helps identify why it should use gevent handler and differentiates it from the other test.

Since the other class has only 1 test in it, you should move the patching to each test so the setUp doesn't impact this test.

Similarly the other class should be named something more like `ZooKeeperHandlerTestsWithPatching`

This is probably one of the few cases that the more "enterprise-y" the name the easier it is for the next person to reason about the test

baseplate/lib/live_data/zookeeper.py
nsheaps: I would agree, if this isn't different from the defaults it doesn't make much sense to pass it.

nsheaps (changes requested):
Few more changes

baseplate/lib/live_data/zookeeper.py
KTAtkinson: :no_bell: Can you add a comment to this effect here?

baseplate/server/monkey.py
```diff
@@ -10,4 +10,9 @@ def patch_stdlib_queues() -> None:
     import queue
     import gevent.queue
 
-    patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+    monkey.patch_module(queue, gevent.queue, items=["Queue", "LifoQueue", "PriorityQueue"])
+
+
+def gevent_is_patched() -> bool:
+    """Has any gevent patching has occurred?"""
```
KTAtkinson:
```suggestion
    """Returns true if gevent is patched, otherwise returns false."""
```

KTAtkinson (changes requested):

tests/integration/live_data/zookeeper_tests.py
RoganMurley: We can't do this because some of what `patch_all` patches we can't "unpatch" (as far as I know). For example, we have no way to unpatch `time.sleep` as it is a builtin and `importlib.reload` doesn't unload it.

RoganMurley: (I'll add a comment to this effect)

RoganMurley: It might make sense to review how we handle gevent tests in a future PR as the current state of things is very fiddly.

tests/integration/live_data/zookeeper_tests.py
RoganMurley: Fair, I suppose the `tearDown` won't affect the other cases so there's no need to be separate.

I've gone with the name `ZooKeeperHandlerWithPatchingTests` to follow the `.*Tests` pattern.

baseplate/lib/live_data/zookeeper.py
RoganMurley: @nsheaps  I can do that, but the patching gets much nastier because default values are not easily patched 🤢 I just pushed a commit with that change so let me know what you think

tests/integration/live_data/zookeeper_tests.py
```diff
@@ -48,3 +60,30 @@ def test_create_client_with_credentials(self):
...
+    def test_create_client_uses_threading_handler_when_not_gevent_patched(self):
+        secrets = mock.Mock(spec=SecretsStore)
+        client = zookeeper_client_from_config(
+            secrets, {"zookeeper.hosts": "%s:%d" % zookeeper_endpoint.address}
+        )
+        assert isinstance(client.handler, SequentialThreadingHandler)
+
+    @mock.patch(
+        "kazoo.retry.KazooRetry.__init__.__defaults__",
+        (1, 0.1, 2, 0.4, 60.0, True, gevent.sleep, None, None),
```
RoganMurley: @nsheaps Here is the yucky patch. There are some tricks I could do with a context manager to only modify the one default arg, but I'm unsure if it's worth it.

tests/integration/live_data/zookeeper_tests.py
nsheaps: https://docs.python.org/3/library/importlib.html#importlib.reload

You might just have to import `time` before patching

tests/integration/live_data/zookeeper_tests.py
nsheaps: 💡  If you're looking to unit test this, why not mock `baseplate.server.monkey.gevent_is_patched` and test which instance is in the handler without looking to actually monkey patch things?

baseplate/lib/live_data/zookeeper.py
nsheaps: if passing time.sleep makes this all easier, feel free to pass it. If we want to be sticklers about it, then max_delayu is also defaulting to 60, and backoff to 2 ( [source](https://github.com/python-zk/kazoo/blob/master/kazoo/retry.py#L45) )

tests/integration/live_data/zookeeper_tests.py
RoganMurley: Unfortunately the Gevent handler throws a configuration error if the sleep function is unpatched, so we'd still need to patch `time.sleep` in this case.

```
kazoo.exceptions.ConfigurationError: Retry handler and event handler  must use the same sleep func
```

We could patch `gevent_is_patched` if we wanted to avoid doing any gevent patching in test though, that might make things cleaner.

RoganMurley:
@nsheaps @KTAtkinson I tidied up the testing a bit:
* Mocked `gevent_is_patched` rather than do a true gevent patched test.
* I worked out how to patch default values in a less brittle way.

tests/integration/live_data/zookeeper_tests.py
```diff
@@ -1,9 +1,15 @@
...
 from unittest import mock
 
+import gevent.socket
+
 try:
     from kazoo.exceptions import NoNodeError
+    from kazoo.handlers.gevent import SequentialGeventHandler
+    from kazoo.handlers.threading import SequentialThreadingHandler
+    from kazoo.retry import KazooRetry
 except ImportError:
```
nsheaps, line 13: Out of the scope of this PR since you didn't change it, but this seems like something that shouldn't be skipped if something is not installed, especially for tests.

tests/integration/live_data/zookeeper_tests.py
```diff
@@ -48,3 +54,23 @@ def test_create_client_with_credentials(self):
...
         self.assertEqual(list(client.auth_data), [("digest", "myzkuser:hunter2")])
+
+    def test_create_client_uses_threading_handler_when_not_gevent_patched(self):
+        secrets = mock.Mock(spec=SecretsStore)
+        client = zookeeper_client_from_config(
+            secrets, {"zookeeper.hosts": "%s:%d" % zookeeper_endpoint.address}
+        )
+        assert isinstance(client.handler, SequentialThreadingHandler)
+
+    @mock.patch("baseplate.lib.live_data.zookeeper.gevent_is_patched", return_value=True)
```
nsheaps, line 65: 🔕 You should patch this on the other test too so that your results will always be consistent when someone else messes up their tests and accidentally patches without cleaning it up.

nsheaps (approved):
Nice job. Thanks for addressing all of our feedback!

KTAtkinson (approved):

PR #669, opened 2022-05-05
fix(metrics): pass by class reference rather than instantiated class DEVX-585

nsheaps:
I pulled the local package into a local pip install into moderators, which exhibited the problem, then added some logging to confirm that the qsize != size at which point it was obvious that they were all sharing one queue. 

## ✅ Checks

### DIFF
--- a/baseplate/lib/thrift_pool.py
+++ b/baseplate/lib/thrift_pool.py
@@ -105,7 +105,7 @@ def thrift_pool_from_config(
     options = parser.parse(prefix[:-1], app_config)
 
     if options.fifo_queue:
-        kwargs.setdefault("queue_cls", queue.Queue())
+        kwargs.setdefault("queue_cls", queue.Queue)
     if options.size is not None:
         kwargs.setdefault("size", options.size)
     if options.max_age is not None:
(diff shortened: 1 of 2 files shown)


RoganMurley:
Good catch, looks like this was broken in #645

RoganMurley (approved):

baseplate/lib/thrift_pool.py
```diff
@@ -170,7 +171,7 @@ def __init__(
 
         self.size = size
 
-        self.pool = queue_cls
+        self.pool = queue_cls()
```
mterwill, line 174: I can't find anywhere internally that `queue_cls` is being passed in as an argument, but just wanted to call out that this is technically a backwards-incompatible change.

mterwill (approved):

a--hoang:
Note that in addition to metrics, this will also possibly cause thrift connection issues in the case that you've instantiated more than one thrift client in your application and used the default params. 

This will result in all thrift clients in your application using the SAME queue, which means that they can accidentally grab each others' connections/sockets and send invalid requests to the wrong server. All clients are also in contention with each other for the pool size, so increased timeouts/connection failures are likely.

Glad we got this fixed - hoping we can do a bump to v2.5.x after this is merged since this potentially has large impact

KTAtkinson (approved):
Good find!

Please make sure to not reference Reddit internal services in PR descriptions or commit messages.

baseplate/lib/thrift_pool.py
nsheaps: You are correct. Only 3 services are on 2.4.x+ and none of them take advantage of this (this is primarily used for tests)

PR #670, opened 2022-05-06
Add a list of supported versions to docs homepage

nsheaps:
Update docs to include a supported versions section

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,6 +1,11 @@
 Baseplate.py
 ============
 
+:Supported Versions: [v2.4.7](https://github.com/reddit/baseplate.py/releases/tag/v2.4.7) [v2.3.3](https://github.com/reddit/baseplate.py/releases/tag/v2.3.3)
+
+*If you are not on one of these versions, please upgrade. See `baseplate.py-upgrader <https://github.com/reddit/baseplate.py-upgrader>`_ for more info*
+
+
 It's much easier to manage a bunch of services when they all have the same
 shape: the way they're developed, the way they interact with the infrastructure
 they run on, and the way they interact with each other. Baseplate is reddit's
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,7 +1,7 @@
 Baseplate.py
 ============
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1,6 +1,11 @@
 Baseplate.py
 ============
 
+:Supported Versions: `v2.4.7 <https://github.com/reddit/baseplate.py/releases/tag/v2.4.7>`_, `v2.3.3 <https://github.com/reddit/baseplate.py/releases/tag/v2.3.3>`_
```
mterwill, line 4: This seems like something that's going to drift over time as new versions are released and this page is forgotten. Why would we want to advertise support for anything other than the latest 2.x?

nsheaps: I don't disagree, so if you have a better suggestion, I'm all ears! 👂 👂 👂 👂 👂 👂 👂  One alternative would be to replace this block with a statement of "We support only the 2 or 3 latest versions of the library. See the release history in baseplate.py to see which versions are supported." as we have discussed that releases with issues should be marked as one that you don't want to upgrade to in the releases on github.

The intention here is to say what we currently support. While we want everyone to be on the latest, this also is intended to imply that previous versions of 2.4.x are not supported, and 2.3.3 is still supported until 2.4.X is considered stable enough to force everyone to upgrade. My intention is to support only two or three of the latest versions, and due to the issues with 2.4.0 through 2.4.6, and statements from others, these two versions seem the most appropriate to outwardly say that we support.

Company initiatives will force upgrades, but until that happens, people still on certain older versions will still be supported with regards to bug reports. Unfortunately, usually the answer will be "that doesn't exist in a later version," or "we've confirmed this bug and will fix it in the latest 2.x release."

mterwill: Nate and I spoke about this offline. My preference would be to close this out – implicitly, we only support the latest version. If a regression is introduced, it should be communicated and patched quickly to ensure the latest release is stable. An official support policy otherwise would add non-trivial overhead for both users of the library and us as maintainers.

PR #671, opened 2022-05-09
add thrift client metrics

krav:
Closes #OBS-34.

## 💸 TL;DR

This adds metrics as per baseplate.spec to the thrift client. `PrometheusClientSpanObserver` is mostly nicked from/shared with the HTTP client metrics, which still is an open PR.

## 📜 Details
https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md

[Jira](https://reddit.atlassian.net/browse/OBS-48)

If anyone could walk me through how to run this with a real thrift client and server, I'd be very grateful.

## ✅ Checks

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -168,12 +168,16 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
         last_error = None
 
         for time_remaining in self.retry_policy:
-            span = self.server_span.make_child(trace_name)
-            span.set_tag("slug", self.namespace)
-            span.start()
-
             try:
                 with self.pool.connection() as prot:
+                    span = self.server_span.make_child(trace_name)
+                    span.set_tag("slug", self.namespace)
+
+                    client = self.client_cls(prot)
+                    method = getattr(client, name)
+                    span.set_tag("method", method.__name__)
+                    span.start()
+
                     baseplate = span.baseplate
                     if baseplate:
                         service_name = baseplate.service_name
@@ -208,9 +212,6 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
                     if edge_context:
                         prot.trans.set_header(b"Edge-Request", edge_context)
 
-                    client = self.client_cls(prot)
-                    method = getattr(client, name)
-                    span.set_tag("method", method)
                     result = method(*args, **kwargs)
             except TTransportException as exc:
                 # the connection failed for some reason, retry if able
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -169,6 +169,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
 
         for time_remaining in self.retry_policy:
             span = self.server_span.make_child(trace_name)
+            span.set_tag("slug", self.namespace)
             span.start()
 
             try:
@@ -209,6 +210,7 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
 
                     client = self.client_cls(prot)
                     method = getattr(client, name)
+                    span.set_tag("method", method)
                     result = method(*args, **kwargs)
             except TTransportException as exc:
                 # the connection failed for some reason, retry if able
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -95,7 +95,7 @@ thrift_client_active_gauge = Gauge(
     "thrift_client_active_requests", "Current in-flight requests", ["thrift_slug", "thrift_method"]
 )
 
-thrift_client_latency_historygram = Histogram(
+thrift_client_latency_histogram = Histogram(
     "thrift_client_latency_seconds",
     "Latency of thrift client requests",
     ["thrift_slug", "thrift_success"],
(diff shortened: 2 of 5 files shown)


baseplate/clients/thrift.py
```diff
@@ -168,11 +168,16 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
         last_error = None
 
         for time_remaining in self.retry_policy:
-            span = self.server_span.make_child(trace_name)
-            span.start()
-
             try:
                 with self.pool.connection() as prot:
+                    span = self.server_span.make_child(trace_name)
+                    span.set_tag("slug", self.namespace)
```
nsheaps, line 174: 🔕  why not call it "namespace"? Slug to me implies a type of data, typically a shortened form of an ID, not a categorization mechanism like a namespace.

Type of data being iamaslug or i-am-a-slug

baseplate/lib/prometheus_metrics.py
```diff
@@ -99,6 +101,66 @@ def get_active_requests_metric(self) -> Gauge:
...
+        pass
+
+    def active_requests_metric(self, tags: Dict) -> Gauge:
+        return thrift_client_active_gauge.labels(
+            thrift_slug=tags.get("slug", ""), thrift_method=tags.get("method", "")
+        )
+
+    def requests_total_metric(self, tags: Dict) -> Counter:
+        return thrift_client_requests_counter.labels(
+            thrift_slug=tags.get("slug", ""),
```
nsheaps, line 139: Sorry for being a newb here, why does this one need the labels explicitly defined when the counter has the labels already called out? Also what makes `request_total_metric` different than `get_requests_total_metric`?

tests/integration/thrift_tests.py
```diff
@@ -589,22 +590,20 @@ def is_healthy(self, context, req=None):
                 self.assertEqual(handler.probe, IsHealthyProbe.LIVENESS)
 
 
-class ThriftErrorReplacementTests(GeventPatchedTestCase):
```
nsheaps: Is this test not accurate anymore?

baseplate/observers/prometheus.py
```diff
@@ -127,7 +131,80 @@ def on_child_span_created(self, span: Span) -> None:
...
+            )
+            return
+
+        self.tags["success"] = "true"
+        if exc_info is not None:
+            exc = exc_info[1]
+            self.tags["exception_type"] = exc.__class__.__name__
+            self.tags["success"] = "false"
+            if self.protocol == "thrift" and isinstance(exc, Error):
+                self.tags["thrift_status_code"] = exc.code
```
nsheaps, line 189: Shouldn't this be done regardless of if it was an exception? What about successful responses?

nsheaps (changes requested):
few questions. Thanks for the contribution!

baseplate/clients/thrift.py
krav: The spec specifies a slug, and I chose the namespace string to be that slug - but I'm not actually completely sure that it's the right id to use. It shows up in the tests as "example-service" which looks good, but it might be used differently other places?

baseplate/lib/prometheus_metrics.py
krav: This function is useful because it translates from the tags to Prometheus labels. The required labels are declared in the metric, and this returns a counter with those labels set to values, that you then can call `.inc()` on it.

The `get_requests_total_metric` was added as a convenience to get the metric without labels and is only used in tests AFAIK. Not sure why we don't just import the metrics from the module in the test, but it's the way it was done in the server metrics.

baseplate/observers/prometheus.py
krav: `exc_info` is always `None` if the response was successful.  The spec says: "thrift_baseplate_status_code - the numeric status code from a baseplate.Error [that's an exception class] as a string if present (e.g. 404), or the empty string"

tests/integration/thrift_tests.py
krav: Ooops, I meant to delete the code that was added in this patch. Seems I missed and hit the preceding block 😓 Fixed.

baseplate/observers/prometheus.py
nsheaps: Sorry, I meant to imply that we should set the status code regardless of an exception, not set the value to the exc.code even if there wasn't an exception

baseplate/lib/prometheus_metrics.py
nsheaps: If this pattern already exists I'll let it slide, but I don't understand why tests need the ones without the labels (other than it maybe being more to test)

baseplate/clients/thrift.py
nsheaps: Good enough for me. Thanks for explaining

baseplate/observers/prometheus.py
krav: Looking at `thrift/ttypes.py`, there doesn't seem to be a success status code defined. They seem to be directly taken from HTTP, except for a few, so I guess we could set this to 200 on success. Someone more thrifty might want to chime in on that?

baseplate/lib/prometheus_metrics.py
krav: It's because the tests don't have values to set for those labels, I think, sp they call `collect()` on the metric to get all the samples. I think this is fine. It defines a common interface to those metrics for the unit test to test all metric classes in one parameterized test.

nsheaps (approved):

PR #672, opened 2022-05-16
feat(prometheus): Implement Local Span metrics

Xaelias:
## 💸 TL;DR
Implement local span metrics. Per the spec, we are tracking "latency" (time spent in the span), number of active spans, and number of spans ever started. All of these have a single label, which is the name of the span.

unit/integration tests are in the PR.
Testing it locally should be as simple as doing something like:
```
with baseplate.server_context("test") as context:
    with context.span.make_child('ipinfo_calls', local=True) as span:
        span.context.external.get('https://ipinfo.io/ip')
        span.context.external.get('https://ipinfo.io/json')
        span.context.external.get('https://ipinfo.io/error')
```
or anything within a child span with `local=True`.

## ✅ Checks

## Sample output

```bash
❯ curl -s localhost:6060/metrics | grep -i span
# HELP local_span_active_requests Multiprocess metric
# TYPE local_span_active_requests gauge
local_span_active_requests{pid="63302",span="ipinfo_calls"} 0.0
# HELP local_span_latency_seconds Multiprocess metric
# TYPE local_span_latency_seconds histogram
local_span_latency_seconds_sum{span="ipinfo_calls"} 0.225517531
local_span_latency_seconds_bucket{le="0.0001",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.00025",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.000625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.0015625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.00390625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.009765625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.0244140625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.06103515625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.152587890625",span="ipinfo_calls"} 0.0
local_span_latency_seconds_bucket{le="0.3814697265625",span="ipinfo_calls"} 1.0
local_span_latency_seconds_bucket{le="0.95367431640625",span="ipinfo_calls"} 1.0
local_span_latency_seconds_bucket{le="2.384185791015625",span="ipinfo_calls"} 1.0
local_span_latency_seconds_bucket{le="5.9604644775390625",span="ipinfo_calls"} 1.0
local_span_latency_seconds_bucket{le="14.901161193847656",span="ipinfo_calls"} 1.0
local_span_latency_seconds_bucket{le="+Inf",span="ipinfo_calls"} 1.0
local_span_latency_seconds_count{span="ipinfo_calls"} 1.0
# HELP local_span_requests_total Multiprocess metric
# TYPE local_span_requests_total counter
local_span_requests_total{span="ipinfo_calls"} 1.0
```

### DIFF
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -343,3 +343,60 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     The HTTP success label is "true" if the status code is 2xx or 3xx, "false" otherwise.
     """
     return str(200 <= httpStatusCode < 400).lower()
+
+
+# local labels and metrics
+local_span_labels = [
+    "span",
+]
+# Latency histogram of local span
+# buckets are defined above (from 100µs to ~14.9s)
+local_span_latency_seconds = Histogram(
+    "local_span_latency_seconds",
+    "Latency histogram of local span",
+    local_span_labels,
+    buckets=default_latency_buckets,
+)
+
+# Counter counting total local spans started
+local_span_requests_total = Counter(
+    "local_span_requests_total",
+    "Total number of local spans started",
+    local_span_labels,
+)
+
+# Gauge showing current number of local spans
+local_span_active_requests = Gauge(
+    "local_span_active_requests",
+    "Number of active local spans",
+    local_span_labels,
+)
+
+
+class PrometheusLocalSpanMetrics:
+    def __init__(self) -> None:
+        pass
+
+    def latency_seconds_metric(self, tags: Dict) -> Histogram:
+        return local_span_latency_seconds.labels(
+            span=tags.get("span_name", ""),
+        )
+
+    def requests_total_metric(self, tags: Dict) -> Counter:
+        return local_span_requests_total.labels(
+            span=tags.get("span_name", ""),
+        )
+
+    def active_requests_metric(self, tags: Dict) -> Gauge:
+        return local_span_active_requests.labels(
+            span=tags.get("span_name", ""),
+        )
+
+    def get_latency_seconds_metric(self) -> Histogram:
+        return local_span_latency_seconds
+
+    def get_requests_total_metric(self) -> Counter:
+        return local_span_requests_total
+
+    def get_active_requests_metric(self) -> Gauge:
+        return local_span_active_requests
--- a/baseplate/observers/prometheus.py
+++ b/baseplate/observers/prometheus.py
@@ -14,6 +14,7 @@ from baseplate import Span
 from baseplate import SpanObserver
 from baseplate.lib.prometheus_metrics import PrometheusHTTPClientMetrics
 from baseplate.lib.prometheus_metrics import PrometheusHTTPServerMetrics
+from baseplate.lib.prometheus_metrics import PrometheusLocalSpanMetrics
 from baseplate.lib.prometheus_metrics import PrometheusThriftClientMetrics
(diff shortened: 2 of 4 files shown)


baseplate/lib/prometheus_metrics.py
```diff
@@ -343,3 +343,60 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     The HTTP success label is "true" if the status code is 2xx or 3xx, "false" otherwise.
     """
     return str(200 <= httpStatusCode < 400).lower()
+
```
nsheaps, line 346: 🔕  extra line returns?

Xaelias: No that's the python spec.
We're outside of a class so everything should basically have two lines between them.

```
--- baseplate/lib/prometheus_metrics.py	2022-05-16 13:53:48.004208 +0000
+++ baseplate/lib/prometheus_metrics.py	2022-05-16 13:53:55.730775 +0000
@@ -342,10 +342,11 @@
     """
     """

+
 # local labels and metrics
 local_span_labels = [
     "span",
 ]
 # Latency histogram of local span
would reformat baseplate/lib/prometheus_metrics.py
Oh no! 💥 💔 💥
1 file would be reformatted, 152 files would be left unchanged.
```
(if you remove that extra line)

Xaelias: I could keep that function at the end of the file if you think that makes more sense though

baseplate/observers/prometheus.py
```diff
@@ -57,9 +58,6 @@ def __init__(self) -> None:
     def on_set_tag(self, key: str, value: Any) -> None:
         self.tags[key] = value
 
-    def get_prefix(self) -> str:
-        return f"{self.protocol}_server"
-
```
KTAtkinson: Why is this method being removed? It would be nice for this to be mentioned in the commit message.

KTAtkinson (changes requested):

baseplate/observers/prometheus.py
nsheaps: Looks like it was removed because it was only used in tests. That said, the `$proto_$actor` syntax is used a bunch in the gauges that are set up, I would actually have preferred to see this used in those gauge definitions (as a static class method that's used in generating the string)

nsheaps (changes requested):
I agree with Katie's feedback on the prefix handling, sorry I missed that in previous reviews. Should be quick to change.

baseplate/observers/prometheus.py
Xaelias: Yeah it was just not used significantly anywhere.

I think Nathan is half right. I think originally that's how it was used, but then it was never fully developed and now it's just there to be tested but doesn't bring any value.

nsheaps: To be clear I'm asking for `half used` => `actually used` instead of removing it.

Xaelias: Will do this in a follow up PR to keep the PR focused on a single change.

KTAtkinson (approved):

nsheaps (approved):
making use of get_prefix will be a future PR.

PR #673, opened 2022-05-18
Prometheus/2022 05 17/use get prefix for metric names

Xaelias:
## 💸 TL;DR
Follow up to https://github.com/reddit/baseplate.py/pull/672
We decided to use static variables / methods for promtheus_metrics.py and standardized how names are generated, as well as how variables themselves are called between http/thrift client/server.

tests still pass
This is exclusively refactoring, moving code around and making some things "static". No logic was touched.

## ✅ Checks

### DIFF
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -26,50 +26,14 @@ default_size_buckets = [
     default_size_start * default_size_factor ** i for i in range(default_size_count)
 ]
 
-# http client labels and metrics
-http_client_latency_labels = [
-    "http_method",
-    "http_success",
-    "http_slug",
-]
-http_client_requests_total_labels = [
-    "http_method",
-    "http_success",
-    "http_response_code",
-    "http_slug",
-]
-http_client_active_requests_labels = [
-    "http_method",
-    "http_slug",
-]
-
-# Latency histogram of HTTP calls made by clients
-# buckets are defined above (from 100µs to ~14.9s)
-http_client_latency_seconds = Histogram(
-    "http_client_latency_seconds",
-    "Latency histogram of HTTP calls made by clients",
-    http_client_latency_labels,
-    buckets=default_latency_buckets,
-)
-
-# Counter counting total HTTP requests started by a given client
-http_client_requests_total = Counter(
-    "http_client_requests_total",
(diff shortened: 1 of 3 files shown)


baseplate/lib/prometheus_metrics.py
```diff
@@ -26,316 +26,348 @@
...
-    def get_latency_seconds_metric(self) -> Histogram:
-        return http_client_latency_seconds
+    @classmethod
+    def get_latency_seconds_metric(cls) -> Histogram:
+        return cls.latency_seconds
 
-    def get_requests_total_metric(self) -> Counter:
-        return http_client_requests_total
+    @classmethod
+    def get_requests_total_metric(cls) -> Counter:
```
nsheaps, line 103: Won't this share the gauge and counter across spans if they're class methods?

Xaelias: They are shared. The tags is the only thing differencing these across spans. The underlying object is the same and a singleton. We only ever created one of them outside of the class to begin with.

nsheaps: doh, thank you for correcting my line of thought

baseplate/lib/prometheus_metrics.py
```diff
@@ -26,316 +26,348 @@
...
-
-# Gauge showing current number of active requests by a given client
-http_client_active_requests = Gauge(
-    "http_client_active_requests",
-    "Number of active requests for a given client",
-    http_client_active_requests_labels,
-)
-
 
 class PrometheusHTTPClientMetrics:
```
krav, line 30: Very pro moving these into the class! 

Are these also instantiated when the file is loaded, or when the class is first used? Worried about metrics disappearing on restart.

Xaelias: they're static variables, they exist even if the class is never instantiated
I guess I don't know for sure if they exist before they're accessed?
but they wouldn't exist without tags provided either way so I don't think we can lose anything here

krav: Ah, I didn't realize but of course they need labels first. Thanks!

nsheaps (approved):
I like this much better, thanks for doing this!

PR #674, opened 2022-05-18
Accept json data that is not a dictionary

TylerLubeck:
This barfed if live data nodes contained non-object jsonified strings. Now it doesn't.

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -85,7 +85,13 @@ class NodeWatcher:
         # }
         # If the load type is 'S3', this format is an indication that we support
         # downloading the contents of encrypted files uploaded to S3.
-        if json_data.get("live_data_watcher_load_type") == "S3":
+        is_s3 = False
+        try:
+            is_s3 = json_data["live_data_watcher_load_type"] == "S3"
+        except (KeyError, TypeError):
+            is_s3 = False
+
+        if is_s3:
             if not S3_FETCHER_ENABLED:
                 logger.error("Attempted to use s3 fetcher but packages are missing")
                 return None


baseplate/sidecars/live_data_watcher.py
```diff
@@ -85,7 +85,13 @@ def get_data_to_write(json_data: dict, data: Optional[bytes]) -> Optional[bytes]
...
         # If the load type is 'S3', this format is an indication that we support
         # downloading the contents of encrypted files uploaded to S3.
-        if json_data.get("live_data_watcher_load_type") == "S3":
+        is_s3 = False
+        try:
+            is_s3 = json_data["live_data_watcher_load_type"] == "S3"
+        except (KeyError, TypeError):
+            is_s3 = False
+
+        if is_s3:
```
nsheaps, line 94: While you're making the change, there's a typo on line 107 that I'd like to see fixed `unble` => `unable`

baseplate/sidecars/live_data_watcher.py
```diff
@@ -85,7 +85,13 @@ def get_data_to_write(json_data: dict, data: Optional[bytes]) -> Optional[bytes]
         # }
         # If the load type is 'S3', this format is an indication that we support
         # downloading the contents of encrypted files uploaded to S3.
-        if json_data.get("live_data_watcher_load_type") == "S3":
+        is_s3 = False
+        try:
+            is_s3 = json_data["live_data_watcher_load_type"] == "S3"
+        except (KeyError, TypeError):
```
nsheaps, line 91: since it's one line it's fine but I probably would have opted for `is_s3 = json_data.get("live_data_watcher_load_type", "zk") == "S3"` to avoid the KeyError

That said, catching an exception is very pythonic.

nsheaps: The type for json_data on line 81 should be updated to indicate it can be more than a dict.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -85,7 +85,13 @@ def get_data_to_write(json_data: dict, data: Optional[bytes]) -> Optional[bytes]
...
+            is_s3 = False
+
+        if is_s3:
             if not S3_FETCHER_ENABLED:
```
nsheaps, line 95: Thinking about this now and based on the discussion we had, I think we should consider requiring boto3 as a dependency so that data writers can move between zk and s3 at will without breaking folks that don't have it installed.

nsheaps (changes requested):
Let's make sure the contract is updated to indicate that it's not always a dict anymore (this is what bit us)

Thanks for the fast turnaround!

nsheaps:
This is being held until after a company event when a proper test plan with real data can be executed before we merge this.

baseplate/sidecars/live_data_watcher.py
TylerLubeck: Ha, it was `.get()` before the change. I switched to the KeyError purely for style reasons, I have no strong technical preference. I can change it back

Xaelias:
~If `json_data` is not a dict, the typing is not accurate. Do we want to track that down / fix it?~
Nvm Nathan already said that ><

nsheaps:
Screen Shot 2022-05-19 at 10 16 04 AM
To help with the typing

baseplate/sidecars/live_data_watcher.py
Xaelias, line 94: Maybe it's just me but this feels like a very long way of doing something like:
```
if isinstance(json_data, dict) and json_data.get("live_data_watcher_load_type", None) == "S3":
    if not S3_FETCHER_ENABLED:
```

If we keep the try/except, your except statement doesn't do anything, that value is already false. So I would just `pass` at this point I think.

baseplate/sidecars/live_data_watcher.py
KTAtkinson: I agree with @nsheaps original comment about using `.get(<key>, <default>)`.


I think we might need to catch `AttributeError` as well, if the data happens to be formatted as a json list, that will parse as JSON but it will be a list not an object, like this:

```
>>> import json
>>> raw_json="""
... ["foo", "bar"]
... """
>>> parsed_json=json.
>>> parsed_json=json.loads(raw_json)
>>> parse_json.get("foo")
  File "<stdin>", line 1, in <module>
NameError: name 'parse_json' is not defined
>>> parsed_json.get("foo")
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'get'
```

KTAtkinson (changes requested):

baseplate/sidecars/live_data_watcher.py
nsheaps: Thanks for looking/checking the exception type @KTAtkinson

TylerLubeck:
#676 got traction, I'm going to opt for that one

PR #675, opened 2022-05-19
Make a generic span observer for when the protocol is not HTTP or Thrift

nsheaps:
Adds generic observers for spans that existed in old metrics style to

* silence errors about metrics not tracking properly
* get the most use out of it on release without too much customization

Note: This also changes `local_span_...` to `local_...`
^ I didn't see anything in baseplate spec or prometheus docs about this. I personally don't like naming it 2x (it's already implied to be a span).

### DIFF
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -332,7 +332,7 @@ class MonitoredMemcacheConnection:
 
         """
         trace_name = f"{self.context_name}.{method_name}"
-        span = self.server_span.make_child(trace_name)
+        span = self.server_span.make_child(trace_name, component_name="memcache_client")
         span.set_tag("method", method_name)
         return span
 
--- a/baseplate/clients/redis_cluster.py
+++ b/baseplate/clients/redis_cluster.py
@@ -430,7 +430,7 @@ class MonitoredClusterRedisConnection(rediscluster.RedisCluster):
         command = args[0]
         trace_name = f"{self.context_name}.{command}"
 
-        with self.server_span.make_child(trace_name):
+        with self.server_span.make_child(trace_name, component_name="rediscluster_client"):
             res = super().execute_command(command, *args[1:], **kwargs)
 
         self.hot_key_tracker.maybe_track_key_usage(list(args))
@@ -487,5 +487,5 @@ class MonitoredClusterRedisPipeline(ClusterPipeline):
 
     # pylint: disable=arguments-differ
     def execute(self, **kwargs: Any) -> Any:
-        with self.server_span.make_child(self.trace_name):
+        with self.server_span.make_child(self.trace_name, component_name="rediscluster_client"):
             return super().execute(**kwargs)
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -210,7 +210,7 @@ class BaseplateSession:
             "http.url": request.url,
             "http.slug": self.client_name if self.client_name is not None else self.name,
         }
-        with self.span.make_child(f"{self.name}.request").with_tags(tags) as span:
+        with self.span.make_child(f"{self.name}.request", component_name="http_client").with_tags(tags) as span:
             self._add_span_context(span, request)
 
             # we cannot re-use the same session every time because sessions re-use the same
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -210,7 +210,9 @@ class BaseplateSession:
             "http.url": request.url,
             "http.slug": self.client_name if self.client_name is not None else self.name,
         }
-        with self.span.make_child(f"{self.name}.request", component_name="http_client").with_tags(tags) as span:
+        with self.span.make_child(f"{self.name}.request", component_name="http_client").with_tags(
+            tags
+        ) as span:
             self._add_span_context(span, request)
 
             # we cannot re-use the same session every time because sessions re-use the same
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -32,6 +32,7 @@ default_size_buckets = [
 
 generic_metrics = {}
 
+
 class PrometheusHTTPClientMetrics:
     prefix = "http_client"
 
@@ -381,6 +382,7 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     """
     return str(200 <= httpStatusCode < 400).lower()
 
+
 class PrometheusGenericSpanMetrics:
     prefix = "generic"
 
@@ -471,6 +473,7 @@ class PrometheusLocalSpanMetrics(PrometheusGenericSpanMetrics):
         labels,
     )
 
+
 def get_metrics_for_prefix(prefix: str) -> PrometheusGenericSpanMetrics:
     if prefix not in generic_metrics:
         # local labels and metrics
@@ -479,9 +482,9 @@ def get_metrics_for_prefix(prefix: str) -> PrometheusGenericSpanMetrics:
         ]
         generic_metrics[prefix] = type(
             f"PrometheusGenericSpanMetrics<{prefix}>",
-            (PrometheusGenericSpanMetrics, ),
+            (PrometheusGenericSpanMetrics,),
             {
-                "prefix":prefix,
+                "prefix": prefix,
                 # Reset the class attributes to avoid having the same prefix in the metric names
                 "labels": labels,
                 "latency_seconds": Histogram(
@@ -499,8 +502,8 @@ def get_metrics_for_prefix(prefix: str) -> PrometheusGenericSpanMetrics:
                     f"{prefix}_active_requests",
                     "Number of active local spans",
                     labels,
-                )
-            }
+                ),
+            },
         )
         logger.info(f"Created new metrics class for prefix {prefix}")
     return generic_metrics[prefix]
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -2,7 +2,6 @@ import logging
 
 from typing import Any
 from typing import Dict
-from typing import Optional
 
 from prometheus_client import Counter
 from prometheus_client import Gauge
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -445,34 +445,6 @@ class PrometheusGenericSpanMetrics:
         return cls.active_requests
 
 
-class PrometheusLocalSpanMetrics(PrometheusGenericSpanMetrics):
-    prefix = "local"
-
-    # local labels and metrics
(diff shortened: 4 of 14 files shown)


baseplate/observers/prometheus.py
```diff
@@ -253,3 +254,14 @@ def on_child_span_created(self, span: Span) -> None:
             observer = PrometheusClientSpanObserver()
 
         span.register(observer)
+
+
+class PrometheusGenericSpanObserver(PrometheusLocalSpanObserver):
+    def __init__(self, span_name: Optional[str] = None) -> None:
+        super(PrometheusGenericSpanObserver, self).__init__(span_name)
+        self.metrics: PrometheusGenericSpanMetrics = PrometheusGenericSpanMetrics(span_name)
```
Xaelias: That's going to create a new metric name per span. You don't want that. I think you would want like a common name `generic_server_` or smth, and the having the span being one of the labels.

Xaelias (dismissed):

baseplate/observers/prometheus.py
nsheaps: For the test case I had this actually ends up being `None`, but I wanted a way to differentiate different generic spans without just using tags

Xaelias: But you pass span_name as an argument that you then set as prefix I believe?
My concern now would be that the variables are using a None prefix whatever you do because they're created at the class definition level, not the instance. Which means that this would never work?

nsheaps:
@Xaelias forget about tests for a second, this is what I'm thinking, which when combined with the PR in the test service, results in the following prometheus metrics (service name obfuscated)

Screen Shot 2022-05-19 at 8 34 44 PM

baseplate/lib/prometheus_metrics.py
```diff
@@ -438,3 +443,66 @@ def get_requests_total_metric(cls) -> Counter:
...
+                    labels,
+                ),
+                "active_requests": Gauge(
+                    f"{prefix}_active_requests",
+                    "Number of active local spans",
+                    labels,
+                ),
+            },
+        )
+        logger.info("Created new metrics class for prefix {}" % prefix)
```
nsheaps: make this debug

baseplate/observers/prometheus.py
```diff
@@ -203,23 +215,26 @@ def on_child_span_created(self, span: Span) -> None:
...
 
 
 class PrometheusLocalSpanObserver(SpanObserver):
+    prefix = "local"
+
     def __init__(self, span_name: Optional[str] = None) -> None:
         self.tags: Dict[str, Any] = {"span_name": span_name if span_name is not None else ""}
         self.start_time: Optional[int] = None
+
         self.metrics: PrometheusLocalSpanMetrics = PrometheusLocalSpanMetrics()
```
nsheaps: Swap this for `get_metrics_for_prefix('local')`

baseplate/clients/memcache/__init__.py
```diff
@@ -332,7 +332,7 @@ def _make_span(self, method_name: str) -> Span:
 
         """
         trace_name = f"{self.context_name}.{method_name}"
-        span = self.server_span.make_child(trace_name)
+        span = self.server_span.make_child(trace_name, component_name="memcache_client")
```
nsheaps, line 335: I wonder if the `_client` part in all of these should be handed by `PrometheusClientSpanObserver` instead, then you can use the component name more directly to route to different metrics classes (when migrating _from_ generic to something more like the http/thrift one)

baseplate/observers/prometheus.py
bjk-reddit: @Xaelias Thanks for calling that out. We definitely don't want service names leaking into metric names. That's the job of Prometheus discovery labels. (i.e. Kubernetes namespace, etc, labels).

baseplate/lib/prometheus_metrics.py
```diff
@@ -438,3 +443,38 @@ def get_requests_total_metric(cls) -> Counter:
...
+                ),
+                "active_requests": Gauge(
+                    f"{prefix}_active_requests",
+                    f"Number of active {prefix} spans",
+                    labels,
+                ),
+            },
+        )
+        logger.debug("Created new metrics class for prefix %s", prefix)
+    return generic_metrics[prefix]  # type: ignore
```
nsheaps, line 480: It thinks the type is `type`

baseplate/observers/prometheus.py
```diff
@@ -71,10 +74,11 @@ def set_metrics_by_protocol(self) -> None:
         elif self.protocol == "http":
             self.metrics = PrometheusHTTPServerMetrics()
         else:
-            logger.warning(
-                "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",
+            logger.debug(
```
nsheaps, line 77: This doesn't print by default at debug level. It would spam a lot if it did.

baseplate/observers/prometheus.py
```diff
@@ -126,18 +130,25 @@ def on_child_span_created(self, span: Span) -> None:
...
         self.start_time: Optional[int] = None
         self.metrics: Optional[
-            Union[PrometheusHTTPClientMetrics, PrometheusThriftClientMetrics]
+            Union[
+                PrometheusHTTPClientMetrics,
+                PrometheusThriftClientMetrics,
+                PrometheusGenericSpanMetrics,
+            ]
         ] = None
+        self.prefix = prefix or "client"
```
nsheaps: doubly sure about the value in case someone passes "None" to this

nsheaps: Note that it also defaults non-local spans to `generic_client`

tests/integration/requests_tests.py
```diff
@@ -269,12 +269,10 @@ def test_prometheus_http_client_metrics_inside_local_span(method, http_server):
...
     # we need to clear metrics between test runs from the parametrize annotations
-    REGISTRY._names_to_collectors["http_client_requests_total"].clear()
-    REGISTRY._names_to_collectors["http_client_latency_seconds"].clear()
-    REGISTRY._names_to_collectors["http_client_active_requests"].clear()
-    REGISTRY._names_to_collectors["local_span_latency_seconds"].clear()
-    REGISTRY._names_to_collectors["local_span_requests_total"].clear()
-    REGISTRY._names_to_collectors["local_span_active_requests"].clear()
+    for collector_name in REGISTRY._names_to_collectors:
+        reg = REGISTRY._names_to_collectors[collector_name]
+        if hasattr(reg, "clear") and callable(getattr(reg, "clear")):
```
nsheaps, line 274: Apparently there's a garbage collector in this list that doesn't have `clear()`

This belongs in a setup method but I didn't feel like refactoring this whole file for a test class that has a setup method. The reason this changed is that `local` spans no longer had a metric created for them at this point. This also helps future runs to ensure they're all cleared.

tests/integration/requests_tests.py
```diff
@@ -320,7 +318,7 @@ def test_prometheus_http_client_metrics_inside_local_span(method, http_server):
 
     # checking that we got the local span metrics we expected
     local_span_latency_seconds = REGISTRY._names_to_collectors[
-        "local_span_latency_seconds"
```
nsheaps, line 323: ⚠️  `local_span` => `local`

baseplate/observers/prometheus.py
nsheaps: Discussed in slack, there are certain metrics that look a heck of a lot like service names but aren't actually. Those metrics (like the ones below) are fine.

baseplate/clients/redis_cluster.py
```diff
@@ -430,7 +430,7 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
         command = args[0]
         trace_name = f"{self.context_name}.{command}"
 
-        with self.server_span.make_child(trace_name):
+        with self.server_span.make_child(trace_name, component_name="rediscluster_client"):
```
nsheaps, line 433: I was guessing at some of these names

baseplate/clients/requests.py
```diff
@@ -210,7 +210,9 @@ def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
             "http.url": request.url,
             "http.slug": self.client_name if self.client_name is not None else self.name,
         }
-        with self.span.make_child(f"{self.name}.request").with_tags(tags) as span:
+        with self.span.make_child(f"{self.name}.request", component_name="http_client").with_tags(
```
nsheaps, line 213: The protocol tag forces the metrics to come with this name, but I thought I'd be consistent

baseplate/clients/memcache/__init__.py
KTAtkinson: What would that look like Nate, can you describe a little more?

baseplate/observers/prometheus.py
```diff
@@ -155,11 +166,11 @@ def set_metrics_by_protocol(self) -> None:
         elif self.protocol == "thrift":
             self.metrics = PrometheusThriftClientMetrics()
         else:
-            logger.warning(
-                "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",
+            logger.debug(
+                "Unknown protocol %s. Expected 'http' or 'thrift' protocol.",
```
KTAtkinson: :no_bell: 
```suggestion
                "Unknown protocol %s. 'http' and 'thrift' are supported.",
```

baseplate/observers/prometheus.py
```diff
@@ -71,10 +74,11 @@ def set_metrics_by_protocol(self) -> None:
         elif self.protocol == "http":
             self.metrics = PrometheusHTTPServerMetrics()
         else:
-            logger.warning(
-                "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",
+            logger.debug(
+                "Unknown protocol %s. Expected 'http' or 'thrift' protocol.",
```
KTAtkinson: :no_bell: 
```suggestion
                "Unknown protocol %s. 'http' and 'thrift' are supported.",
```

KTAtkinson (approved):

baseplate/clients/memcache/__init__.py
nsheaps: Right now, make_server_span goes through the server span observer where it looks at protocol and determines the metrics. Lass to use for that. In make_child it then treats them as client or local sounds (depending on an input var) but the metric names themselves come from component_name in the span.

When I used datadog, spans had component names too but it was more in reference to the mechanism used to collect the metrics than the metrics it was tracking (so you may have multiple request libs and differentiating the metrics from them would be via the component even if they're all http spans as an example). The thought here was to separate out the component name from the prefix from metrics to be able to identify that better, letting it come potentially from an extension of the client metrics class

PR #676, opened 2022-05-20
Refactor and test alternate live data loaders

TylerLubeck:
This is an alternate version of #674 .

In this version, I've pulled some things apart so I can more thoroughly test them. I split them up because this is a more complex change, but I prefer this one over the other.

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -85,7 +85,13 @@ class NodeWatcher:
         # }
         # If the load type is 'S3', this format is an indication that we support
         # downloading the contents of encrypted files uploaded to S3.
-        if json_data.get("live_data_watcher_load_type") == "S3":
+        is_s3 = False
+        try:
+            is_s3 = json_data["live_data_watcher_load_type"] == "S3"
+        except (KeyError, TypeError):
+            is_s3 = False
+
+        if is_s3:
             if not S3_FETCHER_ENABLED:
                 logger.error("Attempted to use s3 fetcher but packages are missing")
                 return None
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -23,12 +23,17 @@ from baseplate.server import EnvironmentInterpolation
 
 import boto3  # type: ignore
 from botocore.client import ClientError  # type: ignore
+from botocore.exceptions import EndpointConnectionError  # type: ignore
 
 
 logger = logging.getLogger(__name__)
 
 
 HEARTBEAT_INTERVAL = 300
+
+
+class LoaderException(Exception):
+    pass
 
 
 class LoaderType(Enum):
@@ -65,7 +70,7 @@ class NodeWatcher:
         if loader_type == LoaderType.S3:
             try:
                 data = _load_from_s3(data)
-            except ValueError:
+            except LoaderException:
                 logger.error("Failed to load data from S3. Not writing to destination file")
                 return
 
@@ -119,7 +124,7 @@ def _load_from_s3(data: bytes) -> bytes:
                 "Failed to update live config: unable to fetch content from s3: source config has invalid or missing keys: %s.",
                 e.args[0]
         )
-        raise ValueError from e
+        raise LoaderException from e
 
     s3_client = boto3.client(
         "s3",
@@ -137,14 +142,17 @@ def _load_from_s3(data: bytes) -> bytes:
             error.response["Error"]["Message"],
         )
 
-        raise ValueError from error
+        raise LoaderException from error
+    except EndpointConnectionError as error:
+        logger.exception("Unable to retrieve object")
+        raise LoaderException from error
     except ValueError as error:
         logger.exception(
             "Failed to update live config: params for loading from S3 are incorrect. Received error: %s",
             error,
         )
 
-        raise ValueError from error
+        raise LoaderException from error
 
 
 def watch_zookeeper_nodes(zookeeper: KazooClient, nodes: Any) -> NoReturn:
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -104,7 +104,7 @@ def _parse_loader_type(data: bytes) -> LoaderType:
 
     try:
         loader_type = json_data["live_data_watcher_load_type"]
-    except (json.decoder.JSONDecodeError, KeyError, AttributeError, TypeError):
+    except (KeyError, AttributeError, TypeError):
         logger.debug(
             "Expected dict (JSON object) but got %s. Loading as PASSTHROUGH", str(type(json_data))
         )
@@ -114,6 +114,10 @@ def _parse_loader_type(data: bytes) -> LoaderType:
 
 
 def _load_from_s3(data: bytes) -> bytes:
+    # While many of the baseplate configurations use an ini format,
+    # we've opted for json in these internal-to-znode-configs because
+    # we want them to be fully controlled by the writer of the znode
+    # and json is an easier format for znode authors to work with.
     loader_config = json.loads(data.decode("UTF-8"))
     try:
         region_name = loader_config["region_name"]
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -32,13 +32,13 @@ HEARTBEAT_INTERVAL = 300
 
 
 class LoaderType(Enum):
-    RAW = 'RAW'
+    PASSTHROUGH = 'PASSTHROUGH'
     S3 = 'S3'
 
     @classmethod
     def _missing_(cls, value):
-        logger.error("Loader Type %s has not been implemented yet. Defaulting to RAW", value)
-        return cls.RAW
+        logger.error("Loader Type %s has not been implemented yet. Defaulting to PASSTHROUGH", value)
+        return cls.PASSTHROUGH
 
 
 class NodeWatcher:
@@ -63,7 +63,15 @@ class NodeWatcher:
 
         loader_type = _parse_loader_type(data)
         if loader_type == LoaderType.S3:
-            data = _load_from_s3(data)
+            try:
+                data = _load_from_s3(data)
+            except ValueError:
+                logger.error("Failed to load data from S3. Not writing to destination file")
+                return
+
+        if data is None:
+            logger.warning("No data to write to destination file. Something is likely misconfigured.")
+            return
 
         # swap out the file atomically so clients watching the file never catch
         # us mid-write.
@@ -73,6 +81,8 @@ class NodeWatcher:
             if self.owner and self.group:
                 os.fchown(tmpfile.fileno(), self.owner, self.group)
             os.fchmod(tmpfile.fileno(), self.mode)
+
+
             tmpfile.write(data)
         os.rename(self.dest + ".tmp", self.dest)
 
@@ -80,20 +90,20 @@ class NodeWatcher:
 def _parse_loader_type(data: Optional[bytes]) -> LoaderType:
     try:
         json_data = json.loads(data.decode("UTF-8"))
-    except json.decoder.JSONDecodeError:
-        logger.debug("Data is not valid JSON. Loading as RAW")
-        return LoaderType.RAW
+    except (UnicodeDecodeError, json.decoder.JSONDecodeError):
+        logger.debug("Data is not parseable as JSON, loading as PASSTHROUGH")
+        return LoaderType.PASSTHROUGH
 
     try:
         loader_type = json_data["live_data_watcher_load_type"]
     except (json.decoder.JSONDecodeError, KeyError, AttributeError, TypeError):
-        logger.debug("Data is not a JSON object like {\"live_data_watcher_load_type\":\"load_type\"}. Loading as RAW")
-        return LoaderType.RAW
+        logger.debug("Expected dict (JSON object) but got %s. Loading as PASSTHROUGH", str(type(json_data)))
+        return LoaderType.PASSTHROUGH
 
     return LoaderType(loader_type)
 
 
-def _load_from_s3(data: bytes) -> Optional[bytes]:
+def _load_from_s3(data: bytes) -> bytes:
     loader_config = json.loads(data.decode("UTF-8"))
     try:
         region_name = loader_config["region_name"]
@@ -109,7 +119,7 @@ def _load_from_s3(data: bytes) -> Optional[bytes]:
                 "Failed to update live config: unable to fetch content from s3: source config has invalid or missing keys: %s.",
                 e.args[0]
         )
-        return None
+        raise ValueError from e
 
     s3_client = boto3.client(
         "s3",
@@ -126,12 +136,15 @@ def _load_from_s3(data: bytes) -> Optional[bytes]:
             error.response["Error"]["Code"],
             error.response["Error"]["Message"],
         )
+
+        raise ValueError from error
     except ValueError as error:
         logger.exception(
             "Failed to update live config: params for loading from S3 are incorrect. Received error: %s",
             error,
         )
-    return None
+
+        raise ValueError from error
 
 
 def watch_zookeeper_nodes(zookeeper: KazooClient, nodes: Any) -> NoReturn:
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -41,7 +41,7 @@ class LoaderType(Enum):
     S3 = "S3"
 
     @classmethod
-    def _missing_(cls, value):
+    def _missing_(cls, value: object) -> "LoaderType":
         logger.error(
             "Loader Type %s has not been implemented yet. Defaulting to PASSTHROUGH", value
         )
@@ -95,7 +95,7 @@ class NodeWatcher:
         os.rename(self.dest + ".tmp", self.dest)
 
 
-def _parse_loader_type(data: Optional[bytes]) -> LoaderType:
+def _parse_loader_type(data: bytes) -> LoaderType:
     try:
         json_data = json.loads(data.decode("UTF-8"))
     except (UnicodeDecodeError, json.decoder.JSONDecodeError):
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -1,6 +1,7 @@
 """Watch nodes in ZooKeeper and sync their contents to disk on change."""
 import argparse
 import configparser
+from enum import Enum
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


baseplate/sidecars/live_data_watcher.py
```diff
@@ -35,78 +31,23 @@
 HEARTBEAT_INTERVAL = 300
 
 
+class LoaderType(Enum):
+    RAW = 'RAW'
```
nsheaps: 🔕  "Raw" to me implies some parsing of the data (like mime type), "default"/"zookeeper"/"direct" might be better terminology.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+        )
+        return None
+
+    s3_client = boto3.client(
+        "s3",
+        region_name=region_name,
+    )
+
+    try:
+        s3_object = s3_client.get_object(**s3_kwargs)
```
KTAtkinson, line 144: What error would be raised here if the object or bucket didn't exist?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+        # We require all of these keys to properly read from S3.
+        logger.exception(
+                "Failed to update live config: unable to fetch content from s3: source config has invalid or missing keys: %s.",
+                e.args[0]
+        )
+        return None
+
+    s3_client = boto3.client(
+        "s3",
+        region_name=region_name,
```
KTAtkinson, line 140: What happens here if the region is invalid?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+    try:
+        json_data = json.loads(data.decode("UTF-8"))
+    except json.decoder.JSONDecodeError:
+        logger.debug("Data is not valid JSON. Loading as RAW")
+        return LoaderType.RAW
+
+    try:
+        loader_type = json_data["live_data_watcher_load_type"]
+    except (json.decoder.JSONDecodeError, KeyError, AttributeError, TypeError):
+        logger.debug("Data is not a JSON object like {\"live_data_watcher_load_type\":\"load_type\"}. Loading as RAW")
```
nsheaps:
```suggestion
        logger.debug("Expected dict (JSON object) but got %s. Loading as RAW" % str(type(json_data)) )
```

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+        # Returns bytes.
+        return s3_object["Body"].read()
+    except ClientError as error:
+        logger.exception(
+            "Failed to update live config: failed to load data from S3. Received error code %s: %s",
+            error.response["Error"]["Code"],
+            error.response["Error"]["Message"],
+        )
+    except ValueError as error:
+        logger.exception(
```
nsheaps, line 159: 🔕  I think technically it's more pythonic to just raise an exception than to log it and return `None`. Doesn't make much of a difference to me.

tests/unit/sidecars/live_data_watcher_loader_tests.py
```diff
@@ -0,0 +1,104 @@
...
+from baseplate.sidecars.live_data_watcher import LoaderType
+
+@pytest.fixture()
+def start_mock_s3():
+    with mock_s3():
+        yield
+
+@pytest.mark.parametrize('data,return_value', (
+    # Non-RAW configs
+    (json.dumps({'live_data_watcher_load_type': 'S3'}).encode('utf-8'), LoaderType.S3),
```
nsheaps: Should probably have a test for this saying something other than S3

tests/unit/sidecars/live_data_watcher_loader_tests.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Everything else should be RAW
+    (b'', LoaderType.RAW),
+    (json.dumps({'key': 'value'}).encode('utf-8'), LoaderType.RAW),
+    (json.dumps([]).encode('utf-8'), LoaderType.RAW),
+    (json.dumps(1234).encode('utf-8'), LoaderType.RAW),
+    (json.dumps(None).encode('utf-8'), LoaderType.RAW),
```
KTAtkinson: What about the positive case? What if the contents don't parse as JSON at all?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+        loader_type = json_data["live_data_watcher_load_type"]
+    except (json.decoder.JSONDecodeError, KeyError, AttributeError, TypeError):
+        logger.debug("Data is not a JSON object like {\"live_data_watcher_load_type\":\"load_type\"}. Loading as RAW")
+        return LoaderType.RAW
+
+    return LoaderType(loader_type)
+
+
+def _load_from_s3(data: bytes) -> Optional[bytes]:
+    loader_config = json.loads(data.decode("UTF-8"))
```
KTAtkinson, line 121: :no_bell: I think it's likely a good idea that we are representing this data as JSON and not ini, but using ini would allow us to use the built in config parsing spec. It might be a good idea to add a comment here about why we are using this json and not ini.

tests/unit/sidecars/live_data_watcher_loader_tests.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert data == contents
+
+
+@pytest.mark.parametrize('exc', (
+    ValueError,
+    ClientError({'Error': {'Code': 'error', 'Message': 'msg'}}, 'operation'),
+))
+def test_unsuccessful_load_from_s3(start_mock_s3, exc):
+    with mock.patch('baseplate.sidecars.live_data_watcher.boto3.client') as client:
```
KTAtkinson: This use of a mock.MagicMock limits what you're able to test here. Did you consider using [`botocore.stub.Stubber`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html) instead?

KTAtkinson (changes requested):

baseplate/sidecars/live_data_watcher.py
nsheaps: 🔕  Do we use ini config formats for anything other than loading ini files?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +73,67 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+            "Failed to update live config: failed to load data from S3. Received error code %s: %s",
+            error.response["Error"]["Code"],
+            error.response["Error"]["Message"],
+        )
+    except ValueError as error:
+        logger.exception(
+            "Failed to update live config: params for loading from S3 are incorrect. Received error: %s",
+            error,
+        )
+    return None
```
nsheaps: I would argue that if it failed to load data, it should return an exception, not None, which is a valid piece of data.

tests/unit/sidecars/live_data_watcher_loader_tests.py
nsheaps: The positive test case is at the top. These are for negative parses, where it defaults back to RAW. Keep in mind that the data in S3 is not parsed, it's just dumped to a file.

baseplate/sidecars/live_data_watcher.py
TylerLubeck: If we raise an exception then we either have to handle it elsewhere or have the data watcher fail and exit.

IMO returning None means "we got no data" and that's a valid thing to write out as data.

I can be convinced otherwise pretty easily

nsheaps: "No data" is valid aka empty file (though I'd expect an empty file locally too)

We tried to get this and got an error (4xx, 5xx), I would not expect that to dump an empty file

nsheaps: If it were to return None, you'd just end up getting an exception when trying to write the file instead:

https://github.com/reddit/baseplate.py/pull/676/files#diff-f08bba0851fd42e4ae2a8d9fb6b983e377e46214ca4fbb35a3738f4d97665630R76

Screen Shot 2022-05-20 at 2 28 50 PM

TylerLubeck: hm ok I can see it. Another edge case to consider then:

Let's say the s3 object was created, we had watchers fetch it, then we deleted it from s3 and fetchers ran again. Now they've got an "object doesn't exist error" (or whatever the error is). Do we delete the existing local copy? Leave it alone but with old data?

tests/unit/sidecars/live_data_watcher_loader_tests.py
TylerLubeck: This is all testing parsing the znode data anyway. Positive case is at the top, still needs a 'non json' test case though. I'll add

baseplate/sidecars/live_data_watcher.py
nsheaps: I'd opt to leave it alone but make that error loud A F. We don't want to bring down a service because someone accidentally deleted something, but we also don't want that to go silently unnoticed.

tests/unit/sidecars/live_data_watcher_loader_tests.py
TylerLubeck: TIL about Stubber, I was looking at moto and couldn't find anything in there so fell back to mock

baseplate/sidecars/live_data_watcher.py
TylerLubeck: Honestly I didn't even consider not using JSON for this, writing out ini from scratch in code is a PITA in my experience. I don't have much of a reason other than that, but wowsers I don't wanna change it either

baseplate/sidecars/live_data_watcher.py
TylerLubeck: oh dang we handled this case in the previous form and I accidentally ripped it out, good catch

baseplate/sidecars/live_data_watcher.py
TylerLubeck: [This](https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3) is an admittedly old reference, but it appears to be a `ClientError`

baseplate/sidecars/live_data_watcher.py
TylerLubeck: `EndpointConnectionError`, I'm adding coverage for it now

baseplate/sidecars/live_data_watcher.py
```diff
@@ -20,93 +25,36 @@
...
+class LoaderException(Exception):
+    pass
+
+
+class LoaderType(Enum):
+    PASSTHROUGH = "PASSTHROUGH"
+    S3 = "S3"
+
+    @classmethod
+    def _missing_(cls, value: object) -> "LoaderType":
```
TylerLubeck, line 44: I don't like this type [but](https://github.com/python/typeshed/blob/master/stdlib/enum.pyi#L176)

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +90,76 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
-                tmpfile.write(data)
+
+            tmpfile.write(data)
         os.rename(self.dest + ".tmp", self.dest)
 
 
+def _parse_loader_type(data: bytes) -> LoaderType:
+    try:
+        json_data = json.loads(data.decode("UTF-8"))
+    except (UnicodeDecodeError, json.decoder.JSONDecodeError):
```
TylerLubeck, line 101: I ran the following against a copy of our zk instance and found `UnicodeDecodeError`, otherwise we're clean to load everything in there (even the things we don't actually intend to load)
 
```py
zk = kazooClient('ourcluster')
def parse_recursive(start='/'):
     print(f'Checking {start}')
     children  = zk.get_children(start)
     for c in children:
         parse_recursive(start + '/' + c)
     data, _ = zk.get(start)
     try:
         _parse_loader_type(data)
     except Exception as e:
         print(f'FAIL FAIL FAIL FAIL FAIL {e.__class__.__file__}: {str(e)} \n {data} \n')
         
   parse_recursive()
   ```

File: tests/unit/sidecars/live_data_watcher_loader_tests.py [hunk 13]
```diff
@@ -0,0 +1,125 @@
...
+        "file_key": "some_file",
+        "sse_key": "some_key",
+    }
+    data.pop(missing_config)
+
+    with pytest.raises(LoaderException):
+        assert _load_from_s3(json.dumps(data).encode("utf-8"))
+
+
+def test_successful_load_from_s3(start_mock_s3):
```
TylerLubeck: I suppose this could also be converted to Stubber from moto, @KTAtkinson do you have strong feelings?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -128,25 +90,76 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+        logger.debug("Data is not parseable as JSON, loading as PASSTHROUGH")
```
KTAtkinson, line 102: Do we know how often this log message might happen?

baseplate/sidecars/live_data_watcher.py
KTAtkinson: > no_bell Do we use ini config formats for anything other than loading ini files?

All application and sidecar configuration is in ini, and the [baseplate config library](https://github.com/reddit/baseplate.py/blob/develop/baseplate/lib/config.py) allows you to define a spec for your configuration that would take care of the validation below. I'm not aware of anyone using this outside of Baseplate application and sidecar configuration though,

KTAtkinson: > Honestly I didn't even consider not using JSON for this, writing out ini from scratch in code is a PITA in my experience. I don't have much of a reason other than that, but wowsers I don't wanna change it either

To be clear I don't should use ini. It might be worth pointing out that we are doing some of the things that the baseplate config library does. To me the most compelling reason to use json, and not ini is it's simpler. It allows someone who is writing data to zk to verify their data a lot easier.

tests/unit/sidecars/live_data_watcher_loader_tests.py
KTAtkinson: II would prefer using `Stubber` because it will catch misuses of the API. The way it works now, it assumes that any method or attribute you use on the client or it's return values is valid.

KTAtkinson (changes requested):

KTAtkinson:
I'm putting this proposal here for lack of a better place to put it. I have an alternate approach that might solve a few of the unknowns here.

What if all nodes whose data we expect to exist in S3 had node names in Zookeeper with an agreed upon prefix or suffix. This would allow us to test to make sure all nodes conform to the configuration spec. It would also reduce the surface area for errors. It would require that we update the Zookeeper node name in configuration files.which is not ideal, but that's easily done with a batch change or other mechanism.

The big unknown for me in this case is that I'm not sure how zk permission work.

nsheaps:
@KTAtkinson I disagree with the approach you mentioned with ZK data that's named specific for referring to where the data actually lives, in part due to the fact that this is being added to provide some flexibility in where the data is being stored as it is expanding in size as time goes on. Keeping the location specific allows the owner of the data to change it's location without having to update services to point to the new ZK node.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -120,6 +68,20 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
...
+                data = _load_from_s3(data)
+            except LoaderException:
+                logger.error("Failed to load data from S3. Not writing to destination file")
+                return
+
+        if data is None:
+            logger.warning(
+                "No data to write to destination file. Something is likely misconfigured."
+            )
+            return
```
nsheaps, line 83: would prematurely returning here cause an application to potentially fail if the file doesn't exist? Should it be an error (maybe tracked with a span/metric of some sort that can alert on an error, sentry would suffice if it's monitored) and self-correct to an empty string to write?

baseplate/sidecars/live_data_watcher.py
nsheaps: Screen Shot 2022-06-02 at 11 13 16 AM

Basically if the zk node is empty or contains a string that isn't wrapped in quotes

nsheaps, line 102: These are two different failure modes, feels like they should be split. When would the data not be utf-8 encoded?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -120,6 +68,20 @@ def on_change(self, data: Optional[bytes], _znode_stat: ZnodeStat) -> None:
             self.handle_empty_data()
             return
 
+        loader_type = _parse_loader_type(data)
```
nsheaps, line 71: 🔕  this may be better formatted as returning a tuple with the first element being the loader type and the second being the location to fetch from, if it's null then zk contains the data, otherwise use the return value in the _load_from_s3 call so you're not parsing `data` twice.

baseplate/sidecars/live_data_watcher.py
TylerLubeck: Only when it's binary data or something else shoved in to zookeeper. I found this one while running this method against all of our znodes - it only triggered for some znodes managed by other sources (flink, I _think_, but I don't remember) so we're unlikely to run in to it in our day-to-day, but I wanted to include it anyway because we've said that before

baseplate/sidecars/live_data_watcher.py
TylerLubeck: IMO that makes the code dirtier and is kind of unnecessary. 
- We know the biggest blob will be 1mb
- Most of our znodes are far far less than 1mb
- The goal of this is that anything approaching 1mb will be converted to s3 and thus the znode drops back to <<<<1mb
- This only triggers on startup or when znodes change, so not that often.

there's perhaps an argument about the startup cost of parsing 1mb of json but I don't have the napkin math on what that'd be.

If we're really worried about it, I'd prefer we go one of two routes
- use ujson or orjson or similar
- Bail on the enums and build out a class tree of loaders. This is a lot for 2 loaders though.

baseplate/sidecars/live_data_watcher.py
TylerLubeck: Potentially, but this matches behavior from before any of our S3 work: https://github.com/reddit/baseplate.py/blob/15b60b7f1100a1942299308e698b3db8993cd11e/baseplate/sidecars/live_data_watcher.py#L43

TylerLubeck: That is to say, I'm more worried about the opposite - are there use cases where we're relying on the fact that an empty znode creates no file?

tests/unit/sidecars/live_data_watcher_loader_tests.py
TylerLubeck: Alright, stubber is in place but we still need some mocking.

Because we create a client based on the znode data, we have to create it at runtime. because we create it at runtime, we can't inject our stubbed session/client easily. So I mock out the client create call to always return our stubbed session/client and then we can use stubber everywhere

nsheaps (approved):

KTAtkinson (approved):

PR #679, opened 2022-06-07
revert(prometheus): generic span metrics DEVX-637 DEVX-641

nsheaps:
Reverts #677
Reverts #675 

https://reddit.atlassian.net/browse/DEVX-641

Also lowers log level for the not-configured warning

## ✅ Checks

### DIFF
--- a/baseplate/observers/prometheus.py
+++ b/baseplate/observers/prometheus.py
@@ -71,7 +71,7 @@ class PrometheusServerSpanObserver(SpanObserver):
         elif self.protocol == "http":
             self.metrics = PrometheusHTTPServerMetrics()
         else:
-            logger.warning(
+            logger.debug(
                 "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",
                 self.protocol,
             )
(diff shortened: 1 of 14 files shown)


baseplate/observers/prometheus.py
```diff
@@ -74,11 +71,10 @@ def set_metrics_by_protocol(self) -> None:
         elif self.protocol == "http":
             self.metrics = PrometheusHTTPServerMetrics()
         else:
-            logger.debug(
-                "Unknown protocol %s. 'http' and 'thrift' are supported.",
+            logger.warning(
```
nsheaps: For when we circle back here, I don't think that doing this by protocol is a good idea, for reasons that include:
* what about things that don't have protocols
* What if someone adds a tag for protocol that matches this but don't intend on it
* design discussion about making spans should pass in the metrics class so it doesn't assume

KTAtkinson: Why are we increasing the log level here?
> what about things that don't have protocols

Maybe we should catch that case and document it and provide specific feedback to the client.

> What if someone adds a tag for protocol that matches this but don't intend on it

We could create constants to represent the protocols, this would necessitate a larger rewrite then we probably want to do here. I think it's ok for this issue to stand, though I think I would default to what Ben thinks here.

KTAtkinson (changes requested):

baseplate/observers/prometheus.py
nsheaps: I missed this one, I changed the others. Thanks for catching it.

KTAtkinson (approved):

PR #681, opened 2022-06-09
[BASE-21] Align the behavior of the prometheus observer with existing statsd behavior

Xaelias:
1/ TException were reported as success under statsd, but as failure for
prometheus. Now both report as success.

2/ Errors with status code other than 5xx were reported as success under
statsd, but as failure for prometheus. now both report as success.

Closes # BASE-21

unit / integration tests

## ✅ Checks

### DIFF
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -57,7 +57,6 @@ class _ContextAwareHandler:
                 span.set_tag("exception_type", "Error")
                 span.set_tag("thrift.status_code", exc.code)
                 span.set_tag("thrift.status", status)
-                span.set_tag("success", "false")
                 # mark 5xx errors as failures since those are still "unexpected"
                 if 500 <= exc.code < 600:
                     span.finish(exc_info=sys.exc_info())
@@ -66,7 +65,6 @@ class _ContextAwareHandler:
                 raise
             except TException as e:
                 span.set_tag("exception_type", type(e).__name__)
-                span.set_tag("success", "false")
                 # this is an expected exception, as defined in the IDL
                 span.finish()
                 raise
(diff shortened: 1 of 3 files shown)


baseplate/frameworks/thrift/__init__.py
```diff
@@ -57,7 +57,6 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 span.set_tag("exception_type", "Error")
                 span.set_tag("thrift.status_code", exc.code)
                 span.set_tag("thrift.status", status)
-                span.set_tag("success", "false")
```
nsheaps, line 60: ~Doesn't this still need to be set for 500-599?~

Nevermind, I see in the observer it will set success=false if there's an exception

Xaelias: No it happens automatically because 5xx use the `span.finish(exc_info=sys.exc_info())` call which sets `success` to `False` based on whether `exc_info` is passed or not.

nsheaps (changes requested):
Not actually rejecting this, keeping our review system happy...

You saw this reported with one of our services, can you do a manual integration test with this to be sure it works as expected and detail the steps in the PR description just to have high confidence that this will actually work?

Xaelias:
> Not actually rejecting this, keeping our review system happy...
> 
> You saw this reported with one of our services, can you do a manual integration test with this to be sure it works as expected and detail the steps in the PR description just to have high confidence that this will actually work?

I don't technically need to. The test that I update in there is a test that returns 404. That's why I had to change the success from false to true.

KTAtkinson (approved):

nsheaps (approved):

Xaelias:
> Not actually rejecting this, keeping our review system happy...
> 
> You saw this reported with one of our services, can you do a manual integration test with this to be sure it works as expected and detail the steps in the PR description just to have high confidence that this will actually work?

Added a 5xx test just to be on the safe side, and also show that we expect the behavior to be different :-)

nsheaps (approved):

PR #683, opened 2022-06-13
Add database client metrics

krav:
Closes [#BASE-30](https://reddit.atlassian.net/browse/BASE-30)

## 💸 TL;DR

Add database client latency and success graph. LMK if this is the correct approach to #BASE-22 or if we're taking a more generic one. 

## 📜 Details
```
...
# HELP database_client_latency_seconds Multiprocess metric
# TYPE database_client_latency_seconds histogram
database_client_latency_seconds_sum{statement="select 1"} 0.001254722
database_client_latency_seconds_bucket{le="0.0001",statement="select 1"} 0.0
database_client_latency_seconds_bucket{le="0.00025",statement="select 1"} 0.0
database_client_latency_seconds_bucket{le="0.000625",statement="select 1"} 0.0
database_client_latency_seconds_bucket{le="0.0015625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.00390625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.009765625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.0244140625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.06103515625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.152587890625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.3814697265625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="0.95367431640625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="2.384185791015625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="5.9604644775390625",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="14.901161193847656",statement="select 1"} 1.0
database_client_latency_seconds_bucket{le="+Inf",statement="select 1"} 1.0
database_client_latency_seconds_count{statement="select 1"} 1.0
...
# HELP bp_sqlalchemy_pool_max_size Multiprocess metric
# TYPE bp_sqlalchemy_pool_max_size gauge
bp_sqlalchemy_pool_max_size{pid="7",pool="sqlalchemy"} 0.0
# HELP bp_sqlalchemy_pool_idle_connections Multiprocess metric
# TYPE bp_sqlalchemy_pool_idle_connections gauge
bp_sqlalchemy_pool_idle_connections{pid="7",pool="sqlalchemy"} 0.0
# HELP bp_sqlalchemy_pool_active_connections Multiprocess metric
# TYPE bp_sqlalchemy_pool_active_connections gauge
bp_sqlalchemy_pool_active_connections{pid="7",pool="sqlalchemy"} 0.0
# HELP bp_sqlalchemy_pool_overflow_connections Multiprocess metric
# TYPE bp_sqlalchemy_pool_overflow_connections gauge
bp_sqlalchemy_pool_overflow_connections{pid="7",pool="sqlalchemy"} 0.0
# HELP thrift_server_active_requests Multiprocess metric
# TYPE thrift_server_active_requests gauge
...
# HELP database_client_active_requests Multiprocess metric
# TYPE database_client_active_requests gauge
database_client_active_requests{pid="7",statement="select 1"} 0.0
...
# HELP database_client_requests_total Multiprocess metric
# TYPE database_client_requests_total counter
database_client_requests_total{statement="select 1"} 1.0```
```

## ✅ Checks

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -226,6 +226,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
 
         trace_name = f"{context_name}.execute"
         span = server_span.make_child(trace_name)
+        span.set_tag("protocol", "sqlalchemy")
         span.set_tag("statement", statement[:1021] + "..." if len(statement) > 1024 else statement)
         span.start()
 
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -438,3 +438,59 @@ class PrometheusLocalSpanMetrics:
     @classmethod
     def get_active_requests_metric(cls) -> Gauge:
         return cls.active_requests
+
+
+class PrometheusSQLAlchemyClientMetrics:
+    prefix = "database_client"
+
+    labels = [
+        "statement",
+    ]
+
+    # Latency histogram of HTTP calls made by clients
+    # buckets are defined above (from 100µs to ~14.9s)
+    latency_seconds = Histogram(
+        f"{prefix}_latency_seconds",
+        "Latency histogram of database calls made by client",
+        labels,
+        buckets=default_latency_buckets,
+    )
+    # Counter counting total HTTP requests started by a given client
+    requests_total = Counter(
+        f"{prefix}_requests_total",
+        "Total number of database requests started by a given client",
+        labels,
+    )
+    # Gauge showing current number of active requests by a given client
+    active_requests = Gauge(
+        f"{prefix}_active_requests",
+        "Number of active requests for a given client",
+        labels,
+    )
+
+    def __init__(self) -> None:
+        pass
+
+    @classmethod
+    def latency_seconds_metric(cls, tags: Dict[str, Any]) -> Histogram:
+        return cls.latency_seconds.labels(statement=tags.get("statement", ""))
+
+    @classmethod
+    def requests_total_metric(cls, tags: Dict[str, Any]) -> Counter:
+        return cls.requests_total.labels(statement=tags.get("statement", ""))
+
+    @classmethod
+    def active_requests_metric(cls, tags: Dict[str, Any]) -> Gauge:
+        return cls.active_requests.labels(statement=tags.get("statement", ""))
+
+    @classmethod
+    def get_latency_seconds_metric(cls) -> Histogram:
+        return cls.latency_seconds
+
+    @classmethod
+    def get_requests_total_metric(cls) -> Counter:
+        return cls.requests_total
+
+    @classmethod
+    def get_active_requests_metric(cls) -> Gauge:
+        return cls.active_requests
--- a/baseplate/observers/prometheus.py
+++ b/baseplate/observers/prometheus.py
@@ -137,7 +137,11 @@ class PrometheusClientSpanObserver(SpanObserver):
         self.tags: Dict[str, Any] = {}
         self.start_time: Optional[int] = None
         self.metrics: Optional[
-            Union[PrometheusHTTPClientMetrics, PrometheusThriftClientMetrics]
+            Union[
+                PrometheusHTTPClientMetrics,
+                PrometheusThriftClientMetrics,
+                PrometheusSQLAlchemyClientMetrics,
+            ]
         ] = None
 
     def on_set_tag(self, key: str, value: Any) -> None:
--- a/baseplate/observers/prometheus.py
+++ b/baseplate/observers/prometheus.py
@@ -15,6 +15,7 @@ from baseplate import SpanObserver
 from baseplate.lib.prometheus_metrics import PrometheusHTTPClientMetrics
 from baseplate.lib.prometheus_metrics import PrometheusHTTPServerMetrics
 from baseplate.lib.prometheus_metrics import PrometheusLocalSpanMetrics
+from baseplate.lib.prometheus_metrics import PrometheusSQLAlchemyClientMetrics
 from baseplate.lib.prometheus_metrics import PrometheusThriftClientMetrics
 from baseplate.lib.prometheus_metrics import PrometheusThriftServerMetrics
 from baseplate.thrift.ttypes import Error
@@ -154,6 +155,8 @@ class PrometheusClientSpanObserver(SpanObserver):
             self.metrics = PrometheusHTTPClientMetrics()
         elif self.protocol == "thrift":
             self.metrics = PrometheusThriftClientMetrics()
+        elif self.protocol == "sqlalchemy":
+            self.metrics = PrometheusSQLAlchemyClientMetrics()
         else:
             logger.debug(
                 "No valid protocol set for Prometheus metric collection, metrics won't be collected. Expected 'http' or 'thrift' protocol. Actual protocol: %s",


baseplate/clients/sqlalchemy.py
```diff
@@ -226,6 +226,7 @@ def on_before_execute(
 
         trace_name = f"{context_name}.execute"
         span = server_span.make_child(trace_name)
+        span.set_tag("protocol", "sqlalchemy")
```
nsheaps, line 229: sqlalchemy isn't a protocol, it's a package providing an API to interface python=>SQL, sqlalchemy is just an implementation detail.

baseplate/lib/prometheus_metrics.py
```diff
@@ -438,3 +438,59 @@ def get_requests_total_metric(cls) -> Counter:
     @classmethod
     def get_active_requests_metric(cls) -> Gauge:
         return cls.active_requests
+
+
+class PrometheusSQLAlchemyClientMetrics:
+    prefix = "database_client"
```
nsheaps, line 444: There are lots of different kinds of databases, please name this more specifically, both the class and the prefix. 

```suggestion
class PrometheusSQLClientMetrics:
    prefix = "sql_client"
```

baseplate/lib/prometheus_metrics.py
```diff
@@ -438,3 +438,59 @@ def get_requests_total_metric(cls) -> Counter:
...
+
+    labels = [
+        "statement",
+    ]
+
+    # Latency histogram of HTTP calls made by clients
+    # buckets are defined above (from 100µs to ~14.9s)
+    latency_seconds = Histogram(
+        f"{prefix}_latency_seconds",
```
nsheaps, line 453: There is still an outstanding question of whether or not we should have metrics for `trace_*` vs `{prefix}_*`.

baseplate/clients/sqlalchemy.py
nsheaps, line 229: The more I think about it, the less I like using the protocol in this way, especially since "protocol" isn't even really descriptive here since "TCP" is just as valid as "SQL". 

Rather than having all metrics classes defined in `prometheus_metrics.py` I would like to see a new function on spans for setting the metrics class and then having it defined with where spans are being opened. In a sense, move the metrics class here, then use `span.set_metrics_cls(PrometheusSqlClientMetrics)`

That said, metrics class are also overly verbose.
1. I think that if you don't call `set_metrics_cls` that it should use a dynamically generated metrics class (as I had in my PR that was reverted out)
2. I like the functionality in metrics_tagged.py where certain tags are whitelisted to become labels on the metrics, which I think would reduce the verbosity here of needing to define a metrics class for every time we want metrics. This has the (perhaps positive) side effect of requiring the metrics classes be instantiated, which offers an initialization point where we can do things like setting the gauges and counters with the prefix rather than having to make new classes with class fields referring to the counters+gauges.

nsheaps (changes requested):
We have open outstanding questions on the naming schema of metrics in one of the attached tickets (`baseplate_client_*` vs `{prefix}_*`), and I think continuing this pattern to complete implementation makes it extremely verbose for the clients we have, and not flexible enough for span creators outside of the baseplate.py lib to track metrics on service-implemented spans. 

I think we need the following changes:

- make span creation take in/have a method for passing the metrics class rather than interpreting from the protocol / the protocol shouldn't be used for keying the metrics class
- the protocol shouldn't be the name of the package used to achieve the interface, it should be the actual protocol
- metrics classes should be instantiated in such a way that tags => labels are a bit more automatic via a whitelist
- Spans that don't pass a metrics class should still have generic metrics tracked (out of scope for this PR I think)

baseplate/lib/prometheus_metrics.py
bradengroom: In our current metrics, there are tags for the client name. This tag is populated using the same name used to attach the client to the baseplate context. This allows us to have metrics for multiple postgres databases in the same service. Will this continue to be possible?

baseplate/clients/sqlalchemy.py
krav: I like something like this a lot more, and should be straight forward to do. I was away when that PR got introduced and reverted, can you describe what went wrong?

krav: Technically neither is sql :P I think "protocol" is a misnomer regardless since the coupling between this class and the prometheus observer is tight: we can't expect any other class implementing "sql" or "database" to expose the same tags as the sqlalchemy client. This supports your idea below of getting rid of this tag.

krav: 🤔 Some thinking out loud alternatives:

1) (Current) Set a "protocol" tag on every span and have special cases in a generic Prometheus observer.

Weird dispatcher in the Prom observer, metric classes need to know about client/server implementation details like what tags there are and what they mean.

2) (This) Ditch the dispatching Prometheus observer and instead have Prom observer classes specific to a client or server (or subset of those). 

This acknowledges the tight coupling between client/server details and prometheus, and gets rid of the dispatcher. 

Would we want to keep the de facto interface we currently have on metric classes of `active_requests_metric`, `latency_seconds_metric` and so on? Are we sure it always makes sense, looking at the current things we have in Wavefront?

We'll probably still end up with lots of classes that are very similar except for some details like what tags to look at, or latency distribution. Instead of passing a metric class here we could pass an instance of a class, and then potentially change things like the distribution buckets from the client/server class without making a new metric class.

```
span.set_metrics_cls(PrometheusSqlClientMetrics(latency_buckets={...}))
```

3) Take it a step further and not use the observer pattern, and just build prom into the client/server classes.

Easy, less code, can target what we actually need. Introduces a specific dependency on prom being imported (even if it does nothing when prom is not enabled) which some people don't like? Spans outside baseplate would need to implement prom themselves (good?!).

4) Keep the dispatching Prometheus observer, remove the protocol tag and instead inspect the call stack to see where the span is from.

Keeps the span/observer abstraction clean, not needing to change every client and server. Still reliant on the prom classes knowing implementation details of the clients, probably a bit too magic.

baseplate/lib/prometheus_metrics.py
krav: Good point, adding a label for the endpoint used is definitely needed.

baseplate/clients/sqlalchemy.py
Xaelias: I don't know that I agree. This is our instrumentation. All database clients for instance should be able to expose the same type of metrics like connections, time to completion, errors etc.
Labels are just what we make them. What is specific to sqlalchemy in the current implementation?

krav: Fair enough. What about if we're targeting a slow database and want different latency buckets?

Xaelias: > metric classes need to know about client/server implementation details like what tags there are and what they mean

interesting I've always seen it the other way

The metric class dictates what you need to expose. The client is responsible to make sure to expose these to you. Not the other way around.

baseplate/lib/prometheus_metrics.py
krav: I like this idea!

baseplate/clients/sqlalchemy.py
nsheaps: The "protocol" here doesn't change the answer to your question, but I think you answered it in an earlier comment: the combination of statement and db/db_host would allow us to identify slower queries.

baseplate/lib/prometheus_metrics.py
nsheaps: Would also add that the db host and db itself might be a good label to have as well (assuming "endpoint" != db connection string which would show both, but I think having it split would be better for the future for "errors by DB host" vs by DB)

The point I was trying to make is that this metrics class is more for SQL than it is for something like a redis to Cassandra DB.

baseplate/lib/prometheus_metrics.py
nsheaps: There's a high likelyhood that we'll be erring on the side of `{prefix}_*` for now and introduce a singular metric down the line. Topic for discussion tomorrow though.

baseplate/clients/sqlalchemy.py
KTAtkinson: > Take it a step further and not use the observer pattern, and just build prom into the client/server classes.
Easy, less code, can target what we actually need. Introduces a specific dependency on prom being imported (even if it does nothing when prom is not enabled) which some people don't like? Spans outside baseplate would need to implement prom themselves (good?!).

I like this a lot it's more consistent with how prometheus thinks about things and it allows clients to define their metrics when they initialize which is typically when the server starts. I'm a little fuzzy on what this would look like in terms of tracking server metrics though. 

A downside to this approach is it would throw out a lot of the work that we've already done, and unless de design an API around providing a default set of metrics, clients will need to each define the same set of metrics explicitly.

KTAtkinson (changes requested):

baseplate/lib/prometheus_metrics.py
```diff
@@ -438,3 +438,59 @@ def get_requests_total_metric(cls) -> Counter:
...
     def get_active_requests_metric(cls) -> Gauge:
         return cls.active_requests
+
+
+class PrometheusSQLAlchemyClientMetrics:
+    prefix = "database_client"
+
+    labels = [
+        "statement",
+    ]
```
Xaelias, line 448: I feel like there should be more labels than this. Like endpoint and stuff.  
What's statement exactly btw?

PR #684, opened 2022-06-16
[base-32] instrument sqlalchemy client

krav:
Closes BASE-32

## 💸 TL;DR
adds latency and rate metrics to database client

## 📜 Details

example:

```
databasebp_latency_seconds_sum{database="alt_text",host="postgresql",success="true"} 0.0013580322265625
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.005",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.01",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.025",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.05",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.075",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.1",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.25",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.5",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="0.75",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="1.0",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="2.5",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="5.0",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="7.5",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="10.0",success="true"} 1.0
databasebp_latency_seconds_bucket{database="alt_text",host="postgresql",le="+Inf",success="true"} 1.0
databasebp_latency_seconds_count{database="alt_text",host="postgresql",success="true"} 1.0
databasebp_pool_max_size{pid="7",pool="sqlalchemy"} 0.0
databasebp_pool_idle_connections{pid="7",pool="sqlalchemy"} 0.0
databasebp_pool_active_connections{pid="7",pool="sqlalchemy"} 0.0
databasebp_pool_overflow_connections{pid="7",pool="sqlalchemy"} 0.0
databasebp_requests_total{database="alt_text",host="postgresql",success="true"} 1.0
```

## ✅ Checks

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -1,6 +1,6 @@
 import re
 
-from time import time
+from time import perf_counter
 from typing import Any
 from typing import Dict
 from typing import Optional
@@ -27,6 +27,7 @@ from baseplate import SpanObserver
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
 from baseplate.lib import metrics
+from baseplate.lib.prometheus_metrics import default_latency_buckets
 from baseplate.lib.secrets import SecretsStore
 
 
@@ -197,7 +198,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         f"{PROM_PREFIX}_latency_seconds",
         "Latency histogram of calls to database",
         PROM_LABELS,
-        # buckets=default_latency_buckets, # TODO: make configurable (would require configuring PROM_PREFIX too?)
+        buckets=default_latency_buckets,
     )
 
     requests_total = Counter(
@@ -246,7 +247,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         executemany: bool,
     ) -> Tuple[str, Parameters]:
         """Handle the engine's before_cursor_execute event."""
-        self.time_started = time()
+        self.time_started = perf_counter()
 
         context_name = conn._execution_options["context_name"]
         server_span = conn._execution_options["server_span"]
@@ -288,7 +289,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         }
 
         self.requests_total.labels(**labels).inc()
-        self.latency_seconds.labels(**labels).observe(time() - self.time_started)
+        self.latency_seconds.labels(**labels).observe(perf_counter() - self.time_started)
 
     def on_error(self, context: ExceptionContext) -> None:
         """Handle the event which happens on exceptions during execution."""
@@ -303,7 +304,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         }
 
         self.requests_total.labels(**labels).inc()
-        self.latency_seconds.labels(**labels).observe(time() - self.time_started)
+        self.latency_seconds.labels(**labels).observe(perf_counter() - self.time_started)
 
 
 class SQLAlchemySessionContextFactory(SQLAlchemyEngineContextFactory):
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -170,28 +170,16 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         PROM_POOL_LABELS,
     )
 
-    checked_in_connections_gauge = Gauge(
-        f"{PROM_POOL_PREFIX}_idle_connections",
-        "Number of available, checked in, connections in this pool",
-        PROM_POOL_LABELS,
-    )
-
     checked_out_connections_gauge = Gauge(
-        f"{PROM_POOL_PREFIX}_active_connections",
-        "Number of connections in use, or checked out, in this pool",
-        PROM_POOL_LABELS,
-    )
-
-    overflow_connections_gauge = Gauge(
-        f"{PROM_POOL_PREFIX}_overflow_connections",
-        "Number of connections over the desired size of this pool",
+        f"{PROM_POOL_PREFIX}_client_connections",
+        "Number of connections in use by this pool (checked out + overflow)",
         PROM_POOL_LABELS,
     )
 
     PROM_LABELS = [
-        "host",
-        "database",
-        "success",
+        "sql_address",
+        "sql_database",
+        "sql_success",
     ]
 
     latency_seconds = Histogram(
@@ -218,9 +206,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         pool = self.engine.pool
         if isinstance(pool, QueuePool):
             self.max_connections_gauge.labels(name).set_function(pool.size)
-            self.checked_in_connections_gauge.labels(name).set_function(pool.checkedin)
-            self.checked_out_connections_gauge.labels(name).set_function(pool.checkedout)
-            self.overflow_connections_gauge.labels(name).set_function(pool.overflow)
+            self.checked_out_connections_gauge.labels(name).set_function(pool.checkedout + pool.overflow)
 
     def report_runtime_metrics(self, batch: metrics.Client) -> None:
         pool = self.engine.pool
@@ -283,9 +269,9 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         conn.info["span"] = None
 
         labels = {
-            "host": conn.engine.url.host,
-            "database": conn.engine.url.database,
-            "success": "true",
+            "sql_address": conn.engine.url.host,
+            "sql_database": conn.engine.url.database,
+            "sql_success": "true",
         }
 
         self.requests_total.labels(**labels).inc()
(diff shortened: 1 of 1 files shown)


baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +159,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
 
     """
 
-    PROM_PREFIX = "bp_sqlalchemy_pool"
-    PROM_LABELS = ["pool"]
+    PROM_PREFIX = "databasebp"
```
KTAtkinson: Maybe this should be `postgres`, if we want to denote with a tag that this is a baseplate postgres client we can do that. I have no strong opinion either way there.

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +159,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
+    PROM_POOL_PREFIX = f"{PROM_PREFIX}_pool"
+    PROM_POOL_LABELS = ["pool"]
```
KTAtkinson, line 165: Are there more labels that should be global? Maybe host or prefix. I say this because there is nothing in this client that would prevent a user from having more than one client in one application.

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +159,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
     )
 
     overflow_connections_gauge = Gauge(
-        f"{PROM_PREFIX}_overflow_connections",
+        f"{PROM_POOL_PREFIX}_overflow_connections",
         "Number of connections over the desired size of this pool",
+        PROM_POOL_LABELS,
+    )
+
+    PROM_LABELS = [
```
KTAtkinson, line 179: If we include prefix above we probably want to include it here too.

KTAtkinson (changes requested):

baseplate/clients/sqlalchemy.py
krav: Hm, it could be sqlite or another dialect, so I'd rather keep it generic. `bp` tacked on because that's what baseplate.go does. 

If the dialect is interesting I'd rather have dialect as a label, since that makes automating dashboards easier.

baseplate/clients/sqlalchemy.py
krav: The label `pool` here is the name of the pool which can be used to differentiate them ([more than what we do for statsd apparently](https://github.com/reddit/baseplate.py/blob/090a39690c02dbdded775c97b78d9c36106f3d6d/baseplate/clients/sqlalchemy.py#L229-L232)). 

These metrics are sourced [from the pool](https://github.com/reddit/baseplate.py/blob/090a39690c02dbdded775c97b78d9c36106f3d6d/baseplate/clients/sqlalchemy.py#L219-L223) and I don't see I way to easily get to an engine to get the host or database. I'm not certain we can always know they're the same for every connection, or what we would send if the pool is empty.

baseplate/clients/sqlalchemy.py
krav: I think differentiating information about the connection pool and the queries is fine, as they are different things. Not sure what you'd use the pool label for? Most of the time it will just be "sqlalchemy".

baseplate/clients/sqlalchemy.py
Xaelias: hm... I feel like `baseplate_db_...` or `baseplate_database_...` or `baseplatedb_...` even would make more sense as a prefix. But if baseplate.go already made that decision for us...

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +159,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
+        "host",
+        "database",
+        "success",
+    ]
+
+    latency_seconds = Histogram(
+        f"{PROM_PREFIX}_latency_seconds",
+        "Latency histogram of calls to database",
+        PROM_LABELS,
+        # buckets=default_latency_buckets, # TODO: make configurable (would require configuring PROM_PREFIX too?)
```
Xaelias: I wouldn't mention this in code.

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +159,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
-        f"{PROM_PREFIX}_active_connections",
+        f"{PROM_POOL_PREFIX}_active_connections",
         "Number of connections in use, or checked out, in this pool",
-        PROM_LABELS,
+        PROM_POOL_LABELS,
     )
 
     overflow_connections_gauge = Gauge(
-        f"{PROM_PREFIX}_overflow_connections",
+        f"{PROM_POOL_PREFIX}_overflow_connections",
```
nsheaps: if a pool is a certain size, how do you end up using more than the size allows for?

baseplate/clients/sqlalchemy.py
krav: Thanks for spotting this! Will remove and go with our standard buckets.

baseplate/clients/sqlalchemy.py
krav: Maybe we should have a step to validate the family names as part of cutting the release? They seem to take up an inordinate amount of time atm. Not a big fan of the `bp` suffix either btw

baseplate/clients/sqlalchemy.py
krav: [That's a question for sqlalchemy heads](https://docs.sqlalchemy.org/en/14/search.html?q=max_overflow&check_keywords=yes&area=default). I know "max overflow" is configurable, but don't know the semantics of it.

maeivysea (approved):

baseplate/clients/sqlalchemy.py
nsheaps: @krav we shouldn't be merging them if the metric name is incorrect, I think a release is too late.

With regards to the prefix, lets go with `sql` based on the current spec iteration

KTAtkinson (changes requested):
Looks good from a Python perspective, please update naming to be consistent with the Baseplate spec as talked about off-line.

baseplate/clients/sqlalchemy.py
nsheaps: Current spec reads as the prefix (or namespace, lets be consistent about what we're calling it) to be `sql`

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +160,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
         "Number of connections over the desired size of this pool",
+        PROM_POOL_LABELS,
+    )
```
nsheaps: The [spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md?plain=1#L185) was updated to have max_size and client_connections (which is active + overflow, meaning client_connections can be > max_connections)

baseplate/clients/sqlalchemy.py
krav: Fixed!

baseplate/clients/sqlalchemy.py
```diff
@@ -156,30 +160,50 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
...
+
+    PROM_LABELS = [
```
nsheaps, line 179: The spec was updated to prefix these and rename host -> address

baseplate/clients/sqlalchemy.py
```diff
@@ -254,12 +282,30 @@ def on_after_execute(
...
         context.connection.info["span"].finish(exc_info=exc_info)
         context.connection.info["span"] = None
 
+        labels = {
+            "host": context.connection.engine.url.host,
+            "database": context.connection.engine.url.database,
+            "success": "false",
+        }
+
+        self.requests_total.labels(**labels).inc()
```
nsheaps, line 294: I like passing a kwarg dict here, it keeps values explicitly tied to their keys so ordering doesn't matter as much.

nsheaps (changes requested):
Names need to update to match the spec

baseplate/clients/sqlalchemy.py
krav: Fixed!

baseplate/clients/sqlalchemy.py
krav: I've changed it to overflow + checked out, PTAL

PR #686, opened 2022-06-16
[BASE-34] Add prometheus metrics for cassandra client

Xaelias:
Closes #BASE-34

Manual testing + unit

## ✅ Checks

## Manual test:
```
[...]
"cass_local": CassandraClient("mykeyspace"),
[...]
with baseplate.server_context('http_client') as context:
    with context.span.make_child('cassandra', local=True, component_name='myKeySpace') as span:
        logger.error(span.context)
        for row in span.context.cass_local.execute("DESCRIBE keyspaces;"):
            logger.error(row)
        try:
            span.context.cass_local.execute("SELECT 1;", query_name="failure")
        except:
            pass
        for row in span.context.cass_local.execute_async("DESCRIBE keyspaces;").result():
            logger.warning(row)

```

metrics
```
# HELP cassandra_client_requests_active Multiprocess metric
# TYPE cassandra_client_requests_active gauge
cassandra_client_requests_active{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",pid="67962"} 0.0
cassandra_client_requests_active{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",pid="67962"} 0.0
# HELP cassandra_client_requests_total Multiprocess metric
# TYPE cassandra_client_requests_total counter
cassandra_client_requests_total{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true"} 2.0
cassandra_client_requests_total{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false"} 1.0
# HELP cassandra_client_latency_seconds Multiprocess metric
# TYPE cassandra_client_latency_seconds histogram
cassandra_client_latency_seconds_sum{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true"} 0.004405794999999824
cassandra_client_latency_seconds_sum{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false"} 0.0018226910000000984
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.0001"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.00025"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.000625"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.0015625"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.00390625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.009765625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.0244140625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.06103515625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.152587890625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.3814697265625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="0.95367431640625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="2.384185791015625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="5.9604644775390625"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="14.901161193847656"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true",le="+Inf"} 2.0
cassandra_client_latency_seconds_count{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="",cassandra_success="true"} 2.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.0001"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.00025"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.000625"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.0015625"} 0.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.00390625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.009765625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.0244140625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.06103515625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.152587890625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.3814697265625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="0.95367431640625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="2.384185791015625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="5.9604644775390625"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="14.901161193847656"} 1.0
cassandra_client_latency_seconds_bucket{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false",le="+Inf"} 1.0
cassandra_client_latency_seconds_count{cassandra_contact_points="localhost",cassandra_keyspace="mykeyspace",cassandra_query_name="failure",cassandra_success="false"} 1.0
```

### DIFF
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -7,6 +7,7 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Mapping
+from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
@@ -32,27 +33,28 @@ from baseplate.lib import config
 from baseplate.lib.prometheus_metrics import default_latency_buckets
 from baseplate.lib.secrets import SecretsStore
 
-cassandra_labels = [
-    "cassandra_contact_points",
-    "cassandra_keyspace",
-    "cassandra_query_name",
-]
-cassandra_labels_w_success = cassandra_labels + ["cassandra_success"]
+
+class CassandraPrometheusLabels(NamedTuple):
+    cassandra_contact_points: str
+    cassandra_keyspace: str
+    cassandra_query_name: str
+
+
 REQUEST_TIME = Histogram(
     "cassandra_client_latency_seconds",
     "Time spent executing cassandra query",
-    cassandra_labels_w_success,
+    CassandraPrometheusLabels._fields + ("cassandra_success",),
     buckets=default_latency_buckets,
 )
 REQUEST_ACTIVE = Gauge(
-    "cassandra_client_requests_active",
+    "cassandra_client_active_requests",
     "Current number of active cassandra queries",
-    cassandra_labels,
+    CassandraPrometheusLabels._fields,
 )
 REQUEST_TOTAL = Counter(
     "cassandra_client_requests_total",
     "Total number of cassandra queries",
-    cassandra_labels_w_success,
+    CassandraPrometheusLabels._fields + ("cassandra_success",),
 )
 
 if TYPE_CHECKING:
@@ -258,32 +260,38 @@ def wrap_future(
     return response_future
 
 
-def _on_execute_complete(
-    _result: Any, args: Tuple[Span, float, Tuple[Any, ...]], event: Event
-) -> None:
+class CassandraCallbackArgs(NamedTuple):
+    span: Span
+    start_time: float
+    prom_labels: CassandraPrometheusLabels
+
+
+def _on_execute_complete(_result: Any, args: CassandraCallbackArgs, event: Event) -> None:
     # TODO: tag with anything from the result set?
     # TODO: tag with any returned warnings
-    span, start_time, prom_labels = args
     try:
-        span.finish()
+        args.span.finish()
     finally:
-        REQUEST_TIME.labels(*prom_labels, "true").observe(time.perf_counter() - start_time)
-        REQUEST_TOTAL.labels(*prom_labels, "true").inc()
-        REQUEST_ACTIVE.labels(*prom_labels).dec()
+        prom_labels = args.prom_labels._asdict()
+        REQUEST_TIME.labels(**prom_labels, cassandra_success="true").observe(
+            time.perf_counter() - args.start_time
+        )
+        REQUEST_TOTAL.labels(**prom_labels, cassandra_success="true").inc()
+        REQUEST_ACTIVE.labels(**prom_labels).dec()
         event.set()
 
 
-def _on_execute_failed(
-    exc: BaseException, args: Tuple[Span, float, Tuple[Any, ...]], event: Event
-) -> None:
-    span, start_time, prom_labels = args
+def _on_execute_failed(exc: BaseException, args: CassandraCallbackArgs, event: Event) -> None:
     try:
         exc_info = (type(exc), exc, None)
-        span.finish(exc_info=exc_info)
+        args.span.finish(exc_info=exc_info)
     finally:
-        REQUEST_TIME.labels(*prom_labels, "false").observe(time.perf_counter() - start_time)
-        REQUEST_TOTAL.labels(*prom_labels, "false").inc()
-        REQUEST_ACTIVE.labels(*prom_labels).dec()
+        prom_labels = args.prom_labels._asdict()
+        REQUEST_TIME.labels(**prom_labels, cassandra_success="false").observe(
+            time.perf_counter() - args.start_time
+        )
+        REQUEST_TOTAL.labels(**prom_labels, cassandra_success="false").inc()
+        REQUEST_ACTIVE.labels(**prom_labels).dec()
         event.set()
 
 
@@ -328,12 +336,13 @@ class CassandraSessionAdapter:
         query_name: Optional[str] = None,
         **kwargs: Any,
     ) -> ResponseFuture:
-        prom_labels = (
-            ",".join(self.cluster.contact_points),
-            self.session.keyspace,
-            query_name if query_name is not None else "",
+        prom_labels = CassandraPrometheusLabels(
+            cassandra_contact_points=",".join(self.cluster.contact_points),
+            cassandra_keyspace=self.session.keyspace,
+            cassandra_query_name=query_name if query_name is not None else "",
         )
-        REQUEST_ACTIVE.labels(*prom_labels).inc()
+
+        REQUEST_ACTIVE.labels(**prom_labels._asdict()).inc()
         start_time = time.perf_counter()
         trace_name = f"{self.context_name}.execute"
         span = self.server_span.make_child(trace_name)
@@ -346,10 +355,10 @@ class CassandraSessionAdapter:
         elif isinstance(query, BoundStatement):
             span.set_tag("statement", query.prepared_statement.query_string)
         future = self.session.execute_async(query, parameters=parameters, timeout=timeout, **kwargs)
-        callback_args = (
-            span,
-            start_time,
-            prom_labels,
+        callback_args = CassandraCallbackArgs(
+            span=span,
+            start_time=start_time,
+            prom_labels=prom_labels,
         )
         future = wrap_future(
             response_future=future,
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -33,27 +33,27 @@ from baseplate.lib.prometheus_metrics import default_latency_buckets
 from baseplate.lib.secrets import SecretsStore
 
 cassandra_labels = [
-    "contact_points",
-    "keyspace",
+    "cassandra_contact_points",
+    "cassandra_keyspace",
+    "cassandra_query_name",
 ]
-cassandra_labels_w_success = cassandra_labels + ["success"]
+cassandra_labels_w_success = cassandra_labels + ["cassandra_success"]
 REQUEST_TIME = Histogram(
     "cassandra_client_latency_seconds",
     "Time spent executing cassandra query",
     cassandra_labels_w_success,
     buckets=default_latency_buckets,
 )
-REQUEST_TOTAL = Counter(
-    "cassandra_client_requests_total",
-    "Total number of cassandra queries",
-    cassandra_labels_w_success,
-)
 REQUEST_ACTIVE = Gauge(
     "cassandra_client_requests_active",
     "Current number of active cassandra queries",
     cassandra_labels,
 )
-
+REQUEST_TOTAL = Counter(
+    "cassandra_client_requests_total",
+    "Total number of cassandra queries",
+    cassandra_labels_w_success,
+)
 
 if TYPE_CHECKING:
     import cqlmapper.connection
@@ -313,20 +313,25 @@ class CassandraSessionAdapter:
         query: Query,
         parameters: Optional[Parameters] = None,
         timeout: Union[float, object] = _NOT_SET,
+        query_name: Optional[str] = None,
         **kwargs: Any,
     ) -> Any:
-        return self.execute_async(query, parameters, timeout, **kwargs).result()
+        return self.execute_async(
+            query, parameters=parameters, timeout=timeout, query_name=query_name, **kwargs
+        ).result()
 
     def execute_async(
         self,
         query: Query,
         parameters: Optional[Parameters] = None,
         timeout: Union[float, object] = _NOT_SET,
+        query_name: Optional[str] = None,
         **kwargs: Any,
     ) -> ResponseFuture:
         prom_labels = (
             ",".join(self.cluster.contact_points),
             self.session.keyspace,
+            query_name if query_name is not None else "",
         )
         REQUEST_ACTIVE.labels(*prom_labels).inc()
         start_time = time.perf_counter()
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -337,7 +337,7 @@ class CassandraSessionAdapter:
         **kwargs: Any,
     ) -> ResponseFuture:
         prom_labels = CassandraPrometheusLabels(
-            cassandra_contact_points=",".join(self.cluster.contact_points),
+            cassandra_contact_points=",".join(sorted(self.cluster.contact_points)),
             cassandra_keyspace=self.session.keyspace,
             cassandra_query_name=query_name if query_name is not None else "",
         )
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 from threading import Event
 from typing import Any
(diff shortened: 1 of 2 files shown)


baseplate/clients/cassandra.py
```diff
@@ -21,12 +22,38 @@
...
+    "Time spent executing cassandra query",
+    cassandra_labels_w_success,
+    buckets=default_latency_buckets,
+)
+REQUEST_TOTAL = Counter(
+    "cassandra_client_requests_total",
+    "Total number of cassandra queries",
+    cassandra_labels_w_success,
+)
+REQUEST_ACTIVE = Gauge(
```
krav, line 49: Do we need this? It's subject to sampling effects, and you can do `rate()` on `requests_total` to get the same data.

baseplate/clients/cassandra.py
```diff
@@ -231,20 +258,36 @@ def wait_for_callbacks_result(self: ResponseFuture) -> Any:
     return response_future
 
 
-def _on_execute_complete(_result: Any, span: Span, event: Event) -> None:
+def _on_execute_complete(
+    _result: Any, args: Tuple[Span, datetime, Tuple[Any, ...]], event: Event
```
krav: Maybe use NewType to make this readable?

baseplate/clients/cassandra.py
```diff
@@ -21,12 +22,38 @@
...
 from cassandra.query import PreparedStatement  # pylint: disable=no-name-in-module
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
+from prometheus_client import Counter
+from prometheus_client import Gauge
+from prometheus_client import Histogram
 
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
+from baseplate.lib.prometheus_metrics import default_latency_buckets
```
krav, line 33: I'm worried about the floats in this list causing buckets to not align, if we keep using it we should probably round it to 4ish digits.

baseplate/clients/cassandra.py
Xaelias: I honestly don't care so... Whatever the larger group tells me to do 😅

baseplate/clients/cassandra.py
Xaelias: They're the same everywhere though. Why would they not align?
Or did you mean between go and python for instance?

baseplate/clients/cassandra.py
Xaelias: Not exactly as Ben pointed out in slack. Which is why I added it back.

KTAtkinson (commented):
Putting this on ice until we've nailed down the metrics and labels we want for Cassandra in the baseplate spec.

baseplate/clients/cassandra.py
```diff
@@ -21,12 +22,38 @@
...
 from baseplate.lib.secrets import SecretsStore
 
+cassandra_labels = [
+    "cassandra_contact_points",
```
KTAtkinson: We don't have this label spec'd out. Is this currently surfaced in statsd metrics? Do you have an idea of what sort of cardinality this will introduce? What value do folks get from this?

baseplate/clients/cassandra.py
```diff
@@ -21,12 +22,38 @@
...
+]
+cassandra_labels_w_success = cassandra_labels + ["cassandra_success"]
+REQUEST_TIME = Histogram(
+    "cassandra_client_latency_seconds",
+    "Time spent executing cassandra query",
+    cassandra_labels_w_success,
+    buckets=default_latency_buckets,
+)
+REQUEST_ACTIVE = Gauge(
+    "cassandra_client_requests_active",
```
KTAtkinson: Should be `cassandra_client_active_requests` according to the spec.

baseplate/clients/cassandra.py
```diff
@@ -21,12 +22,38 @@
...
+    "Current number of active cassandra queries",
+    cassandra_labels,
```
KTAtkinson: :no_bell: Would it make sense to include the `cassandra_query_name` label here as well?

baseplate/clients/cassandra.py
```diff
@@ -231,20 +258,32 @@ def wait_for_callbacks_result(self: ResponseFuture) -> Any:
     return response_future
 
 
-def _on_execute_complete(_result: Any, span: Span, event: Event) -> None:
+def _on_execute_complete(
+    _result: Any, args: Tuple[Span, float, Tuple[Any, ...]], event: Event
```
KTAtkinson: Why receive a tuple of args, rather than a set of named arguments?

baseplate/clients/cassandra.py
```diff
@@ -231,20 +258,32 @@ def wait_for_callbacks_result(self: ResponseFuture) -> Any:
...
+) -> None:
     # TODO: tag with anything from the result set?
     # TODO: tag with any returned warnings
+    span, start_time, prom_labels = args
     try:
         span.finish()
     finally:
+        REQUEST_TIME.labels(*prom_labels, "true").observe(time.perf_counter() - start_time)
```
KTAtkinson: IMO, this should enforce the labels it wants to be added, not apply an arbitrary set of labels provided by the caller. This will make the code easier to debug. It would also make it much easier to explicitly see what metrics are being produced.

tests/unit/clients/cassandra_tests.py
```diff
@@ -67,6 +77,13 @@ def test_credentials(self):
 
 class CassandraSessionAdapterTests(unittest.TestCase):
     def setUp(self):
+        # cleaning up prom registry
+        collectors = list(REGISTRY._collector_to_names.keys())
```
KTAtkinson: :no_bell: This is an internal method which makes me worried that this test will be fragile. Is there any way we can do this without using the `_collector_to_names` attribute?

tests/unit/clients/cassandra_tests.py
```diff
@@ -86,3 +103,212 @@ def test_prepare(self):
...
+
+    def test_execute_async_prom_metrics(self):
+        self.session.keyspace = "keyspace"  # mocking keyspace name
+        self.session.cluster.contact_points = [
+            "hostname1",
+            "hostname2",
+        ]  # mocking cluster contact points
+        self.adapter.execute_async("SELECT foo from bar;")
+
+        requests_active = REGISTRY._names_to_collectors[
```
KTAtkinson: Could you [use `REGISTRY.get_sample_value` instead](https://sourcegraph.com/github.com/prometheus/client_python/-/blob/prometheus_client/registry.py?L131) here?

KTAtkinson (changes requested):

baseplate/clients/cassandra.py
nsheaps: Discussed in slack and have an update to the spec pending to add this (with this exact name). @bjk-reddit said "if it's the cluster hostname then that's probably okay". This essentially is "address" but uses the Cassandra terminology because it's not a single hostname:port.

baseplate/clients/cassandra.py
nsheaps: it is included, [L38](https://github.com/reddit/baseplate.py/pull/686/files#diff-b4aa11d610ce8132e3d92cb92a1eeacff2b410797dbfcadd60376fac0c0ceb73R38)

baseplate/clients/cassandra.py
nsheaps: This tripped me up too, and runs the risk of the third argument not being the expected length (which would result in label values being put on the wrong labels)

baseplate/clients/cassandra.py
nsheaps: Agree, this is hard to reason about in comparison with the labels at the top of the file.

baseplate/clients/cassandra.py
Xaelias: It is. `cassandra_labels` has everything but `success` (which obviously doesn't make sense for active requests)

tests/unit/clients/cassandra_tests.py
nsheaps: I think I'm missing why we can't just import the metrics from the Cassandra client and clear them directly. We're not writing tests that check the values of any other metrics other than the ones in use by Cassandra, I'm not sure it makes sense to reset them all (if you can)

baseplate/clients/cassandra.py
Xaelias: The only way I can think of to do that would be to make both `_on_execute_complete` and `_on_execute_failed` class method of `CassandraSessionAdapter` at which point they would just maybe re-compute the labels the same way we currently do them here.

Can you explain to me how it makes things better/easier?

tests/unit/clients/cassandra_tests.py
Xaelias: Thanks I couldn't find this function and nobody raised it when we used this approach in other places.
Will update this PR and do a follow up update for the rest of the codebase.

tests/unit/clients/cassandra_tests.py
Xaelias: "lazyness"
it makes writing tests easier, it doesn't matter if we add or remove metrics, if we check something else etc. we're sure that we get a clean slate for every test.

baseplate/clients/cassandra.py
nsheaps, line 262:
```suggestion
    _result: Any, args: Tuple[Span, float, Tuple[Any, ...]], event: Event

File: baseplate/clients/cassandra.py [hunk 11]
```diff
@@ -231,20 +258,32 @@ def wait_for_callbacks_result(self: ResponseFuture) -> Any:
...
+        REQUEST_TOTAL.labels(*prom_labels, "true").inc()
+        REQUEST_ACTIVE.labels(*prom_labels).dec()
         event.set()
```
nsheaps:
```suggestion
from typing import NamedTuple

class CassandraCallbackPrometheusLabels(NamedTuple):
    contact_points: str
    keyspace: str
    query_name: Optional[str]

class CassandraExecuteCallbackArgs(NamedTuple):
    span: Span
    start_time: float
    labels: CassandraCallbackPrometheusLabels

def _on_execute_complete(
    _result: Any, args: CassandraExecuteCallbackArgs, event: Event 
) -> None:
    # TODO: tag with anything from the result set?
    # TODO: tag with any returned warnings
    span = args["span"]
    start_time = args["start_time"]
    prom_labels = args["labels"]
    try:
        span.finish()
    finally:
        labels = [
            prom_labels["contact_points"], 
            prom_labels["keyspace"], 
            prom_labels["query_name"] or ""
        ]
        labels_with_success = labels + ["true"]
        REQUEST_TIME.labels(*labels_with_success).observe(time.perf_counter() - start_time)
        REQUEST_TOTAL.labels(*labels_with_success).inc()
        REQUEST_ACTIVE.labels(*labels).dec()
        event.set()
```

baseplate/clients/cassandra.py
Xaelias: I believe this comment was addressed, please confirm

baseplate/clients/cassandra.py
```diff
@@ -274,17 +321,29 @@ def execute(
...
     def execute_async(
         self,
         query: Query,
         parameters: Optional[Parameters] = None,
         timeout: Union[float, object] = _NOT_SET,
+        query_name: Optional[str] = None,
         **kwargs: Any,
     ) -> ResponseFuture:
+        prom_labels = CassandraPrometheusLabels(
+            cassandra_contact_points=",".join(self.cluster.contact_points),
```
nsheaps: 🔕 is this ordering deterministic? Perhaps they should be alphabetically ordered so the same set doesn't introduce more values when they're ordered differently.

nsheaps (approved):
Thanks for all these changes!

baseplate/clients/cassandra.py
Xaelias: Done just in case :-)

KTAtkinson (approved):

PR #685, opened 2022-06-16
[base-22] Add prometheus tagged span observer

krav:
## 💸 TL;DR
Prometheus version of the StatsD metrics_tagged. If tagged metrics are enabled it will emit basic metrics, then more if you explicitly enable a tags (like "statement" for the sqlalchemy client). Configuration is shared with StatsD metrics_tagged.  Should have parity with what's currently sent to wavefront for spans.

## 📜 Details

example output:

```
baseplate_server_span_latency_seconds_sum{client="",endpoint="is_healthy",success=""} 0.012034416198730469
baseplate_server_span_latency_seconds_sum{client="",endpoint="example",success=""} 0.08535456657409668
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.005",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.01",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.025",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.05",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.075",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.1",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.25",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.5",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="0.75",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="1.0",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="2.5",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="5.0",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="7.5",success=""} 78.0
baseplate_server_span_latency_seconds_bucket{client="",endpoint="is_healthy",le="10.0",success=""} 78.0
[...]

PR #687, opened 2022-06-17
refactor: prom instrument pyramid, request and thrift directly

krav:
Instrument directly without using observer pattern. Deletes the old prom local span observer, superseded by #685.


baseplate/clients/requests.py
```diff
@@ -101,6 +109,42 @@ def http_adapter_from_config(
     return ValidatingHTTPAdapter(**kwargs)
 
 
+PROM_NAMESPACE = "http_client"
```
KTAtkinson, line 110: :no_bell: I would call this `metric_name_prefix`.

baseplate/clients/thrift.py
```diff
@@ -162,22 +167,54 @@ def retrying(self, **policy: Any) -> Iterator["_PooledClientProxy"]:
         )
 
 
+NANOSECONDS_PER_SECOND = 1e9
+PROM_NAMESPACE = "thriftbp_client"
```
KTAtkinson: This should be `thrift_client`, to allow for all thrift metrics to be surfaced in this metric even if they aren't produced by baseplate.

baseplate/clients/requests.py
```diff
@@ -101,6 +109,42 @@ def http_adapter_from_config(
...
+
+# Latency histogram of HTTP calls made by clients
+# buckets are defined above (from 100µs to ~14.9s)
+LATENCY_SECONDS = Histogram(
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Latency histogram of HTTP calls made by clients",
+    [
+        "http_method",
```
KTAtkinson: Based on the [Baseplate Spec ](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md)there should be a `http_endpoint` label here as well.

baseplate/clients/thrift.py
```diff
@@ -162,22 +167,54 @@ def retrying(self, **policy: Any) -> Iterator["_PooledClientProxy"]:
...
+    "Latency of thrift client requests",
+    PROM_COMMON_LABELS,
+    buckets=default_latency_buckets,
+)
+
+PROM_REQUESTS = Counter(
+    f"{PROM_NAMESPACE}_requests_total",
+    "Total number of outgoing requests",
+    PROM_REQUESTS_LABELS,
+)
```
KTAtkinson: It would allow some of the code below to read more smoothly if these were named similar to the way the metrics constants are named for HTTP. Maybe something like this:
`
```suggestion
REQUEST_LATENCY = Histogram(
    f"{PROM_NAMESPACE}_latency_seconds",
    "Latency of thrift client requests",
    PROM_COMMON_LABELS,
    buckets=default_latency_buckets,
)

REQUESTS_TOTAL = Counter(
    f"{PROM_NAMESPACE}_requests_total",
    "Total number of outgoing requests",
    PROM_REQUESTS_LABELS,
)
```

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -68,10 +75,56 @@ def close(self) -> None:
...
 
+NANOSECONDS_PER_SECOND = 1e9
+PROM_NAMESPACE = "http_server"
+
+HISTOGRAM_LABELS = [
+    "http_method",
+    "http_endpoint",
+    "http_success",
+]
+PROM_LATENCY = Histogram(
```
KTAtkinson: Same as above

KTAtkinson (changes requested):

baseplate/clients/requests.py
krav: According to the [metric and label naming](https://prometheus.io/docs/practices/naming/) doc this is idiomatic: "The prefix is sometimes referred to as namespace by client libraries". It's a bit better explained in Prometheus Up and Running.

baseplate/clients/requests.py
krav: The actual endpoint would have way too high cardinality. The spec says "as defined above" and then above it is the definition for http _servers_, so this might be a mistake in the spec. We could maybe put the calling function in there or something, but that's not very "endpoint" :/

Maybe include this in the spec discussion?

baseplate/clients/thrift.py
krav: Fixed!

baseplate/clients/thrift.py
krav: I agree, fixed!

baseplate/frameworks/pyramid/__init__.py
krav: Done!

baseplate/clients/requests.py
nsheaps: The [spec has been updated](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md?plain=1#L89) since the initial comment 6 days ago for this to be reflective of the name of the endpoint being hit rather than the full URL

baseplate/clients/requests.py
```diff
@@ -101,6 +107,42 @@ def http_adapter_from_config(
...
+        "http_slug",
+    ],
+    buckets=default_latency_buckets,
+)
+# Counter counting total HTTP requests started by a given client
+REQUESTS_TOTAL = Counter(
+    f"{PROM_NAMESPACE}_requests_total",
+    "Total number of HTTP requests started by a given client",
+    [
+        "http_method",
```
nsheaps: 🔕 I like the pattern in the other PRs about using a constant array for defining consistent metric labels between metrics. Can that be added here?

baseplate/clients/requests.py
krav: Let's make a new issue out of that, since this is a refactor and that would introduce something new. Also because I have no idea how to do it :) 

Btw the spec says "may be empty" which AFAIK prom treats the same way as not set, so this superficially fulfills the spec :D

nsheaps: https://reddit.atlassian.net/browse/BASE-85

baseplate/clients/requests.py
```diff
@@ -101,6 +107,42 @@ def http_adapter_from_config(
...
+        "http_success",
+        "http_response_code",
+        "http_slug",
```
nsheaps: I have an open PR for renaming this to client_name since slug doesn't really tell me what it's an identifier for

nsheaps: [PR](https://github.snooguts.net/reddit/baseplate.spec/pull/46) has been merged, this should change to `http_client_name`

baseplate/clients/requests.py
```diff
@@ -204,13 +246,20 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
...
+            "http_method": request.method or "",
+            "http_slug": self.client_name if self.client_name is not None else self.name,
         }
-        with self.span.make_child(f"{self.name}.request").with_tags(tags) as span:
+        start_time = time.perf_counter()
+        ACTIVE_REQUESTS.labels(**active_request_label_values).inc()
+
+        with self.span.make_child(f"{self.name}.request").with_tags(
+            {
+                "http.url": request.url,
```
nsheaps, line 251: was the protocol tag used in wavefront?

baseplate/clients/requests.py
```diff
@@ -204,13 +246,20 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
...
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        tags = {
-            "protocol": "http",
-            "http.method": request.method,
-            "http.url": request.url,
-            "http.slug": self.client_name if self.client_name is not None else self.name,
+        active_request_label_values = {
+            "http_method": request.method or "",
+            "http_slug": self.client_name if self.client_name is not None else self.name,
```
nsheaps: same http_client_name change here please

baseplate/clients/thrift.py
```diff
@@ -71,7 +76,7 @@ class ThriftContextFactory(ContextFactory):
 
     """
 
-    PROM_PREFIX = "bp_thrift_pool"
+    PROM_PREFIX = "thrift_pool"
```
nsheaps: 🔕 the name here is a tad confusing since there are other labels that aren't prefixed with `_pool_` (and it's not clear that this is for labels, not metric names)

baseplate/clients/thrift.py
```diff
@@ -162,22 +167,53 @@ def retrying(self, **policy: Any) -> Iterator["_PooledClientProxy"]:
         )
 
 
+PROM_NAMESPACE = "thrift_client"
+
+PROM_COMMON_LABELS = [
+    "namespace",
```
nsheaps: not sure exactly what this is since I've never used Thrift, is this akin to `grpc_service`? Happy to add it to the spec if it was a miss

baseplate/clients/thrift.py
```diff
@@ -162,22 +167,53 @@ def retrying(self, **policy: Any) -> Iterator["_PooledClientProxy"]:
         )
 
 
+PROM_NAMESPACE = "thrift_client"
+
+PROM_COMMON_LABELS = [
```
nsheaps: The labels here should be prefixed with `thrift_` according to the spec

nsheaps: this is also missing `thrift_method`

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -68,11 +75,60 @@ def close(self) -> None:
...
+    "http_endpoint",
+    "http_success",
+]
+REQUEST_LATENCY = Histogram(
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Time spent processing requests",
+    HISTOGRAM_LABELS,
+    buckets=default_latency_buckets,
+)
+REQUEST_SIZE = Histogram(
```
nsheaps, line 91: Is there a usecase for these request/response size metrics? not against them (not in the spec) but want to be sure they'll actually be useful for folks

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -68,11 +75,60 @@ def close(self) -> None:
...
+
+
 def _make_baseplate_tween(
     handler: Callable[[Request], Response], _registry: Registry
 ) -> Callable[[Request], Response]:
     def baseplate_tween(request: Request) -> Response:
+        time_started = time.perf_counter()
+
+        response: Optional[Response] = None
+        endpoint = request.matched_route.pattern if request.matched_route else ""
```
nsheaps, line 128: 🔕 my assumption is that if someone requests a URL where there isn't a matched_route that the server would 404 right? maybe the matched route should be `__not_found` or something instead of `""`?

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -244,7 +318,6 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         span.set_tag("http.route", request.matched_route.pattern)
         span.set_tag("http.method", request.method)
         span.set_tag("peer.ipv4", request.remote_addr)
-        span.set_tag("http.request_length", request.content_length)
```
nsheaps, line 247: Where did this go? I don't see it anywhere else

baseplate/frameworks/thrift/__init__.py
```diff
@@ -18,10 +22,41 @@
...
 from baseplate.thrift.ttypes import Error
 from baseplate.thrift.ttypes import ErrorCode
 
 
+PROM_NAMESPACE = "thrift_server"
+
+PROM_COMMON_LABELS = ["thrift_method"]
+PROM_LATENCY = Histogram(
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "RPC latencies",
```
nsheaps:
```suggestion
    "Time spent processing requests",
```

baseplate/clients/requests.py
krav: Fixed

baseplate/clients/requests.py
krav: No, it was introduced [here](https://github.com/reddit/baseplate.py/commit/caa214ba5013d4701245da52bcefcd3b718e2bc2#diff-0e5f12c68b4ed54bdd294f03f64c3520a48583db98c928508c61d9f56dd35ffbL203-R213) two months ago

baseplate/clients/thrift.py
krav: The equivalent tag is called `slug` below, so I guess it's the name of the client? Deleting it.

baseplate/clients/thrift.py
krav: Fixed

baseplate/frameworks/pyramid/__init__.py
krav: Not sure actually, as far as I remember these were introduced with the draft pr from spladug

baseplate/frameworks/pyramid/__init__.py
krav: I don't know pyramid well enough that I want to make any assumptions about when or when not matched_route exists. There should be a 404 response code label that's probably a better signal.

baseplate/frameworks/pyramid/__init__.py
krav: It was introduced only to serve the Prometheus metric REQUEST_SIZE (see blame). Figure that it's OK to remove from statsd since it's never been used and only been enabled for a few months.

baseplate/clients/thrift.py
krav: It's used for the pool metric names, renamed it to make it clearer.

baseplate/clients/requests.py
krav: Fixed

baseplate/clients/thrift.py
```diff
@@ -71,19 +103,19 @@ class ThriftContextFactory(ContextFactory):
 
     """
 
-    PROM_PREFIX = "bp_thrift_pool"
-    PROM_LABELS = ["client_cls"]
+    POOL_PREFIX = "thrift_pool"
```
nsheaps: I didn't see these in the spec so I added them [here](https://github.snooguts.net/reddit/baseplate.spec/pull/49). To match the other connection pool metrics, the prefix should be `thrift_client_pool` with the label `thrift_pool` (matches redis)

baseplate/frameworks/pyramid/__init__.py
nsheaps: I'm fine with it as long as @bjk-reddit is. I don't think it's worth adding to the spec at the moment though. Happy to disagree here. I'm not against it since I've used this data from LB logs in the past and this makes it a bit easier

baseplate/frameworks/pyramid/__init__.py
nsheaps: 🔕 I'm making a mental note that we should probably come back to this, since a blank label and a label that says "404" might be a helpful differentiator (beyond success=false and the response code). That said this is already helpful for figuring out 404s on matched routes (eg the route decided to respond 404) vs totally random URL requests.

baseplate/frameworks/thrift/__init__.py
```diff
@@ -83,6 +137,16 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 # a normal result
                 span.finish()
                 return result
```
nsheaps, line 139: if you're returning here wouldn't that prevent the `finally` block from being called?

baseplate/lib/prometheus_metrics.py
```diff
@@ -27,414 +19,8 @@
 ]
 
 
-class PrometheusHTTPClientMetrics:
```
nsheaps, line 30: 👏 👏 👏 👏 👏

baseplate/observers/prometheus.py
```diff
@@ -1,255 +0,0 @@
-import logging
```
nsheaps, line 1: 👏 👏 👏 👏 👏 👏

tests/integration/requests_tests.py
```diff
@@ -170,179 +168,3 @@ def test_external_client_doesnt_send_headers(http_server):
         assert "X-Parent" not in http_server.requests[0].headers
         assert "X-Span" not in http_server.requests[0].headers
         assert "X-Edge-Request" not in http_server.requests[0].headers
-
-
-@pytest.mark.parametrize("client_cls", [InternalRequestsClient, ExternalRequestsClient])
```
nsheaps, line 175: Are there any tests to replace the stuff that was deleted here?

nsheaps (changes requested):
Very close! Thank you for the refactoring!

tests/integration/requests_tests.py
krav: The happy paths are covered by existing unit tests now (caught a few misteaks like that), noodling on how to test errors and exceptions :thinking:

baseplate/frameworks/thrift/__init__.py
krav: No, the finally block is ran before the return here.

"A finally clause is executed whether or not an exception has occurred in the try clause. When an exception has occurred, it is re-raised after the finally clause is executed. The finally clause is also executed ``on the way out'' when the try statement is left via a break or return statement. " (https://documentation.help/Python-PEP/node101.html)

baseplate/clients/thrift.py
krav: Fixed!

baseplate/clients/requests.py
```diff
@@ -101,6 +107,35 @@ def http_adapter_from_config(
     return ValidatingHTTPAdapter(**kwargs)
 
 
+PROM_NAMESPACE = "http_client"
+HTTP_LABELS_COMMON = [
+    "http_method",
```
nsheaps, line 112: ~~This isn't in the spec but is fine for now, will add a PR for it after I finish reviewing this~~

Spec just says "everything server has"

baseplate/clients/requests.py
```diff
@@ -204,13 +239,20 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
 
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        tags = {
-            "protocol": "http",
-            "http.method": request.method,
-            "http.url": request.url,
-            "http.slug": self.client_name if self.client_name is not None else self.name,
+        active_request_label_values = {
+            "http_method": request.method or "",
```
nsheaps, line 243: to ensure consistency between libraries and http protocol versions can you please lowercase this value?

baseplate/clients/requests.py
```diff
@@ -101,6 +107,35 @@ def http_adapter_from_config(
     return ValidatingHTTPAdapter(**kwargs)
 
 
+PROM_NAMESPACE = "http_client"
+HTTP_LABELS_COMMON = [
+    "http_method",
+    "http_client_name",
+]
+HTTP_LABELS_TERMINAL = [*HTTP_LABELS_COMMON, "http_success"]
```
nsheaps, line 115: ~~I'm surprised that http_sucess got missed in the spec as well~~

Nevermind, the spec just says "Everything server has"

baseplate/clients/requests.py
```diff
@@ -101,6 +107,35 @@ def http_adapter_from_config(
...
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Latency histogram of HTTP calls made by clients",
+    HTTP_LABELS_TERMINAL,
+    buckets=default_latency_buckets,
+)
+# Counter counting total HTTP requests started by a given client
+REQUESTS_TOTAL = Counter(
+    f"{PROM_NAMESPACE}_requests_total",
+    "Total number of HTTP requests started by a given client",
+    [*HTTP_LABELS_TERMINAL, "http_response_code"],
```
nsheaps, line 129: I am assuming this isn't on the histogram because of high cardinality?

baseplate/clients/thrift.py
```diff
@@ -216,20 +252,27 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
                     continue
-                except (TApplicationException, TProtocolException):
+                except (TApplicationException, TProtocolException) as exc:
                     # these are subclasses of TException but aren't ones that
                     # should be expected in the protocol. this is an error!
+                    success = "false"
+                    exception = exc
                     span.finish(exc_info=sys.exc_info())
                     raise
                 except Error as exc:
```
nsheaps, line 270: are 4xx/client failures included in the above or this block?

baseplate/clients/thrift.py
```diff
@@ -240,12 +283,30 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                     raise
                 except:  # noqa: E722
                     # something unexpected happened
```
nsheaps, line 285: looks like `success` isn't set here or the block above. Based on how we're exception handling, the finally block can likely just say `success="true" if exception is None else "false"`

baseplate/frameworks/thrift/__init__.py
```diff
@@ -58,19 +102,29 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
...
                 span.set_tag("thrift.status", status)
                 span.set_tag("success", "false")
+
+                success = "false"
+                exception = exc.__class__.__name__
+                status_code = exc.code
+                status_name = status
+
                 # mark 5xx errors as failures since those are still "unexpected"
                 if 500 <= exc.code < 600:
```
nsheaps, line 112: 4xx should be success=false too, not sure if that's covered in this block

baseplate/frameworks/thrift/__init__.py
```diff
@@ -58,19 +102,29 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 span.set_tag("thrift.status_code", exc.code)
                 span.set_tag("thrift.status", status)
                 span.set_tag("success", "false")
+
+                success = "false"
```
nsheaps, line 106: A lot of this is repeated in the except blocks, can we take a similar approach here as the finally block in the client setting some of this?

tests/integration/requests_tests.py
nsheaps: We should make sure we have tests covering the 4xx use cases for http/thrift server/client since that is a topic of contention and want to be sure we don't deviate from the spec.

nsheaps (changes requested):
Few comments on success=false handling for thrift and HTTP plus needs some more tests around the unhappy path.

tests/integration/requests_tests.py
Xaelias: the problem is we don't actually know how the client behave
The behavior we expect is tested. It doesn't mean it's the real behavior. Previous maintainer of the code made the assumption that `baseplate.smth.Error` would be raised and the code/test are structured around that.
Real world errors might be diff.

baseplate/clients/requests.py
Xaelias: Yeah that would explode the cardinality. Each histogram has 15-ish metrics IIRC. Multiply this by get/post/update/...

baseplate/clients/thrift.py
Xaelias: No they would be caught by the generic `TExeption` below it is my current understanding

baseplate/clients/requests.py
KTAtkinson: It looks like we're still missing 'http_endpoint'.

baseplate/clients/requests.py
```diff
@@ -101,6 +107,35 @@ def http_adapter_from_config(
...
+    "http_client_name",
+]
+HTTP_LABELS_TERMINAL = [*HTTP_LABELS_COMMON, "http_success"]
+
+# Latency histogram of HTTP calls made by clients
+# buckets are defined above (from 100µs to ~14.9s)
+LATENCY_SECONDS = Histogram(
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Latency histogram of HTTP calls made by clients",
+    HTTP_LABELS_TERMINAL,
```
KTAtkinson, line 122: Why aren't we tracking response_code here?

baseplate/clients/thrift.py
```diff
@@ -216,20 +252,27 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                     result = method(*args, **kwargs)
                 except TTransportException as exc:
                     # the connection failed for some reason, retry if able
+                    success = "false"
```
KTAtkinson, line 255: :no_bell: I don't see any similar handling for HTTP clients. Does this work properly if TCP connections drop unexpectedly or DNS can't resolve?

KTAtkinson (changes requested):

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -85,6 +141,24 @@ def baseplate_tween(request: Request) -> Response:
...
+            ACTIVE_REQUESTS.labels(http_method=request.method, http_endpoint=endpoint).dec()
+            REQUEST_LATENCY.labels(**histogram_labels).observe(time.perf_counter() - time_started)
+            REQUESTS_TOTAL.labels(
+                **{
+                    **histogram_labels,
+                    "http_response_code": response.status_code if response else "",
+                }
+            ).inc()
+            REQUEST_SIZE.labels(**histogram_labels).observe(request.content_length or 0)
+            RESPONSE_SIZE.labels(**histogram_labels).observe(
```
nsheaps, line 158:
```suggestion
            response_size_metric = RESPONSE_SIZE.labels(**histogram_labels)
            if response and response.content_length:
                response_size_metric.observe(response.content_length)
```

baseplate/clients/requests.py
Xaelias: The spec says histograms don't get `http_response_code`. Cardinality would be the reason. I think I already said this in a comment above but these histograms are 15 or 16 series each (because of the buckets). Times all the http response codes it explodes really quickly.

Xaelias:
Closed in favor of https://github.com/reddit/baseplate.py/pull/704

PR #690, opened 2022-06-23
instrument redis calls

krav:
## 💸 TL;DR
Adds instrumentation for Redis client calls

## ✅ Checks

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -1,11 +1,10 @@
 from math import ceil
+from time import perf_counter
 from typing import Any
 from typing import Dict
 from typing import Optional
 
 import redis
-
-from prometheus_client import Gauge
 
 # redis.client.StrictPipeline was renamed to redis.client.Pipeline in version 3.0
 try:
@@ -13,11 +12,42 @@ try:
 except ImportError:
     from redis.client import Pipeline
 
+from prometheus_client import Counter
+from prometheus_client import Gauge
+from prometheus_client import Histogram
+
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
 from baseplate.lib import message_queue
 from baseplate.lib import metrics
+
+from baseplate.lib.prometheus_metrics import default_latency_buckets
+
+PROM_NAMESPACE = "redis"
+PROM_SHARED_LABELS = [
+    f"{PROM_NAMESPACE}_command",
+    f"{PROM_NAMESPACE}_database",
+    f"{PROM_NAMESPACE}_startup_host",
+]
+LATENCY_SECONDS = Histogram(
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Latency histogram for calls made by clients",
+    [*PROM_SHARED_LABELS, f"{PROM_NAMESPACE}_success"],
+    buckets=default_latency_buckets,
+)
+
+REQUESTS_TOTAL = Counter(
+    f"{PROM_NAMESPACE}_requests_total",
+    "Total number of requests made by client",
+    [*PROM_SHARED_LABELS, f"{PROM_NAMESPACE}_success"],
+)
+
+ACTIVE_REQUESTS = Gauge(
+    f"{PROM_NAMESPACE}_active_requests",
+    "Number of active requests for a given client",
+    PROM_SHARED_LABELS,
+)
 
 
 def pool_from_config(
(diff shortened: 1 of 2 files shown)


baseplate/clients/redis.py
```diff
@@ -1,24 +1,54 @@
...
+
 from baseplate import Span
 from baseplate.clients import ContextFactory
 from baseplate.lib import config
 from baseplate.lib import message_queue
 from baseplate.lib import metrics
 
+from baseplate.lib.prometheus_metrics import default_latency_buckets
+
+PROM_NAMESPACE = "redis"
```
KTAtkinson: :no_bell: I would call this `PROM_PREFIX` personally, I think that's easier to understand what it is.

baseplate/clients/redis.py
```diff
@@ -1,24 +1,54 @@
...
+PROM_SHARED_LABELS = [
+    f"{PROM_NAMESPACE}_command",
+    f"{PROM_NAMESPACE}_database",
+    f"{PROM_NAMESPACE}_startup_host",
```
KTAtkinson: IRRC, `startup_host` is only a concept in clustered redis, I believe this client is only non-clustered redis. The spec indicates that non-clustered redis should have the label `redis_host`.

KTAtkinson (changes requested):
Are you able to add tests here?

baseplate/clients/redis.py
krav: Fixed!

baseplate/clients/redis.py
```diff
@@ -1,24 +1,54 @@
...
 from baseplate.lib import message_queue
 from baseplate.lib import metrics
 
+from baseplate.lib.prometheus_metrics import default_latency_buckets
+
+PROM_NAMESPACE = "redis"
+PROM_SHARED_LABELS = [
+    f"{PROM_NAMESPACE}_command",
+    f"{PROM_NAMESPACE}_database",
+    f"{PROM_NAMESPACE}_host",
```
nsheaps: This is supposed to be `_address` according to the spec

baseplate/clients/redis.py
```diff
@@ -164,7 +194,28 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+                f"{PROM_NAMESPACE}_host": self.connection_pool.connection_kwargs.get("host", ""),
+                f"{PROM_NAMESPACE}_database": self.connection_pool.connection_kwargs.get("db", ""),
+            }
+            ACTIVE_REQUESTS.labels(**labels).inc()
+
+            try:
+                res = super().execute_command(command, *args[1:], **kwargs)
+                if isinstance(res, redis.RedisError):
+                    success = "false"
+                return res
```
nsheaps, line 215: if you `return` here wouldn't that prevent the labels from being set below?

nsheaps (changes requested):
Few changes. Thanks for getting these done!

baseplate/clients/redis.py
krav: Fixed, should it then be `startup_address` instead of `startup_host` for cluster?

baseplate/clients/redis.py
krav: Nope, `finally` is always runs before returning, and as long as the `finally` block doesn't call return itself python jumps back here and returns `res`.

nsheaps: TIL thanks

baseplate/clients/redis.py
nsheaps: I didn't realize this before, this is supposed to be `redis_client_` and the labels are just supposed to be `redis_`

baseplate/clients/redis_cluster.py
```diff
@@ -104,6 +108,31 @@
 #  of key value [key value ...]
 MULTI_KEY_BATCH_WRITE_COMMANDS = frozenset(["MSET", "MSETNX"])
 
+PROM_NAMESPACE = "redis_cluster"
```
nsheaps, line 111: same here, metrics are `redis_cluster_client_` and labels are just `redis_cluster_`

baseplate/clients/redis_cluster.py
```diff
@@ -347,7 +376,7 @@ class ClusterRedisContextFactory(ContextFactory):
     :param connection_pool: A connection pool.
     """
 
-    PROM_PREFIX = "bp_redis_cluster_pool"
+    PROM_PREFIX = f"{PROM_NAMESPACE}_pool"
```
nsheaps, line 379: 🔕 this is correct as `redis_cluster_pool`

nsheaps (changes requested):
Sorry I missed the prefix stuff before! That's all I see remaining.

baseplate/clients/redis.py
krav: renamed; ptal

krav:
Superseded by #690

PR #693, opened 2022-06-24
[BASE-83] - Add Configurable Log Level Support

isugimpy:
Closes BASE-83

## 💸 TL;DR
Historically, we've only had INFO and DEBUG level logging available.  This lack of flexibility has been problematic and increased our log volume considerably over time.  This change enables configurable log level support, while retaining the old behavior if a level is unspecified or if `--debug` is provided.

## 📜 Details
[Jira](https://reddit.atlassian.net/browse/BASE-83)

## ✅ Checks

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -157,11 +157,17 @@ def read_config(config_file: TextIO, server_name: Optional[str], app_name: str) 
 def configure_logging(config: Configuration, debug: bool) -> None:
     logging.captureWarnings(capture=True)
 
+    level = config.get("log_level", "INFO")
+    if level.upper() in logging._levelToName.values():
+        numeric_level = logging.getLevelName(level.upper())
+    else:
+        numeric_level = logging.getLevelName("INFO")
+
     if debug:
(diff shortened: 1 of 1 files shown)


maeivysea (approved):

KTAtkinson (commented):
Currently you can configure log level by configuring [the default logger](https://docs.python.org/3/library/logging.config.html#configuration-file-format) to have a different log level. This configuration is very verbose, but I do hesitate to have two interfaces that are so similar and so close together to configure log level. It could create confusion if the clash.

I need to think about this over the weekend though. I'll look at this again on Monday.

KTAtkinson (changes requested):
Based on [this thread](https://reddit.slack.com/archives/CCM109VKM/p1656345223795939?thread_ts=1655426571.400059&cid=CCM109VKM) in Slack, it seems like we've decided that the log level is already sufficiently configurable from the config file. We did find a bug though, using the `--debug` flag with custom loggers does nothing.

@isugimpy I will leave it up to you if you want to tackle this issue in this PR or just close this PR.

PR #696, opened 2022-06-29
Add prometheus tagged span observer

krav:
## 💸 TL;DR
Just going to leave this here in case we want to instrument spans to Prometheus. Gated behind the `metrics.tagged.prometheus` config attribute.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/observers/prom_metrics_tagged.py
+++ b/baseplate/observers/prom_metrics_tagged.py
@@ -1,0 +1,197 @@
+from random import random
+from time import time
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
+from prometheus_client import Counter
+from prometheus_client import Histogram
+
+from baseplate import _ExcInfo
+from baseplate import BaseplateObserver
+from baseplate import LocalSpan
+from baseplate import RequestContext
+from baseplate import Span
+from baseplate import SpanObserver
+from baseplate.lib import config
+
+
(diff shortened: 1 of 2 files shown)


baseplate/observers/prom_metrics_tagged.py
```diff
@@ -0,0 +1,197 @@
...
+from baseplate import RequestContext
+from baseplate import Span
+from baseplate import SpanObserver
+from baseplate.lib import config
+
+
+class _PrometheusMetrics:
+    def __init__(self, allowlist: List[str]):
+        self.server_rate = Counter(
+            "baseplate_server_span_rate", "Baseplate Server Span Counter", allowlist
```
nsheaps, line 23: These were temporarily taken out but would end up matching this commented block
https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md?plain=1#L442

nsheaps (changes requested):
rejecting to get this out of the review queue until spec can be finalized for this stuff, in particular:

```
active_client_requests - name tbd
active_server_requests - name tbd
local_span_total_requests
```

baseplate/observers/prom_metrics_tagged.py
krav: Can you explain this again? This document changed so there's no longer a line 442.

PR #700, opened 2022-07-04
Redis client upgrade

tomasz-bogun-1:
Closes # SEO-4032

Upgrades the version of the redis python client so that new redis functionality can be used. In my case, we would like to have the count aegument in the lpop function which was available since redis python client version 4.0.0

https://reddit.atlassian.net/browse/SEO-4032

New tests have been added that test the new feture.


salomon-smekecohen:
https://github.com/reddit/baseplate.py/issues/914

PR #703, opened 2022-07-11
Mark InOrderConsumerFactory as deprecated 

sydjryan:
This is not a great way to interact with Kafka - committing the offset after every message puts load on the cluster, especially with many high volume consumers. Additionally, kafka itself doesn't guarantee ordering across partitions. Kafka will deliver messages at least once, but achieving at-most-once should be done at the application layer since publishing to the cluster in many cases does not guarantee exactly-once. 

In the future, we can revisit consumer best practices and potentially create multiple new consumer types for different use cases. But for now, we should at least discourage people from using this one.


mterwill (approved):

mterwill (changes requested):
After chatting with the team, @maeivysea is going to set up some time to discuss this further, sorry for the confusion!

maeivysea (changes requested):
Harold still wants me to review even though draft (oops I lied), will re-review after the meeting!

maeivysea:
Chatted with Syd (thanks for raising!), DevX is currently focused on Snoodev Code Yellow so we're paused on other things. We'll come back to this when we have bandwidth to support!

maeivysea (changes requested):
RC for now so this comes back to me when we're ready

sydjryan:
Closing this for now, it's very old - we can come back when ready

PR #699, opened 2022-06-30
BASE-38] Add component label to local span prometheus metrics

Xaelias:
spec still being updated to reflect expectations
will update PR if spec is different than current implementation

## ✅ Checks

- [x] CI tests (if present) are passing
- [x] Adheres to code style for repo
- [x] Contributor License Agreement (CLA) completed if not a Reddit employee

PR #704, opened 2022-07-13
Re-implement prometheus metrics http/thrift server/client directly outside of the observers

Xaelias:
Follow up to https://github.com/reddit/baseplate.py/pull/687

## 💸 TL;DR

Refactoring thrift/http client/server prom metrics to instrument the "clients" directly instead of relying on observers.

## Related issues
Getting the content-length before the content of the thrift response was read returned None.
This is related to https://github.com/reddit/baseplate.py/issues/575
As a workaround, this change re-sets `self.content_length` to the appropriate value after the `SpanFinishingAppIterWrapper` is configured.

## ✅ Checks


baseplate/clients/requests.py
```diff
@@ -101,6 +108,35 @@ def http_adapter_from_config(
     return ValidatingHTTPAdapter(**kwargs)
 
 
+PROM_NAMESPACE = "http_client"
+HTTP_LABELS_COMMON = [
```
nsheaps, line 112: 🔕  note to self, we should update the spec to imply that endpoint is not included here.

This is fine 👍

baseplate/clients/requests.py
```diff
@@ -101,6 +108,35 @@ def http_adapter_from_config(
...
+    f"{PROM_NAMESPACE}_latency_seconds",
+    "Latency histogram of HTTP calls made by clients",
+    HTTP_LABELS_TERMINAL,
+    buckets=default_latency_buckets,
+)
+# Counter counting total HTTP requests started by a given client
+REQUESTS_TOTAL = Counter(
+    f"{PROM_NAMESPACE}_requests_total",
+    "Total number of HTTP requests started by a given client",
+    [*HTTP_LABELS_TERMINAL, "http_response_code"],
```
nsheaps, line 130: 🔕 note to self, we should update the spec to call this label out separately in the client as well.

This is fine 👍

baseplate/clients/requests.py
```diff
@@ -204,28 +240,57 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
...
     def send(self, request: PreparedRequest, **kwargs: Any) -> Response:
         """Send a :py:class:`~requests.PreparedRequest`."""
-        tags = {
-            "protocol": "http",
-            "http.method": request.method,
-            "http.url": request.url,
-            "http.slug": self.client_name if self.client_name is not None else self.name,
+        active_request_label_values = {
+            "http_method": request.method.lower() if request.method else "",
+            "http_client_name": self.client_name if self.client_name is not None else self.name,
```
nsheaps, line 245: 🔕 huh.... didn't realize we had two names for it. Out of scope for this PR but my brain is wondering how they differ.

baseplate/clients/requests.py
```diff
@@ -204,28 +240,57 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
...
+                http_success = ""
+
+            latency_label_values = {**active_request_label_values, "http_success": http_success}
+            requests_total_label_values = {
+                **latency_label_values,
+                "http_response_code": str(status_code),
+            }
+
+            LATENCY_SECONDS.labels(**latency_label_values).observe(time.perf_counter() - start_time)
+            REQUESTS_TOTAL.labels(**requests_total_label_values).inc()
```
nsheaps, line 292: 🔕 note to self, we should update the spec to imply that these total count metrics should say "total **completed** requests"

This is fine 👍

baseplate/clients/requests.py
nsheaps, line 112: 🔕 Just as a call out, [pending request for _address](https://github.snooguts.net/reddit/baseplate.spec/pull/48/files)

Need to discuss the cardinality further before doing this though

baseplate/clients/requests.py
```diff
@@ -204,28 +240,57 @@ def _add_span_context(self, span: Span, request: PreparedRequest) -> None:
...
+        finally:
+            if sys.exc_info()[0] is not None:
+                status_code = ""
+                http_success = "false"
+            elif response and response.status_code:
+                http_success = getHTTPSuccessLabel(response.status_code)
+                status_code = str(response.status_code)
+            else:
+                status_code = ""
+                http_success = ""
```
KTAtkinson, line 283: When would we arrive at this case? It seems like if we didn't get a response back then that does not indicate that he request was successful.

tests/unit/clients/requests_tests.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+
+class TestBaseplateSessionProm:
+    def setup(self):
+        ACTIVE_REQUESTS.clear()
+        LATENCY_SECONDS.clear()
+        REQUESTS_TOTAL.clear()
+
+    # need to patch Session() somehow...
```
KTAtkinson: It seems like these docs are maybe out of date? It seems like you figured out how to patch the session.

tests/unit/clients/requests_tests.py
```diff
@@ -0,0 +1,125 @@
...
+            REGISTRY.get_sample_value(
+                "http_client_requests_total",
+                {
+                    **prom_labels,
+                    "http_response_code": str(status_code),
+                    "http_success": http_success,
+                },
+            )
+            == 1
+        )
```
KTAtkinson, line 124: I think this can be simplified like this by using a single `with` block with the subject being the `mock_manager`. Something like this:
```
with mock.Mock() as mock_manager:
   # add mocks
  with mock.patch("baseplate.clients.requests.Session", spec=Session) as session:
      # do request
  # assert mock state
```

tests/unit/clients/thrift_tests.py
```diff
@@ -46,3 +64,160 @@ class ExampleClient:
...
+            "thrift_method": "handle",
+            "thrift_client_name": "",
+        }
+        requests_total_prom_labels = {
+            "thrift_exception_type": exc_type,
+            "thrift_baseplate_status": status,
+            "thrift_baseplate_status_code": status_code,
+        }
+
+        mock_manager = mock.Mock()
```
KTAtkinson, line 161: Same as above.

KTAtkinson (changes requested):

tests/unit/clients/requests_tests.py
Xaelias: yup leaving comments to myself 😅

baseplate/clients/requests.py
Xaelias: I don't think it should. But... I figured having the catch-all for it didn't hurt... I'm happy to remove it though.

tests/unit/clients/requests_tests.py
Xaelias: I'm not sure how that's simplified. Don't you just end up replacing `mock_manager = mock.Mock()` with `with mock.Mock() as mock_manager:` and then changing the indent of things?

(also the Mock itself is not a context manager so it can't be used directly like this in a with statement)

baseplate/clients/thrift.py
```diff
@@ -71,19 +112,19 @@ class ThriftContextFactory(ContextFactory):
 
     """
 
-    PROM_PREFIX = "bp_thrift_pool"
-    PROM_LABELS = ["client_cls"]
+    POOL_PREFIX = "thrift_client_pool"
```
nsheaps, line 115: 🔕  Looks good https://github.snooguts.net/reddit/baseplate.spec/blob/thrift-connection-pool-metrics/component-apis/prom-metrics.md?plain=1#L157-L164

baseplate/clients/thrift.py
```diff
@@ -166,89 +207,128 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
+                thrift_success = "false"
+                exception_type = exc_info[0].__name__
+                current_exc = exc_info[1]
+                if isinstance(current_exc, Error):
+                    baseplate_status_code = current_exc.code
+                    baseplate_status = ErrorCode()._VALUES_TO_NAMES.get(current_exc.code, "")
+
+            REQUEST_LATENCY.labels(
+                thrift_method=name,
+                thrift_client_name="",
```
nsheaps, line 316: 🔕 Assuming this is not set anywhere by clients/in the interface. For now, would the pool name be appropriate?

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -68,11 +75,60 @@ def close(self) -> None:
...
+
+
 def _make_baseplate_tween(
     handler: Callable[[Request], Response], _registry: Registry
 ) -> Callable[[Request], Response]:
     def baseplate_tween(request: Request) -> Response:
+        time_started = time.perf_counter()
+
+        response: Optional[Response] = None
+        endpoint = request.matched_route.pattern if request.matched_route else ""
```
nsheaps, line 128: 🔕 for pyramid I'm assuming this URL is something like `/users/:user_id/details`

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -83,8 +139,43 @@ def baseplate_tween(request: Request) -> Response:
...
                 response.app_iter = SpanFinishingAppIterWrapper(request.span, response.app_iter)
+                response.content_length = content_length
         finally:
+            http_endpoint = endpoint
+            http_method = request.method.lower()
+            http_response_code = ""
+
+            if sys.exc_info() == (None, None, None):
+                http_success = (
+                    getHTTPSuccessLabel(int(response.status_code)) if response else "false"
```
nsheaps, line 152: Why the casting to an `int` here? is it not already an `int`?

Mainly wondering if we need any safety around here for when there is no response code (eg if the connection is closed, is this `None` or -1?)

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -83,8 +139,43 @@ def baseplate_tween(request: Request) -> Response:
...
+                    **histogram_labels,
+                    "http_response_code": http_response_code,
+                }
+            ).inc()
+            REQUEST_SIZE.labels(**histogram_labels).observe(request.content_length or 0)
+            if response:
+                try:
+                    if response.content_length is not None:
+                        RESPONSE_SIZE.labels(**histogram_labels).observe(response.content_length)
+                except Exception:
```
nsheaps, line 176: I believe this is the handling of the case where it errors and content_length is not set. This is fine to me, but a comment would be nice

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -244,7 +335,6 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         span.set_tag("http.route", request.matched_route.pattern)
         span.set_tag("http.method", request.method)
         span.set_tag("peer.ipv4", request.remote_addr)
-        span.set_tag("http.request_length", request.content_length)
```
nsheaps, line 247: Is there a reason to remove this?

baseplate/frameworks/thrift/__init__.py
```diff
@@ -70,7 +107,7 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
                 # this is an expected exception, as defined in the IDL
                 span.finish()
                 raise
-            except:  # noqa: E722
+            except Exception:  # noqa: E722
```
nsheaps, line 110: :no_bell: isn't a generic exception implied without this?

baseplate/lib/prometheus_metrics.py
```diff
@@ -27,414 +19,8 @@
...
-    def get_active_requests_metric(cls) -> Gauge:
-        """Return the active_requests metrics"""
-        return cls.active_requests
-
-
 def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     """
     The HTTP success label is "true" if the status code is 2xx or 3xx, "false" otherwise.
     """
     return str(200 <= httpStatusCode < 400).lower()
```
nsheaps, line 26: 🔕 totally out of scope for this PR, but why isn't 1xx considered success?

baseplate/observers/prometheus.py
```diff
@@ -1,255 +0,0 @@
-import logging
```
nsheaps, line 1: odds are #696 / #699 will reintroduce this

tests/integration/pyramid_tests.py
```diff
@@ -279,70 +277,3 @@ class StreamingTestRequest(webtest.TestRequest):
         self.assertTrue(self.server_observer.on_finish.called)
 
         response.app_iter.close()
-
-    def test_prometheus_metrics(self):
-        self.baseplate.register(PrometheusBaseplateObserver())
```
nsheaps, line 284: Is there a reason these tests wouldn't still work without the observer?

baseplate/frameworks/pyramid/__init__.py
nsheaps: Answered my own question with the tests.

tests/integration/pyramid_tests.py
nsheaps: Answered my own question later (sorry, can't resolve till I submit)

tests/unit/frameworks/thrift_tests.py
```diff
@@ -0,0 +1,159 @@
...
+                "dec",
+                wraps=PROM_ACTIVE.labels(thrift_method="handle").dec,
+            ) as active_dec_spy_method:
+                mock_manager.attach_mock(active_dec_spy_method, "dec")
+                with expectation:
+                    context_handler.handle()
+
+        common_prom_labels = {
+            "thrift_method": "handle",
+            "thrift_success": str(exc is None).lower(),
```
nsheaps, line 131: 🔕 👌

nsheaps (approved):
Great tests. Nothing blocking though there are some questions in the review.

baseplate/clients/requests.py
Xaelias: `self.name` is the name from the context.
So if you declare your client like so:
```
baseplate.configure_context(
   {
      ...
      # see above for when to use which of these
      "foo": ExternalRequestsClient(),
      "bar": InternalRequestsClient(),
      ...
   }
)
```
we'll surface `foo` and `bar` if you don't specify another client name

Happy to change the implementation if we don't want / like that.

baseplate/clients/thrift.py
Xaelias: 👀  I forgot about this I need to do some more research

baseplate/frameworks/thrift/__init__.py
Xaelias: The linter screams if you do a bare except

baseplate/observers/prometheus.py
Xaelias: That's fine. Either it's re-introduced and it's not a big deal. Or the linter will scream.

baseplate/clients/requests.py
Xaelias: I'm not sure what you mean with "completed" here but even failed requests will show up in the current implementation. Just with maybe no status code.

baseplate/frameworks/pyramid/__init__.py
Xaelias: Yeah that cast is technically not required indeed.

This does make the assumption that this is not None. If the connection is unexpectedly closed it should be an exception which means that's the `else` code path.

baseplate/frameworks/pyramid/__init__.py
Xaelias: It waws added originally with a94f21f7d4d229757572d2f2fc714b543f9a386e for the original implementation using observers. So It's just not used is all.
Which means we should actually remove http.route too probably

KTAtkinson (approved):

baseplate/clients/thrift.py
Xaelias: It looks like originally Kris intended to use `self.namespace` back in 49f14a2bb24098024713d019994b035ee4e2bf7b

baseplate/clients/requests.py
nsheaps: No this is fine, it matches what we had before as far as I could tell, just seems strange to give it two names that are both unique to that instance. Maybe in a future revision.

baseplate/clients/thrift.py
nsheaps: I'm not sure if that's appropriate so happy to punt on that part for now, but we should get to the bottom of the value we want here.

nsheaps (approved):

baseplate/frameworks/pyramid/__init__.py
Xaelias: It's really just guarding against the case content_length is not a number... I can add a comment but I'm not 100% what I would say outside of "we can't observe None / not an integer so we just don't"

PR #706, opened 2022-07-14
Instrument redis calls (#690) patch

nataliest:
Closes # 

## 💸 TL;DR

Patch for https://github.com/reddit/baseplate.py/pull/690:
- separated prometheus prefix from labels prefix
```
PROM_PREFIX = "redis[_cluster]_client"
PROM_LABELS_PREFIX = "redis[_cluster]"
```
- rename `_startup_host` to `_startup_address` to match the spec.

## 📜 Details
[Design Doc]()

[Jira]()

in progress

## ✅ Checks

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -28,7 +28,7 @@ PROM_NAMESPACE = "redis"
 PROM_SHARED_LABELS = [
     f"{PROM_NAMESPACE}_command",
     f"{PROM_NAMESPACE}_database",
-    f"{PROM_NAMESPACE}_host",
+    f"{PROM_NAMESPACE}_address",
 ]
 LATENCY_SECONDS = Histogram(
     f"{PROM_NAMESPACE}_latency_seconds",
@@ -196,7 +196,7 @@ class MonitoredRedisConnection(redis.StrictRedis):
             success = "true"
             labels = {
                 f"{PROM_NAMESPACE}_command": command,
-                f"{PROM_NAMESPACE}_host": self.connection_pool.connection_kwargs.get("host", ""),
+                f"{PROM_NAMESPACE}_address": self.connection_pool.connection_kwargs.get("host", ""),
                 f"{PROM_NAMESPACE}_database": self.connection_pool.connection_kwargs.get("db", ""),
             }
             ACTIVE_REQUESTS.labels(**labels).inc()
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -28,7 +28,7 @@ PROM_NAMESPACE = "redis"
 PROM_SHARED_LABELS = [
     f"{PROM_NAMESPACE}_command",
     f"{PROM_NAMESPACE}_database",
-    f"{PROM_NAMESPACE}_startup_host",
+    f"{PROM_NAMESPACE}_host",
 ]
 LATENCY_SECONDS = Histogram(
     f"{PROM_NAMESPACE}_latency_seconds",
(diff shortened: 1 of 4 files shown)


KTAtkinson (changes requested):
Looks good so far! Thanks for putting this together.

Send this back my way when you have some tests!

baseplate/clients/redis_cluster.py
```diff
@@ -347,7 +378,7 @@ class ClusterRedisContextFactory(ContextFactory):
     :param connection_pool: A connection pool.
     """
 
-    PROM_PREFIX = "bp_redis_cluster_pool"
```
nsheaps, line 350: based on https://github.snooguts.net/reddit/baseplate.spec/pull/49/files#diff-ebe820b9ffcd96083b5e0695835dbada2bb1d2b3b158092ccbb947c95779129b this will need some changes to match the `redis_client` implementation

nsheaps (changes requested):

baseplate/clients/redis_cluster.py
nataliest: since `PROM_PREFIX = "redis_cluster_client"` and then `PROM_PREFIX = f"{PROM_PREFIX}_pool"`, we get `redis_cluster_client_pool`, so _I think_ my diff matches your PR changes. lmk if I missed something

baseplate/clients/redis.py
```diff
@@ -1,24 +1,56 @@
...
 
+from baseplate.lib.prometheus_metrics import default_latency_buckets
+
+PROM_PREFIX = "redis_client"
+PROM_LABELS_PREFIX = "redis"
+
+PROM_SHARED_LABELS = [
+    f"{PROM_LABELS_PREFIX}_command",
+    f"{PROM_LABELS_PREFIX}_database",
+    f"{PROM_LABELS_PREFIX}_address",
```
nsheaps: when this is set later, does the `host` include the port too?

nataliest: no, just the host, so for `"redis.url": "redis://localhost:1234/0"` we'll only get `localhost` because we call `self.connection_pool.connection_kwargs.get("host", "")`

baseplate/clients/redis.py
```diff
@@ -1,24 +1,76 @@
...
+    "Number of active requests for a given client",
+    PROM_SHARED_LABELS,
+)
+
+PROM_POOL_PREFIX = f"{PROM_PREFIX}_pool"
+PROM_LABELS = ["pool"]
+
+MAX_CONNECTIONS = Gauge(
+    f"{PROM_POOL_PREFIX}_max_size",
+    "Maximum number of connections allowed in this redisbp pool",
```
nsheaps: 🔕  might be better to replace `redisbp` with `redis client connection`

tests/unit/clients/redis_cluster_tests.py
```diff
@@ -1,9 +1,119 @@
...
 
+from unittest import mock
+
 import fakeredis
+import pytest
+
+from prometheus_client import REGISTRY
+
+try:
+    import redis
```
nsheaps: Shouldn't this be trying to import rediscluster?

Not happy about this style of skipping tests just because something the lib provides is not installed but you didn't introduce the pattern so I'm not going to block the PR for now. It runs the risk of running a test pass on baseplate and tests pass but not these, which you wouldn't know because you forgot to install something.

nsheaps (approved):

tests/unit/clients/redis_cluster_tests.py
nataliest: i didn't introduce the pattern but i'll try to remove this

KTAtkinson (approved):

PR #709, opened 2022-07-15
Replace kafka_address label with user-defined kafka_client_name

Xaelias:


nsheaps (changes requested):
Spec was updated for a service-provided client_name instead to reduce cardinality

nsheaps (approved):

PR #711, opened 2022-07-19
Add pre commit hooks

Xaelias:


KTAtkinson (changes requested):
Why add the hooks?

These steps will take quite a while and will block a commit from being submitted. This feels like it would be a bad developer experience.

Xaelias:
> Why add the hooks?
> 
> These steps will take quite a while and will block a commit from being submitted. This feels like it would be a bad developer experience.

You don't have to install / use them if you think they're detrimental. The fmt is also fairly quick in my experience, and that's the only one running on commit. The other two are running on push only, and, for me at least, I'd rather they fail locally before I push rather than pushing, and then receiving a notification that something went wrong that I may or may not see for a while (while reviewers get notified and review code that may or may not be entirely functional).

KTAtkinson:
> > Why add the hooks?
> > These steps will take quite a while and will block a commit from being submitted. This feels like it would be a bad developer experience.
> 
> You don't have to install / use them if you think they're detrimental. The fmt is also fairly quick in my experience, and that's the only one running on commit. The other two are running on push only, and, for me at least, I'd rather they fail locally before I push rather than pushing, and then receiving a notification that something went wrong that I may or may not see for a while (while reviewers get notified and review code that may or may not be entirely functional).

As long as the checks are optional I think this looks great! one last request though, could you provide docs in the README.md about these check and how to enable/disable them for those who are interested in using them?

KTAtkinson (approved):

PR #714, opened 2022-07-25
test memcached prom instrumentation

krav:
Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/tests/unit/clients/memcache_tests.py
+++ b/tests/unit/clients/memcache_tests.py
@@ -18,16 +18,17 @@ from baseplate.clients.memcache import lib as memcache_lib
 
 
 class PrometheusInstrumentationTests(unittest.TestCase):
-    # test the @_prom_instrument decorator by calling one of the decorated functions
-    def test_prometheus_decorator(self):
+    def setUp(self):
         # stub out the server span, each call to MonitoredMemcacheConnection
         # will create a new span - but we don't use that for prom
-        span = mock.Mock()
         span_inner = mock.Mock()
         span_inner.__enter__ = mock.Mock(return_value=mock.Mock())
         span_inner.__exit__ = mock.Mock(return_value=None)
-        span.make_child = lambda trace_name: span_inner
+        self.span = mock.Mock()
+        self.span.make_child = lambda trace_name: span_inner
 
+    # test the @_prom_instrument decorator by calling one of the decorated functions
+    def test_prometheus_decorator(self):
         # stub out the client, we only use the server name
         pooled_client = mock.Mock()
         pooled_client.server = "server name"
@@ -37,7 +38,7 @@ class PrometheusInstrumentationTests(unittest.TestCase):
             "memcached_command": "incr",
         }
 
-        mmc = MonitoredMemcacheConnection("test", span, pooled_client)
+        mmc = MonitoredMemcacheConnection("test", self.span, pooled_client)
         mmc.incr("blah", 1)
 
         assert (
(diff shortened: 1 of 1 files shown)


tests/unit/clients/memcache_tests.py
```diff
@@ -10,11 +10,61 @@
...
 
+from prometheus_client import REGISTRY
 from baseplate.lib.config import ConfigurationError
 from baseplate.clients.memcache import pool_from_config
+from baseplate.clients.memcache import MonitoredMemcacheConnection
 from baseplate.clients.memcache import lib as memcache_lib
 
 
+class PrometheusInstrumentationTests(unittest.TestCase):
+    def test(self):
```
KTAtkinson: Could this test have a better name?

tests/unit/clients/memcache_tests.py
```diff
@@ -10,11 +10,61 @@
...
+                prom_labels,
+            )
+            == 0
+        )
+        assert (
+            REGISTRY.get_sample_value(
+                "memcached_client_requests_total",
+                {
+                    **prom_labels,
+                    "memcached_success": "true",
```
KTAtkinson, line 56: Should we have a test where success is false as well?

tests/unit/clients/memcache_tests.py
```diff
@@ -10,11 +10,61 @@
...
+from prometheus_client import REGISTRY
 from baseplate.lib.config import ConfigurationError
 from baseplate.clients.memcache import pool_from_config
+from baseplate.clients.memcache import MonitoredMemcacheConnection
 from baseplate.clients.memcache import lib as memcache_lib
 
 
+class PrometheusInstrumentationTests(unittest.TestCase):
+    def test(self):
+        span = mock.Mock()
```
KTAtkinson: It would be great to mock out the span more precisely here if possible. Ideally this test would fail if the span API changed in such a way that the metrics could not be collected.

KTAtkinson (changes requested):

tests/unit/clients/memcache_tests.py
krav: done!

tests/unit/clients/memcache_tests.py
krav: Good call, added!

tests/unit/clients/memcache_tests.py
krav: the prom metrics get collected regardless of the span, but each function creates a new span so it needs to be stubbed out. added a comment to that effect

nsheaps (approved):

tests/unit/clients/memcache_tests.py
```diff
@@ -10,11 +10,87 @@
...
+            == 1
+        )
+
+        class FailClient(mock.Mock):
+            server = "hello"
+
+            def incr(self, _a, _b, noreply=None):
+                raise TypeError
+
+        fail_mmc = MonitoredMemcacheConnection("test", span, FailClient())
```
KTAtkinson: Can your failure test case actually be a different test case. As written if the success case fails so does this one.

KTAtkinson (changes requested):

tests/unit/clients/memcache_tests.py
krav: done

KTAtkinson (approved):

PR #716, opened 2022-07-25
Touch-ups from testing

nsheaps:
Testing revealed some merge issues and misses that are being addressed in the PR. Originally it was scoped to one fix but since they're all small and tested together, they're batched together in this PR.

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -213,9 +213,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         pool = self.engine.pool
         if isinstance(pool, QueuePool):
             self.max_connections_gauge.labels(name).set_function(pool.size)
-            self.checked_out_connections_gauge.labels(name).set_function(
-                pool.checkedout + pool.overflow
-            )
+            self.checked_out_connections_gauge.labels(name).set_function((lambda: pool.checkedout() + pool.overflow()))
 
     def report_runtime_metrics(self, batch: metrics.Client) -> None:
         pool = self.engine.pool
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -213,7 +213,9 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
         pool = self.engine.pool
         if isinstance(pool, QueuePool):
             self.max_connections_gauge.labels(name).set_function(pool.size)
-            self.checked_out_connections_gauge.labels(name).set_function((lambda: pool.checkedout() + pool.overflow()))
+            self.checked_out_connections_gauge.labels(name).set_function(
+                (lambda: pool.checkedout() + pool.overflow())
+            )
 
     def report_runtime_metrics(self, batch: metrics.Client) -> None:
         pool = self.engine.pool
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -171,7 +171,7 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
     )
 
     checked_out_connections_gauge = Gauge(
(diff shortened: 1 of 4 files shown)


baseplate/frameworks/pyramid/__init__.py
```diff
@@ -332,7 +332,8 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         )
         span.set_tag("protocol", "http")
         span.set_tag("http.url", request.url)
-        span.set_tag("http.route", request.matched_route.pattern)
+        # not sure this is accurate, as this is based on pyramid events 
+        # span.set_tag("http.route", request.matched_route.name)
```
nsheaps: Happy to remove this but harold does something funky here that makes this not the Pyramid request object (and therefore there is no `pattern`. Considering this block is talking about events, I'm not sure this is in reaction to an actual request.

baseplate/clients/sqlalchemy.py
```diff
@@ -214,7 +214,7 @@ def __init__(self, engine: Engine, name: str = "sqlalchemy"):
         if isinstance(pool, QueuePool):
             self.max_connections_gauge.labels(name).set_function(pool.size)
             self.checked_out_connections_gauge.labels(name).set_function(
```
krav: set_function doesn't work in multiprocess mode, why was this reintroduced?

baseplate/frameworks/pyramid/__init__.py
krav: This should be removed, it was added for the previous implementation that used spans. This is `endpoint = request.matched_route.pattern if request.matched_route else ""` in prom now btw.

baseplate/clients/sqlalchemy.py
Xaelias: Probably the fact we had to rebase, re-create PRs etc. etc.
It was a mistake and should be removed.

nataliest (approved):

Xaelias (approved):

KTAtkinson (approved):

baseplate/clients/sqlalchemy.py
```diff
@@ -209,14 +209,6 @@ def __init__(self, engine: Engine, name: str = "sqlalchemy"):
...
         self.time_started = 0.0
 
-        # Prometheus pool metrics
-        pool = self.engine.pool
-        if isinstance(pool, QueuePool):
-            self.max_connections_gauge.labels(name).set_function(pool.size)
-            self.checked_out_connections_gauge.labels(name).set_function(
-                pool.checkedout + pool.overflow
-            )
-
```
krav, line 219: You may want to reintroduce these in report_runtime_metrics, they're not called atm

nsheaps: They are there already :)

krav: oh they are, and checkedout _includes_ overflow, nevermind

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -325,14 +329,22 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
             if self.edge_context_factory:
                 request.edge_context = self.edge_context_factory.from_upstream(edge_payload)
 
+        endpoint = ""
+        if request.matched_route:
+            endpoint = (
+                request.matched_route.pattern
+                if request.matched_route.pattern
```
nsheaps, line 336: _one time_ I ran into an attribute error saying that `pattern` was not available on types.SimpleNamespace. Since then, it's been impossible for me to reproduce which is slightly worrisome to me. I noticed it with the init job in harold. Maybe it was old code that I caught in the middle of a rebuild, but if it wasn't, then this should be changed to a hasattr check.

KTAtkinson (approved):

PR #718, opened 2022-07-29
correct pool label

nsheaps:
label is supposed to be prefixed for redis pool

memcache pool is supposed to be idle_connections not free_connections


maeivysea (approved):
lgtm! one thought, do we know if people are relying on the now-outdated names (aka should we send comms so people know - if these are relatively new then prob not)

nsheaps:
These are new, not released yet, but good thought @maeivysea

PR #720, opened 2022-08-01
fix(metrics): make http_server http_endpoint use more general matching

nsheaps:
see: https://github.com/reddit/baseplate.py/pull/716#discussion_r931175277

```alembic -c infrared/development.ini upgrade head
XXXXX
AttributeError: 'types.SimpleNamespace' object has no attribute 'pattern'
make: *** [Makefile:41: make-test-data] Error 1
```

Tested with <service-name> initdb job, and only 2 services do this.

Even though the simplenamespace isn't the best option since it doesn't match the same contract, I don't want to break things as part of the upgrade

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -333,7 +333,7 @@ class BaseplateConfigurator:
         if request.matched_route:
             endpoint = (
                 request.matched_route.pattern
-                if request.matched_route.pattern
+                if hasattr(request.matched_route, "pattern")
                 else request.matched_route.name
             )
         ACTIVE_REQUESTS.labels(http_method=request.method.lower(), http_endpoint=endpoint).inc()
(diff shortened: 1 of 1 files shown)


Xaelias (approved):

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -333,7 +333,7 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
         if request.matched_route:
             endpoint = (
                 request.matched_route.pattern
-                if request.matched_route.pattern
+                if hasattr(request.matched_route, "pattern")
```
bradengroom: Should this be?
```
if hasattr(request.matched_route, "pattern") and request.matched_route.pattern:
```

With the old behavior, we would use `request.matched_route.name` if `request.matched_route.pattern` were `None`. Now, we will use `request.matched_route.pattern` any time it's present (even if it's `None`).

bradengroom (changes requested):
Questions about when `request.matched_route.pattern` is `None`

baseplate/frameworks/pyramid/__init__.py
nsheaps: Good call, thanks for the extra thought!

bradengroom (approved):

PR #722, opened 2022-08-02
Apply same logic for request.matched_route.pattern to middleware

nsheaps:
I'm not confident that this fixes it, but the request was to match the logic between the two and add a test to reduce the risk, which I do agree with.


tests/unit/frameworks/pyramid/http_server_prom_tests.py
```diff
@@ -126,3 +126,106 @@ def test_http_server_metric_collection_method(self, response):
             )
             == expected_response_size_count
         )
+
+    @pytest.mark.parametrize(
```
nsheaps, line 132: I debated adding a parameterized arg to this test for controlling this rather than wholly duplicating this test. This was easier, but happy to adjust

KTAtkinson (approved):

PR #725, opened 2022-08-03
fix(metrics): track 404 on new request better BASE-125 BASE-126

nsheaps:
Screen Shot 2022-08-03 at 3 54 42 PM
On the internal service we're using for testing, the negative metric decreases further when hitting a page that has no matched route

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -150,7 +150,7 @@ def _make_baseplate_tween(
                 )
             else:
                 http_endpoint = "404"
-                
+
             http_method = request.method.lower()
             http_response_code = ""
 
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -208,6 +208,9 @@ def manually_close_request_metrics(request: Request, response: Optional[Response
                 time.perf_counter() - request.reddit_start_time
             )
 
+        if hasattr(request, "content_length"):
+            REQUEST_SIZE.labels(**histogram_labels).observe(request.content_length or 0)
+
         if response:
             try:
                 if response.content_length is not None:
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -123,8 +123,8 @@ def _make_baseplate_tween(
 ) -> Callable[[Request], Response]:
     def baseplate_tween(request: Request) -> Response:
         time_started = (
-            request.start_time
-            if (hasattr(request, "start_time") and request.start_time is not None)
+            request.reddit_start_time
+            if (hasattr(request, "start_time") and request.reddit_start_time is not None)
             else time.perf_counter()
         )
 
@@ -172,7 +172,7 @@ def _make_baseplate_tween(
                 "http_success": http_success,
             }
 
-            if getattr(request, "prom_metrics_enabled", False):
+            if getattr(request, "reddit_prom_metrics_enabled", False):
                 ACTIVE_REQUESTS.labels(http_method=http_method, http_endpoint=http_endpoint).dec()
                 REQUEST_LATENCY.labels(**histogram_labels).observe(
                     time.perf_counter() - time_started
@@ -194,7 +194,7 @@ def _make_baseplate_tween(
                         pass
 
             # avoid a reference cycle
-            request.prom_metrics_enabled = False
+            request.reddit_prom_metrics_enabled = False
             request.start_server_span = None
         return response
 
@@ -202,8 +202,13 @@ def _make_baseplate_tween(
 
 
 def _manually_close_request_metrics(request: Request) -> None:
-    """Close the request metrics manually when using wsgi to make trackable requests that
-    are triggered by pyramid scripting"""
+    """
+    Close the request metrics manually when using wsgi to make trackable requests that
+    are triggered by pyramid scripting.
+
+    This is duplicated from the tween because it has specific handling around not having
+    access to a response to determine the appropriate status code and success criteria.
+    """
     http_endpoint = ""
     if request.matched_route:
         http_endpoint = (
@@ -222,7 +227,6 @@ def _manually_close_request_metrics(request: Request) -> None:
         http_response_code = "200"
     else:
         http_success = "false"
-        http_response_code = "500"
 
     histogram_labels = {
         "http_method": http_method,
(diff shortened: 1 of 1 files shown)


baseplate/frameworks/pyramid/__init__.py
```diff
@@ -307,7 +310,7 @@ def _on_new_request(self, event: pyramid.events.ContextFound) -> None:
 
         # this request didn't match a route we know
         if not request.matched_route:
-            # TODO: some metric for 404s would be good
+            ACTIVE_REQUESTS.labels(http_method=request.method.lower(), http_endpoint="404").inc()
```
nsheaps: not against making this an else clause later

KTAtkinson (approved):

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -148,9 +148,10 @@ def baseplate_tween(request: Request) -> Response:
                     if (hasattr(request.matched_route, "pattern") and request.matched_route.pattern)
                     else request.matched_route.name
                 )
-            http_method = request.method.lower()
+            else:
+                http_endpoint = "404"
```
Xaelias: are we sure 404 is the only reason we end up in this case?

nsheaps: Pretty sure, about 85% confidence. [Docs](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#the-matched-route)

For you to get a 5xx it has to already be in there.

My guess is that technically a middleware could mess with it and unset it or something strange, but I don't think there's benefit to supporting those strange use cases

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -306,10 +365,22 @@ def _on_application_created(self, event: pyramid.events.ApplicationCreated) -> N
...
+        if request.matched_route:
+            endpoint = (
+                request.matched_route.pattern
+                if (hasattr(request.matched_route, "pattern") and request.matched_route.pattern)
+                else request.matched_route.name
+            )
+        else:
+            endpoint = "404"
+        request.prom_metrics_enabled = True
+        request.start_time = time.perf_counter()
```
Xaelias: I'm not entirely a fan of starting to put our own fields inside of the request field. But if we're going to do that I would do two things:
- give them very reddit names (baseplate_prometheus_start_time or smth)
- why not put the endpoint in there as well?

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -166,29 +171,83 @@ def baseplate_tween(request: Request) -> Response:
...
 
             # avoid a reference cycle
+            request.prom_metrics_enabled = False
             request.start_server_span = None
         return response
 
     return baseplate_tween
 
 
+def _manually_close_request_metrics(request: Request) -> None:
```
Xaelias: This looks like a lot of code duplication. What prevents us from calling this in the tween above?

baseplate/frameworks/pyramid/__init__.py
nsheaps: I agree, not a huge fan either, but I think it's better and easier than most of the alternatives since the request is the thing that's consistently passed between funcs

> give them very reddit names

because they're not specific to reddit, which does open us more to collisions but I'm not really concerned about that. how about a `reddit_...` prefix on them so it's not too long, since the data itself isn't actually specific to prom for the start time?

> put the endpoint in there

it already is via request.matched_route

baseplate/frameworks/pyramid/__init__.py
nsheaps: it is partially, but this has specific handling around not having a response to process things like sizes and response codes from

baseplate/frameworks/pyramid/__init__.py
Xaelias: > it already is via request.matched_route

well except we have this string of 3 different logic tests because we can't trust it's there

That's why I was saying that :-)

Xaelias: > because they're not specific to reddit, which does open us more to collisions but I'm not really concerned about that.

Well you threw me for a loop when I read `request.start_time` because I didn't understand where it was coming from at first.  
If it had an obnoxiously long name related to reddit/baseplate/prometheus it would have been more obvious is all. Also yes name collision :-)

baseplate/frameworks/pyramid/__init__.py
Xaelias: But that would be true coming from the tween too correct?

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -141,52 +139,90 @@ def baseplate_tween(request: Request) -> Response:
...
+    # ensure any active counters have been incremented before decrementing them and tracking the
+    # rest of the request
+    if getattr(request, "reddit_prom_metrics_enabled", False):
+        http_endpoint = ""
+        if (
+            hasattr(request, "reddit_tracked_endpoint")
+            and request.reddit_tracked_endpoint is not None
+        ):
+            http_endpoint = request.reddit_tracked_endpoint
+        elif request.matched_route:
```
nsheaps, line 167: I doubt we'll hit this block but I wanted to be safe

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -141,52 +139,90 @@ def baseplate_tween(request: Request) -> Response:
...
             else:
-                http_success = "false"
-
-            histogram_labels = {
-                "http_method": http_method,
-                "http_endpoint": http_endpoint,
-                "http_success": http_success,
+                http_success = "true"
+                http_response_code = "200"
+        else:
```
nsheaps, line 188: worth noting that any exception will not track a response code

baseplate/frameworks/pyramid/__init__.py
nsheaps: Consolidated the logic

baseplate/frameworks/pyramid/__init__.py
nsheaps: prefixed with reddit_ and tried to re-use the endpoint where possible

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -141,52 +139,93 @@ def baseplate_tween(request: Request) -> Response:
...
 
-            # avoid a reference cycle
-            request.start_server_span = None
-        return response
+        if response:
+            try:
+                if response.content_length is not None:
+                    RESPONSE_SIZE.labels(**histogram_labels).observe(response.content_length)
+            except Exception:
+                pass
```
Xaelias: Maybe do the same logic between the two? It's "weird" that we use `hasattr` for one of them and use a `try/except` for the other.

Xaelias (dismissed):
Feel free to ignore my request for change

baseplate/frameworks/pyramid/__init__.py
nsheaps: resolved, though a part of me thinks anything and everything analytics should be wrapped in a giant try so metrics breaking doesn't break the response

Xaelias: not against the idea FWIW

Xaelias (approved):

PR #726, opened 2022-08-04
Remove assert statement from config.py

KTAtkinson:
## 💸 TL;DR
Asserts in Python make errors hard to track down and can be disabled at runtime if particular flags are provided. These things together make asserts a bad pattern in production code. This PR is an example PR for a bug to get rid of all asserts.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/lib/config.py
+++ b/baseplate/lib/config.py
@@ -495,7 +495,8 @@ class SpecParser(Parser[ConfigNamespace]):
     def parse(self, key_path: str, raw_config: RawConfig) -> ConfigNamespace:
         parsed = ConfigNamespace()
         for key, spec in self.spec.items():
-            assert "." not in key, "dots are not allowed in keys"
+            if "." in key:
+                raise AssertionError("dots are not allowed in keys")
 
             if key_path:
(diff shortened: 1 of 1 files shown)


baseplate/lib/config.py
```diff
@@ -495,7 +495,8 @@ def __init__(self, spec: ConfigSpec):
     def parse(self, key_path: str, raw_config: RawConfig) -> ConfigNamespace:
         parsed = ConfigNamespace()
         for key, spec in self.spec.items():
-            assert "." not in key, "dots are not allowed in keys"
+            if "." in key:
+                raise AssertionError("dots are not allowed in keys")
```
RyanBalfanz, line 499: Now that this is being changed is there a better/more semantic exception that may be raised?

RyanBalfanz:
> Asserts in Python make errors hard to track down and can be disabled at runtime if particular flags are provided. These things together make asserts a bad pattern in production code. This PR is an example PR for a bug to get rid of all asserts.

Assertions have their place and are 👌 if used correctly, but I agree this is not one of them.

PR #728, opened 2022-08-04
[BASE-127] Fix thrift client connection reuse (introduced in 2.4.13)

Xaelias:
Emit prometheus metrics for individual retry calls instead of whole
"transaction"
Do not emit metrics if the connection cannot be established to begin
with. (self.pool.connection() raises an exception)

The documentation for thrift_pool.py:ThriftConnectionPool.connection() reads:
> When the context is exited, the connection is returned to the pool.
> However, if it was exited via an unexpected Thrift exception, the
> connection is closed instead because the state of the connection is
> unknown.

As of 2.4.13, TTransportException (used to decide whether we should
retry or not) is raised, but caught before the context manager handling
pool.connection() can catch it. Which means that on retries, the
connections would end up in an unknown state, returned to the pool, and
yield potential bad results on subsequent uses. This change fixes that.

## ✅ Checks

### DIFF
--- a/baseplate/clients/thrift.py
+++ b/baseplate/clients/thrift.py
@@ -207,15 +207,15 @@ def _build_thrift_proxy_method(name: str) -> Callable[..., Any]:
         trace_name = f"{self.namespace}.{name}"
         last_error = None
 
-        try:
-            for time_remaining in self.retry_policy:
-                start_time = time.perf_counter()
-                ACTIVE_REQUESTS.labels(
-                    thrift_method=name,
-                    thrift_client_name=self.namespace,
-                ).inc()
-
+        for time_remaining in self.retry_policy:
+            try:
                 with self.pool.connection() as prot:
+                    start_time = time.perf_counter()
+                    ACTIVE_REQUESTS.labels(
+                        thrift_method=name,
+                        thrift_client_name=self.namespace,
+                    ).inc()
+
                     span = self.server_span.make_child(trace_name)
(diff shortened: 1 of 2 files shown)


KTAtkinson (approved):

baseplate/clients/thrift.py
```diff
@@ -266,7 +266,7 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
                         last_error = str(exc)
                         if exc.inner is not None:
                             last_error += f" ({exc.inner})"
-                        continue
+                        raise  # we need to raise all exceptions so that self.pool.connect() self-heals
```
krav, line 269: Oof! Good catch.

tests/unit/clients/thrift_tests.py
```diff
@@ -175,25 +176,26 @@ def handle(*args, **kwargs):
...
             )
-            == 1
+            == tries
         )
         assert REGISTRY.get_sample_value("thrift_client_active_requests", prom_labels) == 0
         assert mock_manager.mock_calls == [
             mock.call.inc(),
             mock.call.dec(),
-        ]  # ensures we first increase number of active requests
+        ]*tries  # ensures we first increase number of active requests
```
krav: This needs the linter

krav (approved):
This function getting so long and deeply nested seems to make it a bit of a liability, it might be good to split it up

nsheaps (approved):

PR #729, opened 2022-08-05
Add warnings and recommendations to docs about timeout-related connection leaks

RoganMurley:
## 💸 TL;DR
This PR updates the docs with a warning about `ServerTimeout` exceptions, and a recommendation to avoid SqlAlchemy connection pool leaks until the issue is resolved.

### DIFF
--- a/docs/api/baseplate/clients/sqlalchemy.rst
+++ b/docs/api/baseplate/clients/sqlalchemy.rst
@@ -53,10 +53,11 @@ Configuration
 -------------
 
 .. warning::
-   :py:class:`sqlalchemy.pool.QueuePool` does not handle :py:class:`~baseplate.observers.timeout.ServerTimeout`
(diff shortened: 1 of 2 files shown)


docs/api/baseplate/clients/sqlalchemy.rst
```diff
@@ -52,6 +52,14 @@ request::
 Configuration
 -------------
 
+.. warning::
+
+   :py:class:`sqlalchemy.pool.QueuePool` does not handle `ServerTimeout`
```
nsheaps, line 57: Is there a code example of how to do this the right way vs wrong way? That'd be great to include here

(generally I prefer it when saying "Don't do it this way" to guide people through the right way of doing it (though the starting pointers of "do this instead" helps a lot))

KTAtkinson (changes requested):
Generally looks good +1 to Nate's comment.

PR #730, opened 2022-08-05
discussion only, please ignore

nsheaps:


baseplate/clients/thrift.py
```diff
@@ -207,14 +207,14 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
-        try:
-            for time_remaining in self.retry_policy:
-                start_time = time.perf_counter()
-                ACTIVE_REQUESTS.labels(
-                    thrift_method=name,
-                    thrift_client_name=self.namespace,
-                ).inc()
+        for time_remaining in self.retry_policy:
+            start_time = time.perf_counter()
+            ACTIVE_REQUESTS.labels(
```
nsheaps: This should probably be after it acquires a connection since it can't make the request without a connection, but I could be swayed either way

PR #731, opened 2022-08-05
Use context manager or annotations for prometheus Gauge when possible /

Xaelias:
applicable using g.track_inprogress()

This avoids having to manually call g.inc() and g.dec(), potentially
introducing bad metric reporting when exceptions are not handled
properly etc.

## ✅ Checks

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -194,20 +194,19 @@ class MonitoredRedisConnection(redis.StrictRedis):
         command = args[0]
         trace_name = f"{self.context_name}.{command}"
 
-        with self.server_span.make_child(trace_name):
+        labels = {
+            f"{PROM_LABELS_PREFIX}_command": command,
+            f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
+                "client_name", ""
+            ),
+            f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get("db", ""),
+            f"{PROM_LABELS_PREFIX}_type": "standalone",
+        }
+        with self.server_span.make_child(trace_name), ACTIVE_REQUESTS.labels(
+            **labels
+        ).track_inprogress():
             start_time = perf_counter()
             success = "true"
-            labels = {
-                f"{PROM_LABELS_PREFIX}_command": command,
-                f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
-                    "client_name", ""
-                ),
-                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
-                    "db", ""
-                ),
-                f"{PROM_LABELS_PREFIX}_type": "standalone",
-            }
-            ACTIVE_REQUESTS.labels(**labels).inc()
 
             try:
                 res = super().execute_command(command, *args[1:], **kwargs)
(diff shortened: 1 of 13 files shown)


nsheaps (commented):
Holding on this until the next release is cut to reduce the scope of potential changes and failures.

baseplate/frameworks/queue_consumer/kombu.py
```diff
@@ -139,14 +139,17 @@ def handle(self, message: kombu.Message) -> None:
...
         try:
             # We place the call to ``baseplate.make_server_span`` inside the
             # try/except block because we still want Baseplate to see and
             # handle the error (publish it to error reporting)
-            with self.baseplate.make_server_span(context, self.name) as span:
+            with self.baseplate.make_server_span(
+                context, self.name
+            ) as span, AMQP_ACTIVE_MESSAGES.labels(
+                **prometheus_labels._asdict()
+            ).track_inprogress():
```
KTAtkinson, line 152: This is a little difficult to read. Maybe something like this is better?

```suggestion
            with ( 
                self.baseplate.make_server_span(
                    context, 
                    self.name,
                ) as span,
                AMQP_ACTIVE_MESSAGES.labels(
                    **prometheus_labels._asdict(),
                ).track_inprogress()
            ):
```

baseplate/clients/redis.py
```diff
@@ -194,20 +194,19 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+            f"{PROM_LABELS_PREFIX}_command": command,
+            f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
+                "client_name", ""
+            ),
+            f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get("db", ""),
+            f"{PROM_LABELS_PREFIX}_type": "standalone",
+        }
+        with self.server_span.make_child(trace_name), ACTIVE_REQUESTS.labels(
+            **labels
+        ).track_inprogress():
```
KTAtkinson, line 207: This is a bit hard to read. How does this look:
```suggestion
        with (
            self.server_span.make_child(trace_name), 
            ACTIVE_REQUESTS.labels(**labels).track_inprogress(),
        ):
```

tests/unit/frameworks/queue_consumer/kombu_tests.py
```diff
@@ -79,57 +79,40 @@ def test_handle(self, context, span, baseplate, name, message):
...
+                mock.call("amqp.routing_key", "routing-key"),
+                mock.call("amqp.consumer_tag", "consumer-tag"),
+                mock.call("amqp.delivery_tag", "delivery-tag"),
+                mock.call("amqp.exchange", "exchange"),
+            ],
+            any_order=True,
+        )
+        handler_fn.assert_called_once_with(context, message.decode(), message)
+        message.ack.assert_called_once()
+        assert (
```
KTAtkinson, line 99: Please add descriptive errors to these asserts. I know this is not the style in most of these tests, but it will be soon.

KTAtkinson (changes requested):

maeivysea (changes requested):
(appeasing harold)

baseplate/frameworks/queue_consumer/kombu.py
Xaelias: I thought the `with (...)` syntax was not introduced until later versions of python? But I can try

tests/unit/frameworks/queue_consumer/kombu_tests.py
Xaelias: All these tests existed before we came in there. I'm just changing the indent here. Can you point me to an example of what you're looking for?

baseplate/frameworks/queue_consumer/kombu.py
Xaelias: This fails the linter. I'm pretty sure this syntax was only introduced in 3.10 too. That's also why we can't do things like `@h.labels(...).time()` because that's only supported in 3.10 as well.

KTAtkinson (approved):
I could have sworn I approved this on Friday! Sorry about the wait.

maeivysea (approved):

maeivysea:
(sorry not sure what happened, this disappeared from harold)

PR #732, opened 2022-08-11
instrument redis pipelined requests

krav:
Closes #BASE-137 

## 💸 TL;DR

We might want to drop the `command` label for these?

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -282,7 +282,6 @@ class MonitoredRedisPipeline(Pipeline):
             }
             num_reqs = len(self.command_stack)
             commands = [args[0] for (args, options) in self.command_stack]
-            print(labels)
             ACTIVE_REQUESTS.labels(**labels).inc()
             try:
                 return super().execute(**kwargs)
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -283,10 +283,9 @@ class MonitoredRedisPipeline(Pipeline):
             num_reqs = len(self.command_stack)
             commands = [args[0] for (args, options) in self.command_stack]
 
-            # Whether chained or pipelined it's one concurrent transaction
-            ACTIVE_REQUESTS.labels(**labels).inc()
             try:
-                return super().execute(**kwargs)
+                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
+                    return super().execute(**kwargs)
             except:  # noqa: E722
                 success = "false"
                 raise
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -282,14 +282,15 @@ class MonitoredRedisPipeline(Pipeline):
             }
             num_reqs = len(self.command_stack)
             commands = [args[0] for (args, options) in self.command_stack]
-
+            print(labels)
+            ACTIVE_REQUESTS.labels(**labels).inc()
             try:
-                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
-                    return super().execute(**kwargs)
+                return super().execute(**kwargs)
             except:  # noqa: E722
                 success = "false"
                 raise
             finally:
+                ACTIVE_REQUESTS.labels(**labels).dec()
                 for command in commands:
                     result_labels = {
                         **labels,
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -267,7 +267,6 @@ class MonitoredRedisPipeline(Pipeline):
     # pylint: disable=arguments-differ
     def execute(self, **kwargs: Any) -> Any:
         with self.server_span.make_child(self.trace_name):
-
             success = "true"
             start_time = perf_counter()
             labels = {
@@ -280,9 +279,9 @@ class MonitoredRedisPipeline(Pipeline):
                 ),
                 f"{PROM_LABELS_PREFIX}_type": "standalone",
             }
-            num_reqs = len(self.command_stack)
-            commands = [args[0] for (args, options) in self.command_stack]
+
             ACTIVE_REQUESTS.labels(**labels).inc()
+
             try:
                 return super().execute(**kwargs)
             except:  # noqa: E722
@@ -290,16 +289,12 @@ class MonitoredRedisPipeline(Pipeline):
                 raise
             finally:
                 ACTIVE_REQUESTS.labels(**labels).dec()
-                for command in commands:
-                    result_labels = {
-                        **labels,
-                        f"{PROM_LABELS_PREFIX}_success": success,
-                        f"{PROM_LABELS_PREFIX}_command": command,
-                    }
-                    REQUESTS_TOTAL.labels(**result_labels).inc()
-                    LATENCY_SECONDS.labels(**result_labels).observe(
-                        (perf_counter() - start_time) / num_reqs
-                    )
+                result_labels = {
+                    **labels,
+                    f"{PROM_LABELS_PREFIX}_success": success,
+                }
+                REQUESTS_TOTAL.labels(**result_labels).inc()
+                LATENCY_SECONDS.labels(**result_labels).observe(perf_counter() - start_time)
 
 
 class MessageQueue:
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -267,7 +267,41 @@ class MonitoredRedisPipeline(Pipeline):
     # pylint: disable=arguments-differ
     def execute(self, **kwargs: Any) -> Any:
         with self.server_span.make_child(self.trace_name):
-            return super().execute(**kwargs)
+
+            success = "true"
+            start_time = perf_counter()
+            labels = {
+                f"{PROM_LABELS_PREFIX}_command": "pipeline",
+                f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
+                    "client_name", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
+                    "db", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_type": "standalone",
+            }
+            num_reqs = len(self.command_stack)
+            commands = [args[0] for (args, options) in self.command_stack]
+
+            # Whether chained or pipelined it's one concurrent transaction
+            ACTIVE_REQUESTS.labels(**labels).inc()
+            try:
+                return super().execute(**kwargs)
+            except:  # noqa: E722
+                success = "false"
+                raise
+            finally:
+                ACTIVE_REQUESTS.labels(**labels).dec()
+                for command in commands:
+                    result_labels = {
+                        **labels,
+                        f"{PROM_LABELS_PREFIX}_success": success,
+                        f"{PROM_LABELS_PREFIX}_command": command,
+                    }
+                    REQUESTS_TOTAL.labels(**result_labels).inc()
+                    LATENCY_SECONDS.labels(**result_labels).observe(
+                        (perf_counter() - start_time) / num_reqs
+                    )
 
 
 class MessageQueue:
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -290,7 +290,6 @@ class MonitoredRedisPipeline(Pipeline):
                 success = "false"
                 raise
             finally:
-                ACTIVE_REQUESTS.labels(**labels).dec()
                 for command in commands:
                     result_labels = {
                         **labels,
--- a/baseplate/clients/redis_cluster.py
+++ b/baseplate/clients/redis_cluster.py
@@ -513,13 +513,14 @@ class MonitoredClusterRedisPipeline(ClusterPipeline):
             num_reqs = len(self.command_stack)
             commands = [cmd.args[0] for cmd in self.command_stack]
 
+            ACTIVE_REQUESTS.labels(**labels).inc()
             try:
-                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
-                    return super().execute(**kwargs)
+                return super().execute(**kwargs)
             except:  # noqa: E722
                 success = "false"
                 raise
             finally:
+                ACTIVE_REQUESTS.labels(**labels).dec()
                 for command in commands:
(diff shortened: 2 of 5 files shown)


baseplate/clients/redis_cluster.py
```diff
@@ -498,4 +498,38 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
+                    "db", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_type": "standalone",
+            }
+            num_reqs = len(self.command_stack)
+            commands = [cmd.args[0] for cmd in self.command_stack]
+
+            # Whether chained or pipelined it's one concurrent transaction
+            ACTIVE_REQUESTS.labels(**labels).inc()
```
Xaelias, line 513: https://github.com/reddit/baseplate.py/pull/731
you can do something like this for the active requests

krav: Done

baseplate/clients/redis_cluster.py
```diff
@@ -498,4 +498,37 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+            commands = [cmd.args[0] for cmd in self.command_stack]
+
+            try:
+                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
+                    return super().execute(**kwargs)
+            except:  # noqa: E722
+                success = "false"
+                raise
+            finally:
+                ACTIVE_REQUESTS.labels(**labels).dec()
```
nsheaps, line 521: the `with` block covers this case

baseplate/clients/redis_cluster.py
```diff
@@ -498,4 +498,37 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
+                    "db", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_type": "standalone",
+            }
+            num_reqs = len(self.command_stack)
+            commands = [cmd.args[0] for cmd in self.command_stack]
+
+            try:
+                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
```
nsheaps: [under the hood](https://sourcegraph.com/github.com/prometheus/client_python/-/blob/prometheus_client/context_managers.py?L37:7), this still uses `inc` and `dec`, can you please add a test for this (happy path and a case where `execute` returns an exception)

tests/unit/clients/redis_tests.py
```diff
@@ -117,6 +127,15 @@ def test_execute_command(self, monitored_redis_connection, expected_labels):
         )
         assert REGISTRY.get_sample_value(f"{REQUESTS_TOTAL._name}_total", expected_labels) == 1
 
+    def test_pipeline_instrumentation(self, monitored_redis_connection, expected_labels):
+        monitored_redis_connection.pipeline("test").set("hello", 42).set("goodbye", 23).execute()
+        labels = {**expected_labels, "redis_command": "SET", "redis_success": "true"}
+        assert REGISTRY.get_sample_value(f"{REQUESTS_TOTAL._name}_total", labels) == 2.0
```
nsheaps: I don't see a test for the `pipeline` part here

nsheaps (changes requested):
Some stuff to address before this looks good to me

tests/unit/clients/redis_tests.py
krav: what do you mean? it's the same metric. the `.pipeline("test")` call instantiates the pipelined connection

baseplate/clients/redis_cluster.py
krav: as i was rewriting this i was thinking i'd leave something behind, and that i did

tests/unit/clients/redis_tests.py
```diff
@@ -117,6 +127,15 @@ def test_execute_command(self, monitored_redis_connection, expected_labels):
...
+        assert (
```
KTAtkinson: Please provide a descriptive error message. This isn't the current style, but it helps if your test fails in the future.

KTAtkinson (approved):

baseplate/clients/redis_cluster.py
krav: @Xaelias Any clue as how to do this? Looking at https://github.com/reddit/baseplate.py/pull/731/files#diff-a91d47420f289905a951c1be7b82eb613e3d3ca01e72871564b122a6a715774a the tests were just removed.

tests/unit/clients/redis_tests.py
krav: Added

baseplate/clients/redis.py
```diff
@@ -269,7 +269,39 @@ def __init__(
...
+            commands = [args[0] for (args, options) in self.command_stack]
+
+            try:
+                with ACTIVE_REQUESTS.labels(**labels).track_inprogress():
+                    return super().execute(**kwargs)
+            except:  # noqa: E722
+                success = "false"
+                raise
+            finally:
+                for command in commands:
```
krav: Since no one's commented on it: This has a linear performance impact with the number of commands in the pipeline. It's probably not a noticeable impact, but I think it's unexpected from instrumentation. An option here is to deviate from the spec and drop the `command` label. The call would then just become a singular `.inc(len(self.command_stack))`.

baseplate/clients/redis_cluster.py
nsheaps: I still don't see a test for this, I think I'd be satisfied by setup resetting the counter for calls and just checking that inc and dec were both called once and the resulting value is 0 (which you already have)

nsheaps (changes requested):
Would like a test for the context manager handled track_inprogress call

baseplate/clients/redis_cluster.py
krav: The old strategy to test doesn't seem to work with this. @Xaelias Did you spend any effort figuring out how to test this for #731? 

Happy to revert to using `inc` and `dec` as this PR originally had.

krav:
> Would like a test for the context manager handled track_inprogress call

Changed it back to inc/dec with tests. Also added tests for the cluster client.

baseplate/clients/redis.py
```diff
@@ -269,7 +269,39 @@ def __init__(
...
+                raise
+            finally:
+                ACTIVE_REQUESTS.labels(**labels).dec()
+                for command in commands:
+                    result_labels = {
+                        **labels,
+                        f"{PROM_LABELS_PREFIX}_success": success,
+                        f"{PROM_LABELS_PREFIX}_command": command,
+                    }
+                    REQUESTS_TOTAL.labels(**result_labels).inc()
```
nsheaps: Why are we doing this per command in the pipeline? Wouldn't the other commands be tracked separately from the pipeline itself via the `super` call?

tests/integration/redis_cluster_tests.py
```diff
@@ -156,3 +157,12 @@ def test_pipeline(self):
...
         self.assertIsNone(span_observer.on_finish_exc_info)
+
+        labels = {
+            "redis_client_name": "",
+            "redis_type": "standalone",
+            "redis_database": "0",
+            "redis_command": "GET",
+            "redis_success": "true",
+        }
+        assert REGISTRY.get_sample_value("redis_client_requests_total", labels) == 5.0
```
nsheaps: Why test only the gets?

I feel like the test here should spell out what we expect to track, at minimum for clarity purposes:

total = 8
gets = 5
sets = 1
delete = 1
pipeline = 1

baseplate/clients/redis.py
nsheaps: Separately I'm assuming that the active requests never track the individual commands within the pipeline since they're sent together?

krav: > Why are we doing this per command in the pipeline? Wouldn't the other commands be tracked separately from the pipeline itself via the `super` call?

If you look at what that `super` call does it never hits any code path we've already instrumented. We assumed it did, which is why we didn't instrument this to begin with. What redis actually does is [pack the commands into one request](https://github.com/redis/redis-py/blob/f9f9d06c9951f8536bf9321dcebc96759eae03e0/redis/asyncio/client.py#L1261-L1266).

Hence its one "active request"/connection, but several commands being executed. See my other comment about including the individual names of commands executed, it's my preference to remove it but that's not compliant with the spec.

tests/integration/redis_cluster_tests.py
krav: Aren't we then just testing the redis library? Every command follows the exact same code path in baseplate. It's passed through redis [pretty much always the same way](https://github.com/redis/redis-py/blob/031b2087b57713d29b98140fb1f0b9f7b4fc7a21/redis/commands/core.py#L1410-L1435), and back to our mocked connection object. 

Happy to add it, but would appreciate this kind of feedback up front.

baseplate/clients/redis.py
krav: Actually, looking at the clustered Redis implementation (only), on retry it [can call `execute_command` on `RedisCluster`](https://github.com/redis/redis-py/blob/fd9fea6bc07bf0970a5a42c5ec1788272446910c/redis/cluster.py#L1998-L2026). As I understand it we want this to count in the metrics for every attempt, but potentially there are pitfalls.

baseplate/clients/redis.py
```diff
@@ -267,7 +267,39 @@ def __init__(
     # pylint: disable=arguments-differ
     def execute(self, **kwargs: Any) -> Any:
         with self.server_span.make_child(self.trace_name):
-            return super().execute(**kwargs)
+
```
KTAtkinson: nit: it looks like there is an additional blank line here

baseplate/clients/redis.py
nsheaps: Cool, thanks for the explanation!

tests/integration/redis_cluster_tests.py
nsheaps: I can't provide feedback on code that hasn't been written yet :)

To be clear, what I'm calling out here is that you changed metrics, did a bunch of things to work through the pipeline code you wrote (which is great), but then only tested part of it. Tests should capture every aspect that we want to track, and in this case, it was a miss from the pipeline tracking. You got a test for tracking the gets within the pipeline, but not the pipeline tracking itself (which would fail because you never increment the total count for pipeline calls, nor are you tracking inc or dec for the active call) which would have been covered if writing assertions for the behavior that the test is performing (that was my original comment [here](https://github.com/reddit/baseplate.py/pull/732#discussion_r945893821) )

I'm not totally clear myself on what we want, which is partially why I'm asking for the tests that cover all the metrics we expect to be emitted from the test case here. Defining it as tests makes it much clearer than is defined elsewhere and easier to validate. Not defining it leads to ambiguity and potential for change without realizing that it's not fully covered by the unit test.

baseplate/clients/redis.py
```diff
@@ -267,7 +267,39 @@ def __init__(
...
+            success = "true"
+            start_time = perf_counter()
+            labels = {
+                f"{PROM_LABELS_PREFIX}_command": "pipeline",
+                f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
+                    "client_name", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
+                    "db", ""
```
KTAtkinson, line 278: Shouldn't this default to `0`?

baseplate/clients/redis.py
KTAtkinson: Ideally, it doesn't make sense to associate the time the entire pipeline with each individual command in the pipeline. If folks want instrumentation per command they can provide a connection with instrumentation. Do you know how this work currently for Wavefront users?

baseplate/clients/redis_cluster.py
```diff
@@ -497,4 +497,37 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
+
+            success = "true"
+            start_time = perf_counter()
+            labels = {
+                f"{PROM_LABELS_PREFIX}_command": "pipeline",
+                f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
+                    "client_name", ""
+                ),
+                f"{PROM_LABELS_PREFIX}_database": self.connection_pool.connection_kwargs.get(
+                    "db", ""
```
KTAtkinson, line 508: Same as above, the default value here should be `0`.

tests/integration/redis_cluster_tests.py
KTAtkinson: TBH, this shouldn't be tacked onto an existing test, you should specifically have a test for metrics separate from any other test so that it's apparent what we're testing and what's not being tested.

KTAtkinson (changes requested):

baseplate/clients/redis.py
krav: This would then deviate from what we do in `MonitoredRedisConnection`.

baseplate/clients/redis.py
krav: Weird this isn't caught by the formatter.

baseplate/clients/redis_cluster.py
krav: Same here, we should change all of the defaults for redis in this case.

tests/integration/redis_cluster_tests.py
krav: Are all these things not all covered by the unit test? This should maybe not have made it into the PR. I added it for some assurance for myself early on and is now pretty redundant.

I don't understand your original comment, and you never followed it up. What does "the pipeline tracking itself" mean?

Looking at it, I think maybe there's a misunderstanding here that `pipeline()` is effectful? It's not, it just instantiates a `Pipeline(Redis)`. It's not until you call `execute()` that the pipeline start command `MULTI` is sent down the wire, so it's never counted against any of our metrics.

nsheaps (approved):

KTAtkinson (approved):

PR #733, opened 2022-08-18
BASE-163 - fix(thrift server) - Have the thrift server populate label…

Xaelias:
…s for prometheus metrics even when the error does not come from baseplate if possible.

hrift_baseplate_status_code - will be set to exc.code if available
hrift_baseplate_status - will be set to the corresponding _baseplate_
corresponding name for this status code. We do not have a way to get the
name for a new exception not defined in baseplate, or for a name that
was changed from the one originally set in baseplate.

## ✅ Checks


bradengroom (approved):

KTAtkinson (approved):
Thanks for the great comments!

PR #734, opened 2022-08-21
[live-data-fetcher] Anonymize s3 client in sidecar to allow unsigned requests

SiddharthManoj:
## 💸 TL;DR

This is a bug-fix for an issue we found with some services unable to access public resources in an s3 bucket from the AWS Kubernetes Prod account.

Some services were seeing the following error: `"An error occurred (AccessDenied) when calling the GetObject operation"` despite the objects/resources being fully public. All of this was despite anonymous local retrieval via an incognito browser being successful.

## 📜 Details

This bug was discovered as part of the [Zookeeper -> S3](https://docs.google.com/document/d/1vaSg31jkIuh5yf-wt2-SdU6N6T7w8Em7X9QKeKphd2c/edit) migration project for Experiments.

See [Migration Timeline](https://docs.google.com/document/d/1U5UUjkMS3l13KZ1xoQHXU1NJHk1t1YcOQgMbQKRB_jU/edit) for time sequence of events, findings, issues we ran into, and our resolutions.

For this migration, we require services to be able to READ public encrypted objects in an S3 bucket on the AWS Kubernetes Prod account. 

However, our first launch attempt saw services such as `reddit-service-admin-action` encountering [AccessDenied](https://app.reddit.logdna.com/bca89a3380/logs/view?t=timestamp%3A1660591916718&a=1660591916718.1512432119740129287&q=%22An%20error%20occurred%20(AccessDenied)%20when%20calling%20the%20GetObject%20operation%22) errors. This issue was especially bizarre given that the resources in the s3 bucket were publicly accessible (verified via incognito browser access).

Thanks to @ckwang8128's magic debugging skills, we discovered that because the s3 requests were signed, credentials were being pulled from local service pods thus making it so that services in a different AWS account than `AWS Kubernetes Prod` were unable to access our public resources. By making the s3 client anonymized, we are now treating the requests as unsigned - accessing the public resources will no longer try to profile for permissions from an AWS account.

This change was sanity checked as a script on a pod - attempted to access a resource in a different cluster (different AWS account). Without this change, access is restricted. After this change, access was allowed. 

Setup of anonymous client and retrieval of object is as follows:

```
import boto3
from botocore import UNSIGNED
from botocore.client import Config

s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED),region_name='us-east-1')
s3.get_object(Bucket="reddit-experiment-config-prod-active-configs", Key="7a9780e27ef4a671a41a3b98478caa43_experiment_config.json", SSECustomerKey="df9ee5aab842a0fa43058bc067afd30d", SSECustomerAlgorithm="AES256")
```

## ✅ Checks

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -15,9 +15,9 @@ from typing import Optional
 
 import boto3  # type: ignore
 
-from botocore import UNSIGNED
+from botocore import UNSIGNED  # type: ignore
 from botocore.client import ClientError  # type: ignore
-from botocore.client import Config
+from botocore.client import Config  # type: ignore
 from botocore.exceptions import EndpointConnectionError  # type: ignore
 from kazoo.client import KazooClient
 from kazoo.protocol.states import ZnodeStat
@@ -137,11 +137,11 @@ def _load_from_s3(data: bytes) -> bytes:
         )
         raise LoaderException from e
 
-    # Client needs to be anonymous/unsigned or boto3 will try to read the local credentials 
-    # on the service pods. And - due to an AWS quirk - any request that comes in signed with 
-    # credentails will profile for permissions for the resource being requested EVEN if the 
-    # resource is public. In other words, this means that a given service could not access 
-    # public resources belonging to another cluster/AWS account unless the request credentials 
+    # Client needs to be anonymous/unsigned or boto3 will try to read the local credentials
+    # on the service pods. And - due to an AWS quirk - any request that comes in signed with
+    # credentails will profile for permissions for the resource being requested EVEN if the
+    # resource is public. In other words, this means that a given service could not access
+    # public resources belonging to another cluster/AWS account unless the request credentials
     # were unsigned.
     s3_client = boto3.client(
         "s3",
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -137,17 +137,23 @@ def _load_from_s3(data: bytes) -> bytes:
         )
         raise LoaderException from e
 
-    # Client needs to be anonymous/unsigned or boto3 will try to read the local credentials
-    # on the service pods. And - due to an AWS quirk - any request that comes in signed with
-    # credentails will profile for permissions for the resource being requested EVEN if the
-    # resource is public. In other words, this means that a given service could not access
-    # public resources belonging to another cluster/AWS account unless the request credentials
-    # were unsigned.
-    s3_client = boto3.client(
-        "s3",
-        config=Config(signature_version=UNSIGNED),
-        region_name=region_name,
-    )
+    if loader_config.get("anon") == True:
+        # Client needs to be anonymous/unsigned or boto3 will try to read the local credentials
+        # on the service pods. And - due to an AWS quirk - any request that comes in signed with
+        # credentails will profile for permissions for the resource being requested EVEN if the
+        # resource is public. In other words, this means that a given service could not access
+        # public resources belonging to another cluster/AWS account unless the request credentials
+        # were unsigned.
+        s3_client = boto3.client(
+            "s3",
+            config=Config(signature_version=UNSIGNED),
+            region_name=region_name,
+        )
+    else:
+        s3_client = boto3.client(
+            "s3",
+            region_name=region_name,
+        )
 
     try:
         s3_object = s3_client.get_object(**s3_kwargs)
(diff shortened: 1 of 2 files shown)


baseplate/sidecars/live_data_watcher.py
```diff
@@ -135,8 +137,15 @@ def _load_from_s3(data: bytes) -> bytes:
...
 
+    # Client needs to be anonymous/unsigned or boto3 will try to read the local credentials
+    # on the service pods. And - due to an AWS quirk - any request that comes in signed with
+    # credentails will profile for permissions for the resource being requested EVEN if the
+    # resource is public. In other words, this means that a given service could not access
+    # public resources belonging to another cluster/AWS account unless the request credentials
+    # were unsigned.
     s3_client = boto3.client(
         "s3",
+        config=Config(signature_version=UNSIGNED),
```
SiddharthManoj, line 148: Note: this makes every s3 client use unsigned requests. Right now the only customer is Experiments, but in the future it could be other services wanting to make use of this pathway.

Do we want to have an explicit check here to see if the use-case is experiments? And if and only if so THEN we make it unsigned?

The con here is that we'd be strict checking some field/name to be able to tell if it's Experiments which goes against the goal of making the sidecar logic more broad...and branching logic is messy in general BUT if there are security concerns at play here, we should definitely do it.

SiddharthManoj: Another alternative I just thought of would be to include another special key-value in the ZK file (e.g. `"anon":true`)) which would let the live-data-fetcher know to instantiate the s3_client this way.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -135,10 +137,23 @@ def _load_from_s3(data: bytes) -> bytes:
         )
         raise LoaderException from e
 
-    s3_client = boto3.client(
-        "s3",
-        region_name=region_name,
-    )
+    if loader_config.get("anon") == True:
```
SiddharthManoj: **Note:** this requires an update to `experiment-config` to start sending over this new `anon` key. I did it this way since I didn't want to assume all services that use this codepath would want to be anonymizing the s3 client. The con here is that we're adding yet another key to check for from the ZK file and adds complexity/branching to this logic.

If we think a) security is not a concern here or b) every service using this pathway would want an anonymized client anyways then I'm fine with always instantiating it as anonymous.

The equivalent update in `experiment-config`: https://github.snooguts.net/reddit/reddit-service-experiment-config/pull/1672

nsheaps (changes requested):
Asked in chat but to keep it in the PR:

~~Is the experiments json public for a reason?~~

~~Why would we be working towards unsigning it rather than authenticating the request (despite its public nature) for audit tracking?~~

Answered the above in a DM, just the test question

tests/unit/sidecars/live_data_watcher_loader_tests.py
```diff
@@ -79,18 +79,18 @@ def test_load_from_s3_missing_config(start_mock_s3, missing_config):
 
 def test_successful_load_from_s3(s3_stub):
     contents = b"my-test-contents"
-    bucket = "my-test-bucket"
-    key = "my-test-key"
+    bucket_name = "my-test-bucket"
     file_key = "my-test-key"
     sse_key = "key"
+    anon = True
```
nsheaps: is there a test for anon=False now?

wuyiqicc (approved):

nsheaps (changes requested):
not sure why I was re-requested, still the outstanding request for the test with `anon = False`

tests/unit/sidecars/live_data_watcher_loader_tests.py
SiddharthManoj: Done

tests/unit/sidecars/live_data_watcher_loader_tests.py
```diff
@@ -77,10 +77,16 @@ def test_load_from_s3_missing_config(start_mock_s3, missing_config):
         assert _load_from_s3(json.dumps(data).encode("utf-8"))
 
 
-def test_successful_load_from_s3(s3_stub):
+@pytest.mark.parametrize(
+    "anon",
+    (
+        True,
+        False,
```
nsheaps, line 84: I guess because this is a unit test, this really changes how it communicates with s3 and therefore the functionality is not any more testable beyond ensuring that you go down that code path.

🔕 would like to see some other mocked s3 responses or checking that swapping this flag changes the config of how the request is made

nsheaps (approved):

tests/unit/sidecars/live_data_watcher_loader_tests.py
SiddharthManoj: Yeah I was thinking about how to go about this, but best I could do is just to make test that both anon=True and anon=False get us down the codepath and return content. 

Testing the actual functionality of an s3 API seems beyond unit testing scope here...overall i agree with your sentiment though

PR #737, opened 2022-09-01
BASE-171 add separate config option to enable Prometheus metrics

0over:
## 💸 TL;DR
Add separate config option to enable Prometheus metrics

[Jira](https://reddit.atlassian.net/browse/BASE-171)

## ✅ Checks
- [X] Adheres to code style for repo
- [X] Contributor License Agreement (CLA) completed if not a Reddit employee

### DIFF
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -43,17 +43,17 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     return str(200 <= httpStatusCode < 400).lower()
 
 
-def are_metrics_enabled(raw_config: Dict[str, str]) -> bool:
-    if "metrics.tagging" in raw_config or "metrics.namespace" in raw_config:
-        return True
-
+def is_metrics_enabled(raw_config: Dict[str, str]) -> bool:
     cfg = config.parse_config(
         raw_config,
         {
             "metrics": {
-                "enabled": config.Optional(config.Boolean, default=False),
+                "enabled": config.Optional(config.Boolean),
             }
         },
     )
 
-    return cfg.metrics.enabled
+    if cfg.metrics.enabled is not None:
+        return cfg.metrics.enabled
+
+    return "metrics.tagging" in raw_config or "metrics.namespace" in raw_config
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -56,4 +56,4 @@ def is_metrics_enabled(raw_config: Dict[str, str]) -> bool:
     if cfg.metrics.enabled is not None:
         return cfg.metrics.enabled
 
-    return "metrics.tagging" in raw_config or "metrics.namespace" in raw_config
+    return True
--- a/baseplate/lib/prometheus_metrics.py
+++ b/baseplate/lib/prometheus_metrics.py
@@ -1,3 +1,8 @@
+from typing import Dict
+
+from baseplate.lib import config
+
+
 # default_latency_buckets creates the default bucket values for time based histogram metrics.
 # we want this to match the baseplate.go default_buckets
 # bp.go v0 ref: https://github.com/reddit/baseplate.go/blob/master/prometheusbp/metrics.go.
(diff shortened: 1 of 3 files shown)


baseplate/lib/prometheus_metrics.py
```diff
@@ -36,3 +41,19 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
     The HTTP success label is "true" if the status code is 2xx or 3xx, "false" otherwise.
     """
     return str(200 <= httpStatusCode < 400).lower()
+
+
+def are_metrics_enabled(raw_config: Dict[str, str]) -> bool:
```
nsheaps: generally, despite the bad grammar, the pattern is to prefix it with `is` instead of `are`

baseplate/lib/prometheus_metrics.py
```diff
@@ -36,3 +41,19 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
...
+    if "metrics.tagging" in raw_config or "metrics.namespace" in raw_config:
```
nsheaps: I think this is a bit confusing, since you could say `metrics.namespace = "harold"` but `metrics.enabled = False`

So that we work towards getting rid of the older settings names, how about the logic as follows (default handled by `or False`). This way you can still explicitly disable it, and the new setting takes priority without affecting those that are currently using the legacy way:

```python
cfg = config.parse_config(raw_config, {})
if hasattr(cfg.metrics, 'enabled'):
  return cfg.metrics.enabled
return "metrics.tagging" in raw_config or "metrics.namespace" in raw_config or False
```

nsheaps (changes requested):
Some minor feedback on the logic for `is_metrics_enabled`

maeivysea (approved):

maeivysea (approved):

KTAtkinson (changes requested):
+1 to Nate's comments.

KTAtkinson (approved):

nsheaps (approved):
No issues with the code, but this should be updated in the baseplate docs as well, which I don't see here, probably somewhere in [here](https://baseplate.readthedocs.io/en/stable/_sources/api/observability-framework.rst.txt)

I don't think it's a blocker for the PR, but should probably be a fast follow.

baseplate/lib/prometheus_metrics.py
```diff
@@ -36,3 +41,19 @@ def getHTTPSuccessLabel(httpStatusCode: int) -> str:
...
+            "metrics": {
+                "enabled": config.Optional(config.Boolean),
+            }
+        },
+    )
+
+    if cfg.metrics.enabled is not None:
+        return cfg.metrics.enabled
+
+    return "metrics.tagging" in raw_config or "metrics.namespace" in raw_config
```
KTAtkinson: If not specified by the client, metrics should be enabled.

KTAtkinson (changes requested):

baseplate/lib/prometheus_metrics.py
0over: In this case we can skip `"metrics.tagging" in raw_config or "metrics.namespace" in raw_config` check and just `return True`, right?

KTAtkinson: Yea, though I think we will merge this the way it is and update this later. We need to get this change out to clients.

KTAtkinson: Actually we want to change this here to make the user upgrade story easier.
```suggestion
    return True
```

KTAtkinson (approved):

docs/cli/serve.rst
```diff
@@ -196,12 +196,16 @@ own copy of the exporter, all listening on the same port. Any given worker can
...
+To enable the exporter, install the ``prometheus-client`` package from PyPI:
 
 .. code-block:: console
 
    $ pip install prometheus-client
 
+and add the following line to the application configuration::
+
+   metrics.enabled = true
+
```
nsheaps:
```suggestion
If for some reason your service doesn't need Prometheus metrics exported, they can be disabled (they're on by default) with:

   metrics.enabled = false
```

PR #742, opened 2022-09-19
exposing prometheus metrics from event-publisher sidecar

ghirsch-reddit:
Closes # 

## 💸 TL;DR

We want to expose Prometheus metrics from our sidecars. This PR sets that up for the event_publisher sidecar, and adds a generic method that can be used by other sidecar types.

## 📜 Details
[Jira](https://reddit.atlassian.net/browse/BASE-122)

I ran this locally, commenting out the bits that actually publish to a real queue. I'm not sure if there's a clean way to do a "realistic" test, but would love to know if there is.

I added the monkey patching based on what I see here, though I'm not confident this is the right way to do this https://github.com/reddit/baseplate.py/blob/develop/bin/baseplate-serve

## ✅ Checks

~~- [ ]Contributor License Agreement (CLA) completed if not a Reddit employee~~

### DIFF
--- a/baseplate/server/prometheus.py
+++ b/baseplate/server/prometheus.py
@@ -78,10 +78,10 @@ def start_prometheus_exporter(address: EndpointConfiguration = PROMETHEUS_EXPORT
 
 
 def start_prometheus_exporter_for_sidecar() -> None:
-    port = os.environ.get("BASEPLATE_SIDECAR_ADMIN_PORT")
+    port = os.environ.get("BASEPLATE_SIDECAR_METRICS_PORT")
     if port is None:
         logger.error(
-            "BASEPLATE_SIDECAR_ADMIN_PORT must be set for sidecar to expose Prometheus metrics."
+            "BASEPLATE_SIDECAR_METRICS_PORT must be set for sidecar to expose Prometheus metrics."
         )
     else:
         endpoint = Endpoint("0.0.0.0:" + port)
--- a/baseplate/server/prometheus.py
+++ b/baseplate/server/prometheus.py
@@ -26,6 +26,7 @@ from prometheus_client import generate_latest
 from prometheus_client import multiprocess
 
 from baseplate.lib.config import Endpoint
+from baseplate.lib.config import EndpointConfiguration
 from baseplate.server.net import bind_socket
 
 
@@ -52,7 +53,7 @@ def export_metrics(environ: "WSGIEnvironment", start_response: "StartResponse") 
     return [data]
 
 
-def start_prometheus_exporter() -> None:
+def start_prometheus_exporter(address: EndpointConfiguration = PROMETHEUS_EXPORTER_ADDRESS) -> None:
     if "PROMETHEUS_MULTIPROC_DIR" not in os.environ:
         logger.error(
             "prometheus-client is installed but PROMETHEUS_MULTIPROC_DIR is not set to a writeable directory."
@@ -61,7 +62,7 @@ def start_prometheus_exporter() -> None:
 
     atexit.register(multiprocess.mark_process_dead, os.getpid())
 
-    server_socket = bind_socket(PROMETHEUS_EXPORTER_ADDRESS)
+    server_socket = bind_socket(address)
     server = WSGIServer(
         server_socket,
         application=export_metrics,
@@ -70,7 +71,18 @@ def start_prometheus_exporter() -> None:
     )
     logger.info(
         "Prometheus metrics exported on server listening on %s%s",
-        PROMETHEUS_EXPORTER_ADDRESS,
+        address,
         METRICS_ENDPOINT,
     )
     server.start()
+
+
+def start_prometheus_exporter_for_sidecar() -> None:
+    port = os.environ.get("BASEPLATE_SIDECAR_ADMIN_PORT")
+    if port is None:
+        logger.error(
+            "BASEPLATE_SIDECAR_ADMIN_PORT must be set for sidecar to expose Prometheus metrics."
+        )
+    else:
+        endpoint = Endpoint("0.0.0.0:" + port)
+        start_prometheus_exporter(endpoint)
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -23,7 +23,7 @@ from typing import Any
 from typing import List
 from typing import Optional
 
-from gevent import sleep
+import gevent
 from prometheus_client import Counter, Histogram
 import requests
 
@@ -240,7 +240,8 @@ def publish_events() -> None:
     publisher = BatchPublisher(metrics_client, cfg)
 
     while True:
-        sleep(0)
+        # allow other routines to execute (specifically handling requests to /metrics)
+        gevent.sleep(0)
         message: Optional[bytes]
 
         try:
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -1,3 +1,16 @@
+# pylint: disable=wrong-import-position,wrong-import-order
+from gevent.monkey import patch_all
+
+from baseplate.server.monkey import patch_stdlib_queues
+
+# In order to allow Prometheus to scrape metrics, we need to concurrently
+# handle requests to '/metrics' along with the sidecar's execution.
+# Monkey patching is used to replace the stdlib sequential versions of functions
+# with concurrent versions. It must happen as soon as possible, before the
+# sequential versions are imported.
+patch_all()
+patch_stdlib_queues()
+
 import argparse
 import configparser
 import email.utils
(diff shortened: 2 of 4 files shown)


requirements-transitive.txt
```diff
@@ -17,14 +17,14 @@ certifi==2020.12.5
 cffi==1.14.5
 chardet==4.0.0
 click==7.1.2
-confluent-kafka==1.6.0
+confluent-kafka==1.9.2
```
KTAtkinson: Are there any notable changes here? This is a pretty bug version jump.

baseplate/sidecars/event_publisher.py
```diff
@@ -1,3 +1,13 @@
+# this is here because we want the gevent monkeypatching to happen before
+# any pieces of baseplate are imported at all, as would happen if we were
+# to have this code be in the baseplate package.
+from gevent.monkey import patch_all
+
+from baseplate.server.monkey import patch_stdlib_queues
+
+patch_all()
+patch_stdlib_queues()
```
KTAtkinson, line 12: Why is monkeypatching required. What issue does Baseplate have if no monkey patching happens.

baseplate/sidecars/event_publisher.py
```diff
@@ -114,8 +132,8 @@ def _sign_payload(self, payload: bytes) -> str:
         return f"key={self.key_name}, mac={digest}"
 
     def publish(self, payload: SerializedBatch) -> None:
-        if not payload.item_count:
-            return
+        # if not payload.item_count:
+        #     return
```
KTAtkinson: Should this be  removed?

baseplate/server/prometheus.py
```diff
@@ -70,7 +70,13 @@ def start_prometheus_exporter() -> None:
...
     logger.info(
         "Prometheus metrics exported on server listening on %s%s",
-        PROMETHEUS_EXPORTER_ADDRESS,
+        address,
         METRICS_ENDPOINT,
     )
     server.start()
+
+
+def start_prometheus_exporter_sidecar() -> None:
```
KTAtkinson:
```suggestion
def start_prometheus_exporter_for_sidecar() -> None:
```

KTAtkinson (changes requested):

requirements-transitive.txt
ghirsch-reddit: Nothing seems particularly notable https://github.com/confluentinc/confluent-kafka-python/releases

I just did a version bump because I was unable to build locally. I see this in the 1.9.2 release notes `Wheel for macOS M1/arm64`, which would explain that.

baseplate/sidecars/event_publisher.py
ghirsch-reddit: I'm not convinced it is _required_, and would believe that there is a better way to do it. My understanding is that `gevent` allows concurrent execution. The monkey patching seems to make a subset of the standard library (which would normally be blocking) work concurrently https://www.gevent.org/api/gevent.monkey.html

Without this monkey patching, the server would be set up properly but all requests to it would hang indefinitely because other execution was happening instead (and then suddenly when the process was interrupted they would complete). After the monkey patching, the requests were handled concurrently as they came in.

baseplate/sidecars/event_publisher.py
ghirsch-reddit: oops, thanks for catching

baseplate/sidecars/event_publisher.py
nsheaps: I'm actually surprised that sidecars are launched by loading this module directly. gevent loops work similar to javascript, so what you're describing is that something is in the event loop is adding more to where it's executing than the end (likely the [while loop here](https://github.com/reddit/baseplate.py/blob/BASE-122/baseplate/sidecars/event_publisher.py#L240)), which gevent solves by patching sockets to be in the middle (which lets it do other things while waiting for a response to socket events...I think?).

This is a bit too magic for me and I think the underlying issue is the while loop as it stands. I'd rather not rely on sockets to force this to happen. This can be resolved by `gevent.sleep(0)` within the loop to ensure gevent gets a chance to run other stuff in the event loop.

setup.cfg
```diff
@@ -30,6 +30,8 @@ exclude_lines =
 [flake8]
 max-line-length = 100
 ignore = W503, E203, E501, D100, D101, D102, D103, D104, D105, D106, D107
+per-file-ignores =
```
nsheaps, line 33: can these be ignored for the specific lines you're targeting rather than the whole file please?

nsheaps (changes requested):
Would like to see a gevent.sleep(0) in the while loop and move flake8 ignores to the specific lines you're targeting.

baseplate/sidecars/event_publisher.py
ghirsch-reddit: > I'm actually surprised that sidecars are launched by loading this module directly

That's fair, it could definitely make sense to make a separate entrypoint. But yeah, that's how it currently works https://github.snooguts.net/reddit/docker-baseplate-sidecars/blob/master/entrypoint.sh

> gevent.sleep(0)

Interesting, this makes sense to me though it does not work as-is. Still working on understanding how it differs from what I expected

ghirsch-reddit: To clarify Nate's comment a bit (since I didn't fully understand it), the point was that _only_ monkey patching just works because there happens to be things with socket events inside of that while loop. Rather than relying on that, better to be explicit and on top of monkey patching, also add the sleep.

setup.cfg
ghirsch-reddit: In case it was not clear, that would mean adding an ignore comment to every one of these lines (and same for the other sidecars) https://github.com/reddit/baseplate.py/blob/34cd60deb263b2da5167eab29d527b60f5ff1870/baseplate/sidecars/event_publisher.py#L12-L42

Plus any time we import things in the future in these files, it will also need to be added to them.

So at that point it seems better to me to just ignore the rules for the file. If you'd like I could mention in the comment at the top that the two linting rules are ignored and that setting can be found in setup.cfg

ghirsch-reddit: And unfortunately it sounds like there's no way in-file to just mark a block to be ignored https://stackoverflow.com/questions/64428794/flake8-disable-linter-only-for-a-block-of-code

baseplate/sidecars/event_publisher.py
```diff
@@ -130,16 +151,18 @@ def publish(self, payload: SerializedBatch) -> None:
         for _ in RetryPolicy.new(budget=MAX_RETRY_TIME, backoff=RETRY_BACKOFF):
             try:
                 with self.metrics.timer("post"):
```
nsheaps, line 153: Since we're no longer publishing statsd metrics, this can probably be removed along with other references to self.metrics.

EDIT: let's make this a second pass for all of baseplate

baseplate/sidecars/event_publisher.py
```diff
@@ -215,6 +240,7 @@ def publish_events() -> None:
     publisher = BatchPublisher(metrics_client, cfg)
 
     while True:
+        sleep(0)
```
nsheaps: gevent does patch this, but I'd feel more comfortable if it was explicitly the gevent sleep and if this was at the end of the loop

requirements-transitive.txt
```diff
@@ -17,14 +17,14 @@ certifi==2020.12.5
 cffi==1.14.5
```
nsheaps: Is there a need for these upgrades to happen as part of this change besides the m1 stuff (which shouldn't be needed when using docker compose to do the testing)? generally I'd opt to separate out upgrades from functional changes in case there's a hidden breaking change to make it easier to roll back.

setup.cfg
nsheaps: Ah, not my favorite but not worth addressing differently. Thanks for looking into it.

nsheaps (approved):

baseplate/sidecars/event_publisher.py
ghirsch-reddit: It is the gevent sleep, but I can switch the import from `from gevent import sleep` to just `import gevent` so it's more obvious inline. And yep, I can move to the end.

requirements-transitive.txt
ghirsch-reddit: True, I can move to a separate PR.

baseplate/sidecars/event_publisher.py
ghirsch-reddit: Oh actually, I noticed that since the while loop contains a `continue`, if we put the `sleep` at the end of the loop then it is harder to predict when it will actually allow the other routine to happen. So I'm going to put it back at the top of the loop

baseplate/sidecars/event_publisher.py
KTAtkinson: Sounds good, it would be nice if this were included in your comment.

KTAtkinson (approved):

PR #741, opened 2022-09-14
fix(RedisClient): Add 'redis_client_name` property

KTAtkinson:
Closes # 

## 💸 TL;DR

This change adds the `redis_client_name` property to the RedisClient. This allows clients that use a Redis proxy to set the `redis_client_name` label on Prometheus metrics.

## 📜 Details

Currently clients that use the `client_name` property to set the Redis
client with a Envoy Redis proxy are not able to connect to Redis. At
it's root this is an issue an issue with the redis-py library (see
https://github.com/redis/redis-py/issues/2384).

Currently Baseplate's wrapper around the Redis client uses the
`client_name` to set the `redis_client_name` label on Prometheus metrics.
Due to the issue mentioned above clients who use a Redis Proxy are not
about to set the `redis_client_name` Prometheus label. The additional
`redis_client_name` property is now used to set the Prometheus label.
For backwards compatibility, if `client_name` is set, but
`redis_client_name` is not, the `redis_client_name` defaults to
`client_name`.

Equivalent change to RedisCluster shortly.

## ✅ Checks

- [X] CI tests (if present) are passing
- [X] Adheres to code style for repo
- [ ] Contributor License Agreement (CLA) completed if not a Reddit employee

PR #743, opened 2022-09-19
fix(RedisClusterClient): Add `redis_client_name` attribute

KTAtkinson:
This updates RedisClusterClient.

Currently clients that use the `client_name` property to set the Redis client with a Envoy Redis proxy are not able to connect to Redis. At it's root this is an issue an issue with the redis-py library (see redis/redis-py#2384).

Currently Baseplate's wrapper around the Redis client uses the `client_name` to set the `redis_client_name` label on Prometheus metrics. Due to the issue mentioned above clients who use a Redis Proxy are not about to set the `redis_client_name` Prometheus label. The additional `redis_client_name` property is now used to set the Prometheus label. For backwards compatibility, if `client_name` is set, but `redis_client_name` is not, the `redis_client_name` defaults to `client_name`.

This makes the same change to #741 does for the RedisCluster client.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/clients/redis_cluster.py
+++ b/baseplate/clients/redis_cluster.py
@@ -328,12 +328,21 @@ def cluster_pool_from_config(
 class ClusterRedisClient(config.Parser):
     """Configure a clustered Redis client.
 
+    :param redis_client_name: The name of this Redis client. Prefer to use `client_name`
+    keyword argument to the `pool_from_config` function. Use this if your using a Redis
+    host or proxy that doesn't support the `CLIENT SETNAME` function.
+
     This is meant to be used with
     :py:meth:`baseplate.Baseplate.configure_context`.
     See :py:func:`cluster_pool_from_config` for available configuration settings.
     """
 
     def __init__(self, redis_client_name: str = "", **kwargs: Any):
+        # This is for backwards compatibility. Originally we asked clients to
+        # set the `client_name` attribute to get the `redis_client_name`
+        # tag to appear on Prometheus metrics. Unfortunately this broke clients
+        # that use a proxy to connect to Redis.
+        # See: https://github.com/redis/redis-py/issues/2384/
         client_name = redis_client_name
         if client_name == "":
             client_name = kwargs.get("client_name", "")
--- a/baseplate/clients/redis_cluster.py
+++ b/baseplate/clients/redis_cluster.py
@@ -333,12 +333,19 @@ class ClusterRedisClient(config.Parser):
     See :py:func:`cluster_pool_from_config` for available configuration settings.
     """
 
-    def __init__(self, **kwargs: Any):
+    def __init__(self, redis_client_name: str = "", **kwargs: Any):
+        client_name = redis_client_name
+        if client_name == "":
+            client_name = kwargs.get("client_name", "")
+
         self.kwargs = kwargs
+        self.redis_client_name = client_name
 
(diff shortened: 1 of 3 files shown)


nsheaps (approved):
Thanks for taking care of this!

KTAtkinson:
I realized that I hadn't passed  the `redis_client_name` keyword argument to Redis pipeline

baseplate/clients/redis_cluster.py
```diff
@@ -333,12 +333,19 @@ class ClusterRedisClient(config.Parser):
     See :py:func:`cluster_pool_from_config` for available configuration settings.
```
nsheaps, line 337: 🔕 on the other one you added some pydoc describing the params and some comments below, did you want to duplicate them?

tests/integration/redis_cluster_tests.py
```diff
@@ -156,3 +170,91 @@ def test_pipeline(self):
...
+            "client_name",
+        ]:
+            with self.subTest():
+                self.setup_baseplate_redis(
+                    redis_cluster_kwargs={
+                        client_name_kwarg_name: client_name,
+                    },
+                )
+                # Clear preometheus metrics
+                ACTIVE_REQUESTS.clear()
```
nsheaps, line 187: This should probably be in setup so each test starts from a clean state

tests/integration/redis_cluster_tests.py
```diff
@@ -156,3 +170,91 @@ def test_pipeline(self):
...
         self.assertIsNone(span_observer.on_finish_exc_info)
+
+    def test_metrics(self):
+        client_name = "redis_test"
+        for client_name_kwarg_name in [
+            "redis_client_name",
+            "client_name",
+        ]:
+            with self.subTest():
+                self.setup_baseplate_redis(
```
nsheaps, line 181: 🔕 documenting our convo: Redis should clear state between tests too

baseplate/clients/redis_cluster.py
```diff
@@ -501,9 +527,7 @@ def execute(self, **kwargs: Any) -> Any:
             start_time = perf_counter()
             labels = {
                 f"{PROM_LABELS_PREFIX}_command": "pipeline",
-                f"{PROM_LABELS_PREFIX}_client_name": self.connection_pool.connection_kwargs.get(
-                    "client_name", ""
-                ),
+                f"{PROM_LABELS_PREFIX}_client_name": self.redis_client_name,
```
nsheaps, line 530: Thanks! As far as I know, the connection pool does have a client name as well but I don't think anyone ever set it. This would technically also be a backwards incompatible change ([example](https://github.com/redis/redis-py/blob/master/tests/test_connection_pool.py#L82) from redis-py tests)

nsheaps (approved):

maeivysea (approved):

baseplate/clients/redis_cluster.py
KTAtkinson: We are trying to avoid setting `client_name` on the connection pool as this causes the issue described in the PR description. That is why we always default the value of  `redis_client_name` to the value of the same value as `connection.client_name`.

tests/integration/redis_cluster_tests.py
KTAtkinson: I moved this to `tearDown` in #744. I will merge this before and then merge and remove these lines in that PR.

PR #745, opened 2022-09-21
Reduce cardinality of http_server_active_requests

SuperQ:
Closes # 

## 💸 TL;DR
Set the multiprocess mode on `http_server_active_requests` to "livesum" to reduce the cardinality to one metric per server.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -50,6 +50,7 @@ REQUEST_ACTIVE = Gauge(
     "cassandra_client_active_requests",
     "Current number of active cassandra queries",
     CassandraPrometheusLabels._fields,
+    multiprocess_mode="livesum",
 )
 REQUEST_TOTAL = Counter(
     "cassandra_client_requests_total",
--- a/baseplate/clients/memcache/__init__.py
+++ b/baseplate/clients/memcache/__init__.py
@@ -195,6 +195,7 @@ ACTIVE_REQUESTS = Gauge(
     f"{PROM_NAMESPACE}_client_active_requests",
     "Number of active requests",
     LABELS_COMMON,
+    multiprocess_mode="livesum",
 )
 
 
--- a/baseplate/clients/redis.py
+++ b/baseplate/clients/redis.py
@@ -50,6 +50,7 @@ ACTIVE_REQUESTS = Gauge(
     f"{PROM_PREFIX}_active_requests",
     "Number of active requests for a given client",
     PROM_SHARED_LABELS,
+    multiprocess_mode="livesum",
 )
 
 PROM_POOL_PREFIX = f"{PROM_PREFIX}_pool"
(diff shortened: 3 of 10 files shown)


KTAtkinson (approved):

KTAtkinson (changes requested):
This change explicitly violates the baseplate spec. I agree that this changes makes sense but the spec should be updated ahead of this change.

SuperQ:
Yup, that's fine, I'll make a proposal to update the spec.

nsheaps (changes requested):
Does this need to be done anywhere else while we're at it?

Also is there a ticket tracking this? I just closed BASE-103 which was a similar thing for statsd

SuperQ:
Probably all of the `active_requests` metrics. I don't think they're useful on a per-pid basis.

garretthoffman:
Hey can we actually get this implemented on all `Gauge` metrics and try to expedite this into a new release? For services that are still deployed on legacy EC2 / puppet infrastructure not having a `live*` multiprocessing mode is accumulating `*.db` files and thousands of metrics due to autoreload on these servers that create new PIDs. This is causing scrape times to balloon up to 1.5-2 seconds and is causing event loop contention with the main application server.

SuperQ:
@garretthoffman I'm not sure this will actually help the situation you're seeing. Even with `livesum`, we still need to read all the `*.db` files in order to produce the aggregation.

I haven't looked closely at the Python implementation, but when we did similar things in Ruby, we avoided using the real system PID value, but a worker number. This way when you roll worker processes, they recycle the same identifier. This avoided the infinite generating of new `*.db` files every time a worker restarted.

I thought we had went with this implementation in baseplate.py.

garretthoffman:
@SuperQ Understood about the `*.db` and well get in the other fix you committed. Just for some more information about 9600 out of 10000 metrics were from these Guage PIDs on some samples I checked so anything to reduce the cardinality on these would be great. Worker number seems extremely reasonable.

SuperQ:
I have the baseplate spec change approved. Just waiting for it to be merged.

nsheaps (approved):

garretthoffman:
Are we going to make this change for the thrift server  and client Gauges as well? Or do we want to handle that in a separate PR?

SuperQ:
I can follow up with more, there are a lot of other Gauges that I need to go through.

KTAtkinson (approved):

SuperQ:
@garretthoffman I've updated the PR to add livesome to most of the Gauge metrics.

garretthoffman:
> @garretthoffman I've updated the PR to add livesome to most of the Gauge metrics.

Appreciate this. The two offending metrics for minsky though are `thrift_server_active_requests` and `cassandra_client_active_requests` which are not updated here. Is there a reason that these can't be updated?

SuperQ:
Done, added those as well.

PR #750, opened 2022-10-06
upgrade flake8 and pylint

ghirsch-reddit:
## 💸 TL;DR
Builds were failing (e.g. [link](https://github.com/reddit/baseplate.py/actions/runs/3200563486/jobs/5227611744)) because importlib_metadata (a dependency of flake8) [removed a feature](https://importlib-metadata.readthedocs.io/en/latest/history.html). Pinning it to `importlib_metadata < 5` works, but instead let's just update flake8.

That had some conflicts, so updating pylint as well (and removing some transitive dependencies that don't seem important to pin since they are already effectively pinned by the existing imports).
## ✅ Checks

### DIFF
--- a/requirements-transitive.txt
+++ b/requirements-transitive.txt
@@ -28,6 +28,7 @@ greenlet==1.0.0
 hupper==1.10.2
 idna==2.10
 imagesize==1.2.0
+importlib-metadata==4.2.0
 iniconfig==1.1.1
 isort==5.7.0
 Jinja2==2.11.3
@@ -35,6 +36,7 @@ kazoo==2.8.0
 kombu==4.6.11
 lazy-object-proxy==1.4.3
 MarkupSafe==1.1.1
+mccabe==0.7.0
 moto==3.1.4
 mypy-extensions==0.4.3
 ndg-httpsclient==0.5.1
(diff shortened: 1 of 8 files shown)


ghirsch-reddit:
Oops, meant to open this as a draft until I confirmed that the build was succeeding

nsheaps (changes requested):
Not an all out rejection just looking for more info:

1. Any idea why this magically broke out of nowhere?
2. I see you're removing some pylint configs, are they not necessary anymore/rolled into other rules?
3. I also see you're removing some dependencies, were they causing issues or just unused? (transitive dependencies don't get locked in python without the requirements files, since most setup.py suggests a range (with potentially untested versions)

ghirsch-reddit:
1. It's because flake8 [declares a dependency](https://github.com/PyCQA/flake8/blob/3.8.4/setup.cfg#L51) on `importlib-metadata` with no version requirement, and `importlib-metadata` released a v5.0 that removes a feature flake8 had been using. As mentioned in the description, adding a requirement ourselves for `importlib-metadata<5` also works, but seems better to just upgrade
2. The ones I removed have been removed or replaced from flake8. They were all either defaults, or a couple of settings to ignore checks. For the latter, I replaced them if possible. If you have questions about any specific ones, happy to answer them.
3. They did cause issues insofar as the version we pinned to was now too old. I can definitely still pin them to the newer versions rather than getting rid of them entirely, but they all seemed not particularly valuable to do that for. All 3 were just for linting/error checking and required by flake8 anyway, so there's no obvious reason to me why we'd want to pin to old versions.

requirements-transitive.txt
```diff
@@ -28,14 +28,15 @@ greenlet==1.0.0
 hupper==1.10.2
 idna==2.10
 imagesize==1.2.0
+importlib-metadata==4.2.0
```
nsheaps, line 31: discussed offline but this is what stops it from breaking again.

nsheaps (approved):

PR #751, opened 2022-10-07
fix(runtime metrics): Use a null statsd client for runtime metrics

KTAtkinson:
Closes # 

## 💸 TL;DR

Currently runtime metrics aren't being reported to Prometheus when StatsD is disabled. This change initialize a null StatD client to report StatsD metrics to.

## 📜 Details
[Design Doc]()

[Jira]()

I tried to get this working on Harold, but wasn't able to get runtime metrics to appear even on Harold running Baseplate v2.5.4. We will need to test this to make sure it works.

## ✅ Checks

### DIFF
--- a/baseplate/server/runtime_monitor.py
+++ b/baseplate/server/runtime_monitor.py
@@ -218,6 +218,14 @@ def _report_runtime_metrics_periodically(
 
 def start(server_config: Dict[str, str], application: Any, pool: Pool) -> None:
     baseplate: Baseplate = getattr(application, "baseplate", None)
+    # As of October 1, 2022 Reddit uses Prometheus to track metrics, not Statsd
+    # this checks to see if Prometheus metrics are enabled and uses this to determine
+    # if runtime metrics should be reported. Prometheus metrics default to "on".
+    # As of October 12, 2022 some teams still have permission to use Statsd for metrics
+    # reporting so we still need runtime metrics to be sent to Statsd sometimes, therefore
+    # we can't remove Statsd from runtime reporter functions all together yet, instead we stub 
+    # out the metrics Statsd client in the case there is none. There will be no Statsd client
+    # in the case of teams that have fully moved from Statsd to Promtheus.
     if not baseplate or not prometheus_metrics.is_metrics_enabled(server_config):
         logger.info("No metrics client configured. Server metrics will not be sent.")
         return
--- a/baseplate/server/runtime_monitor.py
+++ b/baseplate/server/runtime_monitor.py
@@ -225,7 +225,7 @@ def start(server_config: Dict[str, str], application: Any, pool: Pool) -> None:
     # reporting so we still need runtime metrics to be sent to Statsd sometimes, therefore
     # we can't remove Statsd from runtime reporter functions all together yet, instead we stub 
     # out the metrics Statsd client in the case there is none. There will be no Statsd client
-    # in the case of teams that have fully moved from Statsd to Promtheus.
+    # in the case of teams that have fully moved from Statsd to Prometheus.
     if not baseplate or not prometheus_metrics.is_metrics_enabled(server_config):
         logger.info("No metrics client configured. Server metrics will not be sent.")
         return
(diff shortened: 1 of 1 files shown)


baseplate/server/runtime_monitor.py
```diff
@@ -217,7 +218,7 @@ def _report_runtime_metrics_periodically(
 
 def start(server_config: Dict[str, str], application: Any, pool: Pool) -> None:
     baseplate: Baseplate = getattr(application, "baseplate", None)
-    if not baseplate or not baseplate._metrics_client:
+    if not baseplate or not prometheus_metrics.is_metrics_enabled(server_config):
         logger.info("No metrics client configured. Server metrics will not be sent.")
         return
```
nsheaps, line 231: This check isn't really checking if it _should_ be on, but rather if it was instantiated properly. Prometheus also doesn't have a "metrics client" so the naming of this now seems weird.

Maybe something like:
```suggestion
    if not baseplate or not baseplate._metrics_client:
        logger.info("No statsd metrics configured, statsd metrics will not be sent. If you're expecting them, check configs for 'metrics.tagging' or 'metrics.namespace'")

    if not prometheus_metrics.is_metrics_enabled(server_config):
        logger.info("No metrics frameworks are enabled, skipping reporter setup"
        return
```

baseplate/server/runtime_monitor.py
```diff
@@ -272,10 +273,18 @@ def start(server_config: Dict[str, str], application: Any, pool: Pool) -> None:
         except Exception as exc:
             logger.info("monitoring.gc.refcycle disabled: %s", exc)
 
+    statsd_metrics_client = baseplate._metrics_client
```
nsheaps, line 284: It _probably_ doesn't make a difference, but I feel like it makes more sense for this to be before any of the reporters are instantiated, _just in case_ it tries to report metrics in the `__init__()` functions

baseplate/server/runtime_monitor.py
```diff
@@ -272,10 +273,18 @@ def start(server_config: Dict[str, str], application: Any, pool: Pool) -> None:
...
+        statsd_metrics_client = metrics.Client(
+            transport=metrics.NullTransport(log_if_unconfigured=False),
+            namespace="noop",
+        )
+
     thread = threading.Thread(
         name="Server Monitoring",
         target=_report_runtime_metrics_periodically,
-        args=(application.baseplate._metrics_client, reporters),
+        args=(statsd_metrics_client, reporters),
```
nsheaps, line 294: I really doubt it's high, but this explicitly means we're spending CPU cycles in production on something that we're just going to throw away.

nsheaps (commented):
Your solution is much more concise, but I'm not sure I want to keep a client around where we're literally throwing it away instead of making it actually optional. That said, the scope of yours is much smaller so I think I'd be slightly more comfortable that it'd work in the absence of a test that covers it. What do you think?

nsheaps (changes requested):
RC to keep the system happy but not an all out rejection.

baseplate/server/runtime_monitor.py
KTAtkinson: I did it here in order to address the specific issue we see. We can move it if we see the issue elsewhere. What I was trying to avoid is context.metrics being set to a nil client. It wouldn't be ideal for large amounts of metrics to be flowing through this client. This is a shim for Baseplate internals only, not for applications to use.

baseplate/server/runtime_monitor.py
KTAtkinson: If baseplate is None, which I'm not clear on when this would happen, accessing `baseplate._metrics_client` will fail. I could check for that case lower down.

nsheaps (approved):

baseplate/server/runtime_monitor.py
ghirsch-reddit: You could also do something like
```suggestion
    statsd_metrics_client = None
    if not baseplate or not baseplate._metrics_client:
        logger.info("No metrics client configured. Server metrics will not be sent.")
        // initialize noop client
    else:
        statsd_metrics_client = baseplate._metrics_client
```

ghirsch-reddit (changes requested):
I don't think this change is doing what you think it is. What is the point of passing around no-op controllers as opposed to just getting rid of all this code?

Or if the goal is to actually send metrics, I don't see why making a no-op controller would help with that.

ghirsch-reddit (approved):
I'd still slightly prefer passing around None, but this works if it unblocks things and is quicker.

baseplate/server/runtime_monitor.py
KTAtkinson: This matches to old code, I believe it's using `baseplate._metrics_client` as a gauge to see if metrics are enabled. This is no longer a valid way to determine if metrics are enabled when we use Prometheus.

PR #752, opened 2022-10-10
add Prometheus to remaining sidecar types

ghirsch-reddit:
## 💸 TL;DR
Reproducing https://github.com/reddit/baseplate.py/pull/742. This makes all sidecar types expose Prometheus metrics.

## 📜 Details
JIRA
- https://reddit.atlassian.net/browse/DEVX-753
- https://reddit.atlassian.net/browse/BASE-122

- [x] tested with a service in snoodev - it worked insofar as it ran, though I'm not 100% sure that all of the behavior is unchanged since the service was not using the sidecars in dev before
- [x] a test in snoodev for services with each of the sidecar types
- [ ] deploy a service for real with an override to use the new versions, and closely monitor it

and only then update infrared to use these by default

## ✅ Checks

### DIFF
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -118,13 +118,7 @@ class V2JBatch(V2Batch):
 
 
 class BatchPublisher:
-    def __init__(self, metrics_client: metrics.Client, cfg: Any):
-        bp = Baseplate()
-        bp.configure_context(
-            {
-                "http_client": ExternalRequestsClient("event_collector"),
-            }
-        )
+    def __init__(self, bp: Baseplate, metrics_client: metrics.Client, cfg: Any):
         self.baseplate = bp
         self.metrics = metrics_client
         self.url = f"{cfg.collector.scheme}://{cfg.collector.hostname}/v{cfg.collector.version}"
@@ -231,10 +225,17 @@ def publish_events() -> None:
         max_message_size=MAX_EVENT_SIZE,
     )
 
+    bp = Baseplate()
+    bp.configure_context(
+        {
+            "http_client": ExternalRequestsClient("event_collector"),
+        }
+    )
+
     # pylint: disable=maybe-no-member
     serializer = SERIALIZER_BY_VERSION[cfg.collector.version]()
     batcher = TimeLimitedBatch(serializer, MAX_BATCH_AGE)
-    publisher = BatchPublisher(metrics_client, cfg)
+    publisher = BatchPublisher(bp, metrics_client, cfg)
 
     while True:
         # allow other routines to execute (specifically handling requests to /metrics)
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -1,3 +1,16 @@
+# pylint: disable=wrong-import-position,wrong-import-order
+from gevent.monkey import patch_all
+
+from baseplate.server.monkey import patch_stdlib_queues
+
+# In order to allow Prometheus to scrape metrics, we need to concurrently
+# handle requests to '/metrics' along with the sidecar's execution.
+# Monkey patching is used to replace the stdlib sequential versions of functions
+# with concurrent versions. It must happen as soon as possible, before the
+# sequential versions are imported.
+patch_all()
+patch_stdlib_queues()
+
 import argparse
 import configparser
 import email.utils
@@ -10,9 +23,13 @@ from typing import Any
 from typing import List
 from typing import Optional
 
+import gevent
 import requests
 
-from baseplate import __version__ as baseplate_version
+from prometheus_client import Counter
+
+from baseplate import Baseplate
+from baseplate.clients.requests import ExternalRequestsClient
 from baseplate.lib import config
 from baseplate.lib import metrics
 from baseplate.lib.events import MAX_EVENT_SIZE
@@ -42,6 +59,8 @@ MAX_RETRY_TIME = 5 * 60
 MAX_BATCH_AGE = 1
 # maximum size (in bytes) of a batch of events
 MAX_BATCH_SIZE = 500 * 1024
+
+PUBLISHES_COUNT_TOTAL = Counter("event_publishes_total", "total count of published events")
 
 
 class MaxRetriesError(Exception):
@@ -100,14 +119,17 @@ class V2JBatch(V2Batch):
 
 class BatchPublisher:
     def __init__(self, metrics_client: metrics.Client, cfg: Any):
+        bp = Baseplate()
+        bp.configure_context(
+            {
+                "http_client": ExternalRequestsClient("event_collector"),
+            }
+        )
+        self.baseplate = bp
         self.metrics = metrics_client
         self.url = f"{cfg.collector.scheme}://{cfg.collector.hostname}/v{cfg.collector.version}"
         self.key_name = cfg.key.name
         self.key_secret = cfg.key.secret
-        self.session = requests.Session()
-        self.session.headers[
-            "User-Agent"
-        ] = f"baseplate.py-{self.__class__.__name__}/{baseplate_version}"
 
     def _sign_payload(self, payload: bytes) -> str:
         digest = hmac.new(self.key_secret, payload, hashlib.sha256).hexdigest()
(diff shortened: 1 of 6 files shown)


baseplate/sidecars/event_publisher.py
```diff
@@ -116,7 +116,9 @@ def serialize(self) -> SerializedBatch:
 
 publishCountTotal = Counter("event_publish_total", "total count of published events")
```
nsheaps: 🔕 these should get added to the spec for consistency reasons

baseplate/sidecars/event_publisher.py
```diff
@@ -116,7 +116,9 @@ def serialize(self) -> SerializedBatch:
...
 publishLatency = Histogram("event_publish_latency", "latency for publishing a batch of events")
-publishErrors = Counter("event_publish_errors", "total count of published events", ["error_type"])
+publishErrors = Counter(
```
nsheaps: Sorry for commenting on this now instead of a few weeks ago, but how are these metrics different than those of the client that the sidecar uses? Not against it...however this sidecar doesn't use the requests lib from baseplate, so effectively those calls would be uninstrumented (they should be though), and we should have those for consistent tracking whenever the requests lib is used (not to mention it gives a bit more data than these metrics do)

baseplate/sidecars/event_publisher.py
```diff
@@ -116,7 +116,9 @@ def serialize(self) -> SerializedBatch:
 
 publishCountTotal = Counter("event_publish_total", "total count of published events")
 publishLatency = Histogram("event_publish_latency", "latency for publishing a batch of events")
```
nsheaps: Should this have a label indicating how many events were in the batch? (would increase cardinality but would also give some identifier why one call was [potentially] slower than another)

baseplate/sidecars/trace_publisher.py
```diff
@@ -87,29 +110,34 @@ def publish(self, payload: SerializedBatch) -> None:
...
                 response.raise_for_status()
             except requests.HTTPError as exc:
+                publishErrors.labels(error_type="http").inc()
                 self.metrics.counter("error.http").increment()
                 response = getattr(exc, "response", None)
                 if response is not None:
                     logger.exception("HTTP Request failed. Error: %s", response.text)
                     # If client error, skip retries
                     if response.status_code < 500:
+                        publishErrors.labels(error_type="http.client").inc()
```
nsheaps: I see this matches the statsd error, but what is an "http.client" error? That's not really descriptive enough for us to take action on.

baseplate/sidecars/trace_publisher.py
```diff
@@ -87,29 +110,34 @@ def publish(self, payload: SerializedBatch) -> None:
...
+                        response = self.session.post(
+                            self.endpoint,
+                            data=payload.serialized,
+                            headers=headers,
+                            timeout=self.post_timeout,
+                            stream=False,
+                        )
                 response.raise_for_status()
             except requests.HTTPError as exc:
+                publishErrors.labels(error_type="http").inc()
```
nsheaps: This seems quite strange to me in comparison with the other metrics we use around requests. Why should we track all errors in one bucket, then a subset in another bucket? Why not have one metric that identifies the type of error (then you can sum by error type) like how the requests lib tracks response code?

nsheaps (changes requested):
While I recognize the prom metrics are basically in parity with the statsd metrics, I think we should be aiming for the proper metrics, rather than replacement metrics. Additionally, these need spec changes (even though the same sidecar are used for go and python), and we should ensure any clients used in the sidecars are the baseplate clients so we get that instrumentation for free

baseplate/sidecars/event_publisher.py
ghirsch-reddit: Oh cool, yeah switching to the baseplate request lib makes sense to me 👍

baseplate/sidecars/trace_publisher.py
ghirsch-reddit: Yeah, that's fair. Maybe makes sense to use a status_code

baseplate/sidecars/event_publisher.py
```diff
@@ -116,7 +116,9 @@ def serialize(self) -> SerializedBatch:
...
-publishErrors = Counter("event_publish_errors", "total count of published events", ["error_type"])
+publishErrors = Counter(
+    "event_publish_errors", "total count of errors when publishing events", ["error_type"]
```
fishy: please follow https://prometheus.io/docs/practices/naming/ for naming convention.

all metrics should be suffixed with the unit (whenever applicable), then counters should be suffixed with `_total` after the unit.

so the 3 here should be: `event_publishes_total` (or maybe `event_publish_attempts_total`?), `event_publish_latency_seconds`, and `event_publish_errors_total` respectively. similar to other sidecars.

baseplate/sidecars/event_publisher.py
```diff
@@ -100,14 +104,15 @@ def serialize(self) -> SerializedBatch:
...
+        })
+        self.baseplate = bp
         self.metrics = metrics_client
         self.url = f"{cfg.collector.scheme}://{cfg.collector.hostname}/v{cfg.collector.version}"
         self.key_name = cfg.key.name
         self.key_secret = cfg.key.secret
-        self.session = requests.Session()
-        self.session.headers[
-            "User-Agent"
-        ] = f"baseplate.py-{self.__class__.__name__}/{baseplate_version}"
```
ghirsch-reddit, line 110: Maybe I'm missing something, but it seems like this was getting overwritten below anyway (and same in the other file)

ghirsch-reddit:
~Added the metrics to the spec here https://github.snooguts.net/reddit/baseplate.spec/pull/74~

nsheaps (approved):
Nice job! Do you have some example metrics from testing?

ghirsch-reddit:
Not good ones 😅 My plan now is to ask a team who actually uses these sidecars in a not-very-critical service if I can do better tests using their service

maeivysea (approved):

ghirsch-reddit:
Okey doke, finally tested it:
- events_publisher: https://github.snooguts.net/reddit/reddit-service-http-request-collector/pull/98
- trace_publisher: https://github.snooguts.net/reddit/reddit-service-realtime/pull/342

Example metrics from the trace_publisher test (forgot to record ones from the other test, but it was pretty similar)
```
# HELP zipkin_trace_publishes_total Multiprocess metric
# TYPE zipkin_trace_publishes_total counter
zipkin_trace_publishes_total 36.0
# HELP http_client_requests_total Multiprocess metric
# TYPE http_client_requests_total counter
http_client_requests_total{http_client_name="trace_collector",http_method="post",http_response_code="202",http_success="true"} 3.0
# HELP http_client_latency_seconds Multiprocess metric
# TYPE http_client_latency_seconds histogram
http_client_latency_seconds_sum{http_client_name="trace_collector",http_method="post",http_success="true"} 0.031773647060617805
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.0001"} 0.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.0005"} 0.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.001"} 0.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.0025"} 0.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.005"} 0.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.01"} 2.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.025"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.05"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.1"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.25"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="0.5"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="1.0"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="5.0"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="15.0"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="30.0"} 3.0
http_client_latency_seconds_bucket{http_client_name="trace_collector",http_method="post",http_success="true",le="+Inf"} 3.0
http_client_latency_seconds_count{http_client_name="trace_collector",http_method="post",http_success="true"} 3.0
# HELP http_client_active_requests Multiprocess metric
# TYPE http_client_active_requests gauge
http_client_active_requests{http_client_name="trace_collector",http_method="post",pid="7"} 0.0
```

PR #754, opened 2022-10-14
Fix accessing non-existent Span object

zerozez:
## 💸 TL;DR
PR fixes an issue of accessing non-existent `Span` object in `on_error` event callback.

## 📜 Details
Even if it is appearing that `Span` object is created in `on_before_execute` in might happened that other exceptions/errors can be raised before that, which triggers `on_error` event but not `on_before_execute`. That leads to `AttributeError('NoneType' object has no attribute 'finish')` exception in the traceback. As an example, if a wrong typed value passed and `ValueError` raised on forming a query, that happens.

PR prevents this situation by checking `Span` object, before doing anything with it. 

- [x] Added test case

## ✅ Checks

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -298,9 +298,10 @@ class SQLAlchemyEngineContextFactory(ContextFactory):
 
     def on_error(self, context: ExceptionContext) -> None:
         """Handle the event which happens on exceptions during execution."""
-        exc_info = (type(context.original_exception), context.original_exception, None)
-        context.connection.info["span"].finish(exc_info=exc_info)
-        context.connection.info["span"] = None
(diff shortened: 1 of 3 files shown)


KTAtkinson (changes requested):
Can you write a test that verifies that this code fixes the bug you mentioned in the PR description? If it's not feasible that's fine, just send the review back to me, otherwise it looks good.

zerozez:
Hey @KTAtkinson I added additional integration test as you asked, with a simple logic: no `on_before_execute` runs - no children spawn. Check that `on_error` doesn't fail added in [c0b97f7](https://github.com/reddit/baseplate.py/pull/754/commits/c0b97f797a28598a9ec681761be29696dba8c3ab)

KTAtkinson (approved):
Thanks of the tests!

PR #759, opened 2022-11-09
Populate Cassandra cluster name from cluster metadata

justinmir:
The cluster name label allows us to map the client queries back to the cluster they are being executed on. It can be used to determine client query error rates and latencies broken down by cluster and map these back to specific queries similar to how `cassandra_client_name` is used.

If the metadata field is not populated, this returns an empty string. The metadata field is populated on connect and refreshed at a regular interval, however, the cluster name will not change.

Closes # 

## 💸 TL;DR
Introduces `cassandra_cluster_name` prometheus label for Cassandra client request metrics. 
It can be used to determine client query error rates and latencies broken down by cluster and map these back to specific queries similar to how `cassandra_client_name` is used.

If the cluster does not return the cluster name metadata field, this returns an empty string. However,
the cluster name is populated as part of the Cassandra driver's `connect` call.

## 📜 Details
[Baseplate.spec PR](https://github.snooguts.net/reddit/baseplate.spec/pull/78) 
[Jira](https://reddit.atlassian.net/browse/STRGS-417)

Unit test and integration test introduced.

## ✅ Checks
- [x] Reddit employee

### DIFF
--- a/baseplate/clients/cassandra.py
+++ b/baseplate/clients/cassandra.py
@@ -38,6 +38,7 @@ class CassandraPrometheusLabels(NamedTuple):
     cassandra_client_name: str
     cassandra_keyspace: str
     cassandra_query_name: str
+    cassandra_cluster_name: str
 
 
 REQUEST_TIME = Histogram(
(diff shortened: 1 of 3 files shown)


kylelemons (changes requested):
This is the original implementaiton, not the one that asks the cluster what its own name is, right?

justinmir:
This is the one that "asks" the cluster for its own name. In this case, the python cassandra driver already queries the cluster for a bunch of metadata on connection, including the cluster name, so we just use that queried cluster name - its the same as the cluster name returned by DESCRIBE CLUSTER.

kylelemons (approved):
Cool!

PR #760, opened 2022-11-21
Adds support for `MultiProcessValue` ids based on env variable

davinci26:
## 💸 TL;DR

When baseplate is running as part of Monoceros process restarts create too many files which make prometheus client extremely slow and eventually it timeouts.

## 📜 Details
[Design Doc]()

We want to be able to provide an alternative id to prometheus client now based on an environment variable. To achieve this we set the process id to:

* `MULTIPROCESS_WORKER_ID` if provided and fallback to PID in case it doesn't exist
* Moved the prometheus client initialization before we make the server.

Tested in the internal reddit development platform.

## ✅ Checks

Signed-off-by: Sotiris Nanopoulos <sotiris.nanopoulos@reddit.com>

### DIFF
--- a/baseplate/server/__init__.py
+++ b/baseplate/server/__init__.py
@@ -265,6 +265,13 @@ def load_app_and_run_server() -> None:
         config = read_config(args.config_file, args.server_name, args.app_name)
     assert config.server
 
+    if is_metrics_enabled(config.app):
+        from baseplate.server.prometheus import start_prometheus_exporter
+
+        start_prometheus_exporter()
+    else:
+        logger.info("Metrics are not configured, Prometheus metrics will not be exported.")
+
     configure_logging(config, args.debug)
 
     app = make_app(config.app)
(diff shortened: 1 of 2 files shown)


davinci26:
@SuperQ PTAL

SuperQ (approved):
LGTM

maeivysea (approved):

baseplate/server/__init__.py
```diff
@@ -265,6 +265,13 @@ def load_app_and_run_server() -> None:
         config = read_config(args.config_file, args.server_name, args.app_name)
     assert config.server
 
+    if is_metrics_enabled(config.app):
```
nsheaps, line 270: What's the reasoning behind moving this logic?

davinci26: We generate metrics even if `is_metrics_enabled` is enabled with the previous logic. So without this move if you `ls` the `prometheus multiprocess dir` you see that there are files generated there because prom is initialized

davinci26: Actually let me answer a bit better:

Somewhere in the app construction we are generating data prometheus stats before  we call `is_metrics_enabled` and `start_prometheus_exporter()`. This can be verified by the presence of files in the prometheus dir.

And because we set  multiprocess client id at `start_prometheus_exporter` these metrics are using the default `id` which defeats the purpose of this PR

nsheaps: Thanks for explaining! A comment to that effect might be good so we don't make that mistake again :)

nsheaps (approved):

PR #771, opened 2023-02-08
initialize gevent loop in baseplate scripts

justincely:
## 💸 TL;DR

The various baseplate-<action> scripts are the entry point to our running servers, and this is were we patch the stdlib to work well with gevent.   When working with gRPC, there is additional patching that needs to to be done in these scripts.  This is a third sentence so as to be technically correct.

## 📜 Details
[Jira](https://reddit.atlassian.net/browse/MESH-1327 )

We've currently been doing the `grpc_gevent.init_gevent()` patching in the various service's `__init__.py` files, but we've seen performance issues and python profiling graphs that indicated this patch wasn't having the intended affect.  The thrift client performance is significantly (2x or more) better and more consistent than the grpc client of the same server.

After some live debugging and profiling with pyspy, it looked like all the grpc calls were still outside the main server's event loop. According to various gh issues and docs, `init_gevent`:

```
This must be called AFTER the python standard lib has been patched, but BEFORE creating and gRPC objects.
```

We suspect calling this in the __init__ file was not sufficiently early to get the patching done right, so moving it here in an attempt to get it done as early and seamlessly as possible for users.  

I think adding this here creates a hard requirement of grpc for all servers, so happy to wrap this in a `try/except` to handle cases where they don't bring it in - but open to suggestions/preference for how to handle this.  

Installed this into services running in snoodev and did see some measurable performance improvements under conditions of high load for the client/servers.  

## ✅ Checks

### DIFF
--- a/bin/baseplate-script
+++ b/bin/baseplate-script
@@ -12,10 +12,15 @@ from baseplate.server.monkey import patch_stdlib_queues
 patch_all()
 patch_stdlib_queues()
 
-# grpc gevent patching
-import grpc.experimental.gevent as grpc_gevent  # noreorder
+# grpc gevent patching.
+# only patch if server does have gRPC as a dependency as this is
+# not needed for thrift or http only servers.
+try:
+    import grpc.experimental.gevent as grpc_gevent  # noreorder
+    grpc_gevent.init_gevent()
+except ImportError:
+    pass
 
-grpc_gevent.init_gevent()
 
 from baseplate.server import load_and_run_script
 
--- a/bin/baseplate-script
+++ b/bin/baseplate-script
@@ -12,6 +12,11 @@ from baseplate.server.monkey import patch_stdlib_queues
 patch_all()
 patch_stdlib_queues()
 
+# grpc gevent patching
+import grpc.experimental.gevent as grpc_gevent  # noreorder
+
+grpc_gevent.init_gevent()
+
 from baseplate.server import load_and_run_script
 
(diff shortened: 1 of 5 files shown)


marcoferrer (approved):

ghirsch-reddit (approved):

KTAtkinson (changes requested):
The link to the design doc doesn't work. It would be great to take a look at that before approving this, to make sure this is required.

This will likely be integrated into the 2.6 release that should happen in the next few weeks. Does that work with your timeline  Justin?

KTAtkinson (approved):

justincely:
sorry, we didn't have a design doc on this one.  I just forgot to remove that line of the template - have removed now.

If this could be released in a patch earlier, that would honestly be ideal for us.  We've been working to track down performance issues around this for many weeks, and this amount of performance latency is slowing/stopping adoption by high-volume clients as well as causing distractions as we try to tune parameters in other parts of the stack.  

That being said - if it can't happen, it can't happen and we'll hold off on this work until it can go out.

marcoferrer:
So this change is blocking a significant milestone in the gRPC migration. I just linked the design docs in the #baseplate-py channel for the overall project.

We've tried multiple options to get around needing to make this change here but due to the current bp-py architecture, the alternatives have been unsuccessful.

justincely:
Closing this at the moment.  Further testing reveals this has not alleviated what we thought it might.

PR #783, opened 2023-04-19
Add date and time parsers

nakulj:
## 💸 TL;DR
Add date and time parsers to config.py

## 📜 Details
None

## ✅ Checks

- [N/A] Contributor License Agreement (CLA) completed if not a Reddit employee

### DIFF
--- a/baseplate/lib/config.py
+++ b/baseplate/lib/config.py
@@ -244,6 +244,18 @@ def File(mode: str = "r") -> Callable[[str], IO]:  # noqa: D401
 
     return open_file
 
+def DateTime(text: str) -> datetime.datetime:
+    if not text:
+        raise ValueError("expected a time")
+
+    return datetime.datetime.fromisoformat(text)
+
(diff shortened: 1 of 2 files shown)


baseplate/lib/config.py
```diff
@@ -245,6 +245,20 @@ def open_file(text: str) -> IO:
     return open_file
 
 
+def DateTime(text: str) -> datetime.datetime:
```
ketralnis, line 248: these produce timezone naive datetimes, which is a big footgun in python's datetime stack. Instead of using these functions directly I might advise a strptime call or something similar where you can specify the resultant timezone

KTAtkinson: It would be nice if the user was required to specify the tz or the API itself  specified the tz to avoid using tz naive dates or times.

KTAtkinson (changes requested):
Can you describe your use case for this feature? What other things have you tried that use the current version of the API, what where the challenges?

PR #785, opened 2023-05-05
feat: Improve shutdown Delay default

davinci26:
When a baseplate server is shutting down in k8s

* baseplate calls func (srv *Server) Shutdown(ctx context.Context) error which does the following:
* Shutdown works by first closing all open listeners

* k8s is attempting to remove the endpoint for the endpoints group and thus make (new) incoming traffic to the pod 0

This is a race condition because some callers are still trying to connect. We have empirically seen that the 5 second default is good enough for most services.

See also https://github.com/reddit/baseplate.go/pull/615


KTAtkinson (approved):

davinci26:
👓 @clayton-gonsalves @davinci26 @shadialtarsha

bjk-reddit (approved):
This seems like a sensible default.

clayton-gonsalves (approved):

PR #792, opened 2023-05-17
Add queue flushing signal handler

atai92:
Closes # 

## 💸 TL;DR

Add a queue flushing signal handler, so when the event publisher receives a SIGINT/SIGTERM signal it'll continue flushing the queue, so events aren't dropped on shutdown.

## 📜 Details
[Design Doc]()

[Jira]()

Tested using the included `Dockerfile` using the following steps:
```
docker run --sysctl fs.mqueue.msg_max=10000 --sysctl fs.mqueue.msgsize_max=102400 -it -v ~/baseplate.py:/src --ulimit msgqueue=-1 c725da7d50d0
python -m baseplate.sidecars.event_publisher --queue-name test --debug test.ini

# In a separate tab, exec into the container
docker exec -it 07e5b9a2bb37 /bin/bash
python
# Within Python, ran the code below
from baseplate.lib.message_queue import MessageQueue
import time

queue_name = "test"

event_queue = MessageQueue(
        "/events-" + queue_name,
        max_messages=10000,
        max_message_size=102400,
    )


message = b'[{"source":"test","action":"test","noun":"test","client_timestamp":1683823393779,"uuid":"20de9dec-6632-44fa-a3a3-ca29d177776d","action_info":{"reason":"test"}}]'

current_time = time.time()
while time.time() - current_time < 5:
    event_queue.put(message)
```

This screenshot below shows the reproduction of a bug where we would exit from a `BatchFull` exception:
Screenshot 2023-05-16 at 5 16 01 PM

This screenshot below shows that the `BatchFull` exception is no longer present using the same steps above and sending a SIGINT to `event_publisher` will still cause it to wait for the POSIX queue to be empty before exiting:
Screenshot 2023-05-16 at 5 21 55 PM

## ✅ Checks

### DIFF
--- a/baseplate/sidecars/event_publisher.py
+++ b/baseplate/sidecars/event_publisher.py
@@ -5,7 +5,10 @@ import gzip
 import hashlib
 import hmac
 import logging
+import signal
+import sys
 
+from types import FrameType
 from typing import Any
 from typing import List
 from typing import Optional
@@ -163,6 +166,14 @@ class BatchPublisher:
 
 SERIALIZER_BY_VERSION = {"2": V2Batch, "2j": V2JBatch}
 
+def publish_serialized_batch(publisher: BatchPublisher, batcher: TimeLimitedBatch) -> None:
+    """Serializes batch, publishes it using the publisher, and then resets the batch for more messages."""
+    serialized_batch = batcher.serialize()
+    try:
+        publisher.publish(serialized_batch)
+    except Exception:
+        logger.exception("Events publishing failed.")
+    batcher.reset()
 
 def publish_events() -> None:
     arg_parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


baseplate/sidecars/event_publisher.py
```diff
@@ -214,6 +227,33 @@ def publish_events() -> None:
...
+    def flush_queue_signal_handler(_signo: int, _frame: FrameType) -> None:
+        """Signal handler for flushing messages from the queue and publishing them."""
+        message: Optional[bytes]
+        logger.info("Shutdown signal received. Flushing events...")
+
+        while True:
+            try:
+                message = event_queue.get(timeout=0.2)
+            except TimedOutError:
+                if len(batcher.serialize()) > 0:
```
KTAtkinson, line 239: The length here will [always be](https://github.com/reddit/baseplate.py/blob/cb60ac69275d4c0c9f2721b38bf6898543169911/baseplate/sidecars/event_publisher.py#L83) greater than `0` because we add a header. We might want to add a `items_count` function to the batcher to count the number of items currently in the queue. It could use the length of [batcher._items](https://github.com/reddit/baseplate.py/blob/cb60ac69275d4c0c9f2721b38bf6898543169911/baseplate/sidecars/event_publisher.py#L89).

KTAtkinson (changes requested):

redloaf (commented):
Is this still needed?

PR #800, opened 2023-07-24
Provide custom 404 view that doesn't echo path

sp3nx0r:
## 💸 TL;DR
We get some bogus bug bounty submissions that talk about HTML injection on our 404 Not Found pages, which are default Pyramid views which echo out the `path` of the Request that isn't found. So you can put some gibberish in there, but it looks real jank and would not be a valid social engineering attack, but it should be simple for us to just default this to be quieter so we don't receive these reports. 

Ref: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html 

## 📜 Details
Example: https://ads-api.reddit.com/this_is_a_test_where_i_could_spoof_whatever_i_guess

TBD

## ✅ Checks

### DIFF
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -412,6 +412,7 @@ class BaseplateConfigurator:
         config.set_request_factory(RequestFactory(self.baseplate))
         config.add_subscriber(self._on_new_request, pyramid.events.ContextFound)
         config.add_subscriber(self._on_application_created, pyramid.events.ApplicationCreated)
+        config.add_notfound_view(notfound_override)
 
         # Position of the tween is important. We need it to cover all code
         # that can written in the app. This means that it should be above
@@ -444,3 +445,7 @@ def get_is_healthy_probe(request: Request) -> int:
             code,
         )
         return IsHealthyProbe.READINESS
+
+
+def notfound_override(_request: Request) -> Response:
+    return Response("Not Found", status="404 Not Found")
--- a/tests/integration/pyramid_tests.py
+++ b/tests/integration/pyramid_tests.py
@@ -194,6 +194,14 @@ class ConfiguratorTests(unittest.TestCase):
         self.assertFalse(self.observer.on_server_span_created.called)
         self.assertFalse(self.context_init_event_subscriber.called)
 
+    def test_not_found_echo_path(self):
+        # confirm that issue #800 isn't reintroduced. This is an issue where we
+        # echo the path to the 404 in the response which probably isn't a
+        # problem, but does show up in automated vuln scans which can cause some
+        # extra work hunting down false positives
+        resp = self.test_app.get("/doesnt_exist", status=404)
+        self.assertNotIn(b"doesnt_exist", resp.body)
+
     def test_exception_caught(self):
         with self.assertRaises(TestException):
             self.test_app.get("/example?error")


maeivysea:
👋 deferring to our baseplate SMEs

KTAtkinson (approved):

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -412,6 +412,7 @@ def includeme(self, config: Configurator) -> None:
         config.set_request_factory(RequestFactory(self.baseplate))
         config.add_subscriber(self._on_new_request, pyramid.events.ContextFound)
         config.add_subscriber(self._on_application_created, pyramid.events.ApplicationCreated)
+        config.add_not_found_view(not_found_override)
```
ketralnis: This fails to run, it's `add_notfound_view` (one fewer underscore). I'll have a fix up with a test in a sec

sp3nx0r: :facepalm: that what I get for reading the doc in markdown and not humanese.

ketralnis:
I'm not able to push up into this branch but here's the whole diff after the fix and the added test. I've confirmed that the test fails before and passes after

```
diff --git a/baseplate/frameworks/pyramid/__init__.py b/baseplate/frameworks/pyramid/__init__.py
index eba262a..842931d 100644
--- a/baseplate/frameworks/pyramid/__init__.py
+++ b/baseplate/frameworks/pyramid/__init__.py
@@ -412,6 +412,7 @@ class BaseplateConfigurator:
         config.set_request_factory(RequestFactory(self.baseplate))
         config.add_subscriber(self._on_new_request, pyramid.events.ContextFound)
         config.add_subscriber(self._on_application_created, pyramid.events.ApplicationCreated)
+        config.add_notfound_view(notfound_override)

         # Position of the tween is important. We need it to cover all code
         # that can written in the app. This means that it should be above
@@ -444,3 +445,7 @@ def get_is_healthy_probe(request: Request) -> int:
             code,
         )
         return IsHealthyProbe.READINESS
+
+
+def notfound_override(_request: Request) -> Response:
+    return Response("Not Found", status="404 Not Found")
diff --git a/tests/integration/pyramid_tests.py b/tests/integration/pyramid_tests.py
index 2f85b7b..b4581ce 100644
--- a/tests/integration/pyramid_tests.py
+++ b/tests/integration/pyramid_tests.py
@@ -194,6 +194,14 @@ class ConfiguratorTests(unittest.TestCase):
         self.assertFalse(self.observer.on_server_span_created.called)
         self.assertFalse(self.context_init_event_subscriber.called)

+    def test_not_found_echo_path(self):
+        # confirm that issue #800 isn't reintroduced. This is an issue where we
+        # echo the path to the 404 in the response which probably isn't a
+        # problem, but does show up in automated vuln scans which can cause some
+        # extra work hunting down false positives
+        resp = self.test_app.get("/doesnt_exist", status=404)
+        self.assertNotIn(b"doesnt_exist", resp.body)
+
     def test_exception_caught(self):
         with self.assertRaises(TestException):
             self.test_app.get("/example?error")
```

ketralnis (approved):
got the code swapped out and added a test as well. but this is now me approving my own code so it could use another look :)

sp3nx0r:
@ketralnis 's code looks good to me, going to be merging before EOD

PR #811, opened 2023-10-24
Don't fail to finalize observers if one fails

bradengroom:
## 💸 TL;DR

When a request finishes, baseplate invokes `on_finish` for each observer. Sometimes these finalizers will fail and raise an exception. This causes the loop to exit early and we fail to finalize observers later in the loop. This could result in a few things including [postgres sessions not getting closed](https://github.com/reddit/baseplate.py/blob/114248987ce0e8a0ddd102c80b00ef43f4dbf14e/baseplate/clients/sqlalchemy.py#L359) timely. Raising from a finalizer also means we fail to [clean up reference cycles](https://github.com/reddit/baseplate.py/blob/114248987ce0e8a0ddd102c80b00ef43f4dbf14e/baseplate/__init__.py#L634) later in the function and leave more work for GC to pick up later.

## 📜 Details
N/A

## ✅ Checks

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -629,7 +629,10 @@ class Span:
 
         """
         for observer in self.observers:
-            observer.on_finish(exc_info)
+            try:
+                observer.on_finish(exc_info)
+            except Exception:
+                logger.exception("Exception raised while finalizing observer")
 
         # clean up reference cycles
         self.context = None  # type: ignore


salomon-smekecohen (commented):
Thank you B money.

spladug:
:fish: good call!

rram:
🐟 lgtm

baseplate/__init__.py
```diff
@@ -629,7 +629,10 @@ def finish(self, exc_info: Optional[_ExcInfo] = None) -> None:
 
         """
         for observer in self.observers:
-            observer.on_finish(exc_info)
+            try:
+                observer.on_finish(exc_info)
+            except Exception:
+                logger.exception("Exception raised while finalizing observer")
```
KTAtkinson, line 635: Please make it so this exception includes all the details the original exception would include. I don't want things to become harder to debug because we're catching this exception.

KTAtkinson (changes requested):

baseplate/__init__.py
bradengroom: `logger.exception` actually adds that information for us
https://docs.python.org/3/library/logging.html#logging.Logger.exception

KTAtkinson (approved):

PR #814, opened 2023-11-29
Migrate handling of headers to use otel propagators

trevorriles:
Use opentelemetry propgators to handle headers.


baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,129 @@
...
+            self.SAMPLED_KEY,
+        }
+
+
+def _extract_first_element(
+    items: typing.Iterable[CarrierT],
+) -> typing.Optional[CarrierT]:
+    if items is None:
+        return None
+    return next(iter(items), None)
```
Xaelias: If you define:
```suggestion
def _extract_first_element(
    items: typing.Iterable[CarrierT],
    default: Optional[Any] = None,
) -> typing.Optional[CarrierT]:
    if items is None:
        return default
    return next(iter(items), default)
```

Then you can just do:
```
        trace_id = _extract_first_element(getter.get(carrier, self.TRACE_ID_KEY), trace_id)
```
which is slightly cleaner IMO.

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,129 @@
...
+            or flags
+        )
+
+        if (
+            trace_id == trace.INVALID_TRACE_ID
+            or span_id == trace.INVALID_SPAN_ID
+            or self._trace_id_regex.fullmatch(trace_id) is None
+            or self._span_id_regex.fullmatch(span_id) is None
+        ):
+            return context
```
Xaelias, line 66: Can you add a comment that explains why / what's happening there for future reference?

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,129 @@
...
+        # flag values set. Since the setting of at least one implies
+        # the desire for some form of sampling, propagate if either
+        # header is set to allow.
+        if sampled in self._SAMPLE_PROPAGATE_VALUES or flags == "1":
+            options |= trace.TraceFlags.SAMPLED
+
+        return trace.set_span_in_context(
+            trace.NonRecordingSpan(
+                trace.SpanContext(
+                    # trace an span ids are encoded in hex, so must be converted
```
Xaelias: You might want to move that comment line 66 or smth. IMO this looks weird because here you don't convert or encode anything.

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,129 @@
...
+        span_context = span.get_span_context()
+        if span_context == trace.INVALID_SPAN_CONTEXT:
+            return
+
+        sampled = (trace.TraceFlags.SAMPLED & span_context.trace_flags) != 0
+        setter.set(
+            carrier,
+            self.TRACE_ID_KEY,
+            # The Reddit B3 format expects a 64bit TraceId not a 128bit ID. This is truncated for compatibility.
+            format_trace_id(span_context.trace_id)[16:],
```
Xaelias:
```suggestion
            format_trace_id(span_context.trace_id)[-16:],
```

If we only want the least significant 16 bytes, I believe this is more accurate.

Xaelias (changes requested):
Yeah I think that mostly looks good

baseplate/lib/propagator_redditb3.py
trevorriles: Good call.

Xaelias (approved):

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+
+def _extract_first_element(
+    items: typing.Iterable[CarrierT],
+    default: typing.Optional[typing.Any] = None,
+) -> typing.Optional[CarrierT]:
+    if items is None:
+        return default
+    return next(iter(items), default)
```
Xaelias, line 131: You don't actually use the default anywhere though :-D

Xaelias:
Manually merged because github was being annoying

PR #817, opened 2023-11-30
[live-data-fetcher] Distribute read load across objects in S3

SiddharthManoj:
See [Experiments - S3 Prefix Sharding Design Doc](https://docs.google.com/document/d/1K81uZiHzzYpVI32NmNzd7aGd9D0OqqxnoZHTvGSvp8g/edit?usp=sharing) for more info.

See [#incident-3788--slower-experiment-changes-due-to-s3-rate-limiting](https://reddit.enterprise.slack.com/archives/C067Q6N14LA) for full incident context.

Problem (as copied from doc)
On the week of November 27th, we started to observe that the rate of requests for experiment config data to the [S3 bucket ](https://s3.console.aws.amazon.com/s3/buckets/reddit-experiment-config-prod-active-configs?region=us-east-1&tab=objects)was breaching the S3 rate limit of 5500 QPS. The main trigger condition is when experiments are changed, the manifest is uploaded (every X min), and a ZooKeeper Data Watch occurs. This subsequently triggers all service pods to refetch a copy of the new experiment config data (with a retry policy of 5 attempts).

When all pods attempt to fetch from the same S3 file at once, we breach the 5500 QPS limit set by AWS sending a large number of pods into a retry state (up to 5 retries).

These retries are done using boto3’s built-in exponential backoff strategy with a factor of 2 (with jitter). Ultimately we observed that while a retrying Pod may not receive the experiment config on the 1st try, 99% of them receive it by at most the 5th try. However, 1% of Pods do not get it even on the 5th retry and thus stop retrying - therefore these pods do not receive the most up-to-date experiment data until another manifest upload and ZK data watch is triggered.

### Solution

Experiments service will send over multiple copies of the manifest - each prefixed numerically. See: https://github.snooguts.net/reddit/reddit-service-experiment-config/pull/3616

In the baseplate sidecar, we just read in the `num_file_shards` key from ZK and randomly generate a number from 1 to num_file_shards. We prepend that prefix to the file name followed by the delimiter ("/") and then attempt to fetch that file from S3.

 ### Benefits

By having these prefixed in the way above, we make use of S3's built in prefix partitioning for read performance. This means that pod reads can be distributed among the files (random load balancing).

See [AWS Design Practices doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) for more info

 ### Is this backwards compatible with older versions of the sidecar?

Yes, old versions of the sidecar won't know about the `num_file_shards` key in ZK and will just pull down the original file provided in the ZK node (as it works today). 

Only new updated versions will attempt to prepend a prefix.

Old versions will always pull a file that looks like:

`2023-10-31T16:13:18Z_3437442509_29cd4048e52252cb_experiment-config.json (original file)`

New versions will pull files that look like:

```
1/2023-10-31T16:13:18Z_3437442509_29cd4048e52252cb_experiment-config.json

2/2023-10-31T16:13:18Z_3437442509_29cd4048e52252cb_experiment-config.json

3/2023-10-31T16:13:18Z_3437442509_29cd4048e52252cb_experiment-config.json

4/2023-10-31T16:13:18Z_3437442509_29cd4048e52252cb_experiment-config.json
```

### DIFF
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -116,7 +116,7 @@ def _parse_loader_type(data: bytes) -> LoaderType:
     return LoaderType(loader_type)
 
 
-def _generate_sharded_file_key_prefix(num_file_shards: Optional[int]) -> str:
+def _generate_sharded_file_key(num_file_shards: Optional[int], file_key: str) -> str:
     # We can't assume that every ZK Node that is being NodeWatched by the live-data-fetcher
     # will make use of S3 prefix sharding - but, we know at least one does (/experiments).
     # If it's not present or the value is 1, set the prefix to empty string ""
@@ -127,7 +127,8 @@ def _generate_sharded_file_key_prefix(num_file_shards: Optional[int]) -> str:
         # uploaded so fetch one randomly using a randomly generated prefix.
         # Generate a random number from 1 to num_file_shards exclusive to use as prefix.
         sharded_file_key_prefix = str(random.randrange(1, num_file_shards)) + "/"
-    return sharded_file_key_prefix
+    # Append prefix (if it exists) to our original file key.
+    return sharded_file_key_prefix + file_key
 
 
 def _load_from_s3(data: bytes) -> bytes:
@@ -138,13 +139,16 @@ def _load_from_s3(data: bytes) -> bytes:
     loader_config = json.loads(data.decode("UTF-8"))
     try:
         num_file_shards = loader_config.get("num_file_shards")
-        # Append prefix (if it exists) to our original file key.
-        file_key = _generate_sharded_file_key_prefix(num_file_shards) + loader_config["file_key"]
+
+        # We expect this key to always be present, otherwise it's an exception.
+        file_key = loader_config["file_key"]
+
+        sharded_file_key = _generate_sharded_file_key(num_file_shards, file_key)
 
         region_name = loader_config["region_name"]
         s3_kwargs = {
             "Bucket": loader_config["bucket_name"],
-            "Key": file_key,
+            "Key": sharded_file_key,
             "SSECustomerKey": loader_config["sse_key"],
             "SSECustomerAlgorithm": "AES256",
         }
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -4,9 +4,9 @@ import configparser
 import json
 import logging
 import os
+import random
 import sys
 import time
-import random
 
 from enum import Enum
 from pathlib import Path
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -121,8 +121,7 @@ def _generate_sharded_file_key(num_file_shards: Optional[int], file_key: str) ->
     # will make use of S3 prefix sharding - but, we know at least one does (/experiments).
     # If it's not present or the value is 0 or 1, set the prefix to empty string ""
     sharded_file_key_prefix = ""
-    no_shard_values = [None, 0, 1]
-    if not num_file_shards in no_shard_values:
+    if num_file_shards is not None and num_file_shards > 1:
         # If the num_file_shards key is present, we may have multiple copies of the same manifest
         # uploaded so fetch one randomly using a randomly generated prefix.
         # Generate a random number from 1 to num_file_shards exclusive to use as prefix.
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -127,12 +127,12 @@ def _load_from_s3(data: bytes) -> bytes:
         # We can't assume that every caller of this method will be using prefix sharding on
         # their S3 objects.
         num_file_shards = loader_config.get("num_file_shards", 1)
-        logger.error(num_file_shards)
+
         # If the num_file_shards key is present, we may have multiple copies of the same manifest
         # uploaded so fetch one randomly using a randomly generated prefix.
         # Generate a random number from 0 to num_file_shards exclusive to use as prefix.
         file_key_prefix = random.randrange(num_file_shards)
-        logger.error(file_key_prefix)
+
         # If 0 is generated, don’t append a prefix, fetch the file with no prefix
         # since we always upload one file without a prefix.
         if file_key_prefix == 0:
@@ -151,7 +151,6 @@ def _load_from_s3(data: bytes) -> bytes:
             "SSECustomerKey": loader_config["sse_key"],
             "SSECustomerAlgorithm": "AES256",
         }
-        logger.error(file_key)
     except KeyError as e:
         # We require all of these keys to properly read from S3.
         logger.exception(
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -6,6 +6,7 @@ import logging
 import os
 import sys
 import time
+import random
 
 from enum import Enum
 from pathlib import Path
@@ -122,10 +123,31 @@ def _load_from_s3(data: bytes) -> bytes:
     # and json is an easier format for znode authors to work with.
     loader_config = json.loads(data.decode("UTF-8"))
     try:
+        # Default to 1 since we generate random numbers from 0 to num_file_shards exclusive.
+        # We can't assume that every caller of this method will be using prefix sharding on
+        # their S3 objects.
+        num_file_shards = loader_config.get("num_file_shards", 1)
+
+        # If the num_file_shards key is present, we may have multiple copies of the same manifest
+        # uploaded so fetch one randomly using a randomly generated prefix.
+        # Generate a random number from 0 to num_file_shards exclusive to use as prefix.
+        file_key_prefix = random.randrange(num_file_shards)
+
+        # If 0 is generated, don’t append a prefix, fetch the file with no prefix
+        # since we always upload one file without a prefix.
+        if file_key_prefix == 0:
+            sharded_file_key_prefix = ""
+        else:
+            # If any other number is generated, fetch one of the copies of the
+            # file which has an included prefix.
+            sharded_file_key_prefix = str(file_key_prefix) + "/"
+
+        # Append prefix (if it exists) to our original file key.
+        file_key = sharded_file_key_prefix + loader_config["file_key"]
         region_name = loader_config["region_name"]
         s3_kwargs = {
             "Bucket": loader_config["bucket_name"],
-            "Key": loader_config["file_key"],
+            "Key": file_key,
             "SSECustomerKey": loader_config["sse_key"],
             "SSECustomerAlgorithm": "AES256",
         }
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -126,7 +126,8 @@ def _load_from_s3(data: bytes) -> bytes:
         num_file_shards = loader_config.get("num_file_shards")
 
         # We can't assume that every caller of this method will be using prefix sharding on
-        # their S3 objects. If it's not present, set the prefix to empty string ""
+        # their S3 objects (but at least one service definitely does - experiments).
+        # If it's not present or the value is 1, set the prefix to empty string ""
         if not num_file_shards or num_file_shards == 1:
             sharded_file_key_prefix = ""
         else:
--- a/baseplate/sidecars/live_data_watcher.py
+++ b/baseplate/sidecars/live_data_watcher.py
@@ -167,16 +167,15 @@ def _load_from_s3(data: bytes) -> bytes:
         # a public resource belonging to another cluster/AWS account unless the request credentials
         # are unsigned.
 
-        # Access S3 with 10 max retries enabled:
+        # Default # of retries in legacy mode (current mode) is 5.
         s3_client = boto3.client(
             "s3",
-            config=Config(signature_version=UNSIGNED, retries={"total_max_attempts": 10}),
+            config=Config(signature_version=UNSIGNED),
             region_name=region_name,
         )
     else:
         s3_client = boto3.client(
             "s3",
-            config=Config(retries={"total_max_attempts": 10}),
             region_name=region_name,
         )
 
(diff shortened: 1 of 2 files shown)


tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -26,13 +28,21 @@ def setUp(self):
...
+            else:
+                # All other copies should include the sharded prefix.
+                sharded_file_key = str(file_shard_num) + "/" + default_file_key
+            s3_client.put_object(
+                Bucket=bucket_name,
+                Key=sharded_file_key,
+                Body=json.dumps(s3_data).encode(),
+                SSECustomerKey="test_decryption_key",
+                SSECustomerAlgorithm="AES256",
+            )
```
SiddharthManoj, line 50: For unit test, we can upload all the copies of the file in the setUp. 

And then in each unit test, we can optionally provide the `num_file_shards` key in the ZK content. If the key is missing, we should expect the unit test to still pass since it should fetch the original file with no prefix.

If the key is present, we should also expect it to pass by fetching one of the prefixed files.

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -53,6 +63,23 @@ def test_s3_load_type_on_change(self):
...
+        inst = NodeWatcher(str(dest), os.getuid(), os.getgid(), 777)
+
+        new_content = b'{"live_data_watcher_load_type":"S3","bucket_name":"test_bucket","file_key":"test_file_key","sse_key":"test_decryption_key","region_name":"us-east-1", "num_file_shards": 5}'
+        expected_content = b'{"foo_encrypted": "bar_encrypted"}'
+
+        # For safe measure, run this 20 times. It should succeed every time.
+        # We've uploaded 5 files to S3 in setUp() and num_file_shards=5 in the
+        # ZK node so we should be fetching one of these 5 files randomly (and successfully)
+        # and all should have the same content.
+        for i in range(20):
```
SiddharthManoj: I couldn't think of a better way to do this than to run `on_change()` X number of times. This is because the only output of the function being unit tested is the contents of the manifest so functionally that's all I can test.

on_change() has the random number generation logic internally.

By running `on_change()` 50 times, i would expect different random numbers to have generated at least once, so we would pull from one of the 5 files uploaded in setUp() randomly and I would expect it to succeed and have the same contents each time.

SiddharthManoj:
@areitz Just wanted to get the PR out there. We can still wait for the [doc](https://docs.google.com/document/d/1K81uZiHzzYpVI32NmNzd7aGd9D0OqqxnoZHTvGSvp8g/edit?usp=sharing) to be reviewed first and also the Experiments Service PR has to go in first: https://github.snooguts.net/reddit/reddit-service-experiment-config/pull/3616

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -11,8 +11,11 @@
 
 from moto import mock_s3
 
+from baseplate.sidecars.live_data_watcher import _generate_sharded_file_key
```
SiddharthManoj, line 15: I know `_generate_sharded_file_key` has an underscore prefix, but we're only using it internally in reality. 

I just need to call it here as part of the unit test

baseplate/sidecars/live_data_watcher.py
```diff
@@ -115,17 +116,39 @@ def _parse_loader_type(data: bytes) -> LoaderType:
...
+    # We can't assume that every ZK Node that is being NodeWatched by the live-data-fetcher
+    # will make use of S3 prefix sharding - but, we know at least one does (/experiments).
+    # If it's not present or the value is 1, set the prefix to empty string ""
+    if not num_file_shards or num_file_shards == 1:
+        sharded_file_key_prefix = ""
+    else:
+        # If the num_file_shards key is present, we may have multiple copies of the same manifest
+        # uploaded so fetch one randomly using a randomly generated prefix.
+        # Generate a random number from 1 to num_file_shards exclusive to use as prefix.
+        sharded_file_key_prefix = str(random.randrange(1, num_file_shards)) + "/"
```
areitz, line 128: I'm a little confused by `random.randrange()` -- it seems like it'll choose a number between 1 and `num_file_shards - 1`? I ran it a bunch in a repl and never got a 9, until I bumped the end range to 10:

```
>>> print(random.randrange(3, 9))
3
>>> print(random.randrange(3, 9))
3
>>> print(random.randrange(3, 9))
6
>>> print(random.randrange(3, 9))
8
>>> print(random.randrange(3, 9))
8
>>> print(random.randrange(3, 9))
4
>>> print(random.randrange(3, 9))
3
>>> print(random.randrange(3, 9))
4
>>> print(random.randrange(3, 9))
4
>>> print(random.randrange(3, 9))
6
>>> print(random.randrange(3, 9))
3
>>> print(random.randrange(3, 9))
5
>>> print(random.randrange(3, 9))
7
>>> print(random.randrange(3, 9))
7
>>> print(random.randrange(3, 9))
6
>>> print(random.randrange(3, 9))
7
>>> print(random.randrange(3, 9))
6
>>> print(random.randrange(3, 9))
8
>>> print(random.randrange(3, 9))
8
>>> print(random.randrange(3, 9))
6
>>> print(random.randrange(3, 9 + 1))
8
>>> print(random.randrange(3, 9 + 1))
9
>>> print(random.randrange(3, 9 + 1))
6
>>> print(random.randrange(3, 9 + 1))
8
>>> print(random.randrange(3, 9 + 1))
4
```

Do you want `num_file_shards + 1` here?

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -53,6 +93,23 @@ def test_s3_load_type_on_change(self):
...
 
+    def test_s3_load_type_on_change_sharding(self):
+        dest = self.output_dir.joinpath("data.txt")
+        inst = NodeWatcher(str(dest), os.getuid(), os.getgid(), 777)
+
+        new_content = b'{"live_data_watcher_load_type":"S3","bucket_name":"test_bucket","file_key":"test_file_key","sse_key":"test_decryption_key","region_name":"us-east-1", "num_file_shards": 5}'
+        expected_content = b'{"foo_encrypted": "bar_encrypted"}'
+
+        # For safe measure, run this 50 times. It should succeed every time.
+        # We've uploaded 5 files to S3 in setUp() and num_file_shards=5 in the
```
areitz, line 105: Out of curiosity, is there a way to enrich this test such that we verify that every one of the five prefixes was read from at least once?

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -42,7 +57,32 @@ def run(self, result: unittest.TestResult = None) -> unittest.TestResult:
             self.output_dir = Path(loc)
             return super().run(result)
 
-    def test_s3_load_type_on_change(self):
+    def test_generate_sharded_file_key_no_sharding(self):
+        original_file_key = "test_file_key"
+        actual_sharded_file_key = _generate_sharded_file_key(None, original_file_key)
```
areitz: Do you want to extend this test to also include this condition?

```suggestion
        actual_sharded_file_key = _generate_sharded_file_key(1, original_file_key)
```

areitz: And I guess also this?

```suggestion
        actual_sharded_file_key = _generate_sharded_file_key(0, original_file_key)
```

Since `not 0` should be truthy?

areitz (changes requested):
This looks super awesome, thanks!

I had a few minor questions, nothing blocking really. I can approve without code changes once I understand the answers to my questions.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Yup, it's meant to be exclusive. When NUM_FILE_SHARDS = 6 for example, here are the files generated

file.json
1_file.json
2_file.json
3_file.json
4_file.json
5_file.json

So we only ever want to go up to 5.

This is because 6 refers to the total num of files which includes the original file as well.

tests/unit/sidecars/live_data_watcher_tests.py
SiddharthManoj: Not sure if I have any ideas since random number generation doesn't guarantee that we hit all the 5 prefixes even if we ran it 1000 times (very unlikely, but certainly possible). It would probably make the unit test flakey (failing a very small percentage of the time).

There might be a different way to do what you're saying though...but let me investigate.

tests/unit/sidecars/live_data_watcher_tests.py
SiddharthManoj: nope, good catch

SiddharthManoj: wait, sorry I misunderstood.

Yeah it should be truthy - so if the value is 0, the prefix is ""

tests/unit/sidecars/live_data_watcher_tests.py
areitz: Yeah, flakiness is a good worry.

If it's possible to check which prefixes get read, maybe we'd be happy if it's at least two unique ones?

tests/unit/sidecars/live_data_watcher_tests.py
```diff
@@ -59,9 +59,11 @@ def run(self, result: unittest.TestResult = None) -> unittest.TestResult:
 
     def test_generate_sharded_file_key_no_sharding(self):
         original_file_key = "test_file_key"
-        actual_sharded_file_key = _generate_sharded_file_key(None, original_file_key)
         expected_sharded_file_key = "test_file_key"
-        self.assertEqual(actual_sharded_file_key, expected_sharded_file_key)
+        possible_no_sharding_values = [0, 1, None]
```
areitz: Nice!

areitz (approved):

tests/unit/sidecars/live_data_watcher_tests.py
SiddharthManoj: @areitz That sounds reasonable! Should I still only run over it 50 times or should I increase that?

baseplate/sidecars/live_data_watcher.py
```diff
@@ -115,17 +116,39 @@ def _parse_loader_type(data: bytes) -> LoaderType:
...
+
+
 def _load_from_s3(data: bytes) -> bytes:
     # While many of the baseplate configurations use an ini format,
     # we've opted for json in these internal-to-znode-configs because
     # we want them to be fully controlled by the writer of the znode
     # and json is an easier format for znode authors to work with.
     loader_config = json.loads(data.decode("UTF-8"))
     try:
+        num_file_shards = loader_config.get("num_file_shards")
```
KTAtkinson, line 140: :no_bell: This will require that applications update their configuration in order to change the number of shards. Could we store this number in ZK?

tests/unit/sidecars/live_data_watcher_tests.py
mrlevitas: you can probably set [seed](https://docs.python.org/3/library/random.html#random.seed)s for `random` to map to specific shard values with some trial & error (e.g. for some seed X, shard Y will always be returned, and you can have a dict mapping them, then iterate over the dict in the test, setting the seed, and asserting your expected/mapped shard #)

mrlevitas (approved):
one suggestion for how to make the random part of the tests deterministic

baseplate/sidecars/live_data_watcher.py
mrlevitas: If I understand the question correctly, the number of shards is stored in ZK as part of the s3 metadata, see [here](https://github.snooguts.net/reddit/reddit-service-experiment-config/pull/3616/files#diff-5457c2c327f90a415894a357ecb8c42b925fdaf5b1a929bca8b6c2a55c2aecd0R142)

baseplate/sidecars/live_data_watcher.py
```diff
@@ -115,17 +116,39 @@ def _parse_loader_type(data: bytes) -> LoaderType:
...
+    # If it's not present or the value is 1, set the prefix to empty string ""
+    if not num_file_shards or num_file_shards == 1:
+        sharded_file_key_prefix = ""
+    else:
+        # If the num_file_shards key is present, we may have multiple copies of the same manifest
+        # uploaded so fetch one randomly using a randomly generated prefix.
+        # Generate a random number from 1 to num_file_shards exclusive to use as prefix.
+        sharded_file_key_prefix = str(random.randrange(1, num_file_shards)) + "/"
+    # Append prefix (if it exists) to our original file key.
+    return sharded_file_key_prefix + file_key
```
kylelemons: 🔕 I know python is python, but I think I would do

```
sharded_file_key_prefix = ""
if ...:
  sharded_file_key_prefix = str(...)
```

so that it's clear that the variable always has a value no matter the code path.

kylelemons (approved):

jerroydmoore (approved):

tests/unit/sidecars/live_data_watcher_tests.py
SiddharthManoj: Oooh thanks for the pointer @mrlevitas  - let me see if I can figure this out quickly - if not, and if this becomes more urgent, I may have to merge without that unit test.

baseplate/sidecars/live_data_watcher.py
SiddharthManoj: Yeah it's stored in ZK ~ so we should be good to go.

baseplate/sidecars/live_data_watcher.py
```diff
@@ -145,16 +167,15 @@ def _load_from_s3(data: bytes) -> bytes:
...
         # are unsigned.
 
-        # Access S3 with 10 max retries enabled:
+        # Default # of retries in legacy mode (current mode) is 5.
         s3_client = boto3.client(
             "s3",
-            config=Config(signature_version=UNSIGNED, retries={"total_max_attempts": 10}),
+            config=Config(signature_version=UNSIGNED),
             region_name=region_name,
         )
```
SiddharthManoj, line 175: Per https://reddit.slack.com/archives/C067Q6N14LA/p1701373678708859?thread_ts=1701370864.667409&cid=C067Q6N14LA we want to revert the changes made in [PR/813](https://github.com/reddit/baseplate.py/pull/813) where we increased retry total_max_attempt from 5->10

cc @areitz @jerroydmoore

KTAtkinson (approved):

PR #819, opened 2023-12-22
Open Telemetry

Xaelias:
This PR implements otel for http and thrift in bp.py.
Because of unsolvable conflicting dependencies, this also removes python 3.7.


trevorriles:
@chriskuehl you mentioned that there is some changes that have made it to develop but not been released yet. Would it be better for us to target a release branch?

.github/workflows/python-package.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
```
chriskuehl: Looks like we duplicated some of the Python 3.7 dropping work; I made similar changes in #823 (isolated in the first commit).

Maybe it makes sense for us to pull that into a separate PR? TODO for me.

Dockerfile
```diff
@@ -12,6 +12,7 @@ WORKDIR /src
 ENV CASS_DRIVER_NO_EXTENSIONS theytaketoolongtobuild
 
 COPY requirements*.txt ./
+RUN apt-get -y update && apt-get -y install librdkafka-dev && rm -rf /var/lib/apt/lists/*
```
chriskuehl: Do you recall what package needed this? I wonder if there's a possibility of choosing a version that has manylinux wheels uploaded to avoid needing to compile it ourselves at install time.

baseplate/clients/requests.py
```diff
@@ -26,6 +27,9 @@
 from baseplate.lib.prometheus_metrics import getHTTPSuccessLabel
 
 
+RequestsInstrumentor().instrument()
```
chriskuehl, line 29: It looks like the way these work is essentially to patch these libraries to inject tracing logic. I don't think that's inherently a problem but it's definitely a little sneaky, especially to do as an import-time side effect.

In general I think it's best if importing code doesn't have any side effects. This matches people's intuition about what importing does, reduces weird surprises (e.g. adding a seemingly-unrelated import suddenly changes behavior), is faster, and plays better with linters and any other tools that need to import code. It also makes it possible to access any configuration the app provides so that you can adjust the instrumentation settings if needed (e.g. if we wanted to add a way to toggle off instrumentation, or make it possible to exempt certain routes).

Do you think there's a way we can move this out of the import path and instead have it happen at startup time?

baseplate/clients/thrift.py
```diff
@@ -191,6 +201,15 @@ def __init__(
...
         self.namespace = namespace
         self.retry_policy = retry_policy or RetryPolicy.new(attempts=1)
+        self.tracer = trace.get_tracer(__name__)
+
+        self.otel_peer_name = None
+        self.otel_peer_ip = None
+        try:
+            self.otel_peer_name = socket.getfqdn()
+            self.otel_peer_ip = socket.gethostbyname(self.otel_peer_name)
+        except Exception:
```
chriskuehl:
```suggestion
        except socket.gaierror:
```

baseplate/clients/thrift.py
```diff
@@ -223,119 +260,150 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
+                                otelspan.set_status(status.Status(status.StatusCode.ERROR))
+                            else:
+                                span.finish()
+                            raise
+                        except TException:
+                            # this is an expected exception, as defined in the IDL
+                            otelspan.set_status(status.Status(status.StatusCode.OK))
+                            span.finish()
+                            raise
+                        except:  # noqa: E722
```
chriskuehl:
```suggestion
                        except BaseException:
```

Explicitly writing `except BaseException:` is the way to avoid the linter error when that's your goal.

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -35,10 +37,12 @@
 from baseplate.thrift.ttypes import IsHealthyProbe
 
 
+PyramidInstrumentor().instrument()
```
chriskuehl, line 41: Same question as before about instrumenting as an import-time side effect.

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -64,10 +68,15 @@ def __next__(self) -> bytes:
...
             return next(self.app_iter)
         except StopIteration:
-            self.span.finish()
+            trace.get_current_span().set_status(trace.status.StatusCode.OK)
+            if self.span:
+                self.span.finish()
             raise
-        except:  # noqa: E722
-            self.span.finish(exc_info=sys.exc_info())
+        except Exception as e:  # noqa: E722
```
chriskuehl: This is a behavior change (bare except catches all exceptions, `except Exception` catches only exceptions that inherit from Exception. So, this no longer catches exceptions like `KeyboardInterrupt`. Is that OK? If you want to keep the old behavior, you can use `except BaseException as e` instead, which is equivalent to the bare except.

You should be able to drop the `noqa` here now that you're not using a bare except.

baseplate/frameworks/pyramid/__init__.py
```diff
@@ -129,15 +138,21 @@ def baseplate_tween(request: Request) -> Response:
             response = handler(request)
             if request.span:
                 request.span.set_tag("http.response_length", response.content_length)
-        except:  # noqa: E722
+        except Exception as e:  # noqa: E722
```
chriskuehl: Ditto

baseplate/frameworks/thrift/__init__.py
```diff
@@ -80,87 +124,149 @@ def call_with_context(*args: Any, **kwargs: Any) -> Any:
...
+                        logger.debug("Processing TException. [exc=%s]" % exc)
+                        span.set_tag("exception_type", type(exc).__name__)
+                        span.set_tag("success", "false")
+
+                        # this is an expected exception, as defined in the IDL
+                        span.finish()
+                        # Set as OK as this is an expected exception
+                        otelspan.set_status(trace.status.Status(trace.status.StatusCode.OK))
+                        raise
+                    except Exception as exc:  # noqa: E722
```
chriskuehl: Probably can drop noqa, same comments as before.

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class RedditB3Format(TextMapPropagator):
+    """Propagator for the Reddit B3 HTTP header format."""
+
+    TRACE_ID_KEY = "X-Trace"
+    SPAN_ID_KEY = "X-Span"
+    SAMPLED_KEY = "X-Sampled"
+    FLAGS_KEY = "X-Flags"
+    _SAMPLE_PROPAGATE_VALUES = {"1", "True", "true", "d"}
```
chriskuehl:
```suggestion
    _SAMPLE_PROPAGATE_VALUES = frozenset(("1", "True", "true", "d"))
```

baseplate/lib/propagator_redditb3.py
```diff
@@ -0,0 +1,158 @@
...
+    # Although Reddit B3 uses 64bit TraceId's (16 char) we will accept 32 and truncate only when writing client headers.
+    _trace_id_regex = re_compile(r"[\da-fA-F]{16}|[\da-fA-F]{32}")
```
chriskuehl: Since there's no anchor at the end, I think it will always just select the first 16 chars even for a longer ID?

```python
>>> re.match(r"[\da-fA-F]{16}|[\da-fA-F]{32}", "aaaabbbbccccddddaaaabbbbccccdddd").group(0)
'aaaabbbbccccdddd'
```

It also allows other character lengths as long as they're at least 16. Not sure exactly what the intention of the regex is here but given that the `|[\da-fA-F]{32}` part isn't doing anything, this seems unexpected?

chriskuehl (commented):
First pass review -- it would be a good idea to have someone more familiar with the tracing bits also take a look.

.github/workflows/python-package.yaml
chriskuehl: Extracted to https://github.com/reddit/baseplate.py/pull/824

Dockerfile
Xaelias: I am not 100% sure anymore but I think it had smth to do with the apple silicon MBPs.

chriskuehl: Ah, interesting. I actually changed it to use x86_64 emulation in my PR (https://github.com/reddit/baseplate.py/pull/823/commits/97ff89d5bca69405e016d799c742fcb5ec697c6a) since there was no Thrift compiler image for ARM and it was failing on my MacBook trying to build a native arch version of the image. This is probably fine to keep as-is for now, though I'm surprised you didn't run into the same Thrift issue I did (maybe I'm doing something wrong?).

Xaelias: ```DOCKER_DEFAULT_PLATFORM=linux/amd64 docker-compose build```
worked locally for me

Xaelias: so two different paths to achieve somewhat similar results :-D

baseplate/clients/requests.py
Xaelias: Honestly I don't know. I think it would make testing much more complex since you'd need to basically load the entire app to test anything.
It's also how otel says you should be using these auto-instrumenter.

KTAtkinson: I agree with Chris that this smells a little funny. I don't see anything in their API docs that suggests that this needs to be instrumented the top level like this. Can you share the link to the instructions/example you are following?

KTAtkinson (changes requested):

baseplate/clients/requests.py
Xaelias: Can you check the commit I just pushed? Is that closer to what y'all had in mind? (we change names etc. I just wanted to get smth out there for you to be able to look at)

trevorriles:
I've got develop merged in and all the tests passing. I'll begin addressing the other PR comments next.

One thing to note is that opentelemetry-python doesn't officially support 3.12 yet, so we may have to wait for that or restrict baseplate to 3.11 for the time being.

baseplate/lib/propagator_redditb3.py
trevorriles: I think your correct here, I lifted this from upstream here and modified it for Reddit's format.
https://github.com/open-telemetry/opentelemetry-python/blob/d6321d627cd0fc9345a689a9a37cf4cb9ea44ba0/propagator/opentelemetry-propagator-b3/src/opentelemetry/propagators/b3/__init__.py#L46-L47

Let me dig in a bit more to make sure we get the best behavior.

baseplate/frameworks/pyramid/__init__.py
trevorriles: In this case I think we are ok using `Exception`. The opentelemetry lib want's an `Exception` for it's `record_exception` function. As in all these cases we just end the span and re-raise, I think it's fine if we miss ending the span for exceptions that inherit from `BaseException` but not `Exception`.

baseplate/lib/propagator_redditb3.py
trevorriles: Ahh it uses the [re.fullmatch](https://docs.python.org/3/library/re.html#re.fullmatch) function, which if I read it correctly means it will check the second option.

trevorriles: Either way this is incorrect as the baseplate spec uses base10 encoded integers... I'll get a fix together.

baseplate/clients/requests.py
trevorriles: I think @Xaelias 's change works for me what are your thoughts @KTAtkinson & @chriskuehl ?

baseplate/lib/propagator_redditb3.py
trevorriles: Got the fix in to look for base10 integers wrapped in a string.

trevorriles:
@chriskuehl I'm curious what your thoughts are on the lack of 3.12 support at the moment?

https://github.com/open-telemetry/opentelemetry-python/issues/3617

chriskuehl:
> @chriskuehl I'm curious what your thoughts are on the lack of 3.12 support at the moment?

I don't think we can merge this until 3.12 support is available, since we already have services which have migrated to 3.12.

KTAtkinson (approved):

Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY --from=thrift /usr/local/bin/thrift /usr/local/bin/thrift
 WORKDIR /src
 
 COPY requirements*.txt ./
+RUN apt-get -y update && apt-get -y install librdkafka-dev && rm -rf /var/lib/apt/lists/*
```
chriskuehl: Can we move this new apt line above the COPY so that we don't need to reinstall apt packages every time a requirement changes? Maybe it can go directly under the `FROM python:3.12` line?

baseplate/clients/thrift.py
```diff
@@ -223,119 +264,150 @@ def _call_thrift_method(self: Any, *args: Any, **kwargs: Any) -> Any:
...
+                    mutable_metadata: OrderedDict = OrderedDict()
+
+                    pool_addr = self.pool.endpoint.address
+                    if isinstance(pool_addr, str):
+                        otel_attributes[SpanAttributes.NET_PEER_IP] = pool_addr
+                    elif pool_addr is not None:
+                        otel_attributes[SpanAttributes.NET_PEER_IP] = pool_addr.host
+                        otel_attributes[SpanAttributes.NET_PEER_PORT] = pool_addr.port
+                    if otel_attributes.get(SpanAttributes.NET_PEER_IP) in ["127.0.0.1", "::1"]:
+                        otel_attributes[SpanAttributes.NET_PEER_NAME] = "localhost"
```
chriskuehl, line 276: When does localhost traffic like this happen? For the IPv4 address I guess in theory it could be all of 127.0.0.0/8 depending on how this is used? Mostly wondering because it's not uncommon to see e.g. `127.0.1.1` used in default Linux setups in `/etc/hosts`.

baseplate/lib/propagator_redditb3_http.py
```diff
@@ -0,0 +1,157 @@
...
+class RedditB3HTTPFormat(TextMapPropagator):
+    """Propagator for the Reddit B3 HTTP header format."""
+
+    TRACE_ID_KEY = "X-Trace"
+    SPAN_ID_KEY = "X-Span"
+    SAMPLED_KEY = "X-Sampled"
+    FLAGS_KEY = "X-Flags"
+    _SAMPLE_PROPAGATE_VALUES = frozenset({"1", "True", "true", "d"})
+    # Reddit B3 trace and span id's are 64bit integers encoded as decimal
+    _id_regex = re_compile(r"[\d]+")
```
chriskuehl:
```suggestion
    _id_regex = re_compile(r"\d+")
```

baseplate/lib/propagator_redditb3_thrift.py
```diff
@@ -0,0 +1,157 @@
...
+class RedditB3ThriftFormat(TextMapPropagator):
+    """Propagator for the Reddit B3 Thrift header format."""
+
+    TRACE_ID_KEY = "Trace"
+    SPAN_ID_KEY = "Span"
+    SAMPLED_KEY = "Sampled"
+    FLAGS_KEY = "Flags"
+    _SAMPLE_PROPAGATE_VALUES = frozenset({"1", "True", "true", "d"})
+    # Reddit B3 trace and span id's are 64bit integers encoded as decimal
+    _id_regex = re_compile(r"[\d]+")
```
chriskuehl:
```suggestion
    _id_regex = re_compile(r"\d+")
```

baseplate/server/__init__.py
```diff
@@ -190,6 +214,95 @@ def configure_logging(config: Configuration, debug: bool) -> None:
...
+                    "sample_rate not specified in tracing config, default to no sampling."
+                )
+            # Load insecure option from tracing config.
+            insecure: bool
+            if "insecure" in config.tracing:
+                if config.tracing["insecure"] == "false":
+                    insecure = False
+            else:
+                # Default to true. At the time of implementation our current tracing pipeline is not setup for TLS.
+                insecure = True
```
chriskuehl:
```suggestion
            if config.tracing.get("insecure") == "false":
                insecure = False
            else:
                # Default to true. At the time of implementation our current tracing pipeline is not setup for TLS.
                insecure = True
```

tests/integration/otel_thrift_tests.py
```diff
@@ -0,0 +1,963 @@
...
+    def reset_metrics(self, metrics):
+        if not metrics:
+            return
+
+        try:
+            metrics.get_active_requests_metric().clear()
+            metrics.get_latency_seconds_metric().clear()
+            metrics.get_requests_total_metric().clear()
+        except Exception:
+            pass
```
chriskuehl: How does this exception happen? Is it possible to catch a specific error?

chriskuehl (approved):
This looks reasonable but given the size it's difficult to catch any bugs during review. I think we should discuss how we want to test this before the final release.

Let's also get a review from someone more familiar with our tracing setup before merging, since I'm not really qualified to review that beyond basic logic checking.

Dockerfile
trevorriles: Good call, I'll move this up.

baseplate/clients/thrift.py
trevorriles: Honestly I think the effort here is to just provide a default `NET_PEER_NAME` for some common scenarios. It's kind of a convenience if you are filtering or grouping by hostnames I guess so that you get a value instead of empty.

tests/integration/otel_thrift_tests.py
trevorriles: I'll look into the exceptions thrown here and clean this up.

trevorriles:
> This looks reasonable but given the size it's difficult to catch any bugs during review. I think we should discuss how we want to test this before the final release.

Fully on board here, here are two testing steps I want to take before we cut an official release:
- [ ] Additional testing in snoodev to verify context propagation general functionality
- [ ] Cut some form of alpha release and work with teams that are still < 3.12. This will make it easier for them to test in snoodev and possibly even some canary testing.

With the 3.12 limitation I'm interested in your thoughts around maintaining an otel branch/release that will be limited to 3.11 and below until we can get the upstream issues resolved?

> Let's also get a review from someone more familiar with our tracing setup before merging, since I'm not really qualified to review that beyond basic logic checking.

I'm sure we can get @Xaelias to give this another pass, since I've taken this over. Not sure if you can approve your own PR though. Ultimately we are the two people who have context to the tracing work here. Pete Loaf has done most of the golang work, but I'm not sure if it makes sense to have him review here as well or not?

baseplate/server/__init__.py
```diff
@@ -190,6 +214,94 @@ def configure_logging(config: Configuration, debug: bool) -> None:
...
+            logger.info("Tracing Configured")
+        else:
+            logger.warning(
+                "Tracing disabled: service_name and endpoint are required fields in the tracing config."
+            )
+
+
+def otel_tracing_autoinstrumentation() -> None:
+    PyramidInstrumentor().instrument()
+    RequestsInstrumentor().instrument()
```
trevorriles: I need to look at moving these back.

This created an infinite loop somehow related to the import timing of the `ssl` lib and our gevent `patch_all()` call in `baseplate-serve`. 

Honestly I don't 100% know why yet, still digging.

trevorriles: Or an alternative that is not this 😓

tests/integration/otel_thrift_tests.py
trevorriles: This method wasn't called so I just removed it.

I also removed it from the `thrift_tests.py` file as well as that's where we lifted most of these tests.

trevorriles (approved):

baseplate/server/__init__.py
```diff
@@ -190,6 +206,54 @@ def configure_logging(config: Configuration, debug: bool) -> None:
...
+
+def configure_tracing() -> None:
+    logger.info("Entering configure tracing function")
+    sample_rps = 10
+    try:
+        sample_rps = DefaultFromEnv(Integer, "BASEPLATE_TRACE_SAMPLER_RPS", 10)
+    except ValueError:
+        logger.warning(
+            "Invalid BASEPLATE_TRACE_SAMPLER_RPS, falling back to the default of 10 RPS."
+        )
```
trevorriles: @chriskuehl I'm trying to use `DefaultFromEnv` as suggested but I think I'm misunderstanding how this is supposed to work. Does this also require implementing these options in the ini based config to get it to work?

baseplate/frameworks/thrift/__init__.py
```diff
@@ -64,11 +83,40 @@ def __init__(
...
+        if headers:
+            header_dict = {}
+            for k, v in headers.items():
+                try:
+                    header_dict[k.decode()] = v.decode()
+                except UnicodeDecodeError:
+                    self.logger.info("Unable to decode header %s, ignoring." % k.decode())
+
+            ctx = propagator.extract(header_dict)
+            logger.debug("Extracted trace headers. [ctx=%s, header_dict=%s]" % (ctx, header_dict))
```
redloaf: 🔕 I believe this formats the log message even if debug logging is disabled.
```suggestion
            logger.debug("Extracted trace headers. [ctx=%s, header_dict=%s]", ctx, header_dict)
```

redloaf (approved):

PR #820, opened 2023-12-29
allow for empty X-Parent

carldunham:
also tighten up optionality of params

## 💸 TL;DR
The `X-Parent` header could be empty, and will be for the root span. This PR allows for that.

## ✅ Checks

### DIFF
--- a/tests/unit/core_tests.py
+++ b/tests/unit/core_tests.py
@@ -348,9 +348,7 @@ class TraceInfoTests(unittest.TestCase):
         self.assertIsNone(span.flags)
 
     def test_from_upstream_handles_no_parent(self):
-        """
-        Verify that `parent_id` can be left out without raising an exception.
-        """
+        """Verify that `parent_id` can be left out without raising an exception."""
         span = TraceInfo.from_upstream(1, None, 3, None, None)
         self.assertIsNone(span.parent_id)
         self.assertIsNotNone(span.trace_id)
--- a/tests/unit/core_tests.py
+++ b/tests/unit/core_tests.py
@@ -348,5 +348,10 @@ class TraceInfoTests(unittest.TestCase):
         self.assertIsNone(span.flags)
 
     def test_from_upstream_handles_no_parent(self):
+        """
+        Verify that `parent_id` can be left out without raising an exception.
+        """
         span = TraceInfo.from_upstream(1, None, 3, None, None)
         self.assertIsNone(span.parent_id)
+        self.assertIsNotNone(span.trace_id)
+        self.assertIsNotNone(span.span_id)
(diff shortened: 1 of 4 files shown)


KTAtkinson (changes requested):
Thanks for putting this together! It looks good. Can you describe the impact this issue is having?

tests/unit/core_tests.py
```diff
@@ -341,3 +346,7 @@ def test_from_upstream_handles_no_sampled_or_flags(self):
         span = TraceInfo.from_upstream(1, 2, 3, None, None)
         self.assertIsNone(span.sampled)
         self.assertIsNone(span.flags)
+
+    def test_from_upstream_handles_no_parent(self):
+        span = TraceInfo.from_upstream(1, None, 3, None, None)
+        self.assertIsNone(span.parent_id)
```
chriskuehl: Just a suggestion: it might be nice to include a bit of context here with a brief comment.

In my experience this is the kind of test that people come back to later on and think "why are we testing this? can this even happen in real code? of course there's always a parent span because of [xyz reason that doesn't apply in your case]" and end up breaking things. Since this is testing a non-obvious but important scenario, you could add a line like

```python
def test_from_upstream_handles_no_parent(self):
    # Ensure that TraceInfo can be constructed with no parent span. This can happen when
    # the root span is created by this service.
    ...
```

(adjust to match the actual reason)

carldunham:
> Thanks for putting this together! It looks good. Can you describe the impact this issue is having?

You bet! So this is coming up where we have bp.go services and a bp.py service. We can see the `X-Trace` header being passed among the Go services, but the Python service seemed to be ignoring it.

In debugging the cause of this, it was noticed that the code would silently fail when no `X-Parent` header was found, which is a normal operating case. The impact was that none of the tracing headers would be honored.

So this PR aligns what is required and what is optional, so that it will work as expected.

KTAtkinson (approved):

tests/unit/core_tests.py
```diff
@@ -341,3 +346,12 @@ def test_from_upstream_handles_no_sampled_or_flags(self):
         span = TraceInfo.from_upstream(1, 2, 3, None, None)
         self.assertIsNone(span.sampled)
         self.assertIsNone(span.flags)
+
+    def test_from_upstream_handles_no_parent(self):
+        """
+        Verify that `parent_id` can be left out without raising an exception.
+        """
```
chriskuehl: nit (no need to fix): I would suggest following PEP257 when possible since it produces better output from pydoc (and other tools that parse docstrings): https://peps.python.org/pep-0257/#one-line-docstrings

chriskuehl (approved):

chriskuehl (approved):

PR #823, opened 2024-01-16
Support Python 3.10, 3.11, and 3.12

chriskuehl:
## 💸 TL;DR
This PR adds support for Python 3.10, 3.11, and 3.12. Support for 3.7 was dropped in #824.

### Dependency updates
You can see an audit of Baseplate's dependencies for known issues [here](https://docs.google.com/spreadsheets/d/1luXIkt9v1Iizl182aw_2odGo35U0WXxIECEMqMVs1eA/edit). As there is no good way to find incompatibilities automatically, this audit was conducted manually by looking at each dependency's issue tracker and changelog to find any known issues, and versions where support for 3.11/3.12 were first mentioned.

The minimum dependency version bumps here are:

* **`boto3` (old >=1.18, new >=1.28.27):**
  * ✅ No expected impact
  * `https://github.com/boto/boto3/pull/3825`
* **`cassandra-driver` (old >=3.13.0, new >=3.29.0):**
  * ✅ No expected impact
  * `https://github.com/datastax/python-driver/pull/1194`
* **`confluent-kafka` (old >=1.3,<2, new >=2.3.0 ⚠️):**
  * ✅ No expected impact
  * 2.x upgrade is unlikely to break things: "OpenSSL 3.0.x upgrade in librdkafka requires a major version bump, as some legacy ciphers need to be explicitly configured to continue working, but it is highly recommended NOT to use them. The rest of the API remains backward compatible."
  * https://github.com/confluentinc/confluent-kafka-python/blob/master/CHANGELOG.md
* **`gevent` (old >=20.5.0, new >=23.9.1 ⚠️):**
  * ✅ No expected impact
  * Versions are **date-based, not semver**.
  * Only breaking changes since 20.5.0 are to drop Python 3.7, so no impact is expected.
  * https://www.gevent.org/changelog.html
* **`kombu` (old >=4.0.0,<5, new >=5.3.3 ⚠️):**
  * ✅ No expected impact
  * 5.0.0's only breaking change is to drop support for Python 2: https://github.com/celery/kombu/blob/main/Changelog.rst
* **`objgraph` (old>=3.0,<4, new >=3.6.0):**
  * ✅ No expected impact
  * No breaking changes other than dropping support for old Python versions: https://github.com/mgedmin/objgraph/blob/master/CHANGES.rst
* **`pyramid` (old>=1.9.0,<2, new >=1.10.8,<2):**
  * ✅ No expected impact
  * Contains a fix for `imp` import, needed for 3.12: `https://github.com/Pylons/pyramid/pull/3659`
* **`python-json-logger` (old>=2.0,<3, new >=2.0.5,<3):**
  * ✅ No expected impact
* **`sentry-sdk` (old >=0.19,<2, new>=1.35.0,<2 ⚠️):**
  * ✅ No expected impact
  * Breaking change in 1.0.0 isn't used by any of our code
  * Breaking change in 0.20 doesn't look relevant
  * https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md
* **`sqlalchemy` (old >=1.0.0,<2, new>=1.4.49,<2):**
  * ✅ No expected impact
  * 1.4.49 has official support: https://www.sqlalchemy.org/blog/2023/07/05/sqlalchemy-1.4.49-released/

### Upgrade process for Baseplate services

The process for a Baseplate service to upgrade is:

1. Update the `FROM` line in the Dockerfile to point to the 3.12 Python image
2. Update the Docker image in `.drone.yml` (if present) to the 3.12 Python image
3. Update the Baseplate version in `requirements.txt`.
4. Now the hard part: we'll need to upgrade a lot of dependencies.

    The easiest way to do this is to build the Docker image, which will try to install the dependencies and surface any conflicts as errors for you to fix. Any conflicts will be caught by pip, like this:

        ERROR: Cannot install baseplate[pyramid,sentry,sql]==2.6.0a5.dev15+ge71e299 and pyramid==1.10.4 because these package versions have conflicting dependencies.
        
        The conflict is caused by:
            The user requested pyramid==1.10.4
            baseplate[pyramid,sentry,sql] 2.6.0a5.dev15+ge71e299 depends on pyramid<2.0 and >=1.10.8; extra == "pyramid"

    This conflict occurs because we updated the minimum pyramid version to 1.10.8 in this PR in order to be compatible with Python 3.12. To fix it, upgrade pyramid in `requirements.txt`, then try building the image again. It may take several iterations to satisfy all requirements, but rest assured that pip is ensuring you won't miss any as it checks for version conflicts.

5. Double-check your project's dependencies against the [list of known dependency issues with 3.11 and 3.12](https://docs.google.com/spreadsheets/d/1luXIkt9v1Iizl182aw_2odGo35U0WXxIECEMqMVs1eA/edit). The previous step caught any issues with dependencies in conflict with Baseplate, but there may be other packages you need to upgrade which aren't actual dependencies of Baseplate.

6. You _should_ now be good to go. Now is a good time to run automated and manual tests to ensure everything is working.

### Why support 3.12?

The initial goal was to support just 3.11, but we'd like to also support 3.12 if it's not too much extra work so that users can upgrade directly to it. This would get consuming services on a supported Python version through late 2028 for little extra effort.

* [x] CI passes
* [x] Tested manually with services in the local integration environment

## ✅ Checks


baseplate/lib/thrift_pool.py
```diff
@@ -116,7 +116,7 @@ def thrift_pool_from_config(
     if options.max_connection_attempts is not None:
         kwargs.setdefault("max_connection_attempts", options.max_connection_attempts)
     if options.max_retries is not None:
-        raise Exception("max_retries was renamed to max_connection_attempts")
+        raise ValueError("max_retries was renamed to max_connection_attempts")
```
KTAtkinson, line 119: Thanks for choosing a more appropriate error class here.

baseplate/sidecars/trace_publisher.py
```diff
@@ -97,11 +97,11 @@ def publish(self, payload: SerializedBatch) -> None:
                 response.raise_for_status()
             except requests.HTTPError as exc:
                 self.metrics.counter("error.http").increment()
-                response = getattr(exc, "response", None)
-                if response is not None:
-                    logger.exception("HTTP Request failed. Error: %s", response.text)
+                exc_response = getattr(exc, "response", None)
```
KTAtkinson, line 100: :no_bell: what's the purpose of the rename here?

KTAtkinson (approved):

baseplate/sidecars/trace_publisher.py
chriskuehl: The new mypy version didn't like that `response`'s type changes throughout this function. It was originally declared with type `Response` from the `response = self.session.post(...)` a few lines above, but the expression on this line has type `Response | None` which is incompatible.

Using a separate variable seemed like a reasonable fix.

baseplate/clients/sqlalchemy.py
```diff
@@ -343,7 +347,7 @@ class SQLAlchemySessionContextFactory(SQLAlchemyEngineContextFactory):
     """
 
     def make_object_for_context(self, name: str, span: Span) -> Session:
-        engine = super().make_object_for_context(name, span)
+        engine = typing.cast(Engine, super().make_object_for_context(name, span))
```
krvss, line 350: 🔕 In theory, `super().make_object_for_context()` can also legitimately return a `Session` object. Will not break anything now, but potentially can in the future.

baseplate/server/__init__.py
```diff
@@ -226,7 +227,7 @@ def make_app(app_config: Dict[str, str]) -> Callable:
 
 
 def register_signal_handlers() -> threading.Event:
-    def _handle_debug_signal(_signo: int, frame: FrameType) -> None:
+    def _handle_debug_signal(_signo: int, frame: FrameType | None) -> None:
```
krvss, line 230: 🔕 Why not `Optional[FrameType]`? Just curious.

docker-compose.yml
```diff
@@ -3,6 +3,8 @@
 version: "3"
 services:
   baseplate:
+    # Needed as the Thrift image isn't available on ARM.
+    platform: linux/x86_64
```
krvss, line 7: 🔕 Maybe add "TODO" here? Also, I see we mostly use `amd64` in such cases

krvss (approved):
A few nits, otherwise looks fine.

MelissaCole (approved):

baseplate/server/__init__.py
chriskuehl: I personally prefer the semantics of `Optional[X]` over `X | None` (I think it suggests intent more clearly) but my impression is that the the latter has "won" the general consensus and is considered more idiomatic, so I always use the `X | None` approach nowadays. For example, [pyupgrade](https://github.com/asottile/pyupgrade) automatically converts type annotations to it. I tried to find a more official source but opinions are kind of mixed: https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945

Happy to change it back if you prefer. They should be functionally equivalent.

.github/workflows/python-package.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
```
kylelemons, line 15: :tada:

kylelemons (approved):

baseplate/clients/sqlalchemy.py
chriskuehl: It was a little tricky to unwind the `SQLAlchemySessionContextFactory` and `SQLAlchemyEngineContextFactory` relationship. mypy didn't like that `SQLAlchemySessionContextFactory.make_object_for_context()` could return a `Session` since that's incompatible with the parent class returning only an `Engine`.

Initially I tried to fix this by splitting out everything in `SQLAlchemyEngineContextFactory` into a new base class that both could inherit from and implement only `make_object_for_context`. This fixed the typing issues but it was a bit awkward with how `SQLAlchemySessionContextFactory` wants to call `SQLAlchemyEngineContextFactory.make_object_for_context()`, which is why I landed on this current approach (amend the parent class to `Engine | Session` even though it always returns an `Engine` and cast it here to make mypy happy).

That said... as far as I know in this case it always will be an `Engine` returned in this scenario, which is why the cast here is (I think) correct? There's no functionality change here, just typing changes.

carldunham (approved):

PR #824, opened 2024-01-19
Drop python3.7 support

chriskuehl:
## 💸 TL;DR
Python 3.7 is end-of-life, so we are dropping support. There are two PRs (#819, #823) which independently made similar changes, so I'm extracting them out here.

No special testing.

## ✅ Checks


docs/pyproject.toml
```diff
@@ -5,4 +5,4 @@ line-length = 74
 
 # no need for legacy stuff here in the docs! always use the latest supported
```
kylelemons: 🔕 this says "latest supported", so either this should be py39 or we actually mean "oldest supported" in which case update the doc comment too

kylelemons (approved):

docs/pyproject.toml
trevorriles: This should probably be the case in `pyproject.yaml`, `setup.cfg`, and `setup.py` as well.

chriskuehl: I think we only want to do it here because this is the black config specifically for code snippets in the docs where we want to use the latest-supported syntax (I think that's what the comment is referencing).

If we set it in `pyproject.toml` to the latest, black could in theory introduce syntax only compatible with the newest Python we support and not the oldest Python. The same goes for mypy; it will tell you to migrate to new type definitions that don't exist on older Python versions.

PR #826, opened 2024-02-07
SQLA: Invalidate sessions to prevent connections leak.

salomon-smekecohen:
## 💸 TL;DR
Sessions can be re-connected by issuing a statement via a reference. This should rarely be an issue given on_finish's place in the server lifecycle.

However, why have the possibility of leaking connections? Especially relevant in how some users of Baseplate have implemented Celery integrations.

## 📜 Details
Documentation here https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.invalidate

This is a very simple change. However we have ran it on reddit-service-campaign-management for a while without issues.

## ✅ Checks

### DIFF
--- a/baseplate/clients/sqlalchemy.py
+++ b/baseplate/clients/sqlalchemy.py
@@ -360,4 +360,4 @@ class SQLAlchemySessionSpanObserver(SpanObserver):
         self.session = session
 
     def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
-        self.session.close()
+        self.session.invalidate()


baseplate/clients/sqlalchemy.py
```diff
@@ -360,4 +360,4 @@ def __init__(self, session: Session):
         self.session = session
 
     def on_finish(self, exc_info: Optional[_ExcInfo]) -> None:
-        self.session.close()
+        self.session.invalidate()
```
chriskuehl, line 363: If I'm understanding the documentation correctly, it seems like the main difference is that `close()` will allow the underlying database connection to remain, while `invalidate()` forces that connection to be closed. Does this mean connections can no longer be re-used at all? Is that desired and could it have repercussions (like reduced performance, more load on the database, etc.)?

https://docs.sqlalchemy.org/en/20/core/pooling.html#more-on-invalidation

salomon-smekecohen: That is exactly what it does! Leaving these connections as `.close` means leftover references to the session can interact with it and then never close it after. Ex:

```
session.close()
session.lock(table)
# now we have a locked table forever/until the server dies or the transaction gets dropped for idle_in_transaction.
```

Do we currently re-use connections? My understanding is that each span makes new ones. https://github.com/reddit/baseplate.py/blob/develop/baseplate/clients/sqlalchemy.py#L349

salomon-smekecohen:
lets table this until I can test it more. too scary.

PR #828, opened 2024-02-27
Vault CSI support v2

pnovotnak:
## 💸 TL;DR

This PR fixes Vault CSI support by reimplementing the CSI driver. The new driver is a bit simpler and has been tested against mocked CSI behavior (provided in test cases), as well as integration tested against a real Vault CSI driver.

## 📜 Details

This driver leverages this convenient tidbit from the Vault CSI code:

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/c697863c35d5431ec048b440d36550eb3ceb338f/pkg/util/fileutil/atomic_writer.go#L60-L62

By taking advantage of this behavior (modification time of the `..data` symlink) we simplify caching logic greatly by observing modification time of the symlink for cache entry invalidation. Furthermore, since the CSI driver initializes the volume before start, we have forgone the filewatcher on this symlink.

Secret files are [rewritten atomically using this algorithm](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/c697863c35d5431ec048b440d36550eb3ceb338f/pkg/util/fileutil/atomic_writer.go#L91-L134). The upshot is that if we don't resolve the `..data` symlink we can use it each time we read files to get the most recent version from disk.

In my testing environment I was observing 2 minutes between refreshes. I haven't hunted down the configuration for this yet but that appears to be the default.

This was tested with some print statements in a test environment where a Baseplate.py thrift service was serving testing traffic with Vault CSI running providing secrets. I monitored cache hits vs failures to ensure that the files weren't being reloaded more than expected.

<details><summary>Cache hits vs misses log</summary>

```
Feb 27 14:35:45 test-environment: INFO     Listening on ('0.0.0.0', 9090)
Feb 27 14:42:58 test-environment: cache miss: secret/example/secret-value-1@mtime=1709073701.164487, cache_entry=None, secret_data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}
Feb 27 14:42:58 test-environment: cache miss: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=None, secret_data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}
Feb 27 14:42:58 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:42:58 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:42:58 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:42:58 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:42:58 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-1@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-1@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-1@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:43:19 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073701.164487, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-1@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache miss: secret/example/secret-value-1@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}, updating=False), secret_data={'current': 'xxxxxxxxxxx==', 'encoding': 'base64', 'type': 'versioned'}
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache miss: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073701.164487, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False), secret_data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073820.1701584, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073820.1701584, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073820.1701584, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
Feb 27 14:44:10 test-environment: cache hit: secret/example/secret-value-2@mtime=1709073820.1701584, cache_entry=VaultCSIEntry(mtime=1709073820.1701584, data={'current': 'xxxxxxxx', 'encoding': 'base64', 'type': 'versioned'}, updating=False)
```

</summary>

## ✅ Checks


baseplate/lib/secrets.py
```diff
@@ -119,7 +122,9 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
     raise CorruptSecretError(path, f"unknown encoding: {encoding!r}")
 
 
-SecretParser = Callable[[Dict[str, Any], str], Dict[str, str]]
+class SecretParser(Protocol):
+    def __call__(self, data: Dict[str, Any], secret_path: Optional[str] = None) -> Dict[str, str]:
+        ...
```
pnovotnak, line 127: Second parameter is optional which can't be expressed with `Callable`

baseplate/lib/secrets.py
```diff
@@ -509,8 +595,9 @@ def secrets_store_from_config(
         backoff = None
 
     if options.provider == "vault_csi":
-        parser = parse_vault_csi
-        return DirectorySecretsStore(options.path, parser, timeout=timeout, backoff=backoff)
+        return VaultCSISecretsStore(
+            options.path, parser=parse_vault_csi, timeout=timeout, backoff=backoff
+        )
```
pnovotnak: @chriskuehl We are a little torn here WRT backward incompatibility. WDYT about removing the old implementation straight away?

baseplate/lib/secrets.py
```diff
@@ -119,7 +120,9 @@ def _decode_secret(path: str, encoding: str, value: str) -> bytes:
     raise CorruptSecretError(path, f"unknown encoding: {encoding!r}")
 
 
-SecretParser = Callable[[Dict[str, Any], str], Dict[str, str]]
```
pnovotnak, line 122: Second parameter has a default which can't be expressed with `Callable`

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,88 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+    # A mutex to prevent possible duplicate work from event loops
+    updating: bool
+
+
+class VaultCSISecretsStore(SecretsStore):
+    """Access to secret tokens with automatic refresh when changed.
+
+    This local vault allows access to the secrets cached on disk given a path to
+    a directory. It will automatically reload the cache when it is changed. Do not
+    cache or store the values returned by this class's methods but rather get
```
KTAtkinson: :no_bell:  I would say, not caching or storing the values from secrets it's a good idea but also not a hard and fast rule. For example, if the secret never gets loaded from disk again due to the way a client system or library works, it may not be correct to say don't store the secret value. 

```suggestion
    a directory. It will automatically reload the cache when it was changed. Generally do not
    cache or store the values returned by this class's methods but rather get
```

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,88 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+    path: Path
+    data_symlink: Path
+    cache: Dict[str, VaultCSIEntry]
+
+    def __init__(
+        self,
+        path: str,
+        parser: SecretParser,
+        timeout: Optional[int] = None,
+        backoff: Optional[float] = None,
```
KTAtkinson: Is there any way we can eliminate these arguments, I don't think they have any meaning in this context.

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,88 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+            raise ValueError(
+                f"Expected {self.data_symlink} to be a directory. Verify {self.path} is the root of the Vault CSI mount."
+            )
+
+    def get_vault_url(self) -> str:
+        raise NotImplementedError
+
+    def get_vault_token(self) -> str:
+        raise NotImplementedError
+
```
KTAtkinson: Please provide a comment regarding why there are unimplemented. I'm primarily concerned that somwhere someone is using one of these functions and adopting this upgrade will cause them to start to see errors. I would like clear notation in code and in docs indicating what's going on:)

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,88 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+        return os.path.getmtime(self.data_symlink)
+
+    def _raw_secret(self, name: str) -> Any:
+        try:
+            with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                return self.parser(json.load(fp))
+        except FileNotFoundError:
+            # Try a second time in case the file was deleted while we tried to read it
+            with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                return self.parser(json.load(fp))
```
TylerLubeck: Do we want to wrap this with some `SecretNotFoundError`? I know we mostly discussed this as a race condition thing but there's also the case of folks trying to access secrets that they didn't mount, and I think that error could be more clear.

baseplate/lib/secrets.py
pnovotnak: Added docstrings. Worth noting that they are deprecated on the superclass, and the original vault implementation had them raising as well.

baseplate/lib/secrets.py
pnovotnak: Seems reasonable. Some information is lost, but the error might be a bit friendlier.

baseplate/lib/secrets.py
pnovotnak: This was copypasta'd from the original implementation. I made this change as well as a few others

KTAtkinson (approved):
Thanks for getting this together!

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,91 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+        try:
+            with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                return self.parser(json.load(fp))
+        except FileNotFoundError:
+            # Try a second time in case the file was deleted while we tried to read it
+            try:
+                with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                    return self.parser(json.load(fp))
+            except FileNotFoundError:
+                raise SecretNotFoundError(name)
```
TylerLubeck:
```suggestion
            except FileNotFoundError as exc:
                raise SecretNotFoundError(name) from exc
```

This'll let us have a nice exception and a useful stack trace

TylerLubeck (approved):

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,91 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+        """Deprecated and will be removed in 3.0.0"""
+        raise NotImplementedError
+
+    def get_vault_token(self) -> str:
+        """Deprecated and will be removed in 3.0.0"""
+        raise NotImplementedError
+
+    def _get_mtime(self) -> float:
+        """Modification time is store-wide for CSI secrets."""
+        return os.path.getmtime(self.data_symlink)
```
kylelemons, line 434: are we sure the vault mounts in kubernetes have mtime enabled?

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,91 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+
+    def _raw_secret(self, name: str) -> Any:
+        try:
+            with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                return self.parser(json.load(fp))
+        except FileNotFoundError:
+            # Try a second time in case the file was deleted while we tried to read it
+            try:
+                with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                    return self.parser(json.load(fp))
```
kylelemons: I think you need to explain more why you think this is necessary and why it's safe and why only one retry would be sufficient.

Maybe something like
```
# Every few minutes, the Vault CSI will create a new directory and swap over to it.
# There is a race condition where we could resolve the symlink and then the file could be deleted.
# Given the timeframe between such deletions, a single retry is sufficient to double-check
# whether the FileNotFoundError was due to this race or an actually-nonexistent path.
```

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,91 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+        mtime = self._get_mtime()
+        if cache_entry := self.cache.get(secret_path):
+            if cache_entry.mtime == mtime or cache_entry.updating is True:
+                return cache_entry.data, mtime
+            # Prevent multiple requests from updating the cache at the same time by setting the
+            # `updating` flag. Unfortunately we can only prevent this when the cache currently
+            # has a valid entry.
+            self.cache[secret_path] = VaultCSIEntry(
+                mtime=cache_entry.mtime, data=cache_entry.data, updating=True
+            )
```
kylelemons: Is there any harm in double-updating the cache?  It should be idempotent, and if it's not, 'first one wins' is not going to lead to better results.

I don't think the complexity of this is worth it.

docs/pyproject.toml
```diff
@@ -5,4 +5,4 @@ line-length = 74
 
 # Always use our latest supported version here since we want code snippets in
 # docs to use the most up-to-date syntax.
-target-version = ['py312']
+target-version = ['py39']
```
kylelemons: ❗ bad merge

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,205 @@
...
+        self.csi_dir = new_fake_csi(EXAMPLE_SECRETS_DATA)
+
+    def tearDown(self):
+        shutil.rmtree(self.csi_dir)
+
+    def test_can_load_credential_secret(self):
+        secrets_store = get_secrets_store(str(self.csi_dir))
+        data = secrets_store.get_credentials("secret/example-service/example-secret")
+        assert data.username == "reddit"
+        assert data.password == "password"
```
kylelemons, line 174: can you test all of the secret types?

(does Python not automatically handle un-base64-ing secrets that have `encoding: base64`?)

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,205 @@
...
+        secrets_store = get_secrets_store(str(self.csi_dir))
+        data = secrets_store.get_credentials("secret/example-service/example-secret")
+        gevent.sleep(0.1)  # prevent gevent from making execution out-of-order
+        assert data.username == "reddit"
+        assert data.password == "password"
+        simulate_secret_update(self.csi_dir)
+        assert original_data_path != self.csi_dir.joinpath("..data").resolve()
+        data = secrets_store.get_credentials("secret/example-service/example-secret")
+        assert data.username == "reddit"
+        assert data.password == "password"
```
kylelemons: It's important to validate _more than one_ secret update -- historically, the main failure mode of our naive implementations that watch the file would work correctly for one update but will fail to notice the second.

kylelemons:
Also, meta question, did we validate this in snoodev with the real Vault CSI yet?

baseplate/lib/secrets.py
pnovotnak: Yes, this was tested in snoodev and the link has an mtime

baseplate/lib/secrets.py
pnovotnak: The only real reason in my mind is to prevent multiple disk hits. Since the cache will expire every 2 minutes or so, I was concerned that high traffic services might see relatively frequent IO / latency spikes. I'd also argue that the absolute complexity increase is not very great–a single extra line of code and an extra boolean in the cache entry tuple.

Does any of that change your mind?

pnovotnak:
Yes! Tested pretty extensively in snoodev

docs/pyproject.toml
pnovotnak: Yes but also our black version doesn't support 3.12 and the formatter / linter doesn't run locally anymore:

```
❯ docker-compose run baseplate make fmt
[+] Creating 5/0
 ✔ Container baseplatepy-memcached-1           Running                                                                                                                                0.0s
 ✔ Container baseplatepy-zookeeper-1           Running                                                                                                                                0.0s
 ✔ Container baseplatepy-redis-cluster-node-1  Running                                                                                                                                0.0s
 ✔ Container baseplatepy-redis-1               Running                                                                                                                                0.0s
 ✔ Container baseplatepy-cassandra-1           Running                                                                                                                                0.0s
...
black baseplate/ tests/
Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: Invalid value for '-t' / '--target-version': invalid choice: py312. (choose from py27, py33, py34, py35, py36, py37, py38, py39)
make: *** [Makefile:55: fmt] Error 2
```

pnovotnak: Yes, but also we need to upgrade black because it appears to not work with py312:

```
❯ docker-compose run baseplate make fmt
[+] Creating 6/6
 ✔ Container baseplatepy-cassandra-1                                                                                                                                 Created          0.1s
 ✔ Container baseplatepy-redis-1                                                                                                                                     Created          0.1s
 ✔ Container baseplatepy-zookeeper-1                                                                                                                                 Created          0.1s
 ✔ Container baseplatepy-redis-cluster-node-1                                                                                                                        Created          0.1s
 ✔ Container baseplatepy-memcached-1                                                                                                                                 Created          0.1s
 ! redis-cluster-node The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested                  0.0s
[+] Running 5/5
 ✔ Container baseplatepy-cassandra-1           Started                                                                                                                                0.3s
 ✔ Container baseplatepy-memcached-1           Started                                                                                                                                0.4s
 ✔ Container baseplatepy-zookeeper-1           Started                                                                                                                                0.4s
 ✔ Container baseplatepy-redis-cluster-node-1  Started                                                                                                                                0.4s
 ✔ Container baseplatepy-redis-1               Started                                                                                                                                0.3s
...
black baseplate/ tests/
Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: Invalid value for '-t' / '--target-version': invalid choice: py312. (choose from py27, py33, py34, py35, py36, py37, py38, py39)
make: *** [Makefile:55: fmt] Error 2
```

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: All of them are now tested. Looks to me like only 2 types support base64 decoding: Simple and versioned.

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: Done, also with secret value updates.

baseplate/lib/secrets.py
pnovotnak: PS If we decide to replace the current implementation, we can probably remove DirectorySecretsStore & associated tests.

tests/unit/lib/secrets/vault_csi_tests.py
kylelemons: Yep, that's correct.

baseplate/lib/secrets.py
kylelemons: The problem is the mutex.  Using a mutex _always_ increases complexity beyond what it appears on the page.

Now and into the future, we now have to be triply careful that we always unlock the mutex when we lock it.  Especially with exceptions that can result in invisible changes to control flow, that can turn out to be incredibly hard to ensure.

I think I'm actually talking myself _more_ into wanting to remove this, rather than less 😅 

If you want to keep it, I think you'll need to add some `finally` guards to reset `updating=False`, otherwise there is a chance that the cached value never expires and silently goes stale.

kylelemons (commented):
Left a comment about the mutex

baseplate/lib/secrets.py
pnovotnak: Removed!

kylelemons (approved):
I think this looks workable to me, but I am definitely neither a python nor baseplate expert so please also wait for the other reviewers.

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,87 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+        return os.path.getmtime(self.data_symlink)
+
+    def _raw_secret(self, name: str) -> Any:
+        try:
+            with open(self.data_symlink.joinpath(name), "r", encoding="UTF-8") as fp:
+                return self.parser(json.load(fp))
+        except FileNotFoundError:
+            # Every few minutes, the Vault CSI will create a new directory and swap over to it.
+            # There is a race condition where we could resolve the symlink and then the file could
+            # There is a race condition where we could resolve the symlink and then the file could
```
chriskuehl: repeated line

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,271 @@
...
+SecretType: typing_extensions.TypeAlias = typing.Dict[str, any]
+
+
+def write_secrets(secrets_data_path: Path, data: typing.Dict[str, SecretType]) -> None:
+    """Write secrets to the current data directory."""
+    for key, value in data.items():
+        secret_path = secrets_data_path.joinpath(key)
+        secret_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(secret_path, "w") as fp:
+            json.dump(value, fp)
```
chriskuehl: nit: another option if you prefer, either is OK with me:

```suggestion
        secret_path.write_text(json.dumps(value))
```

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,271 @@
...
+    # This path can be monitored for changes
+    # https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/c697863c35d5431ec048b440d36550eb3ceb338f/pkg/util/fileutil/atomic_writer.go#L60-L62
+    data_link = Path(csi_path, "..data")
+    if data_link.exists():
+        # Simulate atomic update
+        new_data_link = Path(csi_path, "..data-new")
+        new_data_link.symlink_to(data_path)
+        new_data_link.rename(data_link)
+    else:
+        data_link.symlink_to(data_path)
```
chriskuehl: Any reason not to always do this?

```suggestion
    # Simulate atomic update
    new_data_link = Path(csi_path, "..data-new")
    new_data_link.symlink_to(data_path)
    new_data_link.rename(data_link)
```

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,271 @@
...
+    human_path = Path(csi_path, "secret")
+    if not human_path.exists():
+        human_path.symlink_to(csi_path.joinpath("..data/secret"))
+
+
+def new_fake_csi(data: typing.Dict[str, SecretType]) -> Path:
+    """Creates a simulated CSI directory with data and symlinks.
+    Note that this would already be configured before the pod starts."""
+    csi_dir = Path(tempfile.mkdtemp())
```
chriskuehl, line 50: nit (optional): since we're using pytest as our test runner already, it might be nice to use pytest's `tmp_path` fixture rather than creating temporary directories manually: https://docs.pytest.org/en/latest/how-to/tmp_path.html

You wouldn't need to manually clean it up this way either, since pytest handles cleanup (by default it leaves the past couple test run outputs around which is helpful in case you want to inspect failures).

baseplate/lib/secrets.py
```diff
@@ -464,6 +467,87 @@ def _get_file_data(self, filename: str) -> Tuple[Dict, float]:
...
+            # be deleted. Given the timeframe between such deletions, a single retry is sufficient
+            # to double-check whether the FileNotFoundError was due to this race or an
+            # actually-nonexistent path.
```
chriskuehl: Can this actually happen? I would expect this to make an `open` syscall on the non-resolved symlink (e.g. `open("<csi_root>/..data/secret")` since `joinpath` doesn't resolve symlinks) which the kernel would handle resolving. Once you get an open file handle it shouldn't matter if it gets deleted afterwards, right?

I tried to roughly test this by running this in one terminal (the idea being to swap the symlink / delete the old secret hundreds of times per second):
```
while :; do
    mkdir -p a
    echo foo > a/secret
    ln -s a ..data-new
    mv -T ..data-new ..data
    rm -rf b
    mkdir -p b
    echo bar > b/secret
    ln -s b ..data-new
    mv -T ..data-new ..data
    rm -rf a
done
```

...and this in another:
```
python3.12 -c 'while True: print(open("..data/secret").read())'
```

and I can never reproduce any issues, even if I add an arbitrary delay between opening the file and reading its contents. (If I have it resolve the path first, there are issues frequently.)

tests/unit/lib/secrets/vault_csi_tests.py
```diff
@@ -0,0 +1,271 @@
...
+        simulate_secret_update(self.csi_dir)
+        assert original_data_path != self.csi_dir.joinpath("..data").resolve()
+        data = secrets_store.get_credentials("secret/bare-secret")
+        assert data.username == "reddit"
+        assert data.password == "password"
+
+    def test_secret_updated(self):
+        secrets_store = get_secrets_store(str(self.csi_dir))
+        data = secrets_store.get_credentials("secret/example-service/example-secret")
+        gevent.sleep(0.1)  # prevent gevent shenanigans
```
chriskuehl, line 218: What's the purpose of these sleeps?

chriskuehl (approved):
Looks reasonable to me, just one question about whether a race condition can actually happen.

baseplate/lib/secrets.py
chriskuehl: This would be transparent to users, right? Looking in Sourcegraph, I don't see any references to `DirectorySecretsStore` outside of this repo, so that seems reasonable to me.

pnovotnak: Internally we seem fine.

The issue is that theoretically there are open source users of baseplate.py that might see breakage. Is it ok to risk that?

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: TIL

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: I tried several versions of this but I couldn't get it to work!

baseplate/lib/secrets.py
pnovotnak: This was all theoretical. I never tried to produce the issue. Your test seems like it ought to have displayed this behavior, if it was something that could actually happen.

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: Tests were flaking if they weren't present. My theory is that gevent is passing control back to the tests. By sleeping, I ensure the IO is complete before the remainder of the test.

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: While I would like to apply this suggestion, I _think_ this would require overhauling the test suite pretty significantly. So I'll omit this change, but TIL about this feature

baseplate/lib/secrets.py
chriskuehl: I think it's OK to remove the old unused implementation. I did some cursory checks (searching GitHub for public users of baseplate) and also checking in with the rest of the team, and our general impression is that nobody is really using Baseplate.py outside of Reddit.

This isn't necessarily great open source hygiene, but I don't think it's worth increasing our maintenance burden for a benefit we're pretty sure isn't actually there. We should call out in the release notes that it's technically a breaking change though.

tests/unit/lib/secrets/vault_csi_tests.py
chriskuehl: Were the flakes locally or in CI? I've been running the tests a bunch locally and can't seem to reproduce any flakes. I'd like to try to dig into this if possible before we merge.

pnovotnak: I only saw it in CI

baseplate/lib/secrets.py
pnovotnak: Ok, I've ripped out the old implementation based on this discussion.

tests/unit/lib/secrets/vault_csi_tests.py
pnovotnak: Example: https://github.com/reddit/baseplate.py/actions/runs/8116056546/job/22185234411

chriskuehl: I looked into this a little bit and was able to replicate the issues frequently in GitHub Actions CI but never locally.

I added some debugging prints and noticed cases where the file mtime did not change but the file contents did, which caused the test failures:
Screenshot 2024-03-07 at 11 53 17 AM

Not sure if this is something weird with the CI environment (maybe like low-resolution timestamps in the filesystem or with the system clock?) or something with gevent I'm not understanding, but I think it's safe to go ahead and merge.

chriskuehl (approved):

PR #829, opened 2024-02-27
CSI Additions

TylerLubeck:
We spoke about dynamically changing the default `config.path` based on the `config.provider`. Unfortunately there does not seem to be a way to tell if the `config.path` is the _default_, or if somebody provided the same value as the default.

My only idea is that we do something akin to this:
```py
default = "**DEFAULT**"
path = config.Optional(config.String, default=default)
provider = config.Optional(config.String, default="vault")
if provider == "vault_csi":
    path = "/mnt/secrets"
else:
   path = "/var/local/secrets.json"
```

But I worry that we're gonna find something somewhere else that ends up trying to use `**DEFAULT**` as a filepath

### DIFF
--- a/baseplate/lib/secrets.py
+++ b/baseplate/lib/secrets.py
@@ -495,6 +495,8 @@ class VaultCSISecretsStore(SecretsStore):
         backoff: Optional[float] = None,
     ):  # pylint: disable=super-init-not-called
         self.path = Path(path)
+        if not self.path.is_dir():
+            raise ValueError(f"{self.path} must be a directory")
         self.cache = {}
         self.data_symlink = self.path.joinpath("..data")
 
--- a/baseplate/lib/secrets.py
+++ b/baseplate/lib/secrets.py
@@ -499,6 +499,7 @@ class VaultCSISecretsStore(SecretsStore):
             raise ValueError(f"{self.path} must be a directory")
         self.cache = {}
         self.data_symlink = self.path.joinpath("..data")
+        self.parser = parser
 
     def get_vault_url(self) -> str:
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


baseplate/lib/secrets.py
```diff
@@ -495,8 +495,11 @@ def __init__(
         backoff: Optional[float] = None,
     ):  # pylint: disable=super-init-not-called
         self.path = Path(path)
+        if not self.path.is_dir():
+            raise ValueError(f"{self.path} must be a directory")
```
pnovotnak: Can we say "root of the CSI volume?" I know it's a little verbose but I think it'll be a helpful pointer

baseplate/lib/secrets.py
```diff
@@ -495,8 +495,11 @@ def __init__(
...
         self.cache = {}
         self.data_symlink = self.path.joinpath("..data")
```
pnovotnak, line 500: Can we check if this exists instead of the dir? That will also detect using a subdir of the CSI volume

We may need a filewatcher for this but I assume and HOPE the volume is fully populated before the pod gets it mounted.

pnovotnak (approved):
Small change requested but not blocking

baseplate/lib/secrets.py
TylerLubeck: I'm nearly certain the volume will exist before the container we care about starts running. I can't find the docs for it right now, but there's something about how these mounts are loaded earlier in the kubernetes lifecycle and pods will fail to start if they can't properly mount

baseplate/lib/secrets.py
KTAtkinson: This might be my go showing but I very much like formatting error messages as "Expected {path} to be a directory, was a {type-of-thing-it-is}"

KTAtkinson (changes requested):

baseplate/lib/secrets.py
TylerLubeck: I think I covered both of these - unless I'm missing something, I don't think we can only ask "is it type X" and not "what type is it", so it's hard to show the {type-of-thing-it-is}

baseplate/lib/secrets.py
pnovotnak: Yea, that's what I'm seeing in snoodev. This is something I can test myself as well.

pnovotnak (approved):

KTAtkinson (approved):

PR #830, opened 2024-02-27
Vault CSI support v2: Fix data race returns null value

pnovotnak:
Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


baseplate/lib/secrets.py
```diff
@@ -524,14 +526,15 @@ def _raw_secret(self, name: str) -> Any:
 
     def get_raw_and_mtime(self, secret_path: str) -> Tuple[Dict[str, str], float]:
         mtime = self._get_mtime()
-        cache_entry = self.cache.get(secret_path, VaultCSIEntry(mtime=0, data=None, updating=False))
```
pnovotnak, line 527: The bug is here: We return this null data value, which we should not have been.

PR #846, opened 2024-04-10
Move instrumentation back to modules instead of __init__

trevorriles:
This is just targeting our working branch for opentelemetry. @chriskuehl and @KTAtkinson asked to have the instrumentation moved out of the top level modules so that they didn't go into effect on import.

I'm honestly not 100% sure what the problem is but it seems to be something around the ssl library getting loaded before it can be monkey patched. This then breaks urllib3.

This PR reverts the change, but if you have a better suggestion I'm all ears and can try to help validate it.

### DIFF
--- a/baseplate/clients/requests.py
+++ b/baseplate/clients/requests.py
@@ -25,6 +25,10 @@ from baseplate.lib import config
 from baseplate.lib.prometheus_metrics import default_latency_buckets
 from baseplate.lib.prometheus_metrics import getHTTPSuccessLabel
 
+from opentelemetry.instrumentation.requests import RequestsInstrumentor
+
+RequestsInstrumentor.instrument()
+
 
 def http_adapter_from_config(
     app_config: config.RawConfig, prefix: str, **kwargs: Any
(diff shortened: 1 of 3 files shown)


trevorriles:
Here's an example error:

```Apr 4 20:43:25 reddit-service-campaign-management-7b797d65cf-8vh9z reddit-service-campaign-management error [00] [01] {"level":"error","ts":1712263404.019811,"msg":"","worker_id":"campaign-management-6","worker_error_msg":"/usr/local/lib/venv/bin/baseplate-serve:12: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util.ssl_ (/usr/local/lib/venv/lib/python3.12/site-packages/urllib3/util/ssl_.py)', 'urllib3.util (/usr/local/lib/venv/lib/python3.12/site-packages/urllib3/util/__init__.py)']. "}```

chriskuehl:
Sorry for the slow review -- I was hoping to dig into this a bit but haven't had time yet. I'll try to check on this probably next week and see if I can find anything about that error.

PR #850, opened 2024-04-10
Update dependabot

SuperQ:
## 💸 TL;DR
* Increase python open PRs from default 5 to 20.
* Add dependabot updates for github actions.
* Update actions and pin versions to hash for supply chain security safety.
* Add a `runtime.txt` file to trick dependabot into thinking this is the
  required version for updates.

## ✅ Checks

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,0 +1,2 @@
+python-3.8
+# Specifies the python version that dependabot should use.
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,8 @@
 ---
 version: 2
 updates:
+  # Note, the file `runtime.txt` is used to hint to dependabot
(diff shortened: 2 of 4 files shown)


runtime.txt
```diff
@@ -0,0 +1,2 @@
+python-3.8
+# Specifies the python version that dependabot should use.
```
chriskuehl, line 2: Just FYI, I'm planning to try out migrating this repo to Poetry sometime soon, which should obviate the need for this (since Poetry solves for all Python versions and can deal with different constraints for each). OK to merge this now but might become unnecessary later.

chriskuehl (approved):

runtime.txt
SuperQ: Yea, I was just looking to get in workarounds to avoid dependabot trying to update things that would fail CI.

PR #891, opened 2024-04-11
Migrate to Poetry

chriskuehl:
This PR migrates Baseplate to use [Poetry](https://python-poetry.org/) for dependency management. Poetry is now responsible for managing Baseplate's dependencies and setting up a local Python environment. The benefits over the previous setup include:

* A local virtualenv which includes linters and other dependencies.
* No more manually resolving version conflicts or doing upgrades; Poetry has a CLI for it.
* No more unpinned requirements being sneakily introduced.
* Poetry solves for all compatible operating systems and Python versions with a single lock file.

Poetry only affects our own development workflow; the files that are published to PyPI are effectively identical. Consumers _do not_ need to switch to Poetry or change anything about their workflow.

## Changes in this PR

This PR upgrades many dependencies, but I've held back a few (pytest, black, reorder-python-imports) which require code changes for now.

Keep in mind that these dependency bumps, just like the ones in the old `requirements*.txt` files, are _only_ used by this repo's own tests. They do not affect consumers or releases we make.

## Validation

I generated a wheel on the current `develop` branch (`python setup.py bdist_wheel`) and a wheel on this branch using Poetry (`poetry build`). You can see a diff of the extracted wheel contents here: https://gist.github.com/chriskuehl/207271df9b590c83557d2d046b4f0913

Most of the changes are due to differences in whitespace or sorting between setuptools and Poetry. The only notable change is that I removed the `all` extra since it's a little harder to support, and it appears to be unused. I don't think there's much use-case for `all` other than developing on the package itself, and `poetry install --all-extras` handles that for us.


baseplate/lib/prometheus_metrics.py
```diff
@@ -32,7 +32,7 @@
 default_size_factor = 2
 default_size_count = 20
 default_size_buckets = [
-    default_size_start * default_size_factor ** i for i in range(default_size_count)
+    default_size_start * default_size_factor**i for i in range(default_size_count)
```
bjk-reddit: Can we move these lint changes to a separate PR?

.github/workflows/python-package.yaml
```diff
@@ -33,16 +33,21 @@ jobs:
       - uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
         with:
           path: ~/.cache/pip
+          # TODO: update for Poetry
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
```
bjk-reddit: This should be as easy as `hashFiles('**/poetry.lock')`

baseplate/lib/prometheus_metrics.py
chriskuehl: I removed the pytest and black upgrades. I'll do them in a followup PR.

bjk-reddit (approved):

.github/workflows/python-publish.yaml
```diff
@@ -15,18 +15,20 @@ jobs:
...
+
+    - name: Update PATH
+      run: 'echo "$HOME/.local/bin" >> "$GITHUB_PATH"'
 
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        python -m pip install setuptools==47.3.0 wheel==0.34.2 twine==3.1.1 requests==2.29.0
+        pipx install poetry==1.8.2
+        pipx install twine
```
chriskuehl: I am not 100% sure this is correct but I believe `pipx` is installed by `setup-python`. This is difficult to test, so I'll pay special attention to this during the next release.

Makefile
```diff
@@ -29,43 +29,43 @@ $(THRIFT_BUILDDIR)/tests/integration/test.thrift_buildstamp: tests/integration/t
...
 
 .PHONY: doctest
 doctest:
-	sphinx-build -M doctest docs/ build/
+	poetry run sphinx-build -M doctest docs/ build/
 
 .PHONY: linkcheck
 linkcheck:
-	sphinx-build -M linkcheck docs/ build/
+	poetry run sphinx-build -M linkcheck docs/ build/
```
chriskuehl: I'm a little mixed on this approach.

The options here are:

* Prefix commands with `poetry run` like this. This always works but is a little noisy.
* Don't prefix commands with `poetry run`. This would require users to manually activate their virtualenv to run the make targets (or we could use direnv to activate it automatically, but this won't have universal adoption).
* Don't prefix commands with `poetry run`, but mess with the PATH in the Makefile itself. This probably works but feels a bit sloppy to me (probably need to unset it for the `make venv` call, for example).

bin/baseplate-healthcheck
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!python
```
chriskuehl, line 1: `#!python` is a special shebang that wheel installers understand. They will rewrite this to the installation Python environment when the wheel is installed: https://packaging.python.org/en/latest/specifications/binary-distribution-format/#recommended-installer-features

This is actually what setuptools was doing already (it automatically rewrites these shebangs to `#!python` when it creates the wheel) which is why you don't see these shebangs in the wheel diff I posted. Poetry doesn't rewrite the shebangs automatically for these "legacy" scripts though, so I had to do it by hand.

pyproject.toml
```diff
@@ -1,3 +1,99 @@
+[tool.poetry]
+name = "baseplate"
+version = "2.7.0a1"
```
chriskuehl, line 3: Notable change: right now, the version is embedded here and not based on SCM tags, so we'll need to manually update this file when making new releases.

We may want to consider adopting https://github.com/mtkennerly/poetry-dynamic-versioning (haven't used it, but looks like what we want)? I'm inclined to wait and do it in a followup PR once we're sure all the other machinery is working.

Makefile
```diff
@@ -29,43 +29,43 @@ $(THRIFT_BUILDDIR)/tests/integration/test.thrift_buildstamp: tests/integration/t
...
 fmt:
-	@if [ ! -f /baseplate-py-dev-docker-image ]; then \
-		echo "Please use the docker-compose environment for consistency. See https://github.com/reddit/baseplate.py/blob/develop/CONTRIBUTING.md."; \
-		exit 1; \
-	fi
-	$(REORDER_PYTHON_IMPORTS) --exit-zero-even-if-changed $(PYTHON_SOURCE)
-	black baseplate/ tests/
-	$(REORDER_PYTHON_IMPORTS) --application-directories /tmp --exit-zero-even-if-changed $(PYTHON_EXAMPLES)
-	black docs/  # separate so it uses its own pyproject.toml
+	poetry run $(REORDER_PYTHON_IMPORTS) --exit-zero-even-if-changed $(PYTHON_SOURCE)
```
mathyourlife-reddit: With `reorder-python-imports 3.12.0`, I get 
```
reorder-python-imports: error: unrecognized arguments: --separate-from-import --separate-relative
```

Removal of those arguments from the Makefile line 1 allows it to run for me.

requirements.txt
mathyourlife-reddit, line 1: A few references to requirements.txt and runtime.txt remain in the repo.  Mostly comments or markdown.

```
git grep 'runtime.*txt'
git grep 'requirements.*txt'
```

Makefile
```diff
@@ -29,43 +29,43 @@ $(THRIFT_BUILDDIR)/tests/integration/test.thrift_buildstamp: tests/integration/t
...
+	poetry run black docs/  # separate so it uses its own pyproject.toml
 
 .PHONY: lint
 lint:
-	$(REORDER_PYTHON_IMPORTS) --diff-only $(PYTHON_SOURCE)
-	black --diff --check baseplate/ tests/
-	flake8 baseplate tests
-	PYTHONPATH=. pylint baseplate/
-	mypy baseplate/
+	poetry run $(REORDER_PYTHON_IMPORTS) --diff-only $(PYTHON_SOURCE)
```
mathyourlife-reddit: I ran into a similar issue to `make fmt` for `make lint`

```
reorder-python-imports: error: unrecognized arguments: --diff-only
```

mathyourlife-reddit (commented):
NICE!

Makefile
chriskuehl: Hmm, how did you end up with that version of reorder-python-imports? I also ran into that issue initially but I pinned it back down to the version we're using prior to this branch since I was trying to avoid making additional changes:
https://github.com/reddit/baseplate.py/blob/81544dee76469a509bbf76b751871b20cc768426/pyproject.toml#L75

Did you install dependencies with Poetry? What does `poetry show reorder-python-imports` produce?

.github/workflows/python-package.yaml
```diff
@@ -32,17 +32,21 @@ jobs:
...
 
       - name: Install dependencies
         env:
           CASS_DRIVER_NO_EXTENSIONS: theytaketoolongtobuild
         run: |
-          python --version
-          python -m pip install -r requirements.txt
+          pip install pipx
+          pipx install poetry==1.8.2
+          make venv
```
salomon-smekecohen: pipx to install poetry? any reason not to use the curl script?

.github/workflows/python-publish.yaml
```diff
@@ -15,18 +15,20 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
       with:
-        python-version: '3.8'
+        python-version: '3.12'
```
salomon-smekecohen, line 18: should this not target our lowest supported version still?

.github/workflows/python-publish.yaml
salomon-smekecohen: probably does but python -m pip works no matter what! why not stick to it?

equivalent in js:

using npm to install npx to install yarn.

.github/workflows/python-publish.yaml
```diff
@@ -15,18 +15,20 @@ jobs:
...
+        pipx install poetry==1.8.2
+        pipx install twine
 
     - name: Build and publish
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
       run: |
-        python setup.py sdist bdist_wheel
+        poetry build
```
salomon-smekecohen: twine seems... no longer necessary. any reason to keep it? poetry can publish and it will allow us to co-locate a lot of config.

Dockerfile
```diff
@@ -6,9 +6,11 @@ COPY --from=thrift /usr/local/bin/thrift /usr/local/bin/thrift
 
 WORKDIR /src
 
-COPY requirements*.txt ./
-RUN pip install -r requirements.txt
+RUN python -m venv /tmp/poetry && \
+    /tmp/poetry/bin/pip install poetry==1.8.2 && \
+    ln -s /tmp/poetry/bin/poetry /usr/local/bin/poetry
```
salomon-smekecohen, line 11: symlink from temp folder to usr will delete the binary but keep the command in the path right? any reason not to use a builder pattern here?

EX:
1. image w poetry builds a venv
2. this venv is copied over to the final image

this will also make the build times much smaller.

also... python 3.12 might be a mistake as we are not automatically testing on lowest supported version. `slim` might be nice too!

Makefile
salomon-smekecohen: My preference is _always_ to have poetry make the venv in the project and never depend on poetry in scripts. instead doing `.venv/bin/python -m COMMAND` you can configure this in poetry.toml, env vars, or pyproject.

bin/baseplate-healthcheck
salomon-smekecohen: does this work in 3.8?

poetry.toml
```diff
@@ -0,0 +1 @@
+virtualenvs.in-project = true
```
salomon-smekecohen, line 1: thank you so much. please add `.venv` to the `.dockerignore` or the image will have the venv copied into it  potentially if you copy `./`

pyproject.toml
```diff
@@ -1,3 +1,99 @@
...
+cqlmapper = ["reddit-cqlmapper"]
+kafka = ["confluent-kafka"]
+memcache = ["pymemcache"]
+prometheus = []  # Kept for backwards compatibility, these are now main requirements
+pyramid = ["pyramid"]
+redis = ["redis"]
+redis-py-cluster = ["redis-py-cluster"]
+refcycle = ["objgraph"]
+requests = ["advocate"]
+s3fetcher = []  # Kept for backwards compatibility, these are now main requirements
```
salomon-smekecohen, line 54: is this needed given minor release?

Makefile
mathyourlife-reddit: Just did a `pipx install reorder-python-imports`

```
$ poetry show reorder-python-imports
 name         : reorder-python-imports
 version      : 2.4.0
 description  : Tool for reordering python imports

dependencies
 - aspy.refactor-imports >=2.1.0
```

pyproject.toml
```diff
@@ -1,3 +1,99 @@
...
+pytest = "7.4.4"
+pytest-cov = "*"
+pytz = "*"
+reorder-python-imports = "2.4.0"
+sphinx = "*"
+sphinx-autodoc-typehints = "*"
+types-redis = "*"
+types-requests = "*"
+types-setuptools = "*"
+webtest = "*"
```
salomon-smekecohen, line 81: dev dependencies can be pinned real strict with no concerns for downstream effects. i worry a bit because too low a version of stuff like `mypy` and `pylint` will _for sure_ conflict w styles.

also: pytz is a direct dev dep?!?!

salomon-smekecohen (commented):
I have some nits and some other comments! But overall this is great.

Makefile
```diff
@@ -29,43 +29,43 @@ $(THRIFT_BUILDDIR)/tests/integration/test.thrift_buildstamp: tests/integration/t
 	rm -f tests/integration/test_thrift/TestService-remote
 	touch $@
 
+.PHONY: venv
+venv:
+	poetry install --all-extras
```
salomon-smekecohen, line 33: you might need `use 3.8` so that people that work on baseplate write 3.8+ code instead of "whatever poetry wants+ code"

.github/workflows/python-package.yaml
chriskuehl: I don't have a strong preference either way. For CI, their official docs recommend either one (although they do list `pipx` first on their official recommendations): https://python-poetry.org/docs/#ci-recommendations

.github/workflows/python-publish.yaml
chriskuehl: I don't think it really matters since the published artifacts will be the same either way. Could change it back if you want.

salomon-smekecohen: right but you want to make sure it works in the lower versions! or do tests cover?

chriskuehl: There's no testing happening in the publish workflow, it doesn't even install any dependencies (other than Poetry itself). I don't think using an older version would catch anything here (not even basic stuff like SyntaxErrors).

The regular tests do run a matrix of all versions 3.8-3.12 though.

salomon-smekecohen: sold.

.github/workflows/python-publish.yaml
chriskuehl: It's not guaranteed to work if there are conflicting packages already installed in global/user site-packages. pipx gives Poetry an isolated virtualenv which is why I like it for stuff like this, though probably not a huge deal for this particular workflow, especially if we get rid of twine.

.github/workflows/python-package.yaml
salomon-smekecohen: i guess its on a github runner and speed isnt really a big concern. im into it.

salomon-smekecohen (approved):
meant to approve last review 😅

redloaf (approved):

trevorriles (approved):

Dockerfile
chriskuehl: I'm pretty wary about the build image pattern for virtualenvs since they aren't really portable between systems. If there's ever a mismatch between image versions or installed packages in the images, then it's not guaranteed to work.

Makefile
chriskuehl: ++ I like that better, not sure why I neglected that option.

bin/baseplate-healthcheck
chriskuehl: Yes it does.

pyproject.toml
chriskuehl: I'll keep this for now since I'm trying to keep the generated wheels 1:1 identical with this PR but we can probably drop this. My impression has been that most people aren't using the extras anyway.

pyproject.toml
chriskuehl: We could set lower bounds (like `>=1.9.0` for mypy or whatever) if you think that's valuable. The lockfile ensures things are pinned, though in theory future upgrades can cause stuff to get downgraded lower than what we might want if there are conflicts. I'm not sure where we'd want to set the bounds though -- it's not uncommon for older Python versions to require older versions of packages than newer versions can use, and one of the benefits of Poetry is that it can handle this correctly (calculating and pinning different dependency trees for each scenario).

Makefile
chriskuehl: Hmm... I guess I would also want people to be able to run `poetry env use <whatever>` and then `make .venv` to be able to switch between versions if they need to test with a specific one.

It's also nice to default to the latest version for interactive development (while still having CI ensure you don't break old versions) since it tends to have better tracebacks, improvements to pdb, etc.

chriskuehl:
@salomon-smekecohen I pushed a new commit addressing some of the main feedback. Let me know if you have thoughts on any of the open threads (or threads I closed)!

salomon-smekecohen (approved):
king

PR #892, opened 2024-04-23
Thrift metrics: omit status codes for exceptions with a non-int code

umbrae:
## 💸 TL;DR
If an exception has a `code` property that is not actually derived from a baseplate ErrorCode, we currently muddy our prometheus labels with invalid values. This goes a step further to reduce the risk of this occurring by checking for integers, which `ErrorCode`'s thrift definition would require.

## 📜 Details
Context available in this thread for employees: https://reddit.slack.com/archives/C070ARBL3M2/p1713888048834589

Unit tested

## ✅ Checks


trevorriles:
@chriskuehl for some context there were the labels we were getting due to the fact that `.code()` is a method on the grpc object.
image

chriskuehl (approved):

PR #913, opened 2024-05-07
InOrderConsumer doesn't work as intended. 

AdrielVelazquez:
## 💸 TL;DR + Details
The InOrderConsumer Kafka Class makes a lot of assumptions that are False and inefficient with how Kafka Processes Events. 

The current logic of using only 1 consumer and commit back the offsets doesn't take into consideration that the upstream messages are sending messages randomly to each partition.  

You would need to enable transactions upstream for the producer, as well as delivering messages with a Key. This would guarantee two things.

1. All messages with a given key are guaranteed to land in a single partition. 

https://docs.confluent.io/platform/current/clients/producer.html#concepts
> The partitioners shipped with Kafka guarantee that all messages with the same non-empty key will be sent to the same partition.
> 
> If the key is provided, the partitioner will hash the key with murmur2 algorithm and divide it by the number of partitions. The result is that the same key is always assigned to the same partition. 

2. Consumers can switch to reading committed transactions instead of uncommitted transactions. 

https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#isolation-level

## 📜 Details

So what is this PR doing so differently? 

Technically, this keeps the same exact APIs in place, but only solves 1 problem. Stop committing back to Kafka so frequently. 

Instead of manually committing every offset which swarms Kafka with requests. Store the offset locally and commit the last max seen offset per partition back to Kafka in the auto.commit frequency. 

This keeps the behavior that you're committing every offset, but only delivering it every couple of seconds. 

## One More Detail

My lilnter went a little crazy when committing this. I can revert that.

### DIFF
--- a/baseplate/frameworks/queue_consumer/kafka.py
+++ b/baseplate/frameworks/queue_consumer/kafka.py
@@ -488,6 +488,7 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
             prometheus_client_name=self.prometheus_client_name,
         )
 
+
 class FastConsumerFactory(_BaseKafkaQueueConsumerFactory):
     """Factory for running a :py:class:`~baseplate.server.queue_consumer.QueueConsumerServer` using Kafka.
 
--- a/baseplate/frameworks/queue_consumer/kafka.py
+++ b/baseplate/frameworks/queue_consumer/kafka.py
@@ -358,14 +358,16 @@ class _BaseKafkaQueueConsumerFactory(QueueConsumerFactory):
 
         # pylint: disable=unused-argument
         def log_assign(
-            consumer: confluent_kafka.Consumer, partitions: List[confluent_kafka.TopicPartition]
+            consumer: confluent_kafka.Consumer,
+            partitions: List[confluent_kafka.TopicPartition],
         ) -> None:
             for topic_partition in partitions:
                 logger.info("assigned %s/%s", topic_partition.topic, topic_partition.partition)
 
         # pylint: disable=unused-argument
         def log_revoke(
-            consumer: confluent_kafka.Consumer, partitions: List[confluent_kafka.TopicPartition]
+            consumer: confluent_kafka.Consumer,
+            partitions: List[confluent_kafka.TopicPartition],
         ) -> None:
             for topic_partition in partitions:
                 logger.info("revoked %s/%s", topic_partition.topic, topic_partition.partition)
@@ -419,6 +421,17 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
     If you need more control, you can create the :py:class:`~confluent_kafka.Consumer`
     yourself and use the constructor directly.
 
+    UPDATE: The InOrderConsumerFactory can NEVER achieve in-order, exactly once message processing.
+
+    Message processing in Kafka to enable exactly once starts at the Producer enabling transactions,
+    and downstream consumers enabling reading exclusively from the committed offsets within a transactions.
+
+    Secondly, without defined keys in the messages from the producer, messages will be sent in a round robin fashion to all partitions in the topic. This means that newer messages could be consumed before older ones if the consumer of those partitions with newer messages are faster.
+
+    Some improvements are made instead that retain the current behaviour, but don't put as much pressure on Kafka by committing every single offset.
+
+
+    Instead of committing every single message's offset back to Kafka, the consumer now commits each offset to it's local offset store, and commits the highest seen value for each partition at a defined interval (auto.commit.interval.ms). "enable.auto.offset.store" is set to false to give our application explicit control of when to store offsets.
     """
 
     # we need to ensure that only a single message handler worker exists (max_concurrency = 1)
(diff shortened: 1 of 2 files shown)


baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -198,9 +202,12 @@ def handle(self, message: confluent_kafka.Message) -> None:
...
 
                 if message_latency is not None:
-                    context.metrics.timer(f"{self.name}.{topic}.latency").send(message_latency)
+                    context.metrics.timer(f"{self.name}.{topic}.latency").send(
+                        message_latency
+                    )
 
-                context.metrics.gauge(f"{self.name}.{topic}.offset.{partition}").replace(offset)
+                context.metrics.gauge(f"{self.name}.{topic}.offset.{
+                                      partition}").replace(offset)
```
chriskuehl: Is this a syntax error? Or at least, Black can't parse it:

```
error: cannot format baseplate/frameworks/queue_consumer/kafka.py: Cannot parse: 209:39:
                 context.metrics.gauge(f"{self.name}.{topic}.offset.{
```

chriskuehl (commented):
This change looks reasonable to me, but I'm not very familiar with Kafka so I can't really provide insight into the decision itself. Could you nominate an additional reviewer (maybe someone from your team) who can double-check that the change looks right before we merge it?

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -419,6 +436,17 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
     If you need more control, you can create the :py:class:`~confluent_kafka.Consumer`
     yourself and use the constructor directly.
 
+    UPDATE: The InOrderConsumerFactory can NEVER achieve in-order, exactly once message processing.
+
+    Message processing in Kafka to enable exactly once starts at the Producer enabling transactions,
+    and downstream consumers enabling reading exclusively from the committed offsets within a transactions.
+
+    Secondly, without a definied keys consumers consume all partitions at the same time, which could have wildly different order.
```
tomncooper:
```suggestion
    Secondly, without defined keys in the messages from the producer, messages will be sent in a round robin fashion to all partitions in the topic. This means that newer messages could be consumed before older ones if the consumer of those partitions with newer messages are faster.
```

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -419,6 +436,17 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
...
+
+    Some improvements are made instead that keeps that "assumed" functionality, but doesn't put pressure on Kafka for committing every single offet.
```
tomncooper:
```suggestion
    Some improvements are made instead that retain the current behaviour, but don't put as much pressure on Kafka by committing every single offset.
```

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -419,6 +436,17 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
...
+
+
+    Instead of committing the offset back to Kafka, commits the offset to it's local offset store, and commits within a certain interval.
```
tomncooper:
```suggestion
    Instead of committing every single message's offset back to Kafka, the consumer now commits each offset to it's local offset store, and commits the highest seen value for each partition at a defined interval (auto.commit.interval.ms).
```

tomncooper (commented):
The linting changes do make it harder to see what has logically changed. I am assuming the main changes are lines: 476, 477 and 496. 

The only comments I have are a few language nits on the doc string and a question over the value of `auto.commit.interval.ms`. If the consumer crashes then there is a now a chance of duplicate message reads. The worst case being crashing just before the end of an interval in which case you would replay `auto.commit.interval.ms` worth of messages. Which, for what it is worth, is pretty standard practice for Kafka consumers.

Do we want to surface that, or is `auto.commit.interval.ms` already set to a low value? Do we want to allow users to tune that?

baseplate/frameworks/queue_consumer/kafka.py
```diff
@@ -419,6 +436,17 @@ class InOrderConsumerFactory(_BaseKafkaQueueConsumerFactory):
...
+
+    Message processing in Kafka to enable exactly once starts at the Producer enabling transactions,
+    and downstream consumers enabling reading exclusively from the committed offsets within a transactions.
+
+    Secondly, without defined keys in the messages from the producer, messages will be sent in a round robin fashion to all partitions in the topic. This means that newer messages could be consumed before older ones if the consumer of those partitions with newer messages are faster.
+
+    Some improvements are made instead that retain the current behaviour, but don't put as much pressure on Kafka by committing every single offset.
+
+
+    Instead of committing every single message's offset back to Kafka, the consumer now commits each offset to it's local offset store, and commits the highest seen value for each partition at a defined interval (auto.commit.interval.ms).
```
skykistler:
```suggestion
    Instead of committing every single message's offset back to Kafka, the consumer now commits each offset to it's local offset store, and commits the highest seen value for each partition at a defined interval (auto.commit.interval.ms). "enable.auto.offset.store" is set to false to give our application explicit control of when to store offsets. 
```

skykistler (commented):
Just one clarification to the comment

baseplate/frameworks/queue_consumer/kafka.py
AdrielVelazquez: Suprisingly other files failed when being formatted (even with black). So I just disabled my formatter in general.

AdrielVelazquez:
@tomncooper and @skykistler This is good to review again. 

I mentioned the config and the default in the docstring, but I would rather not be explicit on the value.

AdrielVelazquez:
@chriskuehl Sorry should have tagged you also. But I cleaned up all the linting formatting my setup had.

tomncooper (approved):

skykistler (approved):

chriskuehl (approved):

PR #915, opened 2024-05-16
Add psycopg2 extra, patch psycogreen in entry points.

### LINKED ISSUE #62: Consider automating Psycogreen monkeypatching
`baseplate-{script,serve}` already hide gevent monkeypatching, so it'd be nice to do the same for psycogreen monkeypatching if sqlalchemy+PG is in use.

salomon-smekecohen:
Closes #62 

## 💸 TL;DR
If you want to use psycopg2, baseplate should automatically patch it for gevent. You can use sqlalchemy with a variety of postgres packages, not just psycopg2. So make the dependencies an extra instead of attaching to sqlalchemy or a "sqlalchemy-postgres" extra.

Pointed my Baseplate.py project to this hash, no issues. Patch works as expected.

## ✅ Checks

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1171,6 +1171,7 @@ files = [
     {file = "lxml-5.2.1-cp37-cp37m-musllinux_1_2_x86_64.whl", hash = "sha256:9e2addd2d1866fe112bc6f80117bcc6bc25191c5ed1bfbcf9f1386a884252ae8"},
     {file = "lxml-5.2.1-cp37-cp37m-win32.whl", hash = "sha256:f51969bac61441fd31f028d7b3b45962f3ecebf691a510495e5d2cd8c8092dbd"},
     {file = "lxml-5.2.1-cp37-cp37m-win_amd64.whl", hash = "sha256:b0b58fbfa1bf7367dde8a557994e3b1637294be6cf2169810375caf8571a085c"},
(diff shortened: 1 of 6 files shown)


poetry.lock
```diff
@@ -2896,6 +2886,7 @@ cqlmapper = ["reddit-cqlmapper"]
 kafka = ["confluent-kafka"]
 memcache = ["pymemcache"]
 prometheus = []
+psycopg2 = []
```
salomon-smekecohen: this is not quite right.

pyproject.toml
```diff
@@ -39,6 +39,8 @@ sentry-sdk = { version = ">=1.35.0,<2.0", optional = true }
 sqlalchemy = { version = ">=1.4.49,<2", optional = true }
 thrift-unofficial = ">=0.19.0,<1.0"
 typing-extensions = "^4.11.0"
+psycopg2 = "^2.0.0"
+psycogreen = "^1.0.0"
```
chriskuehl, line 43: Hmm, I'm surprised this doesn't need to be marked with `optional = true` like the other extra dependencies, but I checked and the built wheel seems to correctly mark it as only needed with the extra:

```
Requires-Dist: psycogreen (>=1.0.0,<2.0.0) ; extra == "psycopg2"
Requires-Dist: psycopg2 (>=2.0.0,<3.0.0) ; extra == "psycopg2"
```

chriskuehl (approved):

pyproject.toml
salomon-smekecohen: Thanks Chris! Yeah maybe I messed this up, I have never added an extra using poetry before. If you have a docs link to how you did it ill give it a read to try and track down differences.

PR #920, opened 2024-05-23
Move baseplate.py sampling and configuration to be more in line with …

trevorriles:
Bring inline with baseplate.go configuration.

- Defaults to sampling at 10 rps.
- Endpoint and service_name are configured via env variables set by infrared.

Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate/lib/tracing.py
+++ b/baseplate/lib/tracing.py
@@ -1,0 +1,50 @@
+import random
+
+from typing import Optional
+from typing import Sequence
+
+from opentelemetry.sdk.trace.sampling import Decision
+from opentelemetry.sdk.trace.sampling import Sampler
+from opentelemetry.sdk.trace.sampling import SamplingResult
+from opentelemetry.trace import SpanKind
+from opentelemetry.util.types import Attributes
+from pyrate_limiter import Duration
+from pyrate_limiter import Limiter
+from pyrate_limiter import Rate
+
+
+class RateLimited(Sampler):
+    """
+    The RateLimited Sampler will sample every request up to a specified requests per second.
+    :param sampler: The parent opentelemetry sampler we are wrapping.
+    :type sampler: opentelemetry.sdk.trace.sampling.Sampler
+    :param rps: The number of requests per second we would like to sample up to.
+    :type rps: int
+    """
+
+    def __init__(self, sampler: Sampler, rps: int):
+        self.rps = rps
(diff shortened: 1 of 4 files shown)


baseplate/server/__init__.py
```diff
@@ -235,66 +229,28 @@ def on_start(self, span: Span, parent_context: Optional[Context] = None) -> None
...
-                )
-            )
-            trace.set_tracer_provider(provider)
-            logger.info("Tracing Configured")
-        else:
-            logger.warning(
-                "Tracing disabled: service_name and endpoint are required fields in the tracing config."
-            )
+    try:
+        sample_rps = int(os.getenv("TRACE_SAMPLER_RPS", 10))
```
redloaf: For things like this, we usually prefix the baseplate env vars with `BASEPLATE_` and use helpers such as `config.DefaultFromEnv`.

redloaf (changes requested):

baseplate/server/__init__.py
trevorriles: Awesome, I'll look into get it updated!

redloaf (approved):

baseplate/lib/tracing.py
```diff
@@ -0,0 +1,50 @@
...
+    def should_sample(
+        self,
+        parent_context: Optional["Context"],
+        trace_id: int,
+        name: str,
+        kind: Optional[SpanKind] = None,
+        attributes: Attributes = None,
+        links: Optional[Sequence["Link"]] = None,
+        trace_state: Optional["TraceState"] = None,
+    ) -> "SamplingResult":
```
chriskuehl: The quoting is odd on these, normally it's only needed for self-references (before typing.Self) or when there is weird import-type type construction or stuff like that. If you remove the quoting (and add imports as needed) does it work, or were you getting some error?

```suggestion
        parent_context: Optional[Context],
        trace_id: int,
        name: str,
        kind: Optional[SpanKind] = None,
        attributes: Attributes = None,
        links: Optional[Sequence[Link]] = None,
        trace_state: Optional[TraceState] = None,
    ) -> "SamplingResult":
```

chriskuehl (approved):

baseplate/lib/tracing.py
trevorriles: I'll update this, PyCharm auto generated with quotes 👍

PR #951, opened 2024-07-23
Remove noisy span debug logging

trevorriles:
Our tracing implementation that was merged has some noisy debug logs emitted as warnings. Removing those to avoid unnecessary log load.


redloaf (approved):
🔕 if these logs are useful for debugging, maybe switch them to `logger.debug` instead of removing them?

chriskuehl (approved):

PR #952, opened 2024-07-24
Move pyramid otel instrumentation to BaseplateConfigurator

trevorriles:
This moves the instrumentation logic to the initialization of the BaseplateConfigurator class instead of being an import side-effect.


trevorriles:
@chriskuehl I believe this is closer to what you were looking for in the main PR. When testing this against some ads services I ran into the case where they didn't import the  module directly and missed instrumentation because of it. Any service integrating pyramid with baseplate should use the BaseplateConfigurator class so I think this is a good landing spot.

chriskuehl (approved):

PR #953, opened 2024-07-30
Patch using opentelemetry-instrumentation-threading

trevorriles:
This solves some context passing issues when crossing thread boundaries. It also upgrades  opentelemetry libs to 1.25.0, which is the release that includes official python 3.12 support.

We apply the instrumentation in this case as a side-effect, similar to how we do the gevent monkey patching.


trevorriles:
Will fix up these errors after lunch. It seems like the exceptions are now stored with the full classpath.

redloaf (approved):

chriskuehl (approved):

PR #964, opened 2024-08-08
Feat: KV store on RequestContext class

trevorriles:
Add a `kv` field to the baseplate `RequestContext` object to allow passing arbitrary data.

This can be used to explicitly pass opentelemetry context across processes, for example with our internal graphql codebase we can do things like:

Attach to context in kv store

```
trace.context_api.attach(context.kv["otel"])
```

Store span in context:
```
trace.set_span_in_context(span, context.kv["otel"])
```

Of course this can be used for other non-tracing purposes as well.

### DIFF
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -192,7 +192,7 @@ class RequestContext:
         self.__prefix = prefix
         self.__wrapped = wrapped
         if kv is None:
-            self.kv = dict()
+            self.kv = {}
         else:
             self.kv = kv
 
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -241,7 +241,7 @@ class RequestContext:
             prefix=self.__prefix,
             span=self.span,
             wrapped=self,
-            kv=self.kv.copy()
+            kv=self.kv.copy(),
         )
 
 
--- a/baseplate/__init__.py
+++ b/baseplate/__init__.py
@@ -193,6 +193,8 @@ class RequestContext:
         self.__wrapped = wrapped
         if kv is None:
             self.kv = dict()
+        else:
+            self.kv = kv
 
         # the context and span reference eachother (unfortunately) so we can't
         # construct 'em both with references from the start. however, we can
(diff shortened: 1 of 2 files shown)


baseplate/__init__.py
```diff
@@ -186,10 +186,15 @@ def __init__(
         prefix: Optional[str] = None,
         span: Optional["Span"] = None,
         wrapped: Optional["RequestContext"] = None,
+        kv: Optional[Dict[str, Any]] = None,
```
chriskuehl, line 189: If I'm understanding correctly, it seems like the request context object is kind of already a key-value store, but with statically-defined fields and better typing. I wonder if it would make sense to add the keys you need to the request context object directly instead?

If it's not possible to statically list the keys as fields on this object, maybe we could limit this to just a key-value store for tracing data? I'm a bit worried that a generic key/value store will have more and more keys added to it over the years, and eventually we'd end up in a situation where we never really know what keys are or aren't present at any point in the codebase, or what types the values are.

trevorriles: Yeah that makes sense, let me switch things up to at least name it otel specific. OTEL uses a kv store for the context, so it'll be sort of the same thing just named for a specific purpose.

trevorriles: This still has the potential to be a bit messy. Generally we are using the global context which is built on top of python ContextVars. This works for most things, but with GQL we're running into an issue with the promise library where execution jobs are put on a queue and are being execute at various times resulting in the context getting screwed up with the global value. Setting the value on the RequestContext object allows us to access it, but for much of the baseplate code we're not using the explicit context value from this object.

I've thought about rewriting the implementation to always use the explicit context value on the the `RequestContext` object but we run into an issue with the `make_object_for_context` function used by the clients. We'd need to pass the context to those calls as well as the "old" server span it currently has as an argument. The problem of updating that function is that it is the interface other clients use to integrate into baseplate.

Either way I can change this to be an otel specific context value, but it should probably only be used in specific scenarios unless we want to rework that make_object_for_context funciton :thinking:

trevorriles:
@chriskuehl I'm going to close this for now. I put this together to solve a specific issue I had within graphql. I'm going to see if I can solve it with out introducing functionality that might increase confusion within bp.py core.

I may be back with something similar as I begin to work with the graphql team, but we don't need this in the upcoming release.

PR #985, opened 2024-09-27
Add support for Python 3.13

chriskuehl:
Python 3.13 was released on 2024-10-07. This PR adds support to Baseplate.py for services running 3.13.

### Merge checklist
* [x] Replace rc2 with final 3.13.0 release once it's available
* [x] Need wheels for some libraries (incomplete list):
  * [x] psycopg2: `https://github.com/psycopg/psycopg2/pull/1695`
  * [x] confluent-kafka: `https://github.com/confluentinc/confluent-kafka-python/issues/1768`
  * [ ] pendulum: `https://github.com/sdispater/pendulum/issues/844`
    * _No wheel yet, so Rust needs to be available at install time. We'll include this in the release notes but don't want to block this branch on it, especially since it can be worked around (e.g. by building your own local wheel ahead of time)._
* [x] gevent is not yet compatible with 3.13, work is in progress: https://github.com/gevent/gevent/commits/py313/
* [ ] Review dependencies manually for issues like we did for 3.11/3.12
  * [x] webob does not yet support Python 3.13 due to a `cgi` import: `https://github.com/Pylons/webob/issues/437`
    * Working around this for now using the `legacy-cgi` package: https://pypi.org/project/legacy-cgi/
  * [x] `grpcio` needs a new release
    * ~[`1.67.0`](https://github.com/grpc/grpc/releases/tag/v1.67.0) has the fix but it's not on PyPI yet: https://pypi.org/project/grpcio/#history (1.67.0rc1 doesn't have the fix)~ Actually we just need 1.66.2.


chriskuehl:
We're hitting `https://github.com/pylint-dev/pylint/issues/10112` now.

.github/workflows/python-package.yaml
```diff
@@ -37,6 +37,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-poetry-
 
+      # TODO(ckuehl|2024-09-26): Remove this once Python 3.13 wheels are
+      # available for our dependencies.
+      - name: Install temporary Python 3.13 build dependencies
+        if: ${{ matrix.python-version == '3.13' }}
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
redloaf, line 45: Call me old fashioned, but it seems like a good idea to install a specific version and verify checksums.

redloaf (approved):

.github/workflows/python-package.yaml
scotthew1: 👍  yeah how long do we expect this workaround to be in place?

scotthew1 (approved):

PR #1001, opened 2024-10-20
Prefer builtins over some typing aliases

RyanBalfanz:
This is an incremental replacement of `typing.Dict` with `builtins.dict`, `typing.List` with `builtins.list`, `typing.Set` with `builtins.set`, and `typing.Tuple` with `builtins.tuple`, though other aliases exist which.

See: https://docs.python.org/3/library/typing.html#deprecated-aliases

> Deprecated since version 3.9: [builtins.dict](https://docs.python.org/3/library/stdtypes.html#dict) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

Deprecated since version 3.9, as such, this change makes sense to follow #1002 which makes version 3.9 the oldest supported version though the aliases are guaranteed to remain in the typing module without deprecation warnings until at least Python 3.14.

Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


RyanBalfanz:
#1002 was moved to #1003

chriskuehl:
@RyanBalfanz it looks like a lot of these were converted by Ruff already, do you think we still need this PR?

RyanBalfanz:
> @RyanBalfanz it looks like a lot of these were converted by Ruff already, do you think we still need this PR?

If Ruff took care of it, then great! I'll take a look to see if anything my have been missed and create a new PR if so.

RyanBalfanz:
I took a look; this is no longer needed.

PR #1003, opened 2024-10-23
Drop python3.8 support

RyanBalfanz:
Closes # 

Python 3.8 EOL date was 2024-10-07.

See https://github.com/reddit/baseplate.py/pull/824 for prior art, dropping Python 3.7.

## 💸 TL;DR

Drops Python v3.8 support.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


.github/workflows/python-package.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
RyanBalfanz: My editor changed the quotes -- I'll switch them back.

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
RyanBalfanz, line 1: Note: I installed the version that is installed explicitly in the `Dockerfile`, hence the version change.

mathyourlife-reddit (approved):
Looks like updates are consistent with the 3.7 drop and test are 🟢 . Seems reasonable to me, would also like a 👍 from @chriskuehl before merging.

RyanBalfanz:
> Looks like updates are consistent with the 3.7 drop and test are 🟢 . Seems reasonable to me, would also like a 👍 from @chriskuehl before merging.

3.7 drop didn't require an updated `poetry.lock` but CI was failing without it, so I am keen to call that out here

chriskuehl (approved):
Thanks for taking care of this!

chriskuehl:
> 3.7 drop didn't require an updated `poetry.lock` but CI was failing without it, so I am keen to call that out here

This looks good to me. Needing to update the `poetry.lock` is expected when you change the `pyproject.toml` since it contains a hash of the important fields, and also it may need to adjust the package versions if you change dependencies. You can use `poetry lock --no-update` to do a minimal update (should be fast as it only fixes the stuff it needs to), or `poetry lock` (which can be slow and pull in a ton of updates).

Worth saying explicitly that `poetry.lock` does **not** affect users of Baseplate in any way. It's only used for Baseplate's own tests and local development, and is not even included in the sdist or wheel. So, even if you run `poetry lock` and pull in more changes than necessary, it has no impact on users so there is zero risk here.

PR #1005, opened 2024-10-23
Emit more shutdown logging by default

chriskuehl:
By default we only print out INFO level logs, so we only get the final `Exiting` message and not the messages about draining or gracefully shutting down. This makes it hard to correlate errors around shutdown time with the server's shutdown sequence.

Promoting these to INFO logs is worth it IMO since errors around shutdown are a common problem.


mathyourlife-reddit (approved):
Sounds good.  It's only a couple lines total (vs a per request line or something) and +1 to providing better visibility to how the server is handling shutdown.

redloaf (approved):

mjbonifacio (approved):
+1 great for visibility, excited to pick up these changes

RyanBalfanz (approved):

PR #1009, opened 2024-10-27
Prevent GreenletExit exceptions (and improve shutdown time) when shutting down with active keepalive connections

chriskuehl:
Internal ticket: SVF-438

This fixes an issue where a Baseplate WSGI server in shutdown doesn't proactively close existing "keepalive" connections and continues to process requests from them. This causes two issues:

1. Requests can be killed in-flight with `GreenletExit` if the server's `stop_time` is reached.
2. Shutdowns are unnecessarily slow because the gevent server waits for connections to be closed (but does nothing to close them itself).

This fix causes gevent to close connections once a server begins graceful shutdown (i.e. once `server.stop()` is called). It does this by adding the `Connection: close` header on responses and by closing connections in the pool.

## Commentary

There doesn't appear to be an easy way to fix this without subclassing gevent's `WSGIHandler` and/or `WSGIServer` classes. These classes are designed to be subclassed according to the documentation, though it would still be better to fix this in gevent.

I believe most gevent users don't hit this as it is fairly rare (and not recommended) to use gevent's built-in WSGI server instead of something like [gunicorn](https://gunicorn.org/).

I've emailed the gevent mailing list to ask about this issue. If the reception is positive I'll submit a patch upstream.

## Validation

* First commit adds a regression test (expected to fail on `develop` branch).
* Second commit adds the fix, which also fixes the regression test.


redloaf (approved):

mjbonifacio (approved):
I'm informed and this appears sensible -- excited to try it. I wouldn't consider my approval authoritative as I don't think I'd be able to foresee other issues this might cause.

baseplate/server/wsgi.py
```diff
@@ -18,6 +24,92 @@
...
+    The default gevent WSGIServer prevents new *connections* once the server
+    enters shutdown, but does not prevent new *requests* over existing
+    keepalive connections. This results in slow shutdowns and in some cases
+    requests being killed mid-flight once the server reaches stop_timeout.
+
+    This server may be used with any gevent WSGIHandler, but the keepalive
+    behavior only works when using BaseplateWSGIHandler.
+    """
+
+    shutdown_event: gevent.event.Event
```
salomon-smekecohen: gevent.event is a module up until recently. So this will be incompatible with a lot of gevent versions, even in fairly up-to-date repositories.

consider:

`from gevent import event`
and
`event.Event`

Screenshot 2024-11-02 at 5 56 15 PM

tests/integration/requests_tests.py
```diff
@@ -37,6 +44,8 @@ def gevent_socket():
 @pytest.fixture
 def http_server(gevent_socket):
     class HttpServer:
+        server: gevent.pywsgi.WSGIServer
```
salomon-smekecohen: `pywsgi` also has the same issue.

salomon-smekecohen (changes requested):
I believe the way the `pywsgi` and `event` modules are accessed will cause issues.
Screenshot 2024-11-02 at 6 00 49 PM

baseplate/server/wsgi.py
chriskuehl: I think some lucky import side effects are preventing this from being an issue with how the code is currently written (it still imports for me with gevent 23.9.1 which is the oldest we support), but good idea to fix this since it's definitely fragile.

salomon-smekecohen:
ty! sorry for not re-reviewing quickly here

PR #1011, opened 2024-10-31
Move healthchecks to it's own package

trevorriles:
This avoids initializing the server packages when only sending a healthcheck.


chriskuehl (approved):
Tested and the performance improvement from cutting out all the unnecessary imports is significant (~1100ms => ~590ms from within the exact same pod talking to the same server). Seems clearly worth it.

PR #1019, opened 2024-11-04
Add secrets fetcher tests where none existed before

RyanBalfanz:
Closes # 

## 💸 TL;DR

Add tests to secrets fetcher sidecar, none existed before. These tests help to capture existing behavior so that changes can be made to secrets fetching with confidence.

I started looking at the atomicity of secrets fetching (i.e. the temporary file), in particular a common failure mode using `json.dump` causing partial writes and malformed JSON files, but want to add some tests before making any changes to the implementation.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,224 @@
...
+        p = pathlib.Path("/var/local/secrets.json")
+        self.assertEqual(p.owner(), whoami)
+
+    def test_sets_group(self):
+        p = pathlib.Path("/var/local/secrets.json")
+        self.assertEqual(p.group(), whoami)
+
+    def sets_mode(self):
+        ...
+        # self.assertEqual(p.stat().st_mode, "vscode")
```
mathyourlife-reddit: 🔕  remove commented out code.

tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,224 @@
...
+                secrets_fetcher.fetch_secrets(cfg, f)
+
+    def test_temporary_file_is_not_deleted(self):
+        p = pathlib.Path("/var/local/secrets.json.tmp")
+        self.assertTrue(p.exists())
+
+    def test_temporary_file_is_partially_written(self):
+        p = pathlib.Path("/var/local/secrets.json.tmp")
+        text = p.read_text()
+        self.assertEqual(text, """{\n  "secrets": {\n    "secret/one": """)
```
mathyourlife-reddit, line 216: should this assert that the file contents either don't include a valid token or that the contents are not valid json rather than matching the malformed content in case the malformed content changes in the future.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ webtest = "*"
 parameterized = "^0.9.0"
 opentelemetry-test-utils = "^0.47b0"
 ruff = "*"
+pyfakefs = "^5.7.1"
```
mathyourlife-reddit, line 92: Probably not really a better option, but just checking if you considered writing the tests without needing to bring in an extra dependency?

tests/unit/sidecars/secrets_fetcher_tests.py
RyanBalfanz: This treatment showcases the general issue and how it precisely manifests in this failure mode. I think the check for invalid JSON is a weaker assertion and the maintenance burden with content changes doesn't outweigh that.

I don't think this partially written file side effect is desirable or beneficial and could be changed. I'd previously done that but didn't want to submit a PR against code that is not tested -- so maybe this test will go away soon anyway.

pyproject.toml
RyanBalfanz: Not strongly, no. Since the current implementation doesn't clean up after itself in the failure mode tested, running tests locally litters with that file cruft. Generally depending on specific fs entities and setup makes this feel like the right choice.

`pyfakefs` seems mature and well maintained by Google.

tests/unit/sidecars/secrets_fetcher_tests.py
RyanBalfanz: I'll do that. I wanted to capture the indented behavior that the secrets fetcher implements but didn't get that working at first. Will remove this for now.

pyproject.toml
mathyourlife-reddit: yeah, the other alternatives seemed like it'd be a bit more headache than it's worth.  Just asking the question before adding dependencies. :)

RyanBalfanz: Yeah, good call out.

mathyourlife-reddit (approved):
Thanks for adding tests!!!

pyproject.toml
chriskuehl: What's your opinion on using [pytest's tmp_path fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html)? This is my usually my preferred way to test functionality that depends on the filesystem. It has a lot of nice properties IMO:

* You can use all the regular filesystem operations.
* It doesn't rely on mocking of any kind.
* It handles cleanup of test files.
* It preserves a small number (last 2?) of test run files on disk, which is IMO very useful for inspecting them to understand what's happening or what went wrong.
* It's easy to drop a breakpoint anywhere in your test and inspect the real filesystem state.

I'm not against in-memory mocks for filesystems but generally don't see much benefit when `tmp_path` exists.

tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,220 @@
...
+configini = f"""
+[secret-fetcher]
+vault.url = https://vault.example.com:8200/
+vault.role = my-server-role
+vault.auth_type = aws
+vault.mount_point = aws-ec2
+
+output.path = /var/local/secrets.json
+output.owner = {getpass.getuser()}
+output.group = {getpass.getuser()}
```
chriskuehl:
```suggestion
output.owner = {whoami}
output.group = {whoami}
```

tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,220 @@
...
+
+        cls.cfg = config.parse_config(fetcher_config, spec)
+
+    def setUp(self):
+        self.setUpPyfakefs()
+        self.fake_fs().create_file("/var/local/secrets.json", contents="initial contents")
+
+        cfg = self.cfg
+        now = datetime.datetime.now(UTC)
+        with unittest.mock.patch("baseplate.sidecars.secrets_fetcher.VaultClientFactory") as mock:
```
chriskuehl: Can you set `autospec=True` here? IMO this should basically always be used when mocking like this in order to avoid replacing the class with a MagicMock that doesn't enforce you call it like the original class it's replacing.

pyproject.toml
RyanBalfanz: I’m not familiar with that option, so I hadn’t considered it. But happy to take a look if we feel strongly that it's a better option here.

Does it make sense here when we are working with specific known file paths like `/var/local/secrets.json`? As in, we're not just testing various operations on arbitrary files, but rather files in specific locations on the filesystem. Where is the boundary for logic and configuration we really want to be under test here?

tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,220 @@
...
+        p = pathlib.Path("/var/local/secrets.json")
+        self.assertTrue(p.is_file())
+
+    def test_sets_owner(self):
+        p = pathlib.Path("/var/local/secrets.json")
+        self.assertEqual(p.owner(), whoami)
+
+    def test_sets_group(self):
+        p = pathlib.Path("/var/local/secrets.json")
+        self.assertEqual(p.group(), whoami)
```
chriskuehl: This seems a bit odd since the username isn't inherently also a group and it's fairly common for there not to be a group with the same name as your username. For example user `nobody` typically has primary group `nogroup`, users on macOS usually have primary group `staff`, etc.

Is this working because of the in-memory filesystem mock or for some other reason?

tests/unit/sidecars/secrets_fetcher_tests.py
```diff
@@ -0,0 +1,220 @@
...
+    # @typing.override  # TODO: Added in version 3.12.
+    def get_secret(self, secret_name: str) -> tuple[typing.Any, datetime.datetime]:
+        """Return a secret value that is not JSON serializable."""
+        s = self.token_expiration
+        return s, self.token_expiration + datetime.timedelta(seconds=30)
+
+
+class Tests(TestCase):
+    @classmethod
+    def setUpClass(cls):
```
chriskuehl, line 75: Just curious, was there a reason to prefer unittest-style class tests rather than pytest style (test functions with fixtures) here? We have a mix of both so I don't think one is more correct than another, just wondering if there was anything motivating this choice over the other.

pyproject.toml
chriskuehl: No worries, I don't feel really strongly about it, but that is how I'd have written these so I figured it was worth asking for your thoughts.

I normally introduce an option to set the output directory and set that when calling it in tests. It feels a little less magical and easier to debug to me since there's no mocking and it's real filesystem operations that you can look at with all the regular filesystem tools you're used to. It has downsides too, of course.

RyanBalfanz: Gotcha, thanks for the notes! I'll take a closer look at it next time I'm working in this space.

tests/unit/sidecars/secrets_fetcher_tests.py
RyanBalfanz: This is because locally I was using a devcontainer under user `vscode`, and in GH CI the user is `root`.

tests/unit/sidecars/secrets_fetcher_tests.py
RyanBalfanz: Nothing except that I was actually able to get this working in this style. There is a [pytest plugin](https://pytest-pyfakefs.readthedocs.io/en/stable/usage.html#patch-using-the-pytest-plugin).

tests/unit/sidecars/secrets_fetcher_tests.py
chriskuehl: Maybe replacing with `grp.getgrgid(os.getegid()).gr_name` would work and also produce the right group name outside of tests?

RyanBalfanz: The docs use `www-data` but that wasn't an easy option in either context. I wasn't concerned much about the specific user, but did want to try and capture the implementation behavior which does change the user, group, and permissions (though this last bit is not under test in this PR).

tests/unit/sidecars/secrets_fetcher_tests.py
RyanBalfanz: done.

chriskuehl (approved):
Thanks for adding this!

PR #1021, opened 2024-11-12
tracing: Only decode the traceparent and tracestate headers

trevorriles:
We all change the UnicodeDecodeError exception to be a debug message. This really shouldn't happen oncee we are restricting the headers to tracestate and traceparent though.

### DIFF
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -84,7 +84,7 @@ class _ContextAwareHandler:
                 try:
                     # this is only for w3c trace headers
                     key = k.decode()
-                    if key == "traceparent" or key == "tracestate":
+                    if key in ["traceparent", "tracestate", "Trace", "Span", "Sampled", "Flags"]:
                         header_dict[key] = v.decode()
                 except UnicodeDecodeError:
                     self.logger.info(f"Unable to decode header {k.decode()}, ignoring.")
--- a/baseplate/frameworks/thrift/__init__.py
+++ b/baseplate/frameworks/thrift/__init__.py
@@ -58,6 +58,16 @@ PROM_ACTIVE = Gauge(
     ["thrift_method"],
     multiprocess_mode="livesum",
 )
+W3C_HEADERS: frozenset[bytes] = frozenset(
+    (
+        b"traceparent",
+        b"tracestate",
+        b"trace",
+        b"span",
+        b"sampled",
+        b"flags",
+    )
+)
 
 
 class _ContextAwareHandler:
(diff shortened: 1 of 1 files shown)


salomon-smekecohen (approved):
thanks Trevor!

baseplate/frameworks/thrift/__init__.py
```diff
@@ -82,7 +82,10 @@ def _set_remote_context(self, request_context: RequestContext) -> Iterator[None]
             header_dict = {}
             for k, v in headers.items():
                 try:
-                    header_dict[k.decode()] = v.decode()
+                    # this is only for w3c trace headers
+                    key = k.decode()
+                    if key == "traceparent" or key == "tracestate":
```
chriskuehl: Just checking, are we sure these are always lowercase? I would expect them to be since WSGI normalizes header capitalization but probably worth checking.

chriskuehl (approved):
lgtm other than test failures

baseplate/frameworks/thrift/__init__.py
```diff
@@ -82,7 +82,10 @@ def _set_remote_context(self, request_context: RequestContext) -> Iterator[None]
             header_dict = {}
             for k, v in headers.items():
                 try:
-                    header_dict[k.decode()] = v.decode()
+                    # this is only for w3c trace headers
+                    key = k.decode()
+                    if key in ["traceparent", "tracestate", "Trace", "Span", "Sampled", "Flags"]:
```
salomon-smekecohen: can we move this slightly magic array to a set somewhere?

W3C_HEADERS = {...} ?

baseplate/frameworks/thrift/__init__.py
```diff
@@ -82,7 +82,10 @@ def _set_remote_context(self, request_context: RequestContext) -> Iterator[None]
...
+                        header_dict[key] = v.decode()
                 except UnicodeDecodeError:
                     self.logger.info(f"Unable to decode header {k.decode()}, ignoring.")
```
chriskuehl: Another possibility if these can contain unsanitized user input would be to ignore encoding errors, e.g.

```python
header_dict[k.decode(errors="replace")] = v.decode(errors="replace")
```

That would replace invalid UTF-8 sequences with �.

Not sure if that is appropriate here or not. I wouldn't want to hide legitimate errors like this, but if these are raw bytes from users, then we're always going to get some quantity of malformed requests from poorly-behaving UAs and we wouldn't want to hide real errors among decode errors that we can't actually do anything about.

baseplate/frameworks/thrift/__init__.py
trevorriles: They should only be being set by the otel lib at this point and the spec is lowercase. I think we should be good.

trevorriles: But maybe we should just do a non case-sensitive test now that we have the old trace headers.

baseplate/frameworks/thrift/__init__.py
trevorriles: For the most part we're building this dictionary so that we can use the otel propagator injection code which expects a dictionary of strings. We' don't really care about any of the other headers which is why I shifted towards making it more specific.

trevorriles: Invalid values should be safe to just drop, if we do encode them with � it would just be considered invalid by otel later.

baseplate/frameworks/thrift/__init__.py
chriskuehl: +1, maybe `W3C_HEADERS = frozenset((...))` as a general best practice so it's not mutable.

salomon-smekecohen (approved):
also slightly smaller perf impact from only decoding relevant headers. i like it.

PR #1022, opened 2024-11-12
Temporarily restore Python 3.8 support

chriskuehl:
This temporarily reverts #1003 as we want to keep the 2.7.1 release small. We will un-revert it in a day or two once we make the 2.7.1 release.

This ended up being a bit tricker than I hoped since Ruff had rewritten a lot of type annotations to make them no longer compatible with 3.8, so I had to unwind that.


chriskuehl:
As usual pylint is giving useless warnings.

chriskuehl:
Giving up, don't have the bandwidth to fight pylint.

chriskuehl:
Draft PR to remove pylint in #1023

PR #1028, opened 2024-11-27
Patch gevent to work with tracing

trevorriles:
Instrument gevent so that it passes opentelemetry context to greenlets. We already have a similar implementation deployed to reddit's internal graphql service. After receiving a bug report from other services missing trace context propagation the fix is being move into baseplate.py.

I've verified that this fix resolves an issue where log lines inside of a greenlet were missing the tracing info.

### DIFF
--- a/baseplate/lib/tracing.py
+++ b/baseplate/lib/tracing.py
@@ -54,7 +54,7 @@ __IMapUnordered = gevent.pool.IMapUnordered
 
 class Runnable(Protocol):
     @property
-    def trace_context(self) -> Context: ...
+    def bp_trace_context(self) -> Context: ...
 
     run: Callable
 
--- a/baseplate/lib/tracing.py
+++ b/baseplate/lib/tracing.py
@@ -1,5 +1,5 @@
-from collections.abc import Sequence
-from typing import Optional
+from collections.abc import Sequence, Callable
+from typing import Optional, Protocol
 
 import gevent.pool
 from opentelemetry import context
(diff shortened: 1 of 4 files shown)


baseplate/lib/tracing.py
```diff
@@ -42,3 +44,72 @@ def should_sample(
...
+
+class Runnable(Protocol):
+    @property
+    def trace_context(self) -> Context: ...
+
+    run: Callable
+
+
+class TracingMixin:
+    def __init__(self: Runnable, *args, **kwargs) -> None:  # type: ignore
```
chriskuehl: Instead of `type: ignore`, can we do

```diff
diff --git a/baseplate/lib/tracing.py b/baseplate/lib/tracing.py
index 2f6ac41..d5aa671 100644
--- a/baseplate/lib/tracing.py
+++ b/baseplate/lib/tracing.py
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Sequence
-from typing import Optional, Protocol
+from typing import Optional, Protocol, Any

 import gevent.pool
 from opentelemetry import context
@@ -60,7 +60,7 @@ class Runnable(Protocol):


 class TracingMixin:
-    def __init__(self: Runnable, *args, **kwargs) -> None:  # type: ignore
+    def __init__(self: Runnable, *args: Any, **kwargs: Any) -> None:
         self.trace_context = context.get_current()
         super().__init__(*args, **kwargs)
```

This seems to pass mypy for me.

baseplate/lib/tracing.py
```diff
@@ -42,3 +44,72 @@ def should_sample(
...
+        self.trace_context = context.get_current()
```
chriskuehl: nit: if there aren't naming restrictions, maybe name it `baseplate_trace_context` to avoid potential future clashes with the gevent class?

chriskuehl (approved):

baseplate/lib/tracing.py
trevorriles: Thanks, I wasn't able to figure this one out!

baseplate/lib/tracing.py
trevorriles: Good call, made the change.

baseplate/lib/tracing.py
```diff
@@ -42,3 +44,72 @@ def should_sample(
...
+
+# Greenlet tracing utils
+__Greenlet = gevent.Greenlet
+__IMap = gevent.pool.IMap
+__IMapUnordered = gevent.pool.IMapUnordered
+
+
+class Runnable(Protocol):
+    @property
+    def trace_context(self) -> Context: ...
```
chriskuehl:
```suggestion
    def bp_trace_context(self) -> Context: ...
```

Missed this one I think.

trevorriles: Yup, sorry. Just pushed the fix

chriskuehl (approved):

PR #1042, opened 2024-12-04
Require sentry-sdk>=1.40.6

chriskuehl:
[sentry-sdk 1.40.6 contains](https://github.com/getsentry/sentry-python/blob/3e43a91b0e7f90f73a4165f7b58d5a10567e19bc/CHANGELOG.md#1406):

> Fix compatibility with greenlet/gevent ([#2756](https://github.com/getsentry/sentry-python/pull/2756)) by `@sentrivana`

@salomon-smekecohen found that this improves performance significantly for Baseplate services, so let's bump up our minimum version to help push users to upgrade.


mathyourlife-reddit (approved):
thanks @salomon-smekecohen !

salomon-smekecohen:
Np! I dont know what minimal version introduced the problem, so this might not help much 😅 but I agree its a good idea to require.

scotthew1 (approved):

PR #1073, opened 2025-02-21
Use processor/client module name as thrift rpc service.

trevorriles:
The goal is to have both ends of the rpc class the same across frameworks, similar to how grpc operates.

This should align us with the upcoming baseplate.go changes. 

https://github.com/reddit/baseplate.go/pull/686

Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/tests/integration/otel_thrift_tests.py
+++ b/tests/integration/otel_thrift_tests.py
@@ -169,7 +169,7 @@ class ThriftTraceHeaderTests(GeventPatchedTestCase, TestBase):
             thrift_client_span,
             {
                 SpanAttributes.RPC_SYSTEM: "thrift",
-                SpanAttributes.RPC_SERVICE: THRIFT_CLIENT_NAME,
+                SpanAttributes.RPC_SERVICE: "tests.integration.test_thrift.TestService",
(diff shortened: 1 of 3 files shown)


tests/integration/otel_thrift_tests.py
```diff
@@ -220,12 +220,14 @@ def example(self, context):
             thrift_client_span.get_span_context().span_id,
         )
 
-        self.assertEqual(thrift_client_span.name, "example_service/example")
+        self.assertEqual(
+            thrift_client_span.name, "tests.integration.test_thrift.TestService/example"
```
marcoferrer, line 224: So this contains the full python package name. To get something close to a "thrift" package, you'll need to only keep the last part of the package. 

The assertion should be against `test_thrift.TestService`.

marcoferrer: The python file name usually maps to the thrift schema name.

trevorriles: Sounds good, is it always only a single depth package? I should be able to just split on `.` and take the last element.

